PR #4, opened 2021-03-04
chore(examples): add example for creating a Http Server

amitksingh1490:
Add Simple HTTP Server in examples

### DIFF
--- a/examples/src/main/scala/app/Main.scala
+++ b/examples/src/main/scala/app/Main.scala
@@ -21,9 +21,8 @@ object Main extends App {
     _ <- console.putStrLn(s"Server started on $PORT")
   } yield ()
 
-  private val env = ServerChannelFactory.Live.auto.toManaged_.toLayer ++ EventLoopGroup.Live.auto(0).toLayer ++
-    Logging.console()
+  private val env = ServerChannelFactory.live ++ EventLoopGroup.live(0)
 
   override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
-    (server.toManaged_).useForever.provideCustomLayer(env).exitCode
+    server.toManaged_.useForever.provideCustomLayer(env).exitCode
 }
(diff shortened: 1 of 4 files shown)


examples/src/main/scala/app/Main.scala
```diff
@@ -0,0 +1,35 @@
+package app
+
+import io.circe.generic.auto._
+import io.circe.syntax._
+import zhttp.domain.http.HttpApp
+import zhttp.domain.http.model.Method.GET
+import zhttp.domain.http.model.Path.{/, Root}
+import zhttp.domain.http.model.Response
+import zhttp.service.netty.EventLoopGroup
+import zhttp.service.netty.server.{Server, ServerChannelFactory}
```
tusharmath, line 10: Too much noise in imports. Can we reduce it to just a one line import from zhttp?

examples/src/main/scala/app/Main.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  private val httpApp = HttpApp.route {
+    case GET -> Root / "plaintext" => Response.text("Hello, World!")
+    case GET -> Root / "json"      => Response.jsonString(Message("Hello, World!").asJson.noSpaces)
+  }
+
+  private def server = for {
+    s <- Server.make(httpApp)
+    _ <- s.start(PORT)
+    _ <- Logging.info(s"Server started on $PORT")
```
tusharmath, line 27: Can drop logging to improve brevity.

examples/src/main/scala/app/Main.scala
```diff
@@ -0,0 +1,35 @@
...
+  } yield ()
+
+  private val env = ServerChannelFactory.Live.auto.toManaged_.toLayer ++ EventLoopGroup.Live.auto(0).toLayer ++
+    Logging.console()
+
```
tusharmath, line 32: Can we make it more concise?

tusharmath (changes requested):

PR #39, opened 2021-03-16
chore(build): crossVersion publishing

amitksingh1490:


build.sbt
```diff
@@ -4,12 +4,15 @@ import scala.concurrent.duration.FiniteDuration
 
 import sbt.enablePlugins
 
-Global / scalaVersion := "2.13.3"
-
```
tusharmath, line 8: I don't think we need to remove this?

build.sbt
```diff
@@ -4,12 +4,15 @@ import scala.concurrent.duration.FiniteDuration
...
 // ZIO Version
 val zioVersion       = "1.0.4-2"
 val zioConfigVersion = "1.0.0"
 val circeVersion     = "0.13.0"
+val scalaVer1        = "2.13.3"
```
tusharmath, line 11:
```suggestion
val scala_2_13        = "2.13.3"
```

tusharmath (changes requested):

build.sbt
amitksingh1490: It's not removed just moved down line no 15

tusharmath (approved):

PR #48, opened 2021-03-17
fix(ServerRequestHandler): Websocket handler attached before handshake in webSocketUpgrade

### LINKED ISSUE #46: Bug: Messages getting lost for some websocket connections.
For some WebSocket apps, the channel handler can get added after the initial message is already sent.

ShrutiVerma97:
fixes #46


tusharmath (approved):

zio-http/src/main/scala/zhttp/service/server/ServerRequestHandler.scala
```diff
@@ -33,17 +33,15 @@ final case class ServerRequestHandler[R](
     if (hh == null) {
       JWebSocketServerHandshakerFactory.sendUnsupportedVersionResponse(ctx.channel, ctx.channel().voidPromise())
     } else {
+      val pl = ctx.channel().pipeline()
+      pl.addLast(WEB_SOCKET_HANDLER, ServerSocketHandler(zExec, res.socket))
+      pl.remove(HTTP_REQUEST_HANDLER)
```
tusharmath, line 38: We might not want to remove the Http Request Handler, because it's possible that the handshake fails and the client retries an HTTP request.

tusharmath (changes requested):

zio-http/src/main/scala/zhttp/service/server/ServerRequestHandler.scala
ShrutiVerma97: We can remove the HTTP Request handler once the handshake succeeds 
```
     val pl = ctx.channel().pipeline()
      pl.addLast(WEB_SOCKET_HANDLER, ServerSocketHandler(zExec, res.socket))
      try {
        // handshake can throw
        hh.handshake(ctx.channel(), jReq).addListener { (future: JChannelFuture) =>
          if (!future.isSuccess) {
            ctx.fireExceptionCaught(future.cause)
            ()
          }else{
            pl.remove(HTTP_REQUEST_HANDLER)
            ()
          }

        }
      } catch {
        case _: JWebSocketHandshakeException =>
          JWebSocketServerHandshakerFactory.sendUnsupportedVersionResponse(ctx.channel, ctx.channel().voidPromise())
      }

tusharmath (approved):

PR #58, opened 2021-03-19
Update sbt to 1.5.0-RC1

sh0hei:


build.sbt
```diff
@@ -15,7 +15,7 @@ Global / scalaVersion := scala_2_13
 
 lazy val root = (project in file("."))
   .settings(
-    skip in publish := true,
+    publish / skip := true,
```
sh0hei, line 18: cf. https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#slash

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.4.7
+sbt.version=1.5.0-RC1
```
tusharmath, line 1: Not sure if we should update the version already. The changes can be incorporated though.

tusharmath (changes requested):

tusharmath:
Closing in favour of #119

PR #83, opened 2021-03-28
Basic Query String Support in URL

davidhoyt:


zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -3,20 +3,20 @@ package zhttp.http
...
-    case URL.Location.Absolute(_, host, _) => Option(host)
+    case URL.Location.Relative      => None
+    case abs: URL.Location.Absolute => Option(abs.host)
   }
 
   val port: Option[Int] = kind match {
-    case URL.Location.Relative             => None
-    case URL.Location.Absolute(_, _, port) => Option(port)
+    case URL.Location.Relative      => None
+    case abs: URL.Location.Absolute => Option(abs.port)
```
tusharmath, line 14: Just out of curiosity is this needed?

davidhoyt: No, but it's more succinct when you have a growing number of fields.

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -45,12 +47,22 @@ object URL                                                             {
...
     } yield url
 
-  def asString(url: URL): String = url.kind match {
-    case Location.Relative                     => url.path.asString
-    case Location.Absolute(scheme, host, port) =>
-      port match {
-        case 80   => s"${scheme.asString}://${host}${url.path.asString}"
-        case port => s"${scheme.asString}://${host}:${port}${url.path.asString}"
+  def asString(url: URL): String = {
+    def withQuery(u: String): String =
```
tusharmath, line 51: can you move `withQuery` out of `asString`. This could cause perf issues.

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -45,12 +47,22 @@ object URL                                                             {
...
+      if (url.query.nonEmpty) {
+        s"$u?${url.query}"
+      } else {
+        u
       }
+
+    url.kind match {
+      case Location.Relative                     => withQuery(url.path.asString)
+      case Location.Absolute(scheme, host, port) =>
+        (scheme, port) match {
```
tusharmath, line 61: Tuple Boxing will incur additional allocations and might affect perf.

zio-http/src/main/scala/zhttp/http/URL.scala
davidhoyt: How does that cause perf issues?

zio-http/src/main/scala/zhttp/http/URL.scala
davidhoyt: It's producing a string. Not typically in a hot path. Did it materially affect benchmarks?

The main point was to convey that the existing logic had a bug. HTTPS and port 80 would have omitted the port but the default port for https is 443.

tusharmath:
Thanks @David-Hoyt-CK! 🙏 🙏 🙏

PR #102, opened 2021-03-31
handle connection reset by peer exception

ravisharnagat:


zio-http/src/main/scala/zhttp/service/server/ServerSocketHandler.scala
```diff
@@ -10,12 +10,12 @@ import zhttp.socket.{Socket, WebSocketFrame}
 final case class ServerSocketHandler[R, E](
   zExec: UnsafeChannelExecutor[R],
   socket: Socket[R, E, WebSocketFrame, WebSocketFrame],
-) extends JSimpleChannelInboundHandler[JWebSocketFrame] {
+) extends JSimpleChannelInboundHandler[JWebSocketFrame]
+    with ServerHttpExceptionHandler { self =>
```
amitksingh1490, line 14: Can we rename `ServerHttpExceptionHandler` to `ServerExceptionHandler` since we are using it in Http as well as websocket handler

amitksingh1490:
Not Needed with new socket API

PR #114, opened 2021-04-02
Multipart formdata

DJLemkes:
Hi!

I created a multipart/form-data request handler to also support those requests. Although I tried to minimize the amount of changes I couldn't get around changing the `Request.Data.content` to `HttpContent[Any, Byte]` instead of `HttpContent[Any, String]`. You probably have some better ideas about this setup than I do so I just want to put this out here and hear your thoughts on this.

Best,

DJ

### DIFF
--- a/example/src/main/scala/HelloWorldAdvanced.scala
+++ b/example/src/main/scala/HelloWorldAdvanced.scala
@@ -2,6 +2,7 @@ import zhttp.http._
 import zhttp.service._
 import zhttp.service.server.ServerChannelFactory
 import zio._
+import zio.stream.ZStream
 
 import scala.util.Try
 
@@ -14,15 +15,29 @@ object HelloWorldAdvanced extends App {
     case Method.GET -> Root / "bar" => Response.text("foo")
   }
 
+  private def streamSize[R, E, A](stream: ZStream[R, E, A]): ZIO[R, E, Long] =
+    stream.mapChunks(c => Chunk(c.size)).fold(0L)(_ + _)
+
   private val app = HttpApp.collectM {
-    case Method.GET -> Root / "random" => random.nextString(10).map(Response.text)
-    case Method.GET -> Root / "utc"    => clock.currentDateTime.map(s => Response.text(s.toString))
+    case Method.GET -> Root / "random"             => random.nextString(10).map(Response.text)
+    case Method.GET -> Root / "utc"                =>
+      clock.currentDateTime.map(s => Response.text(s.toString))
+    case req @ Method.POST -> Root / "file-upload" =>
+      req.content match {
+        case fd: HttpData.MultipartFormData =>
+          val attributes = fd.attributes.map { case (name, attr) =>
+            streamSize(attr.content).map(size => s"$name: $size")
+          }
+          val files      = fd.files.map { case (name, file) => streamSize(file.content).map(size => s"$name: $size") }
+          ZIO.collectAllParN(3)(attributes ++ files).map(sizes => Response.text(sizes.mkString(",")))
+        case _                              => UIO.succeed(Response.status(Status.BAD_REQUEST))
+      }
   }
 
   private val server =
-    Server.port(PORT) ++              // Setup port
-      Server.paranoidLeakDetection ++ // Paranoid leak detection (affects performance)
-      Server.app(fooBar +++ app)      // Setup the Http app
+    Server.port(PORT) ++             // Setup port
+      Server.disableLeakDetection ++ // Paranoid leak detection (affects performance)
+      Server.app(fooBar +++ app)     // Setup the Http app
 
   override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] = {
     // Configure thread count using CLI
--- a/zio-http/src/main/scala/zhttp/http/HttpData.scala
+++ b/zio-http/src/main/scala/zhttp/http/HttpData.scala
@@ -2,6 +2,7 @@ package zhttp.http
 
 import io.netty.buffer.{ByteBuf, ByteBufUtil}
 import zio.Chunk
+import zio.blocking.Blocking
 import zio.stream.ZStream
 
 /**
@@ -13,6 +14,21 @@ object HttpData {
   case object Empty                                            extends HttpData[Any, Nothing]
   final case class CompleteData(data: Chunk[Byte])             extends HttpData[Any, Nothing]
   final case class StreamData[R, E](data: ZStream[R, E, Byte]) extends HttpData[R, E]
+
+  sealed trait FormDataContent extends Product with Serializable
+
+  final case class FileData(name: String, contentType: String, content: ZStream[Blocking, Throwable, Byte])
+      extends FormDataContent
+  final case class AttributeData(name: String, content: ZStream[Blocking, Throwable, Byte]) extends FormDataContent
+
+  object MultipartFormData {
+    def empty: MultipartFormData = MultipartFormData(Map.empty, Map.empty)
+  }
+
+  final case class MultipartFormData(
+    attributes: Map[String, AttributeData],
+    files: Map[String, FileData],
+  ) extends HttpData[Any, Nothing]
 
   /**
    * Helper to create CompleteData from ByteBuf
--- a/zio-http/src/main/scala/zhttp/http/Request.scala
+++ b/zio-http/src/main/scala/zhttp/http/Request.scala
@@ -15,7 +15,6 @@ final case class Request(
     case HttpData.CompleteData(data) => Option(data.map(_.toChar).mkString)
     case _                           => Option.empty
   }
-
 }
 
 object Request {}
--- a/zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
+++ b/zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
@@ -1,19 +1,33 @@
 package zhttp.service.server
 
+import io.netty.handler.codec.http.HttpObjectDecoder.{
+  DEFAULT_MAX_CHUNK_SIZE,
+  DEFAULT_MAX_HEADER_SIZE,
+  DEFAULT_MAX_INITIAL_LINE_LENGTH,
+}
 import io.netty.handler.codec.http.HttpServerKeepAliveHandler
 import zhttp.core._
-import zhttp.service.{HTTP_KEEPALIVE_HANDLER, HTTP_REQUEST_HANDLER, OBJECT_AGGREGATOR, SERVER_CODEC_HANDLER}
+import zhttp.service._
 
 /**
  * Initializes the netty channel with default handlers
  */
 @JSharable
-final case class ServerChannelInitializer(httpH: JChannelHandler, maxSize: Int) extends JChannelInitializer[JChannel] {
+final case class ServerChannelInitializer(httpH: JChannelHandler, hhtpMfdH: JChannelHandler, maxSize: Int)
+    extends JChannelInitializer[JChannel] {
   override def initChannel(channel: JChannel): Unit = {
     channel
       .pipeline()
-      .addLast(SERVER_CODEC_HANDLER, new JHttpServerCodec)
+      // Without increasing the DEFAULT_MAX_CHUNK_SIZE uploads are rather slow. Probably because Netty is chunking up messages
+      // itself. With a file upload, that causes lots of overhead. A factor >8 (resulting in 65K chunks) is ignored so it seems.
+      .addLast(
+        SERVER_CODEC_HANDLER,
+        new JHttpServerCodec(DEFAULT_MAX_INITIAL_LINE_LENGTH, DEFAULT_MAX_HEADER_SIZE, DEFAULT_MAX_CHUNK_SIZE * 8),
+      )
       .addLast(HTTP_KEEPALIVE_HANDLER, new HttpServerKeepAliveHandler)
+      // This handler should always be in front of the `JHttpObjectAggregator` or we won't be able to
+      // decode multipart/form-data requests correctly.
+      .addLast(MULTIPART_FORMDATA_HANDLER, hhtpMfdH)
       .addLast(OBJECT_AGGREGATOR, new JHttpObjectAggregator(maxSize))
       .addLast(HTTP_REQUEST_HANDLER, httpH)
     ()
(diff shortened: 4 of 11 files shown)


zio-http/src/main/scala/zhttp/http/Request.scala
```diff
@@ -8,15 +10,16 @@ final case class Request(endpoint: Endpoint, data: Request.Data = Request.Data.e
...
     case _                          => Option.empty
   }
-
 }
 
 object Request {
-  val emptyContent: HttpContent.Complete[String] = HttpContent.Complete("")
-  final case class Data(headers: List[Header], content: HttpContent[Any, String])
+  val emptyContent: HttpContent.Chunked[Any, Byte] = HttpContent.Chunked(data = ZStream.empty)
+  final case class Data(headers: List[Header], content: HttpContent[Any, Byte])
```
tusharmath, line 22: I am guessing just changing the content to `HttpContent[Any, ByteBuf]` should allow us to make any kind of request. The parsing can be left to the developer or specific codecs can be provided as a part of the library.

tusharmath:
`io.netty.Buffer.Bytebuf` is a mutable, unsafe data structure and also exposes implementation details. It would be great if we could come up with a performant immutable indirection.

DJLemkes:
@tusharmath `io.netty.Buffer.Bytebuf` is indeed mutable and pretty tricky to work with. Take the [messagecodec](https://github.com/dream11/zio-http/blob/597f5d5baa8d9dc9d8c4720692b4bd358358c8f4/zio-http/src/main/scala/zhttp/service/HttpMessageCodec.scala#L34) for example: you need to use this utility function because just using `.content.array()` will throw exceptions caused by [mixed reader and writer indices](https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html).

Personally, I like offering `HttpContent[Any, Byte]` because it allows for lots of flexibility, also for non-text `content-type`. With some added utils like `asString` it still pretty easy to use. I do understand though that this train of thought conflicts with the current API.  Also, you probably already have ideas on how to offer streaming request bodies as I saw some discussion on Discord? I am happy to wait for those design details and then adjust this PR to that interface? The handler doesn't have to change that much; just the way the data is offered to the `zio-http` API.

DJLemkes:
@tusharmath what are your thoughts on my comment?

Also another question: I see that `headers` is part of `Request.Data` and not of `Request` directly. In the `Response`, it's directly part of `Response` and not of its data. I'm probably missing something; what's the reason for that?

tusharmath:
I'll take a look at this PR today.

> Also another question: I see that headers is part of Request.Data and not of Request directly. In the Response, it's directly part of Response and not of its data. I'm probably missing something; what's the reason for that?

Yes, I also think we don't need `Request.Data` anymore.

zio-http/src/main/scala/zhttp/http/HttpContent.scala
```diff
@@ -10,4 +11,20 @@ sealed trait HttpContent[-R, +A] extends Product with Serializable
 object HttpContent {
   final case class Complete[A](data: A)                        extends HttpContent[Any, A]
   final case class Chunked[R, A](data: ZStream[R, Nothing, A]) extends HttpContent[R, A]
+
+  sealed trait FormDataContent extends Product with Serializable
+
+  // TODO: Make the environment R instead of hardcoded Blocking.
+  // It's set to Blocking for now because we need it when reading from a tmp file.
+  final case class FileData(name: String, contentType: String, content: ZStream[Blocking, Throwable, Byte])
+      extends FormDataContent
```
tusharmath, line 20:
```suggestion
  final case class FileData[R](name: String, contentType: String, content: ZStream[Blocking, Nothing, Chunk[A]])
      extends FormDataContent
```

Why should the stream throw?

zio-http/src/main/scala/zhttp/http/HttpContent.scala
```diff
@@ -10,4 +11,20 @@ sealed trait HttpContent[-R, +A] extends Product with Serializable
...
+  final case class AttributeData(name: String, content: ZStream[Blocking, Throwable, Byte]) extends FormDataContent
```
tusharmath, line 21:
```suggestion
  final case class AttributeData[R](name: String, content: ZStream[R, Nothing, Chunk[A]]) extends FormDataContent
```

zio-http/src/main/scala/zhttp/http/HttpContent.scala
```diff
@@ -10,4 +11,20 @@ sealed trait HttpContent[-R, +A] extends Product with Serializable
...
+
+  object MultipartFormData {
+    def empty: MultipartFormData = MultipartFormData(Map.empty, Map.empty)
+  }
+  final case class MultipartFormData(
+    attributes: Map[String, AttributeData],
+    files: Map[String, FileData],
```
tusharmath, line 28: since the name is there in both `FileData` and `AttributeData` do we need a `Map` here?

DJLemkes: Thought the `Map` would be convenient for by-name lookup. Can make it a `Seq` as well but then you have to do a `find`. I personally like keeping the `name` in the `Data` types as well such that you can pass these around and keep everything together.

DJLemkes:
Rebased and updated. Removing `Draft` as more structural changes required at the time have landed in `main`. 

I think this [Tapir PR](https://github.com/softwaremill/tapir/pull/1150) is waiting for this.

Note that it seems we're also affected by [this](https://github.com/netty/netty/issues/11175) Netty bug that should be resolved in the next patch release.

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -14,6 +15,21 @@ object HttpData {
...
+  final case class AttributeData(name: String, content: ZStream[Blocking, Throwable, Byte]) extends FormDataContent
+
+  object MultipartFormData {
+    def empty: MultipartFormData = MultipartFormData(Map.empty, Map.empty)
+  }
+
+  final case class MultipartFormData(
+    attributes: Map[String, AttributeData],
+    files: Map[String, FileData],
+  ) extends HttpData[Any, Nothing]
```
tusharmath, line 31: I think we should add `Blocking` and `Throwable` to HttpData.

```suggestion
  ) extends HttpData[Blocking, Throwable]
```

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
```diff
@@ -1,19 +1,33 @@
...
+      // Without increasing the DEFAULT_MAX_CHUNK_SIZE uploads are rather slow. Probably because Netty is chunking up messages
+      // itself. With a file upload, that causes lots of overhead. A factor >8 (resulting in 65K chunks) is ignored so it seems.
+      .addLast(
+        SERVER_CODEC_HANDLER,
+        new JHttpServerCodec(DEFAULT_MAX_INITIAL_LINE_LENGTH, DEFAULT_MAX_HEADER_SIZE, DEFAULT_MAX_CHUNK_SIZE * 8),
+      )
       .addLast(HTTP_KEEPALIVE_HANDLER, new HttpServerKeepAliveHandler)
+      // This handler should always be in front of the `JHttpObjectAggregator` or we won't be able to
+      // decode multipart/form-data requests correctly.
+      .addLast(MULTIPART_FORMDATA_HANDLER, hhtpMfdH)
```
tusharmath, line 30: We should add this handler conditionally ie. 
1. If the request is of multipart type (this is easy to check thru headers)
2. The user actually wants to parse it as a multipart request. It's possible that the user wants to send a 403 error or something else even before the data is uploaded (There is a way to check this also).

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
```diff
@@ -1,19 +1,33 @@
...
   override def initChannel(channel: JChannel): Unit = {
     channel
       .pipeline()
-      .addLast(SERVER_CODEC_HANDLER, new JHttpServerCodec)
+      // Without increasing the DEFAULT_MAX_CHUNK_SIZE uploads are rather slow. Probably because Netty is chunking up messages
+      // itself. With a file upload, that causes lots of overhead. A factor >8 (resulting in 65K chunks) is ignored so it seems.
+      .addLast(
+        SERVER_CODEC_HANDLER,
+        new JHttpServerCodec(DEFAULT_MAX_INITIAL_LINE_LENGTH, DEFAULT_MAX_HEADER_SIZE, DEFAULT_MAX_CHUNK_SIZE * 8),
+      )
```
tusharmath, line 26: I think we should add this handler also conditionally similar to the `MULTIPART_FORMDATA_HANDLER`

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -14,6 +15,21 @@ object HttpData {
   final case class CompleteData(data: Chunk[Byte])             extends HttpData[Any, Nothing]
   final case class StreamData[R, E](data: ZStream[R, E, Byte]) extends HttpData[R, E]
 
+  sealed trait FormDataContent extends Product with Serializable
+
+  final case class FileData(name: String, contentType: String, content: ZStream[Blocking, Throwable, Byte])
+      extends FormDataContent
+  final case class AttributeData(name: String, content: ZStream[Blocking, Throwable, Byte]) extends FormDataContent
```
tusharmath, line 22: Is there a better way building a stream without depending on `Blocking`?

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
DJLemkes: Thanks for reviewing!

I see your points. Regarding 1), we let Netty detect whether it's a multipart because they have a built-in decoder already which probably also covers some edge cases. The multipart handler is then removed as soon as we know it's not a multipart request [here](https://github.com/dream11/zio-http/blob/99f61ab80488c458a25462166fc2507323ca5d93/zio-http/src/main/scala/zhttp/service/server/ServerMultipartRequestHandler.scala#L72) and [here](https://github.com/dream11/zio-http/blob/99f61ab80488c458a25462166fc2507323ca5d93/zio-http/src/main/scala/zhttp/service/server/ServerMultipartRequestHandler.scala#L84).

Removing a handler is much easier than adding at exactly the right position in the chain. Especially the latter is important as long as we have the aggregator handler.

Curious to see the details on 2)!

zio-http/src/main/scala/zhttp/http/HttpData.scala
DJLemkes: Couldn't find one because [we're doing a `ZStream.fromFile`](https://github.com/dream11/zio-http/blob/99f61ab80488c458a25462166fc2507323ca5d93/zio-http/src/main/scala/zhttp/service/DecodeJRequest.scala#L54) when decoding the multipart. But please let me know if we can provide the `R` earlier already.

example/src/main/scala/HelloWorldAdvanced.scala
```diff
@@ -14,15 +15,29 @@ object HelloWorldAdvanced extends App {
...
+    case req @ Method.POST -> Root / "file-upload" =>
+      req.content match {
+        case fd: HttpData.MultipartFormData =>
+          val attributes = fd.attributes.map { case (name, attr) =>
+            streamSize(attr.content).map(size => s"$name: $size")
+          }
+          val files      = fd.files.map { case (name, file) => streamSize(file.content).map(size => s"$name: $size") }
+          ZIO.collectAllParN(3)(attributes ++ files).map(sizes => Response.text(sizes.mkString(",")))
+        case _                              => UIO.succeed(Response.status(Status.BAD_REQUEST))
+      }
```
tusharmath, line 34: Can you create a separate example to showcase multipart usage.

zio-http/src/main/scala/zhttp/service/server/ServerMultipartRequestHandler.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  self =>
+
+  private val factory                                          = new DefaultHttpDataFactory(DefaultHttpDataFactory.MINSIZE)
+  private val decoderKey: AttributeKey[HttpPostRequestDecoder] = AttributeKey.valueOf("decoder")
+  private val uriKey: AttributeKey[String]                     = AttributeKey.valueOf("uri")
+  private val headersKey: AttributeKey[HttpHeaders]            = AttributeKey.valueOf("headers")
+  private val protocolVersionKey: AttributeKey[HttpVersion]    = AttributeKey.valueOf("protocolVersion")
+
+  // TODO: nearly the samve as the one in `ServerRequestHandler`. Generalize and lift?
```
tusharmath, line 28: Keep it as a separate refactor PR and can have this merged first :)

tusharmath:
**Response.scala**
```scala
sealed trait Response[-R, +E]

object Response {
  case class HttpResponse(status: Status, headers: List[Header], content: HttpData[R, E]) extends Response[R, E]
  case class SocketResponse[-R, +E](socket: SocketApp[R, E]) extends Response[R, E]

  // Responses with the prefix `ReadAs` is used to signal the server to change the channel handler.
  // Based on the response various channel handlers can be created.
  case class ReadAsComplete(cb: HttpData.CompleteData[] => HttpApp[R, E]) extends Response[R, E]
  case class ReadAsMultipart(cb: HttpData.MultipartFormData => HttpApp[R, E]) extends Response[R, E]
}
```


This is how we will respond to a multipart request.
**MultipartExample.scala**
```scala
Http.collect[Request] {
  case GET -> Root / "upload" => Response.AsMultipart { fd => 
    HttpApp.response(Response.HttpResponse(Status.OK, Nil, size(fd)))
  }
}
```

This is how we can process the responses
**ServerRequestHandler.scala**

```scala
def channelRead(ctx: ChannelHandlerContext, msg: HttpRequest) = {
  ctx.channel.config.setAutoRead(false) // Stop reading bytes for now.
  executeAsync(app(msg)) {
     case HttpResponse() => ??? /// continue as before
     case ReadAsMultipart() => 
       /// Remove the current handler
       /// Add the multipart handler
       ctx.channel.read() // Read the channel in chunks now.
     case ReadAsComplete() =>
       /// Add HttpObjectAggregator
      ctx.channel.read()
  }
}
```

We can further cleanup the `MultipartExample` as follows —

>> TODO: (will add an example)

DJLemkes:
Nice approach! Have been busy but I expect to be able to give this some time later this week.

tusharmath:
We will support this via the content-decoder API. Thanks @DJLemkes !

PR #128, opened 2021-04-08
refactor(server): update handlers

amitksingh1490:


.scalafmt.conf
```diff
@@ -18,6 +18,6 @@ rewrite.rules = [SortImports, RedundantBraces, SortModifiers]
 docstrings.wrap = yes
 docstrings.style = Asterisk
 
-newlines.afterInfix = keep
+newlines.afterInfix = many
```
tusharmath, line 21: I think we should keep this.

zio-http/src/main/scala/zhttp/service/server/ServerSocketHandler.scala
```diff
@@ -49,4 +50,15 @@ final case class ServerSocketHandler[R](
...
+  override def userEventTriggered(ctx: JChannelHandlerContext, event: AnyRef): Unit = event match {
+    case _: HandshakeComplete => {
+      executeAsync(ctx, ss.onOpen(ctx.channel().remoteAddress()).uninterruptible)
+      ()
+    }
+    case event                => {
+      ctx.fireUserEventTriggered(event)
+      ()
+    }
+  }
```
tusharmath, line 63:
```suggestion
  override def userEventTriggered(ctx: JChannelHandlerContext, event: AnyRef): Unit = event match {
    case _: HandshakeComplete => executeAsync(ctx, ss.onOpen(ctx.channel().remoteAddress()).uninterruptible)    
    case event                =>  ctx.fireUserEventTriggered(event)    
  }
  ()
```

zio-http/src/main/scala/zhttp/socket/Socket.scala
```diff
@@ -4,7 +4,7 @@ import zio._
 import zio.stream.ZStream
 
 import java.net.{SocketAddress => JSocketAddress}
-
+// Todo add unit tests
```
tusharmath, line 7:
```suggestion
```

Added it here — https://github.com/dream11/zio-http/pull/123#issue-609527733

tusharmath (approved):

Issue #137, opened 2021-04-11
Streaming request body for server


DJLemkes:
It would be really nice if we can get the server incoming request body as a stream of bytes instead of a string. If done properly, this would allow to (among other things):

- extract all kinds of content from the request instead of only string content-types in `HttpContent.Complete`
- provide back pressure to the socket instead of loading all content into memory at once.

In doing so, I would propose merging [`HttpContent.Complete` and `HttpContent.Chunked` ](https://github.com/dream11/zio-http/blob/026f430af34dccfe009cd10d2e98f98f0915aad1/zio-http/src/main/scala/zhttp/http/HttpContent.scala#L10) into something like `HttpContent.Raw`. By adding some utilities on `HttpContent.Raw` (or on the `Request` itself) like `asText` and possible `asJSON` you have nearly the same ergonomics as with `HttpContent.Complete` while also providing access to the raw stream of bytes. More content types can then be added such as `HttpContent.MultipartFormData` in #114 .

On the Netty level, the challenge would be to provide the request body in a `ZStream[R, E, Byte]`. In a first iteration this can be done like [here](https://github.com/dream11/zio-http/blob/597f5d5baa8d9dc9d8c4720692b4bd358358c8f4/zio-http/src/main/scala/zhttp/service/HttpMessageCodec.scala#L34) but it would be nicer if we can let Netty provide backpressure to the socket instead of loading everything in memory first. This can be done by removing the [Netty `HttpObjectAggregator`](https://github.com/dream11/zio-http/blob/026f430af34dccfe009cd10d2e98f98f0915aad1/zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala#L17) and writing or own handler in which we process individual `HttpContent` objects and take control of the channel read config ourselves. Roughly that could look like

```
// for every bit of `HttpContent` that we receive in the handler do
ctx.channel().config().setAutoRead(false)
// offer content bytes to the ZStream (via a bounded ZQueue of size 1?)
ctx.channel().config().setAutoRead(true)
```

Let me know what you think. I can create a small prototype in you think it's worth trying.

tusharmath:
Thanks for opening this issue @DJLemkes 🙏 I have been thinking about it for a long time and am kinda split between multiple approaches.

[...]

PR #147, opened 2021-04-15
Server request content streams

DJLemkes:
Handler that adds streaming requests support as proposed in #137.

This handler only creates streams when strictly necessary. Besides that, it allows the user to not consume the entire request before sending out a response.

In terms of performance it is roughly equal to that of non-streaming for small <=6KB requests. For larger requests it tends to be faster than non-streaming for larger request bodies. These observations are only based on my machine and ran with the `HelloWorldAdvanced` example app.

Lastly, it also fixes a bug currently in the `main` branch where subsequent responses were causing Netty to throw exceptions because of `LAST_EMPTY_CONTENT` not being written correctly.

### DIFF
--- a/example/src/main/scala/HelloWorldAdvanced.scala
+++ b/example/src/main/scala/HelloWorldAdvanced.scala
@@ -13,15 +13,25 @@ object HelloWorldAdvanced extends App {
   }
 
   private val app = Http.collectM {
-    case Method.GET -> Root / "random" => random.nextString(10).map(Response.text)
-    case Method.GET -> Root / "utc"    =>
-      clock.currentDateTime.map(s => Response.text(s.toString))
+    case req @ Method.GET -> Root / "random" => req.discardContent &> random.nextString(10).map(Response.text)
+    case req @ Method.GET -> Root / "utc"    =>
+      req.discardContent &> clock.currentDateTime.map(s => Response.text(s.toString))
+
+    case req @ Method.POST -> Root / "upload" =>
+      req.headers.find(h => h.lowercaseEquals(Header.contentTypeJson)) match {
+        case Some(_) =>
+          req.getBodyAsString.map(echoText => echoText.map(Response.text).getOrElse(Response.status(Status.OK)))
+        case None    =>
+          req.contentSize.map(length => Response.text(s"done: ${length}"))
+      }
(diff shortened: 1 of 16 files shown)


zio-http/src/main/scala/zhttp/service/server/ServerRequestHandler.scala
```diff
@@ -67,8 +67,10 @@ final case class ServerRequestHandler[R](
               } yield ()
             }
           case HttpData.CompleteData(data) =>
-            ctx.writeAndFlush(JUnpooled.copiedBuffer(data.toArray), ctx.channel().voidPromise())
-          case HttpData.Empty              => ctx.flush()
+            ctx.write(JUnpooled.copiedBuffer(data.toArray), ctx.channel().voidPromise())
+            ctx.writeAndFlush(JLastHttpContent.EMPTY_LAST_CONTENT)
+          case HttpData.Empty              =>
+            ctx.writeAndFlush(JLastHttpContent.EMPTY_LAST_CONTENT)
```
amitksingh1490, line 73: Please move this to separate PR

amitksingh1490:
> Lastly, it also fixes a bug currently in the `main` branch where subsequent responses were causing Netty to throw exceptions because of `LAST_EMPTY_CONTENT` not being written correctly.
I am moving this to separate PR as this is a blocker to run benchmarks @DJLemkes

DJLemkes:
Let me know what you think. More then happy to make changes if required :).

tusharmath:
This is great @DJLemkes! Will have look at it.

DJLemkes:
@tusharmath sorry to tag you again :). I can do another rebase when you're ready to review. And again, happy to make changes. Would it help if I added a description of my benchmark setup (used `go-wrk` to POST request bodies as well).

tusharmath:
@DJLemkes Update: I am still doing some POCs on how to incorporate streaming capabilities to ZIO Http without compromising on performance. Thought I should post it here to keep you also updated :)

tusharmath:
Will get this out soon. We have a design that is promising and reasonably performant.

tusharmath:
Will get this out soon. We have a design that is promising and reasonably performant.

PR #156, opened 2021-04-20
Add query string parsing

### LINKED ISSUE #153: QueryParams multimap
Tapir requires to extract from queryParams: `Seq[(String, Seq[String])]`

Inspiration could be extract http4s, akka-http

Fristi:
Resolves #153

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -1,9 +1,19 @@
 package zhttp.http
 
+import io.netty.handler.codec.http.{
+  QueryStringDecoder => JQueryStringDecoder,
+  QueryStringEncoder => JQueryStringEncoder,
+}
+
 import java.net.URI
+import scala.jdk.CollectionConverters._
 import scala.util.Try
 
-case class URL(path: Path, kind: URL.Location = URL.Location.Relative, query: String = "") { self =>
+case class URL(
+  path: Path,
+  kind: URL.Location = URL.Location.Relative,
+  queryParams: Map[String, List[String]] = Map.empty,
+)          { self =>
   val host: Option[String] = kind match {
     case URL.Location.Relative      => None
     case abs: URL.Location.Absolute => Option(abs.host)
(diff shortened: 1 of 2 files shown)


amitksingh1490:
Run `sbt fmt` for formatting @Fristi

zio-http/src/test/scala/zhttp/http/URLSpec.scala
```diff
@@ -4,7 +4,29 @@ import zio.test.Assertion._
...
+      )(
+        isRight(
+          equalTo(Map("user_id" -> List("1", "2"), "order" -> List("ASC"), "text" -> List("zio-http is awesome!"))),
+        ),
+      )
+    },
+  )
+
+  def spec =
+    suite("URL")(fromString)
```
tusharmath, line 31: Any particular reason to create a variable? Can you inline it?
#nitpick

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -1,7 +1,10 @@
 package zhttp.http
 
+import io.netty.handler.codec.http.QueryStringDecoder
+
 import java.net.URI
 import scala.util.Try
+import scala.jdk.CollectionConverters._
 
 case class URL(path: Path, kind: URL.Location = URL.Location.Relative, query: String = "") { self =>
```
tusharmath, line 9:
```suggestion
case class URL(path: Path, kind: URL.Location = URL.Location.Relative, query: Map[String, List[String]] = Map.empty) { self =>
```

Fristi: Fixed in code!

zio-http/src/test/scala/zhttp/http/URLSpec.scala
Fristi: Usually I have multiple tests around one method in a class, I favor splitting them out like this

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -1,9 +1,16 @@
 package zhttp.http
 
+import io.netty.handler.codec.http.{QueryStringDecoder, QueryStringEncoder}
```
tusharmath, line 3:
```suggestion
import io.netty.handler.codec.http.{
 QueryStringDecoder => JQueryStringDecoder, 
 QueryStringEncoder => JQueryStringEncoder}
```

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -53,21 +76,20 @@ object URL
...
-      } else {
-        u
+
+    def path = {
+      val encoder = new QueryStringEncoder(url.path.asString)
+      url.queryParams.foreach { case (key, values) =>
+        values.foreach { value => encoder.addParam(key, value) }
       }
+      encoder.toString
+    }
```
tusharmath, line 86: Unit test for `asString`

tusharmath:
An alternative and some would argue a more expressive way to implement query params would be —

```scala
sealed trait Query[+K, +A] { self =>
  // for combining multiple query params
  def ++[K1 >: K, A1 >: A](other: Query[K1, A1]): Query[K1, A1] = Query.Concat(self, other)

  def asMap: Map[K, List[A]] = ???

  def asString: String = ???
}

object Query {
  case object Empty extends Query[Nothing, Nothing]
  case class Param[K, A](key: K, value: A) extends Query[K, A]
  case class Concat[K, A](qa: Query[K, A], qb: Query[K, A]) extends Query[K, A]

  fromURI(uri: URI): Query[String, String] = ???
}

```

We can do this in the same PR, or get the current changes merged as phase one, and work on query params separately?

Fristi:
What benefit would we gain from it at the moment? If there is not direct need for it, my pragmatic sense kicks in sorry ;)

tusharmath:
Ill try my best to list down a few —

**Expressiveness**
- You could customize how you encode and decode, eg: `foo=1&a=2` vs `foo[]=1,2` are both valid ways to send query strings.
- You could compose multiple params easily to create a larger query string, eg:
  ```scala
  val queryParam = ("foo", "1") && ("bar", 200) && ("baz", true)
  ```
- You could derive a query params codec for case classes 
  ```scala
  case class FooBar(foo: String, bar: Int)
  implicit val qp = Query.derive
  
  val fooBar = FooBar("AAA", 100)
  val qs = fooBar.asQueryString  // "foo=AAA&bar=10"

  Right(fooBar) == Query.fromString(qs) // FooBar("AAA", 100)
  ```
- You could perform pattern matches on query params.
  ```scala
  Http.collect {
     Get -> Root / "home" ?? ("foo", foo) && ("bar", bar) => ???
  }
  ```
**Performance**
- Using an `immutable.Map` is going to update for each query param which is going to be slow.
- Performance will also be better while doing `Query => String` because no intermediary Map will be required.

As I said before we can do this in a separate PR. The current implementation is still miles ahead of the query string support we have right now and I am excited to have it merged for the next release!

zio-http/src/main/scala/zhttp/http/URL.scala
Fristi: It's in `URLSpec`

tusharmath (approved):

tusharmath:
@amitksingh1490 can you help here? Not sure why the build isn't passing.

amitksingh1490:
> @amitksingh1490 can you help here? Not sure why the build isn't passing.

Looks like some issue in making a call to https://api.github.com/users/zio/repos from the client I have added a timeout of 20 secs in https://github.com/dream11/zio-http/pull/160/files this PR the test is getting timed out

amitksingh1490:
@Fristi can you please rebase this with main. The build issue is resolved in Main

Fristi:
Squashed and rebased

amitksingh1490 (approved):

PR #181, opened 2021-05-06
feat: Add Http.route

frekw:
This adds a simple helper to `Http` which allows you do build your overall apps out of sub-apps such as;

```scala
Http.route[Request] {
    case _ -> Root / "graphql" =>
        ZHTTPAdapter.makeHttpService(api)
    case _ -> Root / "ws" =>
        ZHTTPAdapter.makeWebSocketService(api)
}
```

in the cases where the sub application needs to specify multiple handlers (such as handling multiple HTTP methods easily).

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -164,6 +164,11 @@ object Http {
     def apply[R, E, B](f: A => ZIO[R, E, B]): Http[R, E, A, B] = Http.FromEffectFunction(f)
   }
 
+  final case class MakeRoute[A](unit: Unit) extends AnyVal {
+    def apply[R, E, B](pf: PartialFunction[A, Http[R, E, A, B]]) =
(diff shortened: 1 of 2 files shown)


tusharmath:
@frekw This could be a super useful addition to the library. Is there a better name though? 

`Http` domain doesn't really care about `requests`, `responses` and `routes` so something a bit more generic may be?

frekw:
@tusharmath fair point! I'll try to think a bit about it, but spontaneously I think delegate could be decent. What do you think about it?

tusharmath:
Actually, on second thoughts, you are "_routing messages to another Http app_". So `route` kinda makes sense :) 

So maybe instead of `routes`, we can use `route` :)

frekw:
Good point, will change it immediately!

tusharmath (approved):

frekw:
Thanks 🙏 

Not sure why the automerge is failing, looks like a missing `GITHUB_TOKEN` (which is weird since it should always be available). 🤷

tusharmath:
Thanks a ton!

PR #189, opened 2021-05-08
add provide methods to Http

nbenns:
In writing a larger Http application, in order to keep things clean, it makes sense to be able to provide local dependencies to the application.  This helps keep logic from getting entangled in the larger application.

Because `Http` wraps the ZIO objects, you loose the ability to provide the dependencies without providing it in each specific `collectM` instance.  It's a lot nicer to provide the dependencies in one place.

This PR adds the ability to provide the dependencies on the HTTP object itself, which in-turn provides it to each underlying ZIO instance where necessary, allowing providing dependencies once after composition of Http objects.

** Note: Not sure what to write for tests at this point, the types are already pretty restrictive.  Definitely up for suggestions here.

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -1,7 +1,7 @@
 package zhttp.http
 
 import zhttp.http.Http.{Chain, Combine, FoldM, FromEffectFunction}
-import zio.{CanFail, Has, NeedsEnv, Tag, ZEnv, ZIO, ZLayer}
+import zio._
 
 import scala.annotation.unused
 
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -1,6 +1,7 @@
 package zhttp.http
 
-import zio.{CanFail, ZIO}
+import zhttp.http.Http.{Chain, Combine, FoldM, FromEffectFunction}
+import zio.{CanFail, Has, NeedsEnv, Tag, ZEnv, ZIO, ZLayer}
 
 import scala.annotation.unused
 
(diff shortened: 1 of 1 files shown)


tusharmath:
This is awesome and a much-needed operator in Http!

A few things that we can do here — 
- [ ] Add a `Provide` operator constructor. And move the logic to the `evaluate` method. This might most probably mean that `HttpResult` should have this operator actually.
- [ ] Some unit tests around this would be great!

remiguittaut:
This is awesome and I need it! I'd be happy to give a shot if I find time at the Provide constructor with the logic In evaluate. Could I? I'm not sure about how to add stuff on your own PR

nbenns:
@remiguittaut I just haven't had a chance to get back to this yet, but I want to take another stab at it based on @tusharmath's asks.

Right now I'm having an issue on the Scala 3 side because the GADT of FoldM requires that not just `E1 >: E`, but also that `E1 >: EE` so that the types line up.  The Scala 2 implementation just gives up and forces you to put `Any`, but it works.
Based on the this fact, I think taking a different approach is necessary, because we just won't have evidence around that this is the case.

I'll take a look at the code again later today and see what I can come up with, @remiguittaut if you want to help out you can fork my version of the repo and PR into it, or just race me to the solution if you want on your own PR, its not gonna hurt my feelings ;)

remiguittaut:
Haha don't worry, the credit always goes to the main author. I didn't think about the scala3 problems. It's true that that's it scala3 support isn't optional anymore 😛. Little chance that I find the time, but if I do I'd be happy to contribute :)

tusharmath:
Though I think this can be a significant addition to the operator list, using `provide` in hot code paths can cause significant performance degradation.

nbenns:
That's a reason why I didn't do this in evaluate in the first place.
The provide was done in the description of the Http app instead so that it was already weaved through into the ZIO effects when it goes to execute it.

I'm not sure that solves ALL the performance impact it might have, but since its now delegated to ZIO already it shouldn't be any different the doing many individual provides in each route instead of after combining them.

I feel this is still a better strategy, however the GADT of FoldM is still giving me problems and I haven't had the time to sit down and resolve it yet.

frekw:
I tried going the route of adding them to `HttpResult` here: #261.

tusharmath:
fixed in #364

PR #196, opened 2021-05-10
Fix: CORS middleware to handle OPTIONS requests completely on its own

SvenW:
If the app that uses the CORS middleware doesn't handle `OPTIONS` requests it will simply return a 404 and the preflight request will fail


amitksingh1490:
Please rebase this PR

SvenW:
> Please rebase this PR

Thanks. I rebased it just now

tusharmath (approved):

PR #197, opened 2021-05-10
feat: http error handler

### LINKED ISSUE #148: Support for Http error handler
Currently, if there is an exception on the channel, there is no way for the HTTP app to handle it.
Typically someone might want to log it and probably close the connection after an exception is thrown.

**Example**
```scala
Server.port(PORT) ++       
  Server.app(app) ++
  Server.error(cause => log.throwable(cause).as(true))
```

dinojohn:
fixes #148


zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -65,7 +69,7 @@ object Server {
       zExec          <- UnsafeChannelExecutor.make[R].toManaged_
       channelFactory <- ZManaged.access[ServerChannelFactory](_.get)
       eventLoopGroup <- ZManaged.access[EventLoopGroup](_.get)
-      httpH           = ServerRequestHandler(zExec, settings.http)
+      httpH           = ServerRequestHandler(zExec, settings.http, settings.errorHandler)
```
tusharmath, line 72:
```suggestion
      httpH           = ServerRequestHandler(zExec, settings)
```

zio-http/src/main/scala/zhttp/service/server/ServerRequestHandler.scala
```diff
@@ -91,7 +92,10 @@ final case class ServerRequestHandler[R](
    */
   override def exceptionCaught(ctx: JChannelHandlerContext, cause: Throwable): Unit = {
     if (self.canThrowException(cause)) {
```
tusharmath, line 94: I think we can deprecate `canThrowException`.

zio-http/src/main/scala/zhttp/service/Server.scala
dinojohn: settings is a private case class for server object

tusharmath: we can make it `private [zhttp]`

tusharmath (approved):

PR #261, opened 2021-06-14
feat: Add provide* combinators to Http and HttpResult

frekw:
This follows up on the work done in #189 and instead adds the `provide*` combinators to the `HttpResult`.

I also added all of the combinators to `Http` itself, just acting as wrappers on top.

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -303,8 +303,7 @@ object Http {
   private final case class ProvideLayer[E, E1 >: E, R, R0, R1 <: R, A, B](
     self: Http[R, E, A, B],
     layer: ZLayer[R0, E1, R1],
-  )(implicit ev1: R1 <:< R, ev2: NeedsEnv[R])
-      extends Http[R0, E1, A, B]
+  ) extends Http[R0, E1, A, B]
 
   // Ctor Help
   final case class MakeCollectM[A](unit: Unit) extends AnyVal {
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -290,21 +290,18 @@ sealed trait Http[-R, +E, -A, +B] { self =>
    */
   final private[zhttp] def execute(a: A): HttpResult[R, E, B] = {
     self match {
-      case Empty                        => HttpResult.empty
-      case Identity                     => HttpResult.succeed(a.asInstanceOf[B])
-      case Succeed(b)                   => HttpResult.succeed(b)
-      case Fail(e)                      => HttpResult.fail(e)
-      case FromEffectFunction(f)        => HttpResult.effect(f(a))
-      case Collect(pf)                  => if (pf.isDefinedAt(a)) HttpResult.succeed(pf(a)) else HttpResult.empty
-      case Chain(self, other)           => HttpResult.suspend(self.execute(a) >>= (other.execute(_)))
-      case FoldM(self, ee, bb, dd)      =>
+      case Empty                   => HttpResult.empty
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -204,6 +266,50 @@ object Http {
     dd: Http[R, EE, A, BB],
   )                                                                             extends Http[R, EE, A, BB]
 
+  private final case class Provide[R, E, A, B](self: Http[R, E, A, B], r: R) extends Http[Any, E, A, B] {
+    override private[zhttp] def execute(a: A): HttpResult[Any, E, B] =
```
tusharmath: `execute` was supposed to be final (miss from my side).
You will loose out on stacksafety if you override it.

frekw: Gotcha.

The only reason why I use it is to get around an (possibly incorrect) exhaustivity warning on Scala3, see d45475f. Would be happy to just inline all impls if someone can point me to another way around this!

frekw: 1a119f3 is the only way around this I've found that also makes the compiler happy. Can't say it's super pretty (and not sure if extracting `doProvide` keeps it stack-safe?). If you have any other ideas of how to address this I'd be very grateful 🙏

frekw: Oh. And that doesn't seem to work for 2.13. :/

frekw: Changing all of them like so:

```scala
+      case p @ ProvideLayer(_, _)     => p.doProvide
+      // case p: ProvideLayer[_, _, _, _, _, _] => p.doProvide
```

works for both versions but emits warnings in Scala 3. 🤷

zio-http/src/main/scala/zhttp/http/HttpResult.scala
```diff
@@ -94,11 +201,13 @@ object HttpResult {
...
-  private[zhttp] def evaluate[R, E, A](result: HttpResult[R, E, A]): Out[R, E, A] = {
+  private[zhttp] def evaluate[R, E, A](
+    result: HttpResult[R, E, A],
+  )(implicit ev: NeedsEnv[R]): Out[R, E, A] = {
     result match {
-      case m: Out[_, _, _]         => m
-      case Suspend(r)              => evaluate(r())
-      case FoldM(self, ee, aa, dd) =>
+      case m: Out[_, _, _]                       => m
+      case Suspend(r)                            => evaluate(r())
```
tusharmath: Similar to `Suspend` you can add a case for `Provide(a, r)`, it should look something like —

```scala
  case Provide(a, r) => 
    a match {
      case a: Out[_, _, _] => a // For `Out` type `provide` is not required.
      case Effect(z)       => evaluate(z.provide(r)) // For `ZIO` use the built-in `provide` operator
      case FoldM(self, ee, aa, dd) => self.foldM(e => ee(e).provide(r), a => (aa).provide(r), dd.provide(r)) // For `foldM` handle env for all three channels — Failure + Success + Default.
    }
```

frekw:
@tusharmath I updated this to be based on #265. I've removed `provideSomeLayer` and `provideCustomLayer` because they were giving me too much of a headache due to missing evidence for `R <: Has[_]` in `evaluate`. I also think these combinators are quite valuable in themselves.

However, as with the code in #265 we're still seeing exhaustivity warnings in Scala3 (that I believe are false) but am unsure how to work around them. Will ask around a bit on the Scala gitter.

frekw:
Could potentially be https://github.com/lampepfl/dotty/issues/9682 that we're running into.

tusharmath:
fixed in #364

PR #263, opened 2021-06-16
Added wrap to Http so that we can control the execution of a Http

SvenW:
We need this in order to get tracing with [zio-telemetry](https://github.com/zio/zio-telemetry) working. Right now the execution of the Http happens too late which makes it impossible to connect our root span with its children. With this we can wrap the Http and make sure that the spans get connected.
Here is an example where we try to wrap our HttpApp in a root span which then leads to the disconnected spans since the execution doesn't happen right here but in a later stage.
```
  def tracedHttp[R <: Tracing, E](app: HttpApp[R, E]): HttpApp[R, E] =
    http.Http.flatten {
      http.Http.fromEffect(
        Tracing.root("rootSpan")(ZIO.succeed(app))
      )
    }
```

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -169,6 +169,14 @@ sealed trait Http[-R, +E, -A, +B] { self =>
     self.flatMap(Http.fromEffect(_))
 
   /**
+   * Wraps an `Http` app in a surrounding effect that controls the execution of the underlying `Http`.
+   */
+  def wrap[R1 <: R, A1 <: A, E1 >: E, C](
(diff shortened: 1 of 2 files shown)


guersam:
I doubt it would work as expected, because zio-telemetry keeps the context in `FiberRef` rather than `Ref`, and zio-http relies on unsafe runtime execution for integration with Netty. Each unsafe execution introduces an empty FiberContext so that the tracing context might not persist during an HTTP session. 

Maybe `HttpRef` #306 can enable integration with fiber-aware libraries with zio-http or there might be more powerful escape hatch on ZIO side in the future, but I have no concrete idea yet.

tusharmath:
I am still not sure why we need this. Feel free to re-open, if you think this is really important for your usecase.

PR #264, opened 2021-06-16
feat: Add optional remoteAddress to Request

jakubjanecek:
This is only applicable in server context where this information can be very useful for example to provide rate-limiting per IP address or logging/debugging purposes.

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Request.scala
+++ b/zio-http/src/main/scala/zhttp/http/Request.scala
@@ -1,10 +1,15 @@
 package zhttp.http
 
+import zhttp.core.JChannelHandlerContext
+
+import java.net.{InetAddress, InetSocketAddress}
+
 // REQUEST
 final case class Request(
   endpoint: Endpoint,
   headers: List[Header] = List.empty,
   content: HttpData[Any, Nothing] = HttpData.empty,
+  private val channelContext: JChannelHandlerContext = null,
 ) extends HasHeaders
     with HeadersHelpers { self =>
   val method: Method = endpoint._1
@@ -14,6 +19,13 @@ final case class Request(
   def getBodyAsString: Option[String] = content match {
     case HttpData.CompleteData(data) => Option(data.map(_.toChar).mkString)
     case _                           => Option.empty
+  }
+
+  def remoteAddress: Option[InetAddress] = {
+    if (channelContext != null && channelContext.channel().remoteAddress().isInstanceOf[InetSocketAddress])
+      Some(channelContext.channel().remoteAddress().asInstanceOf[InetSocketAddress].getAddress)
+    else
+      None
   }
 
 }
--- a/zio-http/src/main/scala/zhttp/service/DecodeJRequest.scala
+++ b/zio-http/src/main/scala/zhttp/service/DecodeJRequest.scala
@@ -1,6 +1,6 @@
 package zhttp.service
 
-import zhttp.core.JFullHttpRequest
+import zhttp.core.{JChannelHandlerContext, JFullHttpRequest}
 import zhttp.http._
 
 trait DecodeJRequest {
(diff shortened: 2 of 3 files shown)


jakubjanecek:
Making this PR made me realize that I am not sure whether `Request` should be identical both for client and server code. I know that in theory it should but in practice there are certain things that differ, for example in this case because of remote address. What do you think about this?

http4s has a field called `attributes` in the request which allows it to package in more data including the remote address. Should that be the way to go for `zio-http`? Unfortunately we'd need some kind of typed map for this and I am not sure it is available.

zio-http/src/main/scala/zhttp/http/Request.scala
```diff
@@ -1,10 +1,13 @@
 package zhttp.http
 
+import java.net.InetAddress
+
 // REQUEST
 final case class Request(
   endpoint: Endpoint,
   headers: List[Header] = List.empty,
   content: HttpData[Any, Nothing] = HttpData.empty,
+  remote: Option[InetAddress] = None,
```
tusharmath: Im not really sure about this API. 
There is going to be a performance penalty even in-case someone isn't reading the IP address.

jakubjanecek: What performance penalty? You mean unnecessary allocation of `Option` instance?

Do you have other solution in mind? The API doesn't have to be like this, I just think that having access to remote address is crucial for certain use cases.

tusharmath: One option is to save `context` inside the request and have a method `localAddress()` that returns the address from the context.

jakubjanecek: Not sure whether it's much better but I've refactored it that way. Can you please look at it again? Thanks.

jakubjanecek:
@tusharmath Could you please have another look at this PR, please? Thank you.

zio-http/src/main/scala/zhttp/http/Request.scala
```diff
@@ -1,10 +1,15 @@
...
+import zhttp.core.JChannelHandlerContext
+
+import java.net.{InetAddress, InetSocketAddress}
+
 // REQUEST
 final case class Request(
   endpoint: Endpoint,
   headers: List[Header] = List.empty,
   content: HttpData[Any, Nothing] = HttpData.empty,
+  private val channelContext: Option[JChannelHandlerContext] = None,
```
tusharmath:
```suggestion
  private val channelContext: JChannelHandlerContext = null,
```

zio-http/src/main/scala/zhttp/http/Request.scala
```diff
@@ -16,6 +21,13 @@ final case class Request(
     case _                           => Option.empty
   }
 
+  def remoteAddress: Option[InetAddress] =
+    channelContext.map(_.channel().remoteAddress()).collect {
```
tusharmath: Can check for null here.

zio-http/src/main/scala/zhttp/service/DecodeJRequest.scala
```diff
@@ -1,18 +1,19 @@
...
    */
-  def decodeJRequest(jReq: JFullHttpRequest): Either[HttpError, Request] = for {
+  def decodeJRequest(jReq: JFullHttpRequest, ctx: JChannelHandlerContext): Either[HttpError, Request] = for {
     url <- URL.fromString(jReq.uri())
     method   = Method.fromJHttpMethod(jReq.method())
     headers  = Header.make(jReq.headers())
     endpoint = method -> url
     data     = HttpData.fromByteBuf(jReq.content())
-  } yield Request(endpoint, headers, data)
+  } yield Request(endpoint, headers, data, Option(ctx))
```
tusharmath:
```suggestion
  } yield Request(endpoint, headers, data, ctx)
```

zio-http/src/main/scala/zhttp/http/Request.scala
jakubjanecek: Alright, refactored.

tusharmath (approved):

tusharmath:
Thanks @jakubjanecek 🙏

PR #265, opened 2021-06-17
Feature: Provide Combinator in HttpResult

tusharmath:
Picking up changes from #261 #189


zio-http/src/main/scala/zhttp/http/HttpResult.scala
```diff
@@ -98,6 +101,14 @@ object HttpResult {
     result match {
       case m: Out[_, _, _]         => m
       case Suspend(r)              => evaluate(r())
+      case Provide(r, env)         =>
+        evaluate(r match {
+          case Effect(z)             => Effect(ZIO.provide(env)(z))
+          case out: Out[_, _, _]     => out
+          case Suspend(r)            => r().provide(env)
+          case FoldM(rr, ee, aa, dd) => rr.foldM(e => ee(e).provide(env), a => aa(a).provide(env), dd.provide(env))
```
frekw, line 122: I think this needs to be

```scala
rr.foldM(e => ee(e), a => aa(a), dd).evaluate.provide(env)
```

otherwise the output seems to be dependent on the execution order and the following happens:

```scala
sealed trait HttpResult[-R, +E, +A] { self =>
  def provideSome[R0](f: R0 => R): HttpResult[R0, E, A] =
    HttpResult.effect(ZIO.access[R0](identity)).flatMap(e => { self.provide(f(e)) })
}

object HttpResultSpec extends DefaultRunnableSpec with HttpResultAssertion {
  def spec: ZSpec[Environment, Failure] = {
    suite("HttpResult")(
      testM("foldM 1") {
          trait HasInt {
            val int: Int
          }

          val needsEnv = effect(for {
            int <- ZIO.environment[HasInt]
          } yield int.int)

          val app = (succeed(1) *> needsEnv).provideSome[Any](_ => {
            new HasInt { val int = 1 }
          })

          for {
            res <- app.evaluate.asEffect
          } yield assert(res)(equalTo(1))
        },
        testM("foldM 2") {
          trait HasInt {
            val int: Int
          }

          val needsEnv = effect(for {
            int <- ZIO.environment[HasInt]
          } yield int.int)

          val app = (needsEnv *> succeed(1)).provideSome[Any](_ => {
            new HasInt { val int = 1 }
          })

          for {
            res <- app.evaluate.asEffect
          } yield assert(res)(equalTo(1))
        },
    )
  }
}
```

will work for `foldM 1` but fail for `foldM 2`.

tusharmath:
fixed in #364

PR #266, opened 2021-06-18
feat: Use implicit class for HttpApp syntax

frekw:
This is a WIP of using the same [implicit trick used by Has](https://github.com/zio/zio/blob/a53fb07d9bb78629e9564b8da92a824b9e2f6d09/core/shared/src/main/scala/zio/Has.scala#L160) to provide the specialized `HttpApp` syntax. I believe this is the best of both worlds since it can provide specialized combinators while being zero allocation. It also allows us to keep all existing combinators from `Http` without having to reimplement them.

Let me know if you think it's worthwhile and I'll continue the work!

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -294,4 +294,18 @@ object Http {
   final class MkTotal[A](val unit: Unit) extends AnyVal {
     def apply[B](f: A => B): Http[Any, Nothing, A, B] = Http.identity[A].map(f)
   }
+
+  implicit final class HttpAppSyntax[R, E](val self: HttpApp[R, E]) extends AnyVal {
+
+    /**
+     * Converts a failing Http into a non-failing one by handling the failure and converting it to a result if possible.
+     */
+    def silent[R1 <: R, E1 >: E](implicit s: CanBeSilenced[E1, Response[R1, E1]]) =
+      self.catchAll(e => Http.succeed(s.silent(e)))
+
+    /**
+     * Converts a ZIO to an Http type
+     */
+    def responseM[R1, E1](res: ResponseM[R1, E1]): Http[R1, E1, Request, Response[R1, E1]] = Http.fromEffect(res)
+  }
 }
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -302,10 +302,5 @@ object Http {
      */
     def silent[R1 <: R, E1 >: E](implicit s: CanBeSilenced[E1, Response[R1, E1]]) =
       self.catchAll(e => Http.succeed(s.silent(e)))
-
-    /**
-     * Converts a ZIO to an Http type
-     */
-    def responseM[R1, E1](res: ResponseM[R1, E1]): Http[R1, E1, Request, Response[R1, E1]] = Http.fromEffect(res)
   }
 }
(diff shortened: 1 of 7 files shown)


frekw:
Ok, wasn't that much remaining, finished it up!

frekw:
Reverts/changes most changes in https://github.com/dream11/zio-http/issues/206

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -294,4 +294,13 @@ object Http {
   final class MkTotal[A](val unit: Unit) extends AnyVal {
     def apply[B](f: A => B): Http[Any, Nothing, A, B] = Http.identity[A].map(f)
   }
+
+  implicit final class HttpAppSyntax[R, E](val self: HttpApp[R, E]) extends AnyVal {
```
tusharmath: can be moved to a helper trait.

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -2,76 +2,56 @@ package zhttp.http
...
-   * Combines two HttpApps into one.
-   */
-  def +++[R1 <: R, E1 >: E](other: HttpApp[R1, E1]) = HttpApp(asHttp +++ other.asHttp)
-
-  /**
-   * Evaluates the app and returns an HttpResult that can be resolved further
-   */
-  private[zhttp] final def execute(r: Request) = asHttp.execute(r)
-}
-
```
tusharmath, line 24: We can add an `apply` function to support backward compatibility?

zio-http/src/main/scala/zhttp/http/Http.scala
frekw: Fixed in e7f69c5.

zio-http/src/main/scala/zhttp/http/HttpApp.scala
frekw: Fixed in 5c90fab.

I also added a deprecated `asHttp` to the syntax to make it backwards compatible as well.

tusharmath (approved):

zio-http/src/main/scala/zhttp/http/Http.scala
tusharmath: I was thinking more like `RequestSyntax` eg: 


**HttpAppSyntax.scala**
```scala
trait HttpAppSyntax {
  implicit final class HttpAppSyntax[R, E] ....
}
```

frekw: I moved it to its own file, but I think the class definition needs to live outside the trait to make `extends AnyVal` work. 👍

tusharmath (approved):

tusharmath:
thank you @frekw !!!

tusharmath:
I thought about this PR randomly and realized that the idea behind not using extension methods was that an `HttpApp[R, E]` should not have operators such as `map` or `flatMap` etc.

frekw:
Hm, why not? Because it's possible to go from an HttpApp to something that's not? Personally I think that tradeoff is worth it (and expected)! 🙂

frekw:
Because I'd much rather have that than needing to remember what I can do with an HttpApp or not and constantly having to convert back and forth between the two.

tusharmath:
Because `HttpApp` is not a monad. If you look at the type signature `HttpApp[R, E]` it only contains an environment and an error type.

frekw:
I think the way I look at it if that Http is always the monad. And the HttpApp is just the special case when the Http accepts a Request and returns a Response. But it's always an Http that we're working with. In that case I don't think it matters much (i.e within the Http monad you can move from/to an HttpApp) but if you want it to be something more concrete then it does. 🤷 

I think my take as a user is that Http and HttpApp are too similar that having them as separate concepts become confusing, but that's of course up to you to decide!

PR #268, opened 2021-06-22
Fix: Query parameters being ignored in client requests

### LINKED ISSUE #267: Add query parameters to Client request
When using `Client.request` we are currently not including the query parameters in the request

Adriani-Furtado:
closes #267

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -24,6 +24,8 @@ final case class URL(
     case abs: URL.Location.Absolute => Option(abs.port)
   }
 
+  lazy val relative: URL = self.copy(kind = URL.Location.Relative)
+
   def asString: String = URL.asString(self)
 }
 object URL {
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -24,7 +24,10 @@ final case class URL(
     case abs: URL.Location.Absolute => Option(abs.port)
   }
 
-  private[zhttp] def relative: URL = self.copy(kind = URL.Location.Relative)
+  private[zhttp] def relative: URL = self.kind match {
+    case URL.Location.Relative => self
+    case _                     => self.copy(kind = URL.Location.Relative)
+  }
 
   def asString: String = URL.asString(self)
 }
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -24,7 +24,7 @@ final case class URL(
     case abs: URL.Location.Absolute => Option(abs.port)
   }
 
-  lazy val relative: URL = self.copy(kind = URL.Location.Relative)
+  private[zhttp] def relative: URL = self.copy(kind = URL.Location.Relative)
 
   def asString: String = URL.asString(self)
 }
--- a/zio-http/src/main/scala/zhttp/service/EncodeRequest.scala
+++ b/zio-http/src/main/scala/zhttp/service/EncodeRequest.scala
@@ -13,7 +13,7 @@ trait EncodeRequest {
     val method      = req.method.asJHttpMethod
     val uri         = req.url.path match {
       case Root => "/"
-      case path => path.asString
+      case _    => req.url.relative.asString
     }
     val content     = req.getBodyAsString match {
       case Some(text) => JUnpooled.copiedBuffer(text, HTTP_CHARSET)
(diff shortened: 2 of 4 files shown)


zio-http/src/main/scala/zhttp/service/EncodeRequest.scala
```diff
@@ -13,7 +14,7 @@ trait EncodeRequest {
     val method      = req.method.asJHttpMethod
     val uri         = req.url.path match {
       case Root => "/"
-      case path => path.asString
+      case _    => URL.asUriString(req.url)
     }
```
tusharmath, line 17: I think we can just do a `req.url.asString`

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -78,21 +78,23 @@ object URL {
...
+    }
+  }
+
+  private def path(url: URL) = {
+    val encoder = new JQueryStringEncoder(url.path.asString)
+    url.queryParams.foreach { case (key, values) =>
+      values.foreach { value => encoder.addParam(key, value) }
     }
+    encoder.toString
   }
```
tusharmath: Are these changes needed?

zio-http/src/main/scala/zhttp/service/EncodeRequest.scala
Adriani-Furtado: If we use `req.url.asString` we will end up with an `url` instead of `uri`.
`http://localhost:8000/0/0?0=0` - url
`/0/0?0=0` - uri

 I dont know if there would be any implications on using `url` instead.
I will try to have a look and find out if we can just use the `url` instead

zio-http/src/main/scala/zhttp/http/URL.scala
Adriani-Furtado: That would depend on the comment above. If we need to use `uri` instead of `url` then I need a way of extracting a `uri` string. If we are okay to just use `url` then this wont be needed

zio-http/src/main/scala/zhttp/service/EncodeRequest.scala
Adriani-Furtado: @tusharmath I had a look at https://github.com/dream11/zio-http/issues/237 and it seems like using `req.url.asString` is not a viable solution, let me know what you think

tusharmath: I think we can add a local method `makeUrl` or something better, inside of `EncodeRequest` to handle this specialized `string`.

Keeping `asUriString` inside of `URL` is a bit confusing considering we already have `asString` method on URL.

Adriani-Furtado: I thought about doing this in the first place, but that would make us end up with this bit of logic copy and pasted in 2 different locations
```scala
    val encoder = new JQueryStringEncoder(url.path.asString)
    url.queryParams.foreach { case (key, values) =>
      values.foreach { value => encoder.addParam(key, value) }
    }
    encoder.toString
```
The best alternative to copy/paste would be to rename the `URL#path` to possibly `URL#uri`, make it public and return an `URI`.
This way we can call this method to from `EncodeRequest` to get the `Uri` and then call `.toString` and get the desired result.
I will push this change in a dedicated commit so you can have a look and let me know if its a good solution

tusharmath: We could also add a method on URL called `queryString` which would contain the logic for building a string using the query params.
Then inside the encode use `path.asString + url.queryString` to produce the final url string.

tusharmath: Alternatively, we can add a method `relative` which always returns a relative url and then call `asString` on it?

Adriani-Furtado: > We could also add a method on URL called `queryString` which would contain the logic for building a string using the query params.
> Then inside the encode use `path.asString + url.queryString` to produce the final url string.

I think the issue with this is that we are using the `JQueryStringEncoder` to deal with appending the query params, which requires a `uri` string to be created.
We could just manually write the conversion from the queryParams `Map[String, List[String]]` to a `String` but now we would be left with 2 ways of encoding the query_params

Adriani-Furtado: @tusharmath Have gone with the `relative` method as it is the simplest and cleanest solution in my opinion

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -24,6 +24,8 @@ final case class URL(
     case abs: URL.Location.Absolute => Option(abs.port)
   }
 
+  lazy val relative: URL = self.copy(kind = URL.Location.Relative)
```
tusharmath: Can make this package-private. It's a bit too low-level API. What do you think?
Lazy vals are also a bit slower for first-time access. I would prefer a `def` and save it in a value in case of hot paths.

tusharmath (approved):

PR #297, opened 2021-07-13
constructor for FromStream

### LINKED ISSUE #291: Add public from stream constructor for Socket
Case class` FromStream` has no accessible constructor.
Could you please add a way to instantiate it, in a similar way to `FromStreamingFunction` which has method `fromFunction`.

dinojohn:
fixes #291


amitksingh1490 (approved):

adamw:
Would it be possible to have a release with that fix? We'd like to try adding websockets using zio-http to tapir and this was blocking us (maybe more blockers are coming though, who knows ;) )

PR #308, opened 2021-07-24
Add basic request testing util

whiskeylips:
The goal for this MR is to add a basic testing utility & package to zio-http. 

The problem:

`Http.execute` is reasonably package-private to zhttp. However, when writing integration tests, it is inconvenient and malperformant to bind a port for every new test. For this reason, this "zio-http-test" package exposes a TestHttp runner that is able to take an `A` and return an `HttpResult[R, E, B]`. This will allow us to write tests in the following pattern:

```scala
    val app = Http.collect[Request] {
      case Method.GET -> Root / "text" => Response.text("Hello World!")
    }


    TestHttp.execute(app, Request(Method.GET -> url, headers, content)).evaluate.asEffect.map(testFn)
```

I believe this will allow us to reach >90% of test cases, without unsealing `Http[-R, +E, -A, +B]`. Open to discussion or architectural input :).


whiskeylips:
Force pushed to remove some crud from build.sbt.

tusharmath:
This could be a great value addition to the library. Here are some thoughts on how one might want to use it.

```scala
import zio.test._
import zhttp.test._

object Spec extends DefaultRunnableSpec {
  def spec = suite("http") (
    testM("should be ok") {
       val app         = Http.ok
       val req         = ???
       val expectedRes = ???
       assertM(app(req))(expectedRes) // an apply method is added via `zhttp.test` package
    }
  )
}
```

I am not sure if `HttpResult` should be exposed publicly. It is only there as a stop-gap solution until ZIO 2.0. I think it's alright to expose the output wrapped in a ZIO.

whiskeylips:
> This could be a great value addition to the library. Here are some thoughts on how one might want to use it.
> 
> ```scala
> import zio.test._
> import zhttp.test._
> 
> object Spec extends DefaultRunnableSpec {
>   def spec = suite("http") (
>     testM("should be ok") {
>        val app         = Http.ok
>        val req         = ???
>        val expectedRes = ???
>        assertM(app(req))(expectedRes) // an apply method is added via `zhttp.test` package
>     }
>   )
> }
> ```
> 
> I am not sure if `HttpResult` should be exposed publicly. It is only there as a stop-gap solution until ZIO 2.0. I think it's alright to expose the output wrapped in a ZIO.

Neat. I implemented the apply method via a class extension. Wrapped the result in a zio

whiskeylips:
~~As long as the 2.0 version maintains the same evaluate method, or we can extend it to add an evaluate method, we should be fine.~~

found a better way to privatize the internals of HttpResult.

zio-http-test/src/main/scala/zhttp/test/test.scala
```diff
@@ -0,0 +1,11 @@
+package zhttp
+
+import zhttp.http.{Http, HttpResult}
+import zio.ZIO
+
+package object test {
+  implicit class HttpWithTest[R, E, A, B](http: Http[R, E, A, B]) {
+    def apply(req: A): zio.UIO[HttpResult[R, E, B]] = ZIO.succeed(http.execute(req))
```
tusharmath:
```suggestion
    def apply(req: A): zio.ZIO[R, E, B] = http.execute(req).evaluate.asEffect
```

Something like this would be better.

whiskeylips: I think we were working on this at the same time, @tusharmath. https://github.com/dream11/zio-http/pull/308/commits/999f5decabb535c2fc1c0d57d4dff03230655ed9 made those exact changes :).

whiskeylips:
The latest changes were added so that my local tests would work. I've been able to confirm that the following is working as intended:

```
sbt publishLocal # results in an artifact produced in the local ivy cache at io.d11/zhttp-test/<SNAPSHOT>
```

This works as intended when depending on the library and `import zhttp.test._`

whiskeylips:
Hey @tusharmath, I think this is ready. I can contribute additional documentation needed.

tusharmath:
> Hey @tusharmath, I think this is ready. I can contribute additional documentation needed.

That would be awesome! You could add some documentation about how to test in the `Readme.md` file.

build.sbt
```diff
@@ -111,6 +108,12 @@ lazy val zhttpBenchmarks = (project in file("./zio-http-benchmarks"))
       ),
   )
 
+// Testing Package
+lazy val zhttpTest = (project in file("./zio-http-test"))
+  .dependsOn(zhttp)
+  .settings(stdSettings("zhttp-test"))
```
tusharmath, line 114:
```suggestion
  .settings(stdSettings("zhttpTest"))
```

To be consistent with `zhttpBenchmarks`

tusharmath:
Can you try taking an update from `main` I am not sure why the CI is failing.

tusharmath:
Got it, just run the following command and push.

```scala
githubWorkflowGenerate
```

whiskeylips:
Hey @tusharmath ,

I added documentation for the testing package. However, I can't change the name.. the project name determines the package that is published to sonatype, and I think it's better for us to keep in line with established scala standards and name the package `zhttp-test`. LMK what you think,.

tusharmath (approved):

tusharmath:
Thank you once again!

PR #337, opened 2021-08-22
Make path extraction right associative

tusharmath:
## Current Behavior
With the current API only exact match is supported and is **impossible** to match a prefix and extract the rest of the path. For eg: 

```scala
Http.collect[Path] {
  case Root / "a" / rest => rest
}
```
In the above example, the path `/a/b/c/d` will not match the route, resulting in a 404 error.

## New Behavior
The new API is now right-associative and can match paths partially, for eg: 

```scala
Http.collect[Path] {
  case "a" /: rest => rest
}
```
With this new API, instead of using `/` we use `/:`. This allows us to start the extraction from the right-hand side of the expression. 
The value of `rest` will be equal to `/b/c/d` in this new API!

## Support for Exact Match

To support an exact match use case, one needs to mark the end of the path explicitly. This is done by a new `!!` operator eg —

```scala
Http.collect[Path] {
  case "a" /: rest /: !! => rest
}
```
This route would simulate the old behavior of exact match but with a different syntax.

## Bonus!
Special route extractors have been added to pull `int` `short` `long` `date` etc.
```scala
Http.collect[Path] {
  case "user" /: int(id) /: "create" /: !! => console.putStrLn("User created!")
  case "weather" /: date(dd) /: !!         => console.putStrLn(s"Get weather for ${dd}")
}
```

### DIFF
--- a/example/src/main/scala/Authentication.scala
+++ b/example/src/main/scala/Authentication.scala
@@ -37,13 +37,13 @@ object Authentication extends App {
 
   // Http app that requires a JWT claim
   def user(claim: JwtClaim): UHttpApp = Http.collect {
-    case Method.GET -> "user" /: name /: "greet" /: _ => Response.text(s"Welcome to the ZIO party! ${name}")
-    case Method.GET -> "user" /: "expiration" /: _ => Response.text(s"Expires in: ${claim.expiration.getOrElse(-1L)}")
+    case Method.GET -> !! / "user" / name / "greet" => Response.text(s"Welcome to the ZIO party! ${name}")
+    case Method.GET -> !! / "user" / "expiration"   => Response.text(s"Expires in: ${claim.expiration.getOrElse(-1L)}")
   }
 
   // App that let's the user login
   // Login is successful only if the password is the reverse of the username
-  def login: UHttpApp = Http.collect { case Method.GET -> "login" /: username /: password /: _ =>
+  def login: UHttpApp = Http.collect { case Method.GET -> !! / "login" / username / password =>
     if (password.reverse == username) Response.text(jwtEncode(username))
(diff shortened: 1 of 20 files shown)


hilcode:
Instead of breaking all existing code and documentation, I would prefer we introduce `/:` as an _additional_ operator. So

```
Http.collect[Path] {
  case "a" / "b" / "c" / "d" => ...
}
```

would continue to work as expected but

```
Http.collect[Path] {
  case "a" / "b" / "c" /: rest => ...
}
```

would capture everything after `a/b/c` in `rest`.

So `/:` could only be used as the _last_ operator, which I believe covers your use case? No breaking change and no need for `!!`?

[I have not tried any of this so I have no idea about potential problems with precedence or otherwise.]

tusharmath:
@hilcode we can either go left to right OR right to left. We can't do both so using them together doesn't like a possiblity.

## Update 1

Based on the feedback from various folks on the discord channel, I have added the `/` operator back! But with some **caveats**.

1. The performance is going to be a bit less when compared to the `/:` operator.
2. `Root` is still **deprecated** in place of `!!` operator. This can be used to mark both the start and the end of a `Path` eg:

    ```scala
    // Old API
    val path = Root / "a" / "b" / "c"

    // Left Associative (new)
    val path = !! / "a" / "b" / "c"

    // Right Associative (new)
    val path = "a" /: "b" /: "c" /: !!
    ```

    Similar syntax can also be used while **pattern matching**
    ```scala
    // Old API
    val path = Http.collect[Path] { case Root / "a" / "b" / "c" => true }

    // Left Associative (new)
    val path = Http.collect[Path] {case !! / "a" / "b" / "c" => true }

    // Right Associative (new)
    val path = Http.collect[Path] { case "a" /: "b" /: "c" /: !! => true }
    ```

amitksingh1490 (approved):

PR #374, opened 2021-09-30
Move compile in HttpApp to separate module

amitksingh1490:


zio-http/src/main/scala/zhttp/service/HttpCompile.scala
```diff
@@ -0,0 +1,245 @@
...
+import io.netty.handler.codec.http._
+import zhttp.experiment.ContentDecoder
+import zhttp.http.HttpApp.InvalidMessage
+import zhttp.http._
+import zio.stream.ZStream
+import zio.{Chunk, Promise, UIO, ZIO}
+
+import java.net.{InetAddress, InetSocketAddress}
+
+final case class HttpCompile[R, E] private[zhttp] (app: HttpApp[R, E], zExec: HttpRuntime[R])
```
ShrutiVerma97: HttpCompile should be renamed as Handler

zio-http/src/main/scala/zhttp/service/HttpCompile.scala
```diff
@@ -0,0 +1,245 @@
...
+
+object HttpCompile {
+  def apply[R, E](app: HttpApp[R, E], zExec: HttpRuntime[R])(implicit
```
ShrutiVerma97: remove apply method, case class should extend ChannelInboundHandlerAdapter directly

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -23,19 +17,6 @@ case class HttpApp[-R, +E](asHttp: Http[R, E, Request, Response[R, E]]) { self =
 
   def +++[R1 <: R, E1 >: E](other: HttpApp[R1, E1]): HttpApp[R1, E1] = self defaultWith other
 
-  /**
-   * Converts a failing Http app into a non-failing one by handling the failure and converting it to a result if
```
amitksingh1490, line 27: why is this deleted?

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -384,10 +140,4 @@ object HttpApp {
   def fromFunction[R, E, B](f: Request => HttpApp[R, E]): HttpApp[R, E] =
     HttpApp(Http.fromFunction[Request](f(_).asHttp).flatten)
 
-  /**
```
amitksingh1490, line 387: Also this? why was it deleted

PR #382, opened 2021-10-06
Add IOUring support

madper:
Howdy Gurus,

I'd using IOUring when possible as it provides better latency. The pr is still WIP. 
However I have a few questions before moving on:

1. Netty-IOUring is still in incubator. It seems we shouldn't use it as default event loop?
2. Should we provide a interface to allow enduser choosing the IOUring manually?

BR


zio-http/src/main/scala/zhttp/service/EventLoopGroup.scala
```diff
@@ -38,8 +39,16 @@ object EventLoopGroup {
...
+      make(UIO(new IOUringEventLoopGroup(nThread)))
+
+    def uring(nThread: Int, executor: Executor): ZManaged[Any, Nothing, channel.EventLoopGroup] =
+      make(UIO(new IOUringEventLoopGroup(nThread, executor)))
+
     def auto(nThreads: Int): ZManaged[Any, Nothing, channel.EventLoopGroup] =
-      if (Epoll.isAvailable)
+      if (IOUring.isAvailable())
+        uring(nThreads)
+      else if (Epoll.isAvailable)
```
tusharmath: I would not keep `IOUring` as default until it becomes stable.

tusharmath:
thanks @madper! Looking forward to merging this PR once completed :)

madper:
Removed iouring from `auto`. Now people who want to use IOUring has to manually use     `.provideCustomLayer(ServerChannelFactory.uring ++ EventLoopGroup.uring(nThreads))` which looks good to me.

madper:
@tusharmath  As you suggested, now people who want to use iouring needs to appoint it manually.

zio-http/src/main/scala/zhttp/service/EventLoopGroup.scala
```diff
@@ -38,6 +39,12 @@ object EventLoopGroup {
     def epoll(nThreads: Int, executor: Executor): ZManaged[Any, Nothing, channel.EventLoopGroup] =
       make(UIO(new channel.epoll.EpollEventLoopGroup(nThreads, executor)))
 
+    def uring(nThread: Int): ZManaged[Any, Nothing, channel.EventLoopGroup] =
```
girdharshubham, line 44: @madper The EventLoopGroup singleton object is missing the layer for `uring`

girdharshubham (changes requested):

zio-http/src/main/scala/zhttp/service/EventLoopGroup.scala
madper: Sorry @girdharshubham . Just fixed.

girdharshubham (approved):
These changes look good to me.
Tested this PR out on a linux machine with kernel version `5.9.0-050900-generic`

girdharshubham (changes requested):
@madper I think we should keep it as Task[_] instead of UIO[_] because there could be a throwable in the case of Unix-based platforms and kernel version requirement.

tusharmath (approved):

PR #393, opened 2021-10-12
Feature: `finalFragment` flag for WebSocketFrame

morozov11:
On Gprs/LTE connections often need continuation frames processing, but current version zio-http don't contains finalization flag for continuation messages.

### DIFF
--- a/example/src/main/scala/WebSocketAdvanced.scala
+++ b/example/src/main/scala/WebSocketAdvanced.scala
@@ -9,13 +9,13 @@ object WebSocketAdvanced extends App {
   // Message Handlers
   private val open = Socket.succeed(WebSocketFrame.text("Greetings!"))
 
-  private val echo = Socket.collect[WebSocketFrame] { case WebSocketFrame.Text(text) =>
+  private val echo = Socket.collect[WebSocketFrame] { case WebSocketFrame.Text(text, _) =>
     ZStream.repeat(WebSocketFrame.text(s"Received: $text")).schedule(Schedule.spaced(1 second)).take(3)
   }
 
   private val fooBar = Socket.collect[WebSocketFrame] {
-    case WebSocketFrame.Text("FOO") => ZStream.succeed(WebSocketFrame.text("BAR"))
-    case WebSocketFrame.Text("BAR") => ZStream.succeed(WebSocketFrame.text("FOO"))
+    case WebSocketFrame.Text("FOO", _) => ZStream.succeed(WebSocketFrame.text("BAR"))
+    case WebSocketFrame.Text("BAR", _) => ZStream.succeed(WebSocketFrame.text("FOO"))
   }
 
   // Setup protocol settings
--- a/example/src/main/scala/WebSocketAdvanced.scala
+++ b/example/src/main/scala/WebSocketAdvanced.scala
@@ -9,13 +9,13 @@ object WebSocketAdvanced extends App {
   // Message Handlers
   private val open = Socket.succeed(WebSocketFrame.text("Greetings!"))
 
-  private val echo = Socket.collect[WebSocketFrame] { case WebSocketFrame.Text(text, _) =>
+  private val echo = Socket.collect[WebSocketFrame] { case WebSocketFrame.Text(text) =>
     ZStream.repeat(WebSocketFrame.text(s"Received: $text")).schedule(Schedule.spaced(1 second)).take(3)
   }
 
   private val fooBar = Socket.collect[WebSocketFrame] {
-    case WebSocketFrame.Text("FOO", _) => ZStream.succeed(WebSocketFrame.text("BAR"))
-    case WebSocketFrame.Text("BAR", _) => ZStream.succeed(WebSocketFrame.text("FOO"))
+    case WebSocketFrame.Text("FOO") => ZStream.succeed(WebSocketFrame.text("BAR"))
+    case WebSocketFrame.Text("BAR") => ZStream.succeed(WebSocketFrame.text("FOO"))
   }
 
   // Setup protocol settings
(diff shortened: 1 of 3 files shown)


tusharmath:
Can you expand more on why you would require `finalFragment`

morozov11:
> Can you expand more on why you would require `finalFragment`

In iOS 15 for Safai, an experimental function was enabled by default: nsurlsession websocket, which is why, when connecting via LTE, large ws frames begin to split with Continuation. Ideally, you need a facade to assemble a large package into one. But for now, just a flag would be enough.

zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala
```diff
@@ -5,15 +5,16 @@ import zhttp.core.ByteBuf
 
 sealed trait WebSocketFrame extends Product with Serializable { self =>
   def toWebSocketFrame: JWebSocketFrame = WebSocketFrame.toJFrame(self)
+  val finalFragment: Boolean            = true
```
tusharmath:
```suggestion
  def isFinal: Boolean
```

example/src/main/scala/WebSocketAdvanced.scala
```diff
@@ -9,13 +9,13 @@ object WebSocketAdvanced extends App {
   // Message Handlers
   private val open = Socket.succeed(WebSocketFrame.text("Greetings!"))
 
-  private val echo = Socket.collect[WebSocketFrame] { case WebSocketFrame.Text(text) =>
+  private val echo = Socket.collect[WebSocketFrame] { case WebSocketFrame.Text(text, _) =>
```
tusharmath: Instead of deploying a breaking change like this I would prefer if we could access the `isFinal` flag using a accessor eg:

```scala
Socket.collect[WebSocketFrame] { case fr @ Text(text) => fr.isFinal }
```

morozov11: ок

zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala
```diff
@@ -5,15 +5,23 @@ import zhttp.core.ByteBuf
 
 sealed trait WebSocketFrame extends Product with Serializable { self =>
   def toWebSocketFrame: JWebSocketFrame = WebSocketFrame.toJFrame(self)
+  def isFinal: Boolean                  = true
```
tusharmath:
```suggestion
  def isFinal: Boolean
```

zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala
```diff
@@ -5,15 +5,23 @@ import zhttp.core.ByteBuf
...
 
-  final case class Binary(buffer: ByteBuf)                    extends WebSocketFrame
-  final case class Text(text: String)                         extends WebSocketFrame
+  case class Binary(buffer: ByteBuf, override val isFinal: Boolean = true) extends WebSocketFrame
+  object Binary { def unapply(frame: WebSocketFrame.Binary) = Some(frame.buffer) }
+
+  final case class Text(text: String, override val isFinal: Boolean = true) extends WebSocketFrame
+  object Text { def unapply(frame: WebSocketFrame.Text) = Some(frame.text) }
+
   final case class Close(status: Int, reason: Option[String]) extends WebSocketFrame
```
tusharmath, line 31: Close, Ping, Pong should all have `isFinal` implemented isn't it?

morozov11: WebSocketFrame trait has isFinal = true by default. For Close, Ping, Pong there will be no Continuation frames. I think there is no point in setting this value manually. Right?

zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala
```diff
@@ -5,15 +5,23 @@ import zhttp.core.ByteBuf
 
 sealed trait WebSocketFrame extends Product with Serializable { self =>
   def toWebSocketFrame: JWebSocketFrame = WebSocketFrame.toJFrame(self)
+  def isFinal: Boolean
```
tusharmath:
```suggestion
  def toWebSocketFrame: JWebSocketFrame = WebSocketFrame.toJFrame(self)
  def isFinal: Boolean                  = true
```

zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala
```diff
@@ -5,15 +5,23 @@ import zhttp.core.ByteBuf
 
 sealed trait WebSocketFrame extends Product with Serializable { self =>
   def toWebSocketFrame: JWebSocketFrame = WebSocketFrame.toJFrame(self)
```
tusharmath:
```suggestion
  final def toWebSocketFrame: JWebSocketFrame = WebSocketFrame.toJFrame(self)
```

tusharmath (approved):

tusharmath (changes requested):
Can you run `fmt` command once.

morozov11:
> Can you run `fmt` command once.

I ran `fmt` command, but no changes happened and nothing added to commit. Locally the `sbt ++2.13.6 fmtCheck` executed fine. I don't understand what i am doing wrong. May be incorrect line endings?

amitksingh1490:
> > Can you run `fmt` command once.
> 
> I ran `fmt` command, but no changes happened and nothing added to commit. Locally the `sbt ++2.13.6 fmtCheck` executed fine. I don't understand what i am doing wrong. May be incorrect line endings?

Now the formatting is fixed could you please update the examples `WebSocketEcho.scala` and `WebSocketAdvanced.scala`

tusharmath (approved):

Issue #394, opened 2021-10-13
Preserve URL fragment


soujiro32167:
The url fragment `/path#fragment` is omitted in `zhttp.http.URL`. It should become part of the model

```scala
import zhttp.http.URL
import java.net.URI

val url = "/path#hash"

URL.fromString(url) // URL(/path,Relative,Map())

val jurl = new URI(url) // java.net.URI = /path#hash
jurl.getFragment // hash
```

PR #397, opened 2021-10-13
feature: #394 URL fragment

soujiro32167:


zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -97,4 +99,13 @@ object URL {
         else s"${scheme.asString}://$host:$port$path"
     }
   }
+
+  case class Fragment private (raw: String, decoded: String)
```
tusharmath, line 103: I think we should only keep decoded value in the fragment.

amitksingh1490:
can you please run command `fmt` on sbt shell, and push the changes it will fix the formatting.

zio-http/src/main/scala/zhttp/http/URL.scala
soujiro32167: Then we would have to re-encode it when going back to string...

soujiro32167: If its an allocation concern, how about we keep it as a `() => decoded`?

soujiro32167: Or the other way around - `case class Fragment private (encoded: () => String, decoded: String)`

tusharmath:
We will look at the performance piece a bit later. Thank you for your contribution!

tusharmath (approved):

PR #406, opened 2021-10-19
feat(HttpApp): added APIs from main

ShrutiVerma97:


zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -127,6 +127,16 @@ object HttpApp {
...
 
+  /**
+   * Creates an HTTP app which always responds with the same status code and empty data.
+   */
+  def status(code: Status): HttpApp[Any, Nothing] = HttpApp(Http.succeed(Response(code)))
+
+  /**
+   * Takes an HTTP app and returns a ZIO effect.
+   */
+  def toEffect[R, E](req: Request, httpApp: HttpApp[R, E]): ZIO[R, Option[E], Response[R, E]] = httpApp.asHttp(req)
```
tusharmath: drop this function.

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -137,6 +147,11 @@ object HttpApp {
         .flatten,
     )
 
+  /**
+   * Creates an HTTP app which always responds with a 200 status code.
+   */
+  def ok: HttpApp[Any, Nothing] = status(Status.OK)
```
tusharmath, line 148: Add all the methods for all the statuses

sumawa: How about adding something like "dsl" in the package like
package zhttp.http.dsl

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -138,11 +143,326 @@ object HttpApp {
     )
 
   /**
-   * Creates an Http app that responds with 403 - Forbidden status code
+   * Creates an HTTP app which always responds with a 200 status code.
+   */
+  def ok: HttpApp[Any, Nothing] = status(Status.OK)
+
+  /**
+   * Creates an HTTP app which always responds with a 100 status code.
```
sumawa: Should status definitions have their own trait, class object? something like Single Responsibility Principle 
https://en.wikipedia.org/wiki/Single-responsibility_principle

HttpApp has a global responsibility helping create a global app delegating individual steps/pieces to other components.

zio-http/src/main/scala/zhttp/http/HttpApp.scala
tusharmath: I think using a `dsl` package is a great idea. However, it's not very common in the ZIO suite of libraries. We should try to keep it close to the rest of the libraries in the ecosystem. That will help onboard users faster.

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -138,11 +143,326 @@ object HttpApp {
...
+  /**
+   * Creates an Http app that responds with 510 status code
+   */
+  def notExtended(msg: String): HttpApp[Any, Nothing] =
+    HttpApp(Http.succeed(Response.fromHttpError(HttpError.NotExtended(msg))))
+
+  /**
+   * Creates an Http app that responds with 511 status code
+   */
+  def networkAuthenticationRequired(msg: String): HttpApp[Any, Nothing] =
```
tusharmath: Too much noise. I think we can stick to the most used Status codes — 
- `200` `ok`
- `404` `notFound(path)`
- `400` `badRequest(message)`
- `500` `error(message)`

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -128,21 +128,41 @@ object HttpApp {
...
 
+  /**
+   * Creates an HTTP app which always responds with a 200 status code.
+   */
+  def ok: HttpApp[Any, Nothing] = status(Status.OK)
+
+  /**
+   * Creates an HTTP app which always responds with a 404 status code.
+   */
+  def notFound(path: Path): HttpApp[Any, Nothing] = HttpApp(
```
tusharmath: There is a `notFound` already.

We can also add `error(cause: HttpError)`

ShrutiVerma97: @tusharmath We are using `error` for  status code 500 also

tusharmath: We can overload it.

tusharmath: They should return `HttpApp[Any, Nothing]`

zio-http/src/main/scala/zhttp/http/HttpApp.scala
```diff
@@ -130,18 +130,37 @@ object HttpApp {
...
+   * Creates an HTTP app which always responds with the same status code and empty data.
+   */
+  def status(code: Status): HttpApp[Any, Nothing] = HttpApp(Http.succeed(Response(code)))
 
   /**
    * Creates an Http app that responds with 403 - Forbidden status code
    */
-  def forbidden(msg: String): HttpApp[Any, Nothing] =
-    HttpApp(Http.succeed(Response.fromHttpError(HttpError.Forbidden(msg))))
+  def forbidden(msg: String): HttpApp[Any, Nothing] = HttpApp.error(HttpError.Forbidden(msg))
```
ShrutiVerma97, line 143: we are using forbidden in examples as well as test cases

zio-http/src/main/scala/zhttp/http/HttpApp.scala
ShrutiVerma97: For now, we have removed all those status APIs. 
https://github.com/dream11/zio-http/pull/406#discussion_r735314528

tusharmath (approved):

tusharmath (approved):

PR #415, opened 2021-10-25
fix(HttpApp): Provide methods

ShrutiVerma97:


zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -10,7 +10,11 @@ case class Response[-R, +E] private (
   headers: List[Header],
   data: HttpData[R, E],
   private[zhttp] val attribute: HttpAttribute[R, E],
-)
+) { self =>
+  def provide[R1 <: R](r: R)(implicit ev: NeedsEnv[R]): Response[Any, E] =
+    self.copy(data = data.provide(r), attribute = attribute.asInstanceOf[HttpAttribute[Any, E]])
```
ShrutiVerma97, line 15: @tusharmath should I add provide method in HttpAttribute?

ShrutiVerma97:
We don't need this anymore.

PR #421, opened 2021-10-28
Fix: do not ignore query params when url path is empty

GrigoriiBerezin:
When we are trying to send request without url path with query parameters, they are just ignored


kaushik143:
#529

PR #444, opened 2021-10-30
Example of middleware usage

### LINKED ISSUE #417: Document an example of HttpApp with timeout middleware
# Description
Need a example of HttpApp which uses timeout middleware and close the server connection if the Client doesn't send a request under a fixed time duration.

# Example usage
```scala
app @@ timeout(5 second)
```

rpiotrow:
Fix #417

### DIFF
--- a/example/src/main/scala/HelloWorldWithMiddlewares.scala
+++ b/example/src/main/scala/HelloWorldWithMiddlewares.scala
@@ -19,24 +19,24 @@ object HelloWorldWithMiddlewares extends App {
     case Method.GET -> !! / "long-running" => ZIO.succeed(Response.text("Hello World!")).delay(5 seconds)
   }
 
-  val currentTimeMiddleware: HttpMiddleware[Clock, Nothing] = patchM((_, _) =>
+  val serverTime: HttpMiddleware[Clock, Nothing] = patchM((_, _) =>
     for {
       currentMilliseconds <- currentTime(TimeUnit.MILLISECONDS)
       withHeader = Patch.addHeaders(List(Header("X-Time", currentMilliseconds.toString)))
     } yield withHeader,
   )
 
-  val appWithMiddlewares: HttpApp[Console with Clock, IOException] = app @@
+  val middlewares: HttpMiddleware[Console with Clock, IOException] =
     // print debug info about request and response
-    debug @@
-    // close connection if request takes more than 3 seconds
-    timeout(3 seconds) @@
-    // add static header
-    addHeader("X-Environment", "Dev") @@
-    // add dynamic header
-    currentTimeMiddleware
+    debug ++
+      // close connection if request takes more than 3 seconds
+      timeout(3 seconds) ++
+      // add static header
+      addHeader("X-Environment", "Dev") ++
+      // add dynamic header
+      serverTime
 
   // Run it like any simple app
   override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
-    Server.start(8090, appWithMiddlewares.silent).exitCode
+    Server.start(8090, (app @@ middlewares).silent).exitCode
 }
--- a/example/src/main/scala/HelloWorldWithMiddlewares.scala
+++ b/example/src/main/scala/HelloWorldWithMiddlewares.scala
@@ -1,0 +1,40 @@
+import zhttp.http._
+import zhttp.http.middleware.HttpMiddleware._
+import zhttp.http.middleware.{HttpMiddleware, Patch}
+import zhttp.service.Server
+import zio._
+import zio.clock.{Clock, currentTime}
+import zio.console.Console
+import zio.duration.durationInt
+
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+object HelloWorldWithMiddlewares extends App {
+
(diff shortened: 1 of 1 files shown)


example/src/main/scala/HelloWorldWithMiddlewares.scala
```diff
@@ -0,0 +1,40 @@
...
+      // close connection if request takes more than 3 seconds
+      timeout(3 seconds) ++
+      // add static header
+      addHeader("X-Environment", "Dev") ++
+      // add dynamic header
+      patchM((_, _) =>
+        for {
+          currentMilliseconds <- currentTime(TimeUnit.MILLISECONDS)
+          withHeader = Patch.addHeaders(List(Header("X-Time", currentMilliseconds.toString)))
+        } yield withHeader,
```
amitksingh1490: Move this to a ```val currentTimeMiddleWare```

example/src/main/scala/HelloWorldWithMiddlewares.scala
```diff
@@ -0,0 +1,40 @@
...
+      )
+
+  // Run it like any simple app
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
+    Server.start(8090, (app @@ middlewares).silent).exitCode
```
amitksingh1490, line 41: could be done like below
```app @@ debug @@ timeout(3 seconds) @@ currentTimeMiddleware```

Comment (rpiotrow) [reply example/src/main/scala/HelloWorldWithMiddlewares.scala:41]: With `addHeader` it will be quite long:
```scala
(app @@ debug @@ timeout(3 seconds) @@ addHeader("X-Environment", "Dev") @@ currentTimeMiddleware).silent
```
and with no comments explaining what each middleware does. Is that okay? We can also remove `addHeader` and it will be exactly as you suggested.

amitksingh1490: agree with your point but with this example we also want to demo how similar attaching a middleware to the App is as compared to attaching TestAspect
```
 val appWithMiddlewares = app @@
    // print debug info about request and response
    debug @@
    // close connection if request takes more than 3 seconds
    timeout(3 seconds) @@
    // add static header
    addHeader("X-Environment", "Dev") @@
    // add dynamic header
    timeMiddleware

  // Run it like any simple app
    Server.start(8090, appWithMiddlewares.silent).exitCode
    
    ```
    What do you think of this ⬆️

File: example/src/main/scala/HelloWorldWithMiddlewares.scala [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  val currentTimeMiddleware: HttpMiddleware[Clock, Nothing] = patchM((_, _) =>
+    for {
+      currentMilliseconds <- currentTime(TimeUnit.MILLISECONDS)
+      withHeader = Patch.addHeaders(List(Header("X-Time", currentMilliseconds.toString)))
+    } yield withHeader,
+  )
+
+  val newApp = app @@
```
amitksingh1490: please rename this to` appWithMiddlewares`

example/src/main/scala/HelloWorldWithMiddlewares.scala
```diff
@@ -0,0 +1,42 @@
...
+object HelloWorldWithMiddlewares extends App {
+
+  val app: HttpApp[Clock, Nothing] = HttpApp.collectM {
+    // this will return result instantly
+    case Method.GET -> !! / "text"         => ZIO.succeed(Response.text("Hello World!"))
+    // this will return result after 5 seconds, so with 3 seconds timeout it will fail
+    case Method.GET -> !! / "long-running" => ZIO.succeed(Response.text("Hello World!")).delay(5 seconds)
+  }
+
+  val currentTimeMiddleware: HttpMiddleware[Clock, Nothing] = patchM((_, _) =>
```
tusharmath:
```suggestion
  val serverTime: HttpMiddleware[Clock, Nothing] = HttpMiddleware.patchM((_, _) =>
```

example/src/main/scala/HelloWorldWithMiddlewares.scala
```diff
@@ -0,0 +1,42 @@
...
+    // close connection if request takes more than 3 seconds
+    timeout(3 seconds) @@
+    // add static header
+    addHeader("X-Environment", "Dev") @@
+    // add dynamic header
+    currentTimeMiddleware
+
+  // Run it like any simple app
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
+    Server.start(8090, appWithMiddlewares.silent).exitCode
```
tusharmath: Actually, I think the previous approach was a bit better. We are able to showcase that you can compose middlewares together using the `++` operator.
```suggestion
  val middlewares: HttpMiddleware[Console with Clock, IOException] = 
    // print debug info about request and response
    debug ++
    // close connection if request takes more than 3 seconds
    timeout(3 seconds) ++
    // add static header
    addHeader("X-Environment", "Dev") ++
    // add dynamic header
    serverTime

  // Run it like any simple app
  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
    Server.start(8090, app @@ middleware.silent).exitCode
```

amitksingh1490 (approved):

PR #446, opened 2021-10-30
feat(CSRF): add middleware for CSRF handling

### LINKED ISSUE #420: CSRF Middleware | Double Cookie Pattern | Stateless
# Description
Cross-site request forgeries are a type of malicious exploit whereby unauthorized commands are performed on behalf of an authenticated user. We want to be able to provide middleware to prevent such attacks. Since most applications written using zio-http would be stateless we want to be able to provide at least the Double Submit Cookie pattern.

# Double Cookie Pattern
In this technique, we send a random value in both a cookie and as a request parameter, with the server verifying if the cookie value and request value match. When a user visits (even before authenticating to prevent login CSRF), the site should generate a (cryptographically strong) pseudorandom value and set it as a cookie on the user's machine separate from the session identifier. The site then requires that every transaction request include this pseudorandom value as a hidden form value (or other request parameter/header). If both of them match on the server-side, the server accepts it as a legitimate request and if they don't, it would reject the request.

# Api Suggestion
```scala
app @@ csrfBuilder @@ cookie
```

# Specs
* Double Cookie Pattern should be implemented as a middleware
* Unit test cases should be at least 80% for the middleware
* No Scalastyle warnings should be there
* A working example of how to use this middleware

smehta91:
Fixes: https://github.com/dream11/zio-http/issues/420

### DIFF
--- a/example/src/main/scala/CSRFExample.scala
+++ b/example/src/main/scala/CSRFExample.scala
@@ -2,17 +2,24 @@ import zhttp.http._
 import zhttp.http.middleware.CSRF.CSRF
 import zhttp.http.middleware.CSRF.CSRF.CookieSetting
 import zhttp.http.middleware.HttpMiddleware
-import zio.UIO
+import zhttp.service.Server
+import zio._
 
-object CSRFExample {
+object CSRFExample extends App {
   val app: HttpApp[Any, Nothing] = HttpApp.collectM {
-    case Method.GET -> !! / "safeEndpoint"    => UIO(Response.ok)
+    case Method.GET -> !! / "safeEndpoint"    =>
+      UIO(Response.ok)
     case Method.POST -> !! / "unsafeEndpoint" => UIO(Response.ok)
   }
   val csrf: CSRF                 = CSRF("x-csrf", CookieSetting("csrf-token"), () => UIO("token"))
   val generateTokenMiddleware: HttpMiddleware[Any, Nothing] = csrf.generateToken
   val ValidateTokenMiddleware: HttpMiddleware[Any, Nothing] = csrf.checkToken
-  app @@ generateTokenMiddleware.when((method, _, _) => method == Method.GET) @@ ValidateTokenMiddleware.when(
-    (method, _, _) => method == Method.POST,
-  )
+
+  val appNew =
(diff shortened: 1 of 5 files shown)


zio-http/src/main/scala/zhttp/http/middleware/CSRF/CSRF.scala
```diff
@@ -0,0 +1,103 @@
...
+  }) {
+    case (_, _, token) => {
+      token match {
+        case Some(t) =>
+          UIO(
+            Patch.addHeaders(
+              List(
+                Header(
+                  "Set-Cookie",
+                  Cookie(
```
amitksingh1490, line 55: can be written using Cookie @@ syntax

zio-http/src/main/scala/zhttp/http/middleware/CSRF/CSRF.scala
```diff
@@ -0,0 +1,103 @@
...
+      if (verified) {
+        Patch.empty
+      } else {
+        Patch.setStatus(Status.UNAUTHORIZED)
+      }
+    }
+  }
+}
+object CSRF                                                                                          {
+  case class CookieSetting(
```
amitksingh1490, line 90: you should use existing Cookie Domain for this, check Cookie.scala

example/src/main/scala/CSRFExample.scala
```diff
@@ -0,0 +1,25 @@
...
+    case Method.GET -> !! / "safeEndpoint"    =>
+      UIO(Response.ok)
+    case Method.POST -> !! / "unsafeEndpoint" => UIO(Response.ok)
+  }
+  val csrf: CSRF                 = CSRF("x-csrf", CookieSetting("csrf-token"), () => UIO("token"))
+  val generateTokenMiddleware: HttpMiddleware[Any, Nothing] = csrf.generateToken
+  val ValidateTokenMiddleware: HttpMiddleware[Any, Nothing] = csrf.checkToken
+
+  val appNew =
+    app @@ generateTokenMiddleware.when((method, _, _) => method == Method.GET) @@ ValidateTokenMiddleware.when(
```
amitksingh1490, line 19: Api too complex can't we simplify this to something like 
```app @@ csrf(name , token)``` ?

smehta91:
Closing in favor of https://github.com/dream11/zio-http/pull/542

PR #450, opened 2021-10-30
Authentication Middleware

### LINKED ISSUE #431: Add Authentication middleware
# Description

#API suggestion

```scala

```

# Spec

dinojohn:
fixes #431

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/middleware/AuthMiddleware.scala
+++ b/zio-http/src/main/scala/zhttp/http/middleware/AuthMiddleware.scala
@@ -1,0 +1,75 @@
+package zhttp.http.middleware
+
+import io.netty.handler.codec.http.HttpHeaderNames
+import pdi.jwt._
+import pdi.jwt.algorithms.JwtHmacAlgorithm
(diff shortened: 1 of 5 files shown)


zio-http/src/main/scala/zhttp/http/middleware/AuthMiddleware.scala
```diff
@@ -0,0 +1,88 @@
...
+import pdi.jwt._
+import pdi.jwt.algorithms.JwtHmacAlgorithm
+import zhttp.http.HeaderExtension.BasicSchemeName
+import zhttp.http.{Header, HeaderExtension, HttpApp, HttpError}
+import zio.ZIO
+
+/**
+ * Authentication Middlewares for HttpApp.
+ */
+sealed trait AuthMiddleware[-R, +E] { self =>
```
tusharmath, line 13: Can we not use `HttpMiddleware` for authentication?

dinojohn: We can, but the orElse operator won't work as expected

dinojohn:
closing in favour of #474

PR #451, opened 2021-10-30
#436: Logging middleware (wip)

### LINKED ISSUE #436: Logging middleware
# Description
In our HttpApp we came across many cases where we want to log requests and responses to understand what's going inside our application.

# API suggestion
```scala

app @@ log(requestLogger, responseLogger, options)

final case class RequestLogger(logMethod: Boolean, logHeaders: Boolean, level: Int)
final case class ResponseLogger(logMethod: Boolean, logHeaders: Boolean, level: Int)

sealed trait Options
object Options {
 case class SKIP(when: (req, res) => Boolean)
}
```

Spec
- Create Domain for RequestLogger, ResponseLogger, and Options taking care of all the basic options
- Add middleware that allows us to configure requestLogger, responseLogger, and options
- Add example showing the use of log

jgoday:
Implement logging middleware with zio-logging integration as described in #436.

A couple of pending issues/questions
1.- As Nabil A. pointed in the discord zio-hackathon zio-http channel, there should be a way to filter the headers that are going to be logged (maybe in another issue, maybe with a includeHeaders: List[String] in the domain object, ...)
2. Should the log method exposed as middleware has some default values (for request, response, options) ...
3. It is correct to have this logic separate in LogMiddleware.scala file or should be integrated in current HttpMiddleware.scala file ?

A working example could be
```scala
import zio._
import zio.console._
import zio.clock._
import zio.logging.{LogFormat, LogLevel, Logging}
import zhttp.endpoint._
import zhttp.http._
import zhttp.http.Method.GET
import zhttp.http.middleware.HttpMiddleware._
import zhttp.http.middleware.LogMiddleware._
import zhttp.service.Server
import java.io.IOException

object HelloWorld extends App {
  val env =
    Logging.console(
      logLevel = LogLevel.Info,
      format = LogFormat.ColoredLogFormat()
    ) >>> Logging.withRootLoggerName("sample.zio-http")

  def echo = GET / *[String] to { a =>
    Response.text(a.params.toString + "\n")
  }

  def test = GET / "test" to { _ =>
    Response.text("TEST")
  }

  val requestLogger = RequestLogger(logMethod=true, logHeaders=true, level=LogLevel.Debug)
  val responseLogger = ResponseLogger(logMethod=true, logHeaders=true, level=LogLevel.Info)
  val options = Options.Skip(
    { case ((_, url, _), _) => url.path.startsWith(Path("test")) }
  )

  def app: HttpApp[Console with Clock with Logging, Throwable] = echo @@ log(
    request = requestLogger,
    response = responseLogger,
    options = options
  )

    Server.start(8090, app).provideCustomLayer(env).exitCode
}
```


987Nabil:
I am not sure, if this logging headers is usable in a real word application without a headers filter. It seems for me not too big of an effort to impl. this directly. Or do you have a good argument against it?

jgoday:
> I am not sure, if this logging headers is usable in a real word application without a headers filter. It seems for me not too big of an effort to impl. this directly. Or do you have a good argument against it?

I think you are right, we can add an includeHeaders list to control which headers should we print
```
  final case class ResponseLogger(
    logMethod: Boolean,
    logHeaders: Boolean,
    includeHeaders: List[String] = Nil,
    level: LogLevel = LogLevel.Debug,
  )
```

987Nabil:
> > I am not sure, if this logging headers is usable in a real word application without a headers filter. It seems for me not too big of an effort to impl. this directly. Or do you have a good argument against it?
> 
> I think you are right, we can add an includeHeaders list to control which headers should we print
> 
> ```
>   final case class ResponseLogger(
>     logMethod: Boolean,
>     logHeaders: Boolean,
>     includeHeaders: List[String] = Nil,
>     level: LogLevel = LogLevel.Debug,
>   )
> ```

Why not just a function, that takes headers and returns headers? That is very dynamic and easy to implement.

jgoday:
> > > I am not sure, if this logging headers is usable in a real word application without a headers filter. It seems for me not too big of an effort to impl. this directly. Or do you have a good argument against it?
> > 
> > 
> > I think you are right, we can add an includeHeaders list to control which headers should we print
> > ```
> >   final case class ResponseLogger(
> >     logMethod: Boolean,
> >     logHeaders: Boolean,
> >     includeHeaders: List[String] = Nil,
> >     level: LogLevel = LogLevel.Debug,
> >   )
> > ```
> 
> Why not just a function, that takes headers and returns headers? That is very dynamic and easy to implement.

Do you mean something like ?
```
mapHeaders: (List[Header]) => List[Header]
```
I can see its usefulness in certain cases (print the size, filter or do some logic on the value) ...

jgoday:
@987Nabil I have changed the RequestLogger/ResponseLogger in the last commit,
to include, as you have said, the possibility of filtering/modifying the request/response headers (added some basic spec too).

```
final case class RequestLogger(
    logMethod: Boolean,
    logHeaders: Boolean,
    mapHeaders: (List[Header]) => List[Header] = identity,
    level: LogLevel = LogLevel.Debug,
  )
```

987Nabil:
Yes exactly. It should fit any case I guess. Maybe make it optional and remove the boolean for logging headers? If it is None, it is not logged. Did not think this through. Just a thought.

build.sbt
```diff
@@ -80,7 +80,7 @@ lazy val zhttp = (project in file("./zio-http"))
           new URL("https://github.com/amitksingh1490"),
         ),
       ),
-    libraryDependencies ++= Seq(`zio`, `zio-streams`, netty, `scala-compact-collection`, `netty-incubator`),
+    libraryDependencies ++= Seq(`zio`, `zio-logging`, `zio-streams`, netty, `scala-compact-collection`, `netty-incubator`),
```
tusharmath: I think we shouldn't have a hard dependency on zio-logging.

jgoday: how we can achieve this ? should we have a subproject zio-http-logging with the LogMiddleware and the zio-logging dependency isolated
or is there any other way to solve it? (something like a conditional dep or feature, I'm a little rusty with scala/sbt)

987Nabil: Logging is becoming a part of zio core with 2.0. So is this hard split necessary?

tusharmath:
@jgoday We have taken this up internally and are hoping to close that implementation soon. Closing this PR for now. Thank you!

PR #452, opened 2021-10-30
Rewrite CORS to use new middleware API

### LINKED ISSUE #427: Rewrite CORS middleware using new Middleware API
# Description
Cross-Origin Resource Sharing (CORS) is an HTTP-header-based mechanism that allows a server to indicate any origins (domain, scheme, or port) other than its own from which a browser should permit loading resources. Currently, we support this using the CORS domain added under the HTTP domain. 

* File: https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/http/CORS.scala
* PR: https://github.com/dream11/zio-http/pull/196/files

# Requirement 
- Implement CORS middleware using new Middleware API
- Add Example of HttpApp using CORS middleware
- Update existing CORSSpec 

# Reading material
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

rpiotrow:
I think it works in the same way as in current implementation.

Not sure if OPTIONS are handled correctly (in the current and new implementation). Http status 204 and CORS headers and returned for OPTIONS for any path (not only for paths that are defined in routes, but for all). Unfortunately I do not know how to return CORS headers only for paths defined in routes.

@amitksingh1490 @tusharmath 

Fix #427

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
+++ b/zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
@@ -10,9 +10,9 @@ import zhttp.http.HeaderExtension.{BasicSchemeName, BearerSchemeName}
 import java.nio.charset.Charset
 import scala.util.control.NonFatal
 
-case class HeadersList(val headers: List[Header]) extends ReadHeaderExtension[HeadersList]
+private[zhttp] trait HeaderExtension[+A] { self =>
+  def headers: List[Header]
 
-private[zhttp] trait HeaderExtension[+A] extends ReadHeaderExtension[A] { self =>
   def addHeaders(headers: List[Header]): A
 
   def removeHeaders(headers: List[String]): A
@@ -26,15 +26,6 @@ private[zhttp] trait HeaderExtension[+A] extends ReadHeaderExtension[A] { self =
 
   def setChunkedEncoding: A =
     addHeader(Header(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED))
-}
-
-object HeaderExtension {
-  val BasicSchemeName  = "Basic"
-  val BearerSchemeName = "Bearer"
-}
-
-private[zhttp] trait ReadHeaderExtension[+A] { self =>
-  def headers: List[Header]
 
   private def equalsIgnoreCase(a: Char, b: Char) = a == b || toLowerCase(a) == toLowerCase(b)
 
@@ -143,4 +134,9 @@ private[zhttp] trait ReadHeaderExtension[+A] { self =>
 
   def hasHeader(name: CharSequence): Boolean =
     getHeaderValue(name).nonEmpty
+}
+
+object HeaderExtension {
+  val BasicSchemeName  = "Basic"
+  val BearerSchemeName = "Bearer"
 }
--- a/zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
+++ b/zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
@@ -10,9 +10,9 @@ import zhttp.http.HeaderExtension.{BasicSchemeName, BearerSchemeName}
 import java.nio.charset.Charset
 import scala.util.control.NonFatal
 
-private[zhttp] trait HeaderExtension[+A] { self =>
-  def headers: List[Header]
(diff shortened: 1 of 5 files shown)


tusharmath:
- To identify that the route exists or not, one must execute the HttpApp.
- Execution can perform side-effects and thus should only be done if the cors middleware succeed.
- So I think its fine, if cors is activated on all routes.
- Alternatively, we can still provide a function (apart from the middleware), for devs to have more low level control over how they want to support cors (This can be a separate ticket/feature)

zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
```diff
@@ -135,8 +144,3 @@ private[zhttp] trait HeaderExtension[+A] { self =>
   def hasHeader(name: CharSequence): Boolean =
     getHeaderValue(name).nonEmpty
 }
```
tusharmath: Why are these changes required?

zio-http/src/test/scala/zhttp/service/CORSSpec.scala
```diff
@@ -11,9 +12,9 @@ object CORSSpec extends HttpRunnableSpec(8089) {
   val env = EventLoopGroup.auto() ++ ChannelFactory.auto ++ ServerChannelFactory.auto
 
   val app: ZManaged[EventLoopGroup with ServerChannelFactory, Nothing, Unit] = serve {
-    CORS(HttpApp.collect { case Method.GET -> !! / "success" =>
+    HttpApp.collect { case Method.GET -> !! / "success" =>
       Response.ok
-    })
+    } @@ HttpMiddleware.cors()
```
tusharmath:
```suggestion
    } @@ cors
```
Some syntax improvements :)

tusharmath:
Overall look good. Great that the old tests were used as it is!

zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
rpiotrow: `getHeader` uses `contentEqualsIgnoreCase` which is private for `HeaderExtension` and it has logic for comparing name of the header that is `AsciiString` (e.g. `HttpHeaderNames.ORIGIN`) with `CharSequence` that is in `Header.name`. I didn't want to duplicate code so I created `HeadersList` which extends `ReadHeaderExtension` which has only read-only methods from `HeaderExtension` (including `getHeader` used in CORS middleware).

zio-http/src/test/scala/zhttp/service/CORSSpec.scala
rpiotrow: We can make it as `cors()` when `zhttp.http.middleware.HttpMiddleware.cors` is in imports. `cors` will not work because there is parameter in this function (it has default value, but still invocation needs `()`)

rpiotrow:
I think implementation of CORS needs some more analysis and changes, which is out of scope of this issue. What I would see next is:
 * identify use cases/requirements for CORS usage
 * write unit test (or pseudo-unit test) for each of them
 * think how to implement CORS middleware (or maybe not only middleware but some other mechanism/features as well) to fulfill all requirements

zio-http/src/main/scala/zhttp/http/HeaderExtension.scala
rpiotrow: Other options to solve it:
 * use `HeadersList` as some kind of value class, move there all methods that are currently in `HeaderExtension`; in `Request` and `Response` use `HeaderList` as val and delegate methods like `getHeader` to this class
  * instead of trait use implicit class which will add methods to `List[Header]`, that will only work for "read" functions, "write" function would still need a trait like current, but it will be much smaller and "read" methods will be available everywhere on `List[Header]` when import of implicit class will be added.

tusharmath: Ok got it. Would you mind duplicating that bit of code for now? I am facing difficulty in understanding what `HeaderList` really stands for and if it should be part of extensions.

rpiotrow: Sure, I can do this, but I need some time now, I should be available in ~2-3h

rpiotrow: Done, You can take a look @tusharmath

tusharmath (approved):

PR #458, opened 2021-10-31
Add setting content-length/transfer-encoding headers globally

### LINKED ISSUE #448: Initialize response with default headers
When a Response is created, we should add at least the following two headers based on the `HttpData` that is provided 

1. Set `content-length` if the size of the HttpData is known.
2. Set `transfer-encoding: chunked` if the size of HttpData is unknown.

This should only happen if no value is already provided by the user.

grouzen:
Resolves #448 

**Summary**
Conditionally adds headers in the "entry" point of HttpApp


grouzen:
@tusharmath I'm not sure if I did this idiomatically. Please, check. If this way is okay I will add tests.

tusharmath:
I think you can update the `Request` constructor to add appropriate headers instead.

grouzen:
Sorry, but I do not see how Request headers correspond to Response headers. Maybe you can add more details? Thanks!

tusharmath:
You could update the `apply` method inside the Response's companion object such that based on the `HttpData` that is passed, `content-length` and `transfer-encoding` is automatically set.

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -45,24 +45,47 @@ case class Response[-R, +E] private (
...
-    Response(status, headers, data, HttpAttribute.empty)
+  ): Response[R, E] = {
+    val resp       = Response(status, headers, data, HttpAttribute.empty)
+    // Since these headers are mutual exclusive, check for the presence of at least one of them
+    val hasHeaders =
+      resp.getHeader(HttpHeaderNames.TRANSFER_ENCODING).exists(_.value.toString.contains(HttpHeaderValues.CHUNKED)) ||
+        resp.getHeader(HttpHeaderNames.CONTENT_LENGTH).isDefined
+
+    if (hasHeaders) resp
+    else resp.setPayloadHeaders
```
tusharmath: You can set the headers to use transfer encoding as chunked if `data.isChunked` is true.
Similarly, if `data.size` is a `Some(Int)` then you can use that also to set the content length header.

grouzen:
Okay, It seems someone already did this in another PR.

tusharmath:
Sorry about that @grouzen.

PR #459, opened 2021-11-01
fix(cookie): multipleDecode method in cookie

### LINKED ISSUE #455: Cookie domain: Decode cookie doesn't decode multiple cookies
Create an operator to decode multiple cookies
```scala
Cookie.decodeMultiple("Idea-d0e689db=da7366a7-2d54-478e-8ca7-9bd1314b226b; csrf-token=token123")
// Should return Either[Throwable, List[Cookie]]
```

ShrutiVerma97:
fixes #455


zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -190,6 +184,29 @@ object Cookie {
...
 
+  private def splitNameContent(kv: String): (String, Option[String]) =
+    kv.split("=", 2).map(_.trim) match {
+      case Array(v1)     => (v1, None)
+      case Array(v1, v2) => (v1, Some(v2))
+      case _             => ("", None)
+    }
+
+  /**
+   * Decodes a string with multiple cookies into a list of cookie
```
```suggestion
   * Decodes a string with multiple cookies into a list of cookies.
   * Used to decode cookies from Request Headers.
```

amitksingh1490 (approved):

PR #474, opened 2021-11-08
Authentication Middleware as an HttpMiddleware

### LINKED ISSUE #431: Add Authentication middleware

dinojohn:
fixes #431


dinojohn:
closing in favour of #506

PR #476, opened 2021-11-08
WIP: Provide method addition

### LINKED ISSUE #477: Add an implementation for provide in HttpApp and downstream types
* Add `provide` method in HttpApp and downstream types
* Add unit test cases for `provide` in HttpApp

### previously titled: Fix provide Method and Add Unit tests for `provide` in `HttpApp`

girdharshubham:
# Changelog
* provide method added for
  * Socket
  * SocketApp
  * HttpAttribute
  * Response
  * HttpApp
* Unit test case added for Socket's provide
* Unit test case added for HttpApp's provide
### Closes #477

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/HttpAttribute.scala
+++ b/zio-http/src/main/scala/zhttp/http/HttpAttribute.scala
@@ -1,6 +1,8 @@
 package zhttp.http
 
+import zhttp.http.HttpAttribute.Empty
 import zhttp.socket.SocketApp
+import zio.NeedsEnv
 
 /**
  * Attribute holder for and Responses
@@ -8,6 +10,11 @@ import zhttp.socket.SocketApp
 private[zhttp] sealed trait HttpAttribute[-R, +E] extends Product with Serializable { self =>
   def exit: HttpAttribute.AttributeExit[R, E]                                   = HttpAttribute.exit(self)
   def ++[R1 <: R, E1 >: E](other: HttpAttribute[R1, E1]): HttpAttribute[R1, E1] = HttpAttribute.Combine(self, other)
+  def provide(r: R)(implicit env: NeedsEnv[R]): HttpAttribute[Any, E]           = self match {
+    case HttpAttribute.Empty                => Empty
+    case HttpAttribute.Combine(self, other) => HttpAttribute.Combine(self.provide(r), other.provide(r))
+    case HttpAttribute.Socket(app)          => HttpAttribute.Socket(app.provide(r))
+  }
 }
 
 object HttpAttribute {
--- a/zio-http/src/main/scala/zhttp/socket/Socket.scala
+++ b/zio-http/src/main/scala/zhttp/socket/Socket.scala
@@ -20,17 +20,8 @@ sealed trait Socket[-R, +E, -A, +B] { self =>
 
   private[zhttp] def execute(a: A): ZStream[R, E, B] = self(a)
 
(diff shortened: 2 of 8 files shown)


zio-http/src/main/scala/zhttp/http/HttpAttribute.scala
```diff
@@ -1,13 +1,20 @@
...
 import zhttp.socket.SocketApp
+import zio.NeedsEnv
 
 /**
  * Attribute holder for and Responses
  */
 private[zhttp] sealed trait HttpAttribute[-R, +E] extends Product with Serializable { self =>
   def exit: HttpAttribute.AttributeExit[R, E]                                   = HttpAttribute.exit(self)
   def ++[R1 <: R, E1 >: E](other: HttpAttribute[R1, E1]): HttpAttribute[R1, E1] = HttpAttribute.Combine(self, other)
+  def provide(r: R)(implicit env: NeedsEnv[R]): HttpAttribute[Any, E]           = self match {
```
tusharmath, line 13: Create a new type for `HttpAttribute.Provide`

zio-http/src/main/scala/zhttp/socket/Socket.scala
```diff
@@ -20,6 +20,9 @@ sealed trait Socket[-R, +E, -A, +B] { self =>
 
   private[zhttp] def execute(a: A): ZStream[R, E, B] = self(a)
 
+  def provide(r: R)(implicit needsEnv: NeedsEnv[R]): Socket[Any, E, A, B] =
+    Socket.FromStreamingFunction(a => self(a).provide(r))
```
tusharmath:
```suggestion
    Socket.Provide(r, self)
```

zio-http/src/main/scala/zhttp/socket/SocketApp.scala
```diff
@@ -49,6 +50,23 @@ sealed trait SocketApp[-R, +E] { self =>
     loop(self, SocketConfig[R, E]())
   }
 
+  def provide(r: R)(implicit env: NeedsEnv[R]): SocketApp[Any, E] =
```
tusharmath, line 53:
```suggestion
  def provide(r: R)(implicit env: NeedsEnv[R]): SocketApp[Any, E] = Provide(r, self)
```

zio-http/src/main/scala/zhttp/socket/Socket.scala
girdharshubham: While providing an implementation for the Provide type and the method, the compiler keeps on warning that the effect has **Any** as its requirement
```scala
sealed trait Socket[-R, +E, -A, +B] { self =>
  def apply(a: A): ZStream[R, E, B] = self match {
    ...
    case Provide(m, r)               => m(a).provide(r)
  }
  def provide(r: R)(implicit needsEnv: NeedsEnv[R]): Socket[Any, E, A, B] = Provide(self, r)
}
object Socket {
  private final case class Provide[R, E, A, B](m: Socket[R, E, A, B], r: R)                 extends Socket[Any, E, A, B]
}
```

I believe the intention behind your comment is that with provide we should return a declaration of what we'd end up with and then the apply method is where the actual environment elimination should be.

girdharshubham: This has been resolved in newer commit(s)

girdharshubham:
Closing this PR in favour of smaller scoped PRs.
Comments will be resolved in them.

PR #506, opened 2021-11-11
Feat: authentication middleware

### LINKED ISSUE #431: Add Authentication middleware

dinojohn:
fixes #431


zio-http/src/main/scala/zhttp/http/middleware/HttpMiddleware.scala
```diff
@@ -187,6 +187,17 @@ object HttpMiddleware {
       List(Header(HttpHeaderNames.WWW_AUTHENTICATE, HeaderExtension.BasicSchemeName)),
     )
 
+  def basicAuth[R, E](u :String, p:String): HttpMiddleware[R, E] =
+    auth(
```
tusharmath: use the original basic auth ctor

PR #515, opened 2021-11-12
Feat/decoder/multipart requests

smehta91:
Add Multipart decoder


zio-http/src/main/scala/zhttp/http/ContentDecoder.scala
```diff
@@ -34,12 +35,25 @@ object ContentDecoder {
...
+      } yield (if (state.isFirst) Option(queue) else None, state.withAcc(queue).withFirst(false))
+    }
+  def multipartDecoder(request: Request): ContentDecoder[Any, Throwable, Chunk[Byte], Queue[Message]] =
+    ContentDecoder.collect(BackPressure[(Parser, Queue[Message])]()) { case (msg, state, _) =>
+      for {
+        (parser, queue) <- state.acc.fold {
+          Queue.bounded[Message](1).map(q => ((new Parser(request), q)))
+        }(UIO(_))
+        messages        <- Task(parser.getMessages(msg))
+        _               <- ZIO.foreach_(messages)(queue.offer).forkDaemon
```
tusharmath: Do we need `forkDaemon` why isn't it working without it?

PR #533, opened 2021-11-18
SignCookie middleware

### LINKED ISSUE #419: Allow creation of signed cookie
**Description**
To protect cookie tempering on the client side we can create signed cookies.

**API Suggestion**

```scala

// Add a sign method on Cookie
cookie.sign(secret = "password")

// To decode
Cookie.decodeRequestCookie(secret: Option[String], content: String): List[Cookie]
```

ShrutiVerma97:
fixes #419


zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -125,6 +129,29 @@ final case class Cookie(
...
+      sha256.init(secretKey)
+      val signed    = sha256.doFinal(self.content.getBytes())
+      val mda       = MessageDigest.getInstance("SHA-512")
+      val content   = self.content + '.' + getEncoder.encodeToString(mda.digest(signed))
+      Some(self.withContent(content))
+    } catch {
+      case _: Exception => None
+    }
+
+  }
```
tusharmath, line 158: Can we move this logic to the cookie encoder? Have a new field in `Cookie` called `secret: Option[String]`. That way if you have it in the cookie, you can use it to encode and sign the content.

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -125,6 +129,29 @@ final case class Cookie(
     cookie.flatten.mkString("; ")
   }
 
+  def sign(secret: String): Option[Cookie] = {
```
```suggestion
  def sign(secret: String): Cookie = {
```

`sign` should add the secret in the cookie.

zio-http/src/main/scala/zhttp/http/Cookie.scala
ShrutiVerma97: what about decodeResponseCookie? We won't be able to decode from content if it's signed or not, or what is the secret value. We can unsign using secret only.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 445115.37**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 427929.46**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 441839.75**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 437723.26**

example/src/main/scala/example/SignCookies.scala
```diff
@@ -0,0 +1,35 @@
...
+      Response.ok.addCookie(cookie @@ path(!! / "cookie") @@ httpOnly).addCookie(cookie2)
+
+    case req @ Method.GET -> !! / "secure-cookie" => {
+      val res = Response.ok.addCookie(cookie @@ secure @@ path(!! / "secure-cookie")).addCookie(cookie2)
+      println(req.getSignedCookies("secret"))
+      res
+    }
+  }
+  // signing cookies with a secret
+  val newApp      = app @@ signCookies("secret")
```
tusharmath: too much of noise in this example. I would drop the other non-core features from the example. Focus on what's important — "Ability to sign cookies"

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -73,6 +77,13 @@ final case class Cookie(
    */
   def withSameSite(v: Cookie.SameSite): Cookie = copy(sameSite = Some(v))
 
+  /**
+   * Adds secret in the cookie
+   */
+  def withSign(s: String): Cookie = {
```
```suggestion
  def withSign(key: String): Cookie = {
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
tusharmath: we can add another method to `decodeResponseSignedCookie` no?

tusharmath: We shouldn't be so eager to sign the cookie, consider this example — 
```scala
val cookie = Cookie("A", "B").sign("secret").withContent("C")
```

The fact that this cookie should be signed is ignored when we set the content again. This IMO is counterintuitive and breaks compositional semantics.

What do you think?

tusharmath:
Closing because of lack of activity.

zio-http/src/main/scala/zhttp/http/Cookie.scala
ShrutiVerma97: Yup, I feel the same. I've tried implementing it here: https://github.com/dream11/zio-http/pull/751/files

PR #537, opened 2021-11-18
Smart content type middleware

### LINKED ISSUE #238: Smart content-type response headers for static files
When serving static files, it would be nice if the `content-type` header was automatically filled for common file types (ex: `.js` files get `application/javascript`, etc). 

http4s does this here: 
https://github.com/http4s/http4s/blob/main/core/src/main/scala/org/http4s/StaticFile.scala#L240
https://github.com/http4s/http4s/blob/main/core/src/main/scala/org/http4s/MimeDB.scala

When the content-type is not filled, browsers might refuse to load javascript files.

olivierdeckers:
This sets the content type header based on the extension of the request path.

Mime type detection is done by leveraging a JVM mechanism. A default set of mime types is added in META-INF. Users can extend these by adding their own mime.types file as well.

Fixes #238


zio-http/src/main/scala/zhttp/http/middleware/HttpMiddleware.scala
```diff
@@ -363,6 +364,19 @@ object HttpMiddleware {
     existingRoutesWithHeaders orElse optionsHeaders
   }
 
+  /**
+   * Determine the content type of the response from the request path and add it as a response header. Commonly used
+   * mime types are included by default. Additional mime types can be registered by adding them in a mime.types file in
+   * your applications META-INF folder.
+   */
+  def smartContentType[R, E]: HttpMiddleware[R, E] = {
+    val mimeTypes = new MimetypesFileTypeMap()
```
kaushik-barodiya: We are creating a map for each request. Better if we create only once as it can degrade the performance.

zio-http/src/main/scala/zhttp/http/middleware/HttpMiddleware.scala
```diff
@@ -363,6 +364,19 @@ object HttpMiddleware {
...
+    HttpMiddleware.make((_, url, _) => {
+      val contentType = mimeTypes.getContentType(url.path.asString)
+      Patch.addHeader(Header.custom("Content-Type", contentType))
+    })((_, _, S) => S)
```
kaushik-barodiya: I found a few problems with the implementation
1. Patch we have added to update the Response. Here you are using Patch as a state. Can we return the contentType as state and patch the response to update Header.
2. For HeaderName you may use HttpHeaderNames class
3. This middleware will update the Header of all the routes. Instead, we should update the header of only the route serving static file.

example/src/main/scala/example/HelloWorldWithMiddlewares.scala
```diff
@@ -14,12 +15,15 @@ import java.util.concurrent.TimeUnit
...
+  // The content type header will be added based on the extension of the requested file by smartContentType
+  val static = Method.GET / "static" / *[String] to { req => Response.text(req.params) }
+
   val app: HttpApp[Clock, Nothing] = HttpApp.collectM {
     // this will return result instantly
     case Method.GET -> !! / "text"         => ZIO.succeed(Response.text("Hello World!"))
     // this will return result after 5 seconds, so with 3 seconds timeout it will fail
     case Method.GET -> !! / "long-running" => ZIO.succeed(Response.text("Hello World!")).delay(5 seconds)
-  }
+  } +++ static
```
kaushik-barodiya: 👍 Really liked the way you used the new Endpoint API to create another app. But for this example do you think it would be extra overkill? Instead, we may just add another route in the existing app?

zio-http/src/test/scala/zhttp/middleware/HttpMiddlewareSpec.scala
```diff
@@ -135,6 +135,15 @@ object HttpMiddlewareSpec extends DefaultRunnableSpec with HttpAppTestExtensions
...
         }
+      } +
+      suite("smartContentType middleware") {
+        val app = (HttpApp.ok @@ smartContentType).getHeader("Content-Type")
+        testM("set content type header based on request path extension") {
+          assertM(app(Request(url = URL(!! / "test.json"))))(isSome(equalTo("application/json")))
+        } +
+          testM("set content type header to application/octet-stream if the extension is not known") {
+            assertM(app(Request(url = URL(!! / "test.abcd"))))(isSome(equalTo("application/octet-stream")))
+          }
```
kaushik-barodiya: Can we add one more test where we have ContentType Header already set in Response? In that case, we should preserve the header value

kaushik143 (changes requested):

zio-http/src/main/scala/zhttp/http/middleware/HttpMiddleware.scala
olivierdeckers: The map is only created once when the middleware is constructed. I verified this in the HelloWorldWithMiddlewares example

example/src/main/scala/example/HelloWorldWithMiddlewares.scala
olivierdeckers: It allowed me to easily test with multiple values for content type, but I agree it might be overkill for this example. I'll change it to a simple route

zio-http/src/test/scala/zhttp/middleware/HttpMiddlewareSpec.scala
olivierdeckers: I added this

zio-http/src/main/scala/zhttp/http/middleware/HttpMiddleware.scala
olivierdeckers: I fixed this

olivierdeckers:
I fixed the build issue, but branch coverage is failing when I run it locally: `Coverage is below minimum [64.29% < 66.00%]: Branch:Total`
I'm a bit confused, because the only part I added is fully covered according to the report?

kaushik143:
@olivierdeckers Can you resolve the conflicts?

olivierdeckers:
> @olivierdeckers Can you resolve the conflicts?

done

tusharmath:
@olivierdeckers I think the static file pr #850 by @ashprakasan should cover this use case more comprehensively.
I don't think we need a middleware for it (unless you think otherwise).

Cheers! Thank you so much for the contribution 🙏

PR #543, opened 2021-11-21
Feature: Add method for creating text/html response

jczuchnowski:
I'm doing some experiments with serving static content and though this could be useful  in addition to `jsonString` and `text` methods on `Response`.

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Response.scala
+++ b/zio-http/src/main/scala/zhttp/http/Response.scala
@@ -92,6 +92,15 @@ object Response {
     }
   }
 
+  /**
+   * Creates a response with content-type set to text/html
+   */
+  def html(data: String): UResponse =
+    Response(
+      data = HttpData.fromChunk(Chunk.fromArray(data.getBytes(HTTP_CHARSET))),
+      headers = List(Header.contentTypeHtml),
+    )
+
   @deprecated("Use `Response(status, headers, data)` constructor instead.", "22-Sep-2021")
   def http[R, E](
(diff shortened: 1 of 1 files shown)


girdharshubham:
Hi @jczuchnowski 
A better api design would be using Zstream instead. In it, the method would take in the path of the HTML file and use Zstream to read it.

jczuchnowski:
@girdharshubham your suggestion makes the use case very narrow. This method could be easily used also to serve some arbitrary html assembled in the application.

But maybe it would make sense to have a couple of overloaded methods that can take a String or a Path? And maybe a ZStream version too?

girdharshubham:
@jczuchnowski  Yes, that would be great!

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -92,6 +92,15 @@ object Response {
     }
   }
 
+  /**
+   * Creates a response with content-type set to text/html
+   */
+  def html(data: String): UResponse =
+    Response(
+      data = HttpData.fromChunk(Chunk.fromArray(data.getBytes(HTTP_CHARSET))),
```
tusharmath:
```suggestion
      data = HttpData.fromText(data),
```

Can keep `charset` as an argument to `html()` defaulting to `HTTP_CHARSET`.

tusharmath:
One small change @jczuchnowski. Looks good other wise :)

tusharmath (approved):

Issue #571, opened 2021-11-26
Improve performance of cookie parser


tusharmath:
Cookie parsing currently involves a lot of interim memory allocations. Create a PR that improves the performance of cookie parsing by reducing such allocations.

jgoday:
Hi @tusharmath, just created this PR #576 changing Cookie.decodeResponseCookie

vigoo:
Closed by #576

PR #576, opened 2021-11-26
Performance: Cookie decoder improvements

jgoday:
See #571.

Reduce Cookie case class allocations in Cookie.decodeResponseCookie.
Right now, it allocates a new Cookie instance with each field (using Cookie.copy through several methods).
For example,
it allocates 3 instances for ```"SUID=123;httponly=true;expires=2007-12-03T10:15:30.00Z"```.

This change parses each field separately and then creates one single Cookie instance.

This simple jhm benchmark show the two approaches
```scala
package sample

import java.util.concurrent.TimeUnit
import org.openjdk.jmh.annotations._
import zhttp.http.Cookie

@State(Scope.Thread)
@BenchmarkMode(Array(Mode.Throughput))
@OutputTimeUnit(TimeUnit.MICROSECONDS)
class CookieDecode {
  @Benchmark
  def benchmarkDecode(): Unit = {
    val value="SUID=123;httponly=true;expires=2007-12-03T10:15:30.00Z"
    val _ = Cookie.decodeResponseCookie(value)
    ()
  }
}
```

Parsing and copying Cookie case class
```

# Warmup Iteration   1: 0,141 ops/us
[info] # Warmup Iteration   2: 0,151 ops/us
[info] # Warmup Iteration   3: 0,206 ops/us
[info] Iteration   1: 0,217 ops/us
[info] Iteration   2: 0,168 ops/us
[info] Iteration   3: 0,147 ops/us
[info] Result "sample.CookieDecode.benchmarkDecode":
[info]   0,177 ┬▒(99.9%) 0,653 ops/us [Average]
[info]   (min, avg, max) = (0,147, 0,177, 0,217), stdev = 0,036
[info]   CI (99.9%): [? 0, 0,831] (assumes normal distribution)
[info] # Run complete. Total time: 00:01:01
```

Parsing fields with mutable vars
```
[info] # Warmup Iteration   1: 0,340 ops/us
[info] # Warmup Iteration   2: 0,375 ops/us
[info] # Warmup Iteration   3: 0,354 ops/us
[info] Iteration   1: 0,369 ops/us
[info] Iteration   2: 0,375 ops/us
[info] Iteration   3: 0,349 ops/us
[info] Result "sample.CookieDecode.benchmarkDecode":
[info]   0,364 ┬▒(99.9%) 0,244 ops/us [Average]
[info]   (min, avg, max) = (0,349, 0,364, 0,375), stdev = 0,013
[info]   CI (99.9%): [0,120, 0,608] (assumes normal distribution)
[info] # Run complete. Total time: 00:01:00
```

### DIFF
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
@@ -1,0 +1,19 @@
+package zhttp.benchmarks
+
+import org.openjdk.jmh.annotations._
+import zhttp.http._
+
+import java.util.concurrent.TimeUnit
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class CookieDecodeBenchmark {
+  private val cookie = "SUID=123;httponly=true;expires=2007-12-03T10:15:30.00Z"
+
+  @Benchmark
+  def benchmarkApp(): Unit = {
+    val _ = Cookie.decodeResponseCookie(cookie)
+    ()
+  }
+}
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
@@ -10,15 +10,22 @@ import java.util.concurrent.TimeUnit
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
 class CookieDecodeBenchmark {
+  val random = new scala.util.Random()
+  val name   = random.alphanumeric.take(100).mkString("")
+  val value  = random.alphanumeric.take(100).mkString("")
+  val domain = random.alphanumeric.take(100).mkString("")
+  val path   = Path((0 to 10).map { _ => random.alphanumeric.take(10).mkString("") }.mkString(""))
+  val maxAge = random.nextLong()
+
   private val cookie    = Cookie(
-    "CookiesName",
-    "CookiesValue",
+    name,
+    value,
     Some(Instant.now()),
-    Some("domainValue"),
-    Some(Path("Some", "Path")),
+    Some(domain),
+    Some(path),
     true,
     true,
-    Some(1200),
+    Some(maxAge),
     Some(Cookie.SameSite.Strict),
   )
   private val cookieStr = cookie.encode
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
@@ -32,7 +32,7 @@ class CookieDecodeBenchmark {
 
   @Benchmark
   def benchmarkApp(): Unit = {
-    val _ = Cookie.decodeResponseCookie(cookieStr)
+    val _ = Cookie.unsafeDecodeResponseCookie(cookieStr)
     ()
   }
 }
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
@@ -3,17 +3,29 @@ package zhttp.benchmarks
 import org.openjdk.jmh.annotations._
 import zhttp.http._
 
+import java.time.Instant
 import java.util.concurrent.TimeUnit
 
 @State(Scope.Thread)
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
 class CookieDecodeBenchmark {
-  private val cookie = "SUID=123;httponly=true;expires=2007-12-03T10:15:30.00Z"
+  private val cookie    = Cookie(
+    "CookiesName",
+    "CookiesValue",
+    Some(Instant.now()),
+    Some("domainValue"),
+    Some(Path("Some", "Path")),
+    true,
+    true,
+    Some(1200),
+    Some(Cookie.SameSite.Strict),
+  )
+  private val cookieStr = cookie.encode
 
   @Benchmark
   def benchmarkApp(): Unit = {
-    val _ = Cookie.decodeResponseCookie(cookie)
+    val _ = Cookie.decodeResponseCookie(cookieStr)
     ()
   }
 }
--- a/zio-http/src/main/scala/zhttp/http/Cookie.scala
+++ b/zio-http/src/main/scala/zhttp/http/Cookie.scala
@@ -153,36 +153,61 @@ object Cookie {
     var httpOnly: Boolean         = false
     var sameSite: Cookie.SameSite = null
 
-    var prev = 0
-    var i    = 0
-    while (i >= 0) {
-      i = headerValue.indexOf(';', prev)
-      val c       = if (i <= 0) headerValue.substring(prev) else headerValue.substring(prev, i)
-      val (n, ct) = splitNameContent(c)
+    val headerLength = headerValue.length
 
-      (n.toLowerCase, ct) match {
-        case ("expires", v)                => expires = parseDate(v).getOrElse(null)
-        case ("max-age", v)                => maxAge = Try(v.toLong).toOption
-        case ("domain", v)                 => domain = v
-        case ("path", v)                   => path = if (!v.isEmpty) Path(v) else null
-        case ("secure", _)                 => secure = true
-        case ("httponly", _)               => httpOnly = true
-        case ("samesite", v)               =>
-          v.trim.toLowerCase match {
+    // iterate over all cookie fields (until next semicolon)
+    var curr = 0
+    var next = 0
+
+    while (next >= 0 && curr < headerLength) {
+      next = headerValue.indexOf(';', curr)
+      if (next < 0) {
+        next = headerLength
+      }
+
+      // skip whitespaces one by one to avoid trim allocations
+      if (headerValue.charAt(curr) == ' ') {
+        curr = curr + 1
+      } else {
+        // decode name and content first
+        if (name == null) {
+          val splitIndex = headerValue.indexOf('=', curr)
+          if (splitIndex >= 0 && splitIndex < next) {
+            name = headerValue.substring(0, splitIndex)
+            content = headerValue.substring(splitIndex + 1, next)
+          } else {
+            name = headerValue.substring(0, next)
+          }
+        } else if (headerValue.regionMatches(true, curr, "expires=", 0, 8)) {
+          expires = parseDate(headerValue.substring(curr + 8, next)).getOrElse(null)
+        } else if (headerValue.regionMatches(true, curr, "max-age=", 0, 8)) {
+          maxAge = Try(headerValue.substring(curr + 8, next).toLong).toOption
+        } else if (headerValue.regionMatches(true, curr, "domain=", 0, 7)) {
+          domain = headerValue.substring(curr + 7, next)
+        } else if (headerValue.regionMatches(true, curr, "path=", 0, 5)) {
+          val v = headerValue.substring(curr + 5, next)
+          if (!v.isEmpty) {
+            path = Path(v)
+          }
+        } else if (headerValue.regionMatches(true, curr, "secure", 0, 6)) {
+          secure = true
+        } else if (headerValue.regionMatches(true, curr, "httponly", 0, 8)) {
+          httpOnly = true
+        } else if (headerValue.regionMatches(true, curr, "samesite=", 0, 9)) {
+          val v = headerValue.substring(curr + 9, next).toLowerCase
+          v match {
             case "lax"    => sameSite = SameSite.Lax
             case "strict" => sameSite = SameSite.Strict
             case "none"   => sameSite = SameSite.None
             case _        => ()
           }
-        case (_, aContent) if name == null =>
-          name = n
-          content = aContent
-        case (_, _)                        => ()
+        }
+
+        curr = next + 1
       }
-      prev = i + 1
     }
 
-    if (!name.isEmpty || !content.isEmpty)
+    if ((name != null && !name.isEmpty) || (content != null && !content.isEmpty))
       Some(
         Cookie(
           name = name,
--- a/zio-http/src/main/scala/zhttp/http/Cookie.scala
+++ b/zio-http/src/main/scala/zhttp/http/Cookie.scala
@@ -156,7 +156,7 @@ object Cookie {
   def decodeResponseCookie(headerValue: String): Option[Cookie] =
     Try(unsafeDecodeResponseCookie(headerValue)).toOption
 
-  private[http] def unsafeDecodeResponseCookie(headerValue: String): Cookie = {
+  private[zhttp] def unsafeDecodeResponseCookie(headerValue: String): Cookie = {
     var name: String              = null
     var content: String           = null
     var expires: Instant          = null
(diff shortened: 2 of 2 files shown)


zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -147,25 +147,46 @@ object Cookie {
...
+          maxAge = maxAge,
+          domain = domain,
+          path = path,
+          isSecure = secure,
+          isHttpOnly = httpOnly,
+          sameSite = sameSite,
+        ),
+      )
+    } else {
+      None
```
tusharmath: This is great, you have almost doubled the perf! 

You can further reduce allocations by — 
1. Get rid of all the `Option` and use `null` instead. 
2. Every `map` operation is creating a new collection (more memory), use a for/while loop if possible
3. Get rid of interim `Function` and `Tuple` also.
4. Don't allocate the `String` inside the method, pull them outside so that thy are allocated only once.

NOTE: You can have a method called - `unsafeDecodeResponseCookie` that returns a null or a Cookie for performance and make it package private. Then also have a `decodeResponseCookie` which calls `unsafeDecodeResponseCookie` and wraps it in an `Option`

jgoday: Made some changes to avoid string split and some allocations.

Marked internal functions _parseDate_ and _splitNameContent_ @inline functions, not sure if it could be a problem ...

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -143,30 +143,61 @@ object Cookie {
...
+    var sameSite: Cookie.SameSite = null
+
+    var prev = 0
+    var i    = 0
+    while (i >= 0) {
+      i = headerValue.indexOf(';', prev)
+      val c       = if (i <= 0) headerValue.substring(prev) else headerValue.substring(prev, i)
+      val (n, ct) = splitNameContent(c)
+
+      (n.toLowerCase, ct) match {
```
tusharmath: unnecessary tuple allocation.

jgoday: @tusharmath I have mode some changes to avoid string and tuple unnecessary allocations, inner loop is not the fanciest but I think that this way you get more performance ... what do you think ?

tusharmath: Can you commit the `Benchmark` code to the benchmarks project.

jgoday: Sure, added CookieDecodeBenchark to zio-http-benchmarks (very naive at the moment)

zio-http/src/main/scala/zhttp/http/Cookie.scala
tusharmath: can you update the benchmark numbers as before and after. I think we can make more changes for performance :)

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -176,24 +232,28 @@ object Cookie {
...
-    Try(Instant.parse(v)).toOption
-
-  private def splitNameContent(kv: String): (String, Option[String]) =
-    kv.split("=", 2).map(_.trim) match {
-      case Array(v1)     => (v1, None)
-      case Array(v1, v2) => (v1, Some(v2))
-      case _             => ("", None)
+  @inline
+  private def parseDate(v: String): Try[Instant] =
+    Try(Instant.parse(v))
```
tusharmath: Try will allocate additional memory. Instead, you could use a `try { } catch {}` block.
Incase of failure return a `null`

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -176,24 +232,28 @@ object Cookie {
...
+
+  @inline
+  private def splitNameContent(str: String): (String, String) = {
+    val i = str.indexOf('=')
+    if (i >= 0) {
+      (str.substring(0, i).trim, str.substring(i + 1).trim)
```
tusharmath, line 260: Extremely expensive computations. Every `substring` and `trim` has a `O(n)` space cmplx. If you can reduce this it will make a big difference.

zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
```diff
@@ -0,0 +1,19 @@
...
+import org.openjdk.jmh.annotations._
+import zhttp.http._
+
+import java.util.concurrent.TimeUnit
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class CookieDecodeBenchmark {
+  private val cookie = "SUID=123;httponly=true;expires=2007-12-03T10:15:30.00Z"
```
tusharmath: May be generate a larger cookie. With a large payload and more key-value pairs.

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -143,30 +143,86 @@ object Cookie {
...
+        // decode name and content first
+        if (name == null) {
+          val splitIndex = headerValue.indexOf('=', curr)
+          if (splitIndex >= 0 && splitIndex < next) {
+            name = headerValue.substring(0, splitIndex)
+            content = headerValue.substring(splitIndex + 1, next)
+          } else {
+            name = headerValue.substring(0, next)
+          }
+        } else if (headerValue.regionMatches(true, curr, "expires=", 0, 8)) {
```
tusharmath: The string literals will also allocate memory every time the method is called. You could move it into a value so that it's allocated only once.

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -143,30 +143,86 @@ object Cookie {
...
+            path = Path(v)
+          }
+        } else if (headerValue.regionMatches(true, curr, "secure", 0, 6)) {
+          secure = true
+        } else if (headerValue.regionMatches(true, curr, "httponly", 0, 8)) {
+          httpOnly = true
+        } else if (headerValue.regionMatches(true, curr, "samesite=", 0, 9)) {
+          val v = headerValue.substring(curr + 9, next).toLowerCase
+          v match {
+            case "lax"    => sameSite = SameSite.Lax
```
tusharmath: I am not really sure but `"lax"` AFAIK will also allocate additional memory.

zio-http/src/main/scala/zhttp/http/Cookie.scala
tusharmath: Instead of using `None` use `null`. This can cause null pointer exceptions but that should be fine because we will rename this method to `unsafeDecodeResponseCookie` 
and make it package-private. And, we will have a separate method `decodeResponseCookie` which calls `unsafeDecodeResponseCookie` and wraps the output in an `Option`. And that  should be a public method.

zio-http/src/main/scala/zhttp/http/Cookie.scala
tusharmath: Here is a crazy idea :P 

What if you keep the original string as it is and at the time of parsing create a cookie that maintains indexes of start and stop on the original string. Thereby not allocating any intermediary memory at all. eg:
```scala
trait Cookie {
  private def raw: String
  private def nameStart: Int = ??? // fill this at the time of parsing
  private def nameEnd: Int   = ??? // fill this at the time of parsing
  def name: String = raw.substring(nameStart, nameEnd)
}
```
The code above is just to depict the idea, actual implementation might be very different. However, this would be dramatically faster than the original solution because you are only allocating when you need that data.

If you would like to do this index-based optimization in a different PR, I wouldn't mind that also :)

jgoday:
@tusharmath
Updated with your latest suggestions:
1. String literals
2. Update benchmark with a larger payload cookie
```
OLD CODE:
[info] CookieDecodeBenchmark.benchmarkApp  thrpt    3  387562,885 ± 52538,364  ops/s

NEW CODE:
[info] CookieDecodeBenchmark.benchmarkApp  thrpt    3  679569,604 ± 27421,427  ops/s
```
3. Avoid scala.util.Try in parseDate
4. Added unsafeDecodeResponseCookie: Cookie
5. Really like the idea of having a cookie with the index-based fields, specially to avoid splitNameContent allocations in 'decodeRequestCookie', but as you said, maybe it's better to do it in another PR ...

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -176,24 +246,24 @@ object Cookie {
...
-  private def parseDate(v: String): Option[Instant] =
-    Try(Instant.parse(v)).toOption
-
-  private def splitNameContent(kv: String): (String, Option[String]) =
-    kv.split("=", 2).map(_.trim) match {
-      case Array(v1)     => (v1, None)
-      case Array(v1, v2) => (v1, Some(v2))
-      case _             => ("", None)
+  @inline
+  private def splitNameContent(str: String): (String, String) = {
```
tusharmath, line 257: Avoid tuple creation.

zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
```diff
@@ -0,0 +1,31 @@
...
+    "CookiesName",
+    "CookiesValue",
+    Some(Instant.now()),
+    Some("domainValue"),
+    Some(Path("Some", "Path")),
+    true,
+    true,
+    Some(1200),
+    Some(Cookie.SameSite.Strict),
+  )
```
tusharmath, line 30: Can we use a strings that are at least 100 chars or more. 

```scala
val random = new scala.util.Random((Math.random * Int.MaxValue).toInt)
val name = random.alphanumeric.take(100).mkString("")
val value = random.alphanumeric.take(100).mkString("")
val domain = random.alphanumeric.take(100).mkString("")
val name = Path((0 to 10).map { _ => random.alphanumeric.take(10).mkString("") })
```

The real improvement in performance will be observed in larger strings.

zio-http/src/main/scala/zhttp/http/Cookie.scala
tusharmath: You can create two separate functions that give you different parts viz. name and value, from the string.

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -176,24 +246,24 @@ object Cookie {
     val cookies: Array[String]  = headerValue.split(";").map(_.trim)
```
```suggestion
    val cookies: Array[String]  = headerValue.split(';').map(_.trim)
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -176,24 +246,24 @@ object Cookie {
...
     val x: List[Option[Cookie]] = cookies.toList.map(a => {
       val (name, content) = splitNameContent(a)
-      if (name.trim == "" && content.isEmpty) None
-      else Some(Cookie(name, content.getOrElse("")))
+      if (name == "" && content.isEmpty) None
```
```suggestion
      if (name.isEmpty && content.isEmpty) None
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
jgoday: splitNameContent is currently only used in decodeRequestCookie, to avoid this tuple allocation we can just chante it to
```scala
val name    =
  if (i >= 0) a.substring(0, i).trim
  else a.trim
val content =
  if (i >= 0) a.substring(i + 1).trim
  else null

if (name.isEmpty && content.isEmpty) None
else Some(Cookie(name, content))
```

however, maybe this is the best candidate to use a index base view, like  you mentioned earlier ...
```
case class CookieView(cookieStr) implements Cookie {
  def name: String = ...
  def content: String = ....
}
```

zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
jgoday: Updated benchmark code with the following results:
```
Old code:
(min, avg, max) = (393771,450, 395268,674, 396456,159), stdev = 1368,893

New code:
(min, avg, max) = (634553,745, 636047,487, 637677,233), stdev = 1566,179

```

tusharmath (approved):

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -142,58 +153,117 @@ object Cookie {
...
+            content = headerValue.substring(splitIndex + 1, next)
+          } else {
+            name = headerValue.substring(0, next)
+          }
+        } else if (headerValue.regionMatches(true, curr, fieldExpires, 0, fieldExpires.length)) {
+          expires =
+            try { Instant.parse(headerValue.substring(curr + 8, next)) }
+            catch { case _: Throwable => null }
+        } else if (headerValue.regionMatches(true, curr, fieldMaxAge, 0, fieldMaxAge.length)) {
+          maxAge = Try(headerValue.substring(curr + 8, next).toLong).toOption
```
tusharmath: We can create a global `try` `catch` block for this method.

jgoday: @tusharmath , a couple of questions here:
1) would it be enough to wrap with a try each attempt to decode a field?

2) What should we do with the exception, discard it completely?

tusharmath: 1. We should have only 1 top level try catch block inside the `safe` implementation.
2. Incase of failure we should throw an exception in the `unsafe` implementation. The safe implementation captures it and then could do either of the two things —
   a. Ignore the error and return a `None`.
   b. Capture the error and return a `Left[Throwable]`

tusharmath:
Really looking forward to this PR @jgoday :)

zio-http/src/main/scala/zhttp/http/Cookie.scala
jgoday: Ok, changed to catch the exception in decodeResponseCookie but keeping the Option[Cookie] signature

zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
```diff
@@ -0,0 +1,38 @@
...
+    true,
+    true,
+    Some(maxAge),
+    Some(Cookie.SameSite.Strict),
+  )
+  private val cookieStr = cookie.encode
+
+  @Benchmark
+  def benchmarkApp(): Unit = {
+    val _ = Cookie.decodeResponseCookie(cookieStr)
```
tusharmath:
```suggestion
    val _ = Cookie.unsafeDecodeResponseCookie(cookieStr)
```

tusharmath: Can you run this one last time and post the results :)

jgoday: Ok, applied your suggestionts.
I have run the benchmark again, here are the results

jmh:run -i 3 -wi 3 -f1 -t4

Old code:
```
[info] Iteration   1: 379730,343 ops/s
[info] Iteration   2: 383230,774 ops/s
[info] Iteration   3: 381724,031 ops/s
[info] Result "sample.CookieDecodeBenchmark.benchmarkApp":
[info]   381561,716 ±(99.9%) 32033,267 ops/s [Average]
[info]   (min, avg, max) = (379730,343, 381561,716, 383230,774), stdev = 1755,851
[info]   CI (99.9%): [349528,449, 413594,983] (assumes normal distribution)
[info] # Run complete. Total time: 00:01:00
```

New code:
```
[info] Iteration   1: 653978,242 ops/s
[info] Iteration   2: 635232,010 ops/s
[info] Iteration   3: 633734,049 ops/s
[info] Result "sample.CookieDecodeBenchmark.benchmarkApp":
[info]   640981,434 ±(99.9%) 205797,553 ops/s [Average]
[info]   (min, avg, max) = (633734,049, 640981,434, 653978,242), stdev = 11280,458
[info]   CI (99.9%): [435183,881, 846778,986] (assumes normal distribution)
[info] # Run complete. Total time: 00:01:00

```

tusharmath:
@jgoday Thanks! this was pretty amazing :) We make it 1.7x faster. It might not seem much right now, but it all adds up when write a proper e2e application.

jgoday:
Thank you @tusharmath  for all your help and support.

PR #636, opened 2021-12-15
Refactor: Client Response Handler

gciuloaica:


amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 435497.64**

zio-http/src/main/scala/zhttp/service/client/content/handlers/DynamicContentHandler.scala
```diff
@@ -0,0 +1,39 @@
+package zhttp.service.client.content.handlers
+
+import io.netty.buffer.{ByteBuf, Unpooled}
+import io.netty.channel.{ChannelHandlerContext, SimpleChannelInboundHandler}
+import io.netty.handler.codec.http.{FullHttpResponse, HttpHeaderNames}
+import zhttp.http.{HTTP_CHARSET, Header, Status}
+import zhttp.service.Client
+import zhttp.service.Client.{ClientResponse, StringContent}
+
+final case class DynamicContentHandler() extends SimpleChannelInboundHandler[FullHttpResponse](true) {
```
tusharmath:
```suggestion
final class DynamicContentHandler() extends SimpleChannelInboundHandler[FullHttpResponse](true) {
```

As a general practice, we don't use case classes unless its immutable.

zio-http/src/main/scala/zhttp/service/client/content/handlers/DynamicContentHandler.scala
```diff
@@ -0,0 +1,39 @@
...
+      case Some(value) =>
+        value match {
+          case Header.contentTypeTextPlain =>
+            ContentHandler.getStringContentResponse(status, headers, content)
+          case Header.contentTypeJson      =>
+            ContentHandler.getStringContentResponse(status, headers, content)
+          case _                           => ContentHandler.getStringContentResponse(status, headers, content)
+        }
+      case None        => ContentHandler.getStringContentResponse(status, headers, content)
+    }
```
tusharmath: - Why are we pattern matching if in all cases we have to do the same thing?
- Decoding the content to String is expensive, and possibly unnecessary in scenarios when the content is not read by the HttpApp. Eg: An unauthenticated POST request etc.

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,12 +174,26 @@ object Client {
     val url: URL       = endpoint._2
   }
 
-  final case class ClientResponse(status: Status, headers: List[Header], private val buffer: ByteBuf)
+  sealed trait Content {
+    type T
+    def getContent: T
+  }
```
tusharmath: It's not really clear to me why we need Path dependant types here.

zio-http/src/main/scala/zhttp/service/client/content/handlers/DynamicContentHandler.scala
gciuloaica: Stil WIP. there should be different content handling for each type of content type that we support. I used the String model only to exemplify the intent.

zio-http/src/main/scala/zhttp/service/client/content/handlers/DynamicContentHandler.scala
gciuloaica: updated.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 441930.31**

zio-http/src/main/scala/zhttp/service/client/content/handlers/DynamicContentHandler.scala
gciuloaica: refactored this part. As we do not want to have dependencies on encoders/decoders libraries ( for json, xml, etc.) this handler is basically just doing the decoding of the response and copy the content into heap. The `ClientResponse` is providing access to raw content, such that users of the client library can decode it to json/xml/ ....

zio-http/src/main/scala/zhttp/service/Client.scala
gciuloaica: indeed, we do not need any of this part.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 433441.78**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 437564.66**

zio-http/src/main/scala/zhttp/service/client/content/handlers/ClientInboundContentHandler.scala
```diff
@@ -0,0 +1,19 @@
+package zhttp.service.client.content.handlers
+
+import io.netty.buffer.Unpooled
+import io.netty.channel.{ChannelHandlerContext, SimpleChannelInboundHandler}
+import io.netty.handler.codec.http.FullHttpResponse
+import zhttp.http.{Header, Status}
+import zhttp.service.Client
+
+final class ClientInboundContentHandler() extends SimpleChannelInboundHandler[FullHttpResponse](true) {
```
tusharmath:
```suggestion
final class ClientResponseHandler() extends SimpleChannelInboundHandler[FullHttpResponse](true) {
```

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,12 +175,16 @@ object Client {
     val url: URL       = endpoint._2
   }
 
-  final case class ClientResponse(status: Status, headers: List[Header], private val buffer: ByteBuf)
+  final case class ClientResponse(status: Status, headers: List[Header], private val content: ByteBuf)
```
tusharmath:
```suggestion
  final case class ClientResponse(status: Status, headers: List[Header], private val unsafeByteBuf: ByteBuf)
```

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,12 +175,16 @@ object Client {
...
       extends HeaderExtension[ClientResponse] { self =>
-    def getBodyAsString: Task[String] = Task(buffer.toString(self.getCharset))
+
+    def getBodyAsString: Task[String] = Task(content.toString(self.getCharset))
+
+    def getRawBody: Task[ByteBuf] = Task(content)
```
tusharmath:
```suggestion
    def getBody: Task[Chunk[Byte]] = getBodyAsByteBuf.flatMap(buf => Task(Chunk.fromArray(buf.array())))
```

tusharmath: we should avoid exposing any netty based types as much as possible.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 436593.44**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 435999.30**

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,12 +175,16 @@ object Client {
...
   }
 
-  final case class ClientResponse(status: Status, headers: List[Header], private val buffer: ByteBuf)
+  final case class ClientResponse(status: Status, headers: List[Header], private val unsafeByteBuf: ByteBuf)
       extends HeaderExtension[ClientResponse] { self =>
-    def getBodyAsString: Task[String] = Task(buffer.toString(self.getCharset))
+
+    def getBodyAsString: Task[String] = Task(unsafeByteBuf.toString(self.getCharset))
+
+    def getRawBody: Task[Chunk[Byte]] = Task(Chunk.fromArray(unsafeByteBuf.array()))
```
tusharmath: Use `ByteBufUtil` to get a `Byte[]` from `ByteBuf`

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 438889.62**

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,12 +175,16 @@ object Client {
...
   }
 
-  final case class ClientResponse(status: Status, headers: List[Header], private val buffer: ByteBuf)
+  final case class ClientResponse(status: Status, headers: List[Header], private val unsafeByteBuf: ByteBuf)
       extends HeaderExtension[ClientResponse] { self =>
-    def getBodyAsString: Task[String] = Task(buffer.toString(self.getCharset))
+
+    def getBodyAsString: Task[String] = Task(unsafeByteBuf.toString(self.getCharset))
+
+    def getRawBody: Task[Chunk[Byte]] = Task(Chunk.fromArray(ByteBufUtil.getBytes(unsafeByteBuf)))
```
tusharmath: Let's maintain consistent naming conventions. In `Request` this method is called as `getBody` and there is another method for `getBodyAsByteBuf` which is private and returns a ByteBuf.

zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -83,5 +83,13 @@ abstract class HttpRunnableSpec(port: Int) extends DefaultRunnableSpec { self =>
       headers: List[Header] = Nil,
     )(name: CharSequence): ZIO[EventLoopGroup with ChannelFactory with HttpAppCollection, Throwable, Option[String]] =
       request(path, method, content, headers).map(_.getHeaderValue(name))
+
+    def requestRawBody(
```
tusharmath:
```suggestion
    def requestBody(
```

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 450721.60**

zio-http/src/main/scala/zhttp/service/client/content/handlers/ClientResponseHandler.scala
```diff
@@ -0,0 +1,19 @@
+package zhttp.service.client.content.handlers
+
+import io.netty.buffer.Unpooled
+import io.netty.channel.{ChannelHandlerContext, SimpleChannelInboundHandler}
+import io.netty.handler.codec.http.FullHttpResponse
+import zhttp.http.{Header, Status}
+import zhttp.service.Client
+
+final class ClientResponseHandler() extends SimpleChannelInboundHandler[FullHttpResponse](true) {
```
tusharmath, line 12: Add a scala doc to explain what it does. This will help contributors later.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 441286.74**

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -72,6 +81,10 @@ object ServerSpec extends HttpRunnableSpec(8088) {
             val res = app.requestBodyAsString(content = "")
             assertM(res)(equalTo(""))
           } +
+          testM("empty raw body") {
+            val res = app.requestRawBody(content = "")
+            assertM(res)(isEmpty)
+          } +
```
tusharmath: I think one suite should be enough where the content is generated using a `Gen` and then asserted once.

```scala
suite("getBody") {
  test("is valid") {
    check(Gen.anyString) { string =>
       // create an app, execute and asssert
    }
  }
}
```

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 444363.14**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 426721.03**

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -116,6 +116,18 @@ object ServerSpec extends HttpRunnableSpec(8088) {
...
+      checkAllM(nonEmptyContent) { case (string, data) =>
+        val res     = Http.data(data).requestBody()
+        val content = res.map(_.toList.map(_.toChar).mkString)
+        assertM(content)(equalTo(string))
+      }
+    } +
+      testM("empty content") {
+        val res = Http.data(HttpData.empty).requestBody()
+        assertM(res)(isEmpty)
+      }
```
tusharmath: move the test to `ClientSpec.scala`

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 430760.80**

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
gciuloaica: done.

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -11,6 +13,33 @@ object ClientSpec extends HttpRunnableSpec(8082) {
...
+      testM("empty content") {
+        val actual          = Client.request("http://api.github.com/users/zio/repos", ClientSSLOptions.DefaultSSL)
+        val responseContent = actual.flatMap(_.getBody)
+        assertM(responseContent)(isEmpty)
+      } +
+      testM("text content") {
+        val actual          = Client.request("https://www.google.com", ClientSSLOptions.DefaultSSL)
+        val responseContent = actual.flatMap(_.getBodyAsString)
+        assertM(responseContent)(containsString("Google"))
+      },
```
tusharmath: The problem with these tests is that — I can't make it work offline. It's not a major problem, however, I feel that tests become less flaky when they don't depend on external envs. 

Unless really required we should use our own server and client for all tests.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 438294.07**

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,45 +1,69 @@
...
+            testM("empty content") {
+              val actual          = Client.request("http://localhost:8082/users/zio/repos", ClientSSLOptions.DefaultSSL)
+              val responseContent = actual.flatMap(_.getBody)
+              assertM(responseContent)(isEmpty)
+            } +
+            testM("text content") {
+              val actual          = Client.request("http://localhost:8082/users/zio", ClientSSLOptions.DefaultSSL)
+              val responseContent = actual.flatMap(_.getBodyAsString)
+              assertM(responseContent)(containsString("user"))
+            },
```
tusharmath: One of the things we do in ServerSpec is that the `Http` app that we want to test is co-located within the test. This makes it much easier for us to maintain and test. 

Keeping a global HttpApp will become a maintenance mess once the suite becomes large.

Take a look at `AppCollection.scala` to get an understanding of how it's being done.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 437204.65**

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
gciuloaica: Updated the ClientSpec accordingly

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 435703.41**

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,65 @@
...
+      testM("text content") {
+        val actual          = request(!! / "users" / "zio")
+        val responseContent = actual.flatMap(_.getBodyAsString)
+        assertM(responseContent)(containsString("user"))
+      },
+  ).provideCustomLayer(env) @@ flaky @@ timeout(5 second)
+
+  val env = EventLoopGroup.nio() ++ ChannelFactory.nio ++ ServerChannelFactory.nio ++ AppCollection.live
+
+  val clientApp: HttpApp[Any, Nothing] = Http.collectM[Request] {
```
tusharmath: Let's keep the `app` inside each test. This is exactly what we are doing in `ServerSpec`.

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,48 @@
...
+        val responseContent = response.requestBody()
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM("empty content") {
+        val actual          = Http.empty
+        val responseContent = actual.requestBody()
+        assertM(responseContent)(isEmpty)
+      } +
+      testM("text content") {
+        val actual = Http.response(Response(status = Status.OK, data = HttpData.fromText("zio user does not exist")))
```
tusharmath:
```suggestion
        val app = Http.text("zio user does not exist")
```

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,48 @@
...
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM("POST request expect non empty response content") {
+        val headers  = Headers.userAgent("zio-http test")
+        val response = Http.response(Response(status = Status.OK, data = HttpData.fromText("abc"), headers = headers))
+        val responseContent = response.requestBody()
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM("empty content") {
+        val actual          = Http.empty
```
tusharmath:
```suggestion
        val app          = Http.empty
```

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,48 @@
...
-  ).provideCustomLayer(env)
+    } +
+      testM("non empty content") {
+        val actual          = Http.response(Response(status = Status.OK, data = HttpData.fromText("abc")))
+        val responseContent = actual.requestBody()
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM("POST request expect non empty response content") {
+        val headers  = Headers.userAgent("zio-http test")
+        val response = Http.response(Response(status = Status.OK, data = HttpData.fromText("abc"), headers = headers))
```
tusharmath:
```suggestion
        val response = Http.text("abc").addHeaders(headers)
```

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,48 @@
...
-    },
-  ).provideCustomLayer(env)
+    } +
+      testM("non empty content") {
+        val actual          = Http.response(Response(status = Status.OK, data = HttpData.fromText("abc")))
+        val responseContent = actual.requestBody()
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM("POST request expect non empty response content") {
+        val headers  = Headers.userAgent("zio-http test")
```
tusharmath: why are headers important in this test?

gciuloaica: it is just a left out from previous version, where the initial call was made externally to Github and GitHub API requires to provide a valid user -agent  header. Maybe we should set up one by default in the zio-http client. For this change I will remove it and improve the testing.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 428836.47**

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,62 @@
...
+          val isExpectedHeaderValue = userAgentHeader.contains(ClientUserAgentValue)
+          if (isExpectedHeaderValue) req.getBodyAsString.map(text => Response.text(text))
+          else
+            Task.succeed(Response.status(Status.FORBIDDEN))
+
+        }
+
+        val userAgentHeader = Headers.userAgent(ClientUserAgentValue)
+        val response        = app.setMethod(Method.POST).addHeader(userAgentHeader)
+        val responseContent = response.requestBody()
```
tusharmath, line 45:
```suggestion
        val responseContent = response.getRequestHeader(Name.UserAgent)
```

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,62 @@
...
+        "POST request expect non empty response content from a server that requires a mandatory user agent header.",
+      ) {
+        val app = Http.text("test data") ++ Http.collectM[Request] { case req =>
+          val userAgentHeader       = req.getHeaderValue(Headers.Literals.Name.UserAgent)
+          val isExpectedHeaderValue = userAgentHeader.contains(ClientUserAgentValue)
+          if (isExpectedHeaderValue) req.getBodyAsString.map(text => Response.text(text))
+          else
+            Task.succeed(Response.status(Status.FORBIDDEN))
+
+        }
```
tusharmath: I think we can simplify the test —

In ZIO Http, we typically want to only have 2-3 steps while writing a unit test

```scala

// Step 1. Setup
val app = Http.collect[Request] { 
  case req => Response.text(req.getUserAgent.getOrElse(""))
}

// Step 2. Execute
val res = app.requestBodyAsString()

// Step 3. Assert
assertM(res)(equalTo(ClientUserAgentValue)
```

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,62 @@
...
+      val app = Http.ok.requestBodyAsString()
+      assertM(app)(anything)
+    } +
+      testM("non empty content") {
+        val app             = Http.text("abc")
+        val responseContent = app.requestBody()
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM(
+        "POST request expect non empty response content from a server that requires a mandatory user agent header.",
```
tusharmath: Is this documented in some spec? Can you provide a link to it here?

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,62 @@
...
+        val app = Http.text("test data") ++ Http.collectM[Request] { case req =>
+          val userAgentHeader       = req.getHeaderValue(Headers.Literals.Name.UserAgent)
+          val isExpectedHeaderValue = userAgentHeader.contains(ClientUserAgentValue)
+          if (isExpectedHeaderValue) req.getBodyAsString.map(text => Response.text(text))
+          else
+            Task.succeed(Response.status(Status.FORBIDDEN))
+
+        }
+
+        val userAgentHeader = Headers.userAgent(ClientUserAgentValue)
```
tusharmath: If use agent is required as per the spec, then it should be handled automatically within the client. We should still have a test for it.

gciuloaica: @tusharmath in the http spec https://httpwg.org/specs/rfc7231.html#header.user-agent, the user agent SHOULD be set in request header, however some REST APIs are rejecting a client request that is not providing a valid User Agent header. One of those APIs is GitHub API: https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required.
I think that it is a good practice to provide user agent request header as part of the client HTTP request, so we might just to add  a default one.

tusharmath: Got it. I think it would be nice to put a comment in the code about this requirement. `Client.scala` would be the best place.

**Side-Note:** I think the scope of this PR is changing quite a lot :) , I would rather just have the handler code merged and have the rest of the issues solved in a different PR.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 425555.59**

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,52 @@
...
+    }.provideCustomLayerShared(env) @@ timeout(5 seconds) @@ sequential
+  }
+  def clientSpec    = suite("ClientSpec") {
     testM("respond Ok") {
-      val actual = Client.request("http://api.github.com/users/zio/repos", ClientSSLOptions.DefaultSSL)
-      assertM(actual)(anything)
-    },
-  ).provideCustomLayer(env)
+      val app = Http.ok.requestBodyAsString()
+      assertM(app)(anything)
```
tusharmath: might want to asset on status, since the test says "OK"

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -1,16 +1,52 @@
...
+        val responseContent = app.requestBody()
+        assertM(responseContent)(isNonEmpty)
+      } +
+      testM(
+        "POST request expect non empty response content from a server.",
+      ) {
+        val res = Http
+          .text("ZIO user")
+          .setMethod(Method.POST)
+          .requestBodyAsString()
```
tusharmath:
```suggestion
.requestBodyAsString(method = POST, content = "ZIO User")
```

tusharmath:
```suggestion
.requestBodyAsString(method = POST, content = "ZIO User")
```

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 419393.24**

tusharmath (approved):

Issue #619, opened 2021-12-13
Make benchmarks run on forks


tusharmath:
**What is not easy to do right now?**
Currently, it's not possible to run benchmarks on forked repositories

**Describe the solution you'd like**
I would like it if the benchmarks could run on forked repos also.

**Why is this important?**
Contributors might propose changes that can affect performance both in a good or a bad way. At the moment it is impossible to identify the impact of such change.

amitksingh1490:
Due to permission restrictions on running github actions from fork, Any token provided will have only read access, Hence commenting on PR is not possible on forks
Reference: https://github.community/t/github-actions-are-severely-limited-on-prs/18179
However there is a workaround in which we can use events on `pull_request_target` instead of `pull_request`, but this is not supported currently in `sbt-github-actions`

amitksingh1490:
**Approach**

1. Event to be listened should be pull_request_target
2. Create a `zhttp-jar` from the fork branch commit.
3. Add jar path to `Techempower` repo's `build.sbt`.
4. Run Benchmarks.
5. Post Comment.

**Challenges**

1. Github actions restrict pull-requests from fork to do any write operation commenting is one of them, so we need to use `pull_request_target`
2. Commits from fork repository whole path is not accessible in github actions hence we need to create a jar.
3. Creating a jar causes issues in de-dupe so merge strategy needs to be defined.
4. `sbt-github-actions` don't have support for therefore need to pull_request_target write `yml` manually

tusharmath:
May be we can disable it completely on forks. The green check from the CI is something that all contributors love :)

amitksingh1490:
This is fixed in #820. It will skip the comment, benchmark result could be checked in the workflow logs

PR #645, opened 2021-12-16
Upgrade to ZIO 2

kitlangton:
@adamgfraser and I worked on upgrading to the RC of ZIO 2. Obviously, you may not want to merge this directly into master 😄 and perhaps would like to open up a new branch instead. Either way, just opening a PR for now for discussion (and CI!).

🥳


kitlangton:
Interesting. The `SSLSpec` passes locally.

tusharmath:
Thanks @kitlangton! I have created a separate branch `main-zio-2` which we will keep updating with `main`. Trying to figure out why the build is failing.

kitlangton:
INCREDIBLE!!!!!
CleanShot 2021-12-16 at 09 07 58@2x

kitlangton:
Getting weird timeouts on these client SSL specs :'( 
Will take a closer look this next week.

adamgfraser:
@tusharmath All tests are passing now. Looks like there is an issue with the benchmarks, not sure what is going on there.

It looked like in a couple of the SSL tests there was an issue where the channel could get closed before the fiber forked in `HttpRuntime#unsafeRun` completed execution. We addressed by implementing an `unsafeRunUninterruptible` operator that doesn't interrupt the fiber when the channel is closed but would be good to understand why that is occurring and if that is expected behavior.

Also blocked on STTP for the tests now. I wonder if there is a way to test with our own implementation instead?

tusharmath:
Hey @adamgfraser this looks great! Apologies for the conflicts, the main is moving fast. 


- I think the ssl tests are a bit flaky I guess we missed adding the flaky annotation in `main` #654 
- Benchmarks don't run on forked pull requests #619 
- STTP will need to be used until we close #650

Issue #616, opened 2021-12-13
Support Middlewares on `Http`


tusharmath:
**Is your feature request related to a problem? Please describe.**
Middleware is a powerful API but is not supported on `Http`. Currently, it's only supported on `HttpApp`, this is very limiting.

**Describe the solution you'd like**
Currently `Middleware` is only supported for `HttpApp` which restricts us from writing middlewares that work on any kind of inputs and outputs. 

Eg: I want to be able to write an `UHttp[GraphQLRequest, GraphQLResponse]` and use a middleware to convert it to a `UHttp[Request, UResponse]`.

```scala
val caliban: UHttp[GraphQLRequest, GraphQLResponse] = Http.collect[GraphQLRequest] {
  //...
}

val web: UHttp[Request, UResponse] = caliban @@ graphQLCodec
```

There are more interesting use cases that can be solved using this technique. We can write entity decoders like this — 

```scala

val user: UHttp[CreateUser, UserCreated] = Http.collectM[CreateUser] {
  case user => 
   uuid.map { id =>
     UserCreated(name = user.name, age = user.age, id = id)
   }
}

val web: UHttp[Request, UResponse] = user @@ entity(Codec.gen[CreateUser], Codec.gen[UserCreated])
```

**Describe alternatives you've considered**
We can always use operators such as `map` `mapM` `contramap` etc. However, it's not possible to reify a set of such operators without using the Middleware API.

PR #658, opened 2021-12-17
Refactor: remove HttpAppSyntax

### LINKED ISSUE #656: Use `IsHttpApp` as a constraint to apply HttpApp specific operators
Currently, there is an implicit conversion available between an `Http` and `HttpApp`. This is not exactly in line with how the ZIO ecosystem libraries are designed.
The main problem is that some of the operators are not discoverable unless and `Http` can be an `HttpApp`. This also causes a penalty in terms of performance.

Instead of using `HttpAppSyntax` to perform an implicit conversion, we can use a constrain type called `IsHttpApp`,
and then defined operators directly on `Http` such as `middleware` and `addHeader` etc. provided an `IsHttpApp` is available.

Eg:

```scala
sealed trait Http[-R, +E, -A, +B] {
  // ... 
  def addHeader(name: String, value: String)(implicit ev: IsHttpApp[R, E, A, B]): HttpApp[R, E] = ???
  // ...
}
```

**Working in _Scala 3_**

https://scastie.scala-lang.org/Q1NRJTrYTFm9Kkt9c3izxQ

```scala
trait IsHttpApp[+R, -E, +A, -B] {
  type ROut
  type EOut
}

object IsHttpApp {
  type Aux[R, E] = IsHttpApp[R, E, Request, Response[R, E]] {
    type ROut = R
    type EOut = E
  }
  
  implicit def make[R, E]: Aux[R, E] = new IsHttpApp[R, E, Request, Response[R, E]] {
    override type ROut = R
    override type EOut = E
  } 
}
```

But this solution doesn't work in Scala 2.

tusharmath:
Fixes #656 (partially)


amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 433886.31**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 428251.63**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 442587.17**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 433406.48**

zio-http/src/main/scala-2/zhttp/http/IsRequest.scala
```diff
@@ -0,0 +1,16 @@
+package zhttp.http
+
+import scala.annotation.implicitNotFound
+import scala.language.implicitConversions
+
+@implicitNotFound(
+  "This operation is only allowed if the input is a type of Request. " +
+    "However your Http instance accepts the type ${A}, on which this operation can not be applied.",
```
amitksingh1490, line 8: this will fail the scaladoc generation during publish, Please refer this issue https://github.com/scala/bug/issues/10333

zio-http/src/test/scala/zhttp/middleware/MiddlewareSpec.scala
```diff
@@ -50,13 +50,13 @@ object MiddlewareSpec extends DefaultRunnableSpec with HttpAppTestExtensions {
         assertM(program)(equalTo(Vector("200 GET /health 1000ms\n")))
       } +
         testM("log 404 status method url and time") {
-          val program = run(Http.empty @@ debug.withEmpty) *> TestConsole.output
+          val program = run(Http.empty.as(Response.ok) @@ debug.withEmpty) *> TestConsole.output
```
amitksingh1490, line 53: this is a breaking change. Is there a way we can support `Http.empty @@ debug.withEmpty` as well?

tusharmath: Agree. Need to think more about this. Ideally it should get fixed with #616.

tusharmath:
The PR is still broken. It doesn't correctly infer the types in this case -
```scala
Http[Console, ClassCastException, Request, Response[Random, IndexOutOfBoundsException]]
```

The error type inferred isn't correct.

tusharmath:
This doesn't work 😭

PR #669, opened 2021-12-21
Enhance/request setbody

### LINKED ISSUE #614: Request: Add operator to replace the requestBody
There is no operator in on `Request` which can help me update the request body.

Currently `Request` has operators to update `method`, `url` and `headers` but for updating the body there is no Operator for updating RequestBody.
we could add an operator
```scala
def setBody(body: HttpData[R, E]): Request = ???
```

We could create a new request for updating the body but we have to pass `headers` `method` and `url` even if we don't want any change in that.

sumawa:
To resolve #614


amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 438940.10**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 442896.54**

PR #672, opened 2021-12-21
fix HttpData unsafeSize for unicode charset

### LINKED ISSUE #641: Encoding problem when using HttpData.fromText and multi byte characters
HttpData.unsafeSize for Text returns number of characters and not bytes. This results in a bad content length and a broken response when using multi byte characters like e.g. `åäö`.

Current work around:

Instead of
`HttpData.fromText("åäö")` or `Response.text("åäö")`

do 
`HttpData.fromByteBuf(Unpooled.copiedBuffer("åäö", HTTP_CHARSET))`

Version: 1.0.0.0-RC18

sumawa:
To resolve #641


amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 441696.76**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 434217.24**

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -54,7 +54,7 @@ sealed trait HttpData[-R, +E] { self =>
    */
   private[zhttp] def unsafeSize: Long = self match {
     case HttpData.Empty               => 0L
-    case HttpData.Text(text, _)       => text.length.toLong
+    case HttpData.Text(text, charSet) => Unpooled.copiedBuffer(text, charSet).readableBytes().toLong
```
tusharmath: this is a very expensive way to compute size.

sumawa: Right. Creating new buffer is expensive. I will re-look.

sumawa: Done.

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 438773.07**

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -54,7 +54,7 @@ sealed trait HttpData[-R, +E] { self =>
    */
   private[zhttp] def unsafeSize: Long = self match {
     case HttpData.Empty               => 0L
-    case HttpData.Text(text, _)       => text.length.toLong
+    case HttpData.Text(text, charSet) => text.getBytes(charSet).size.toLong
     case HttpData.BinaryChunk(data)   => data.size.toLong
```
tusharmath: We can drop `Text` and `BinaryChunk`

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -54,7 +54,7 @@ sealed trait HttpData[-R, +E] { self =>
...
     case HttpData.BinaryByteBuf(data) => data.readableBytes().toLong
```
tusharmath, line 59:
```suggestion
    case HttpData.Complete(data) => data.readableBytes().toLong
```

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -115,6 +115,11 @@ object ServerSpec extends HttpRunnableSpec(8088) {
         val path = getClass.getResource("/TestFile").getPath
         val res  = Http.data(HttpData.fromStream(ZStream.fromFile(Paths.get(path)))).requestBodyAsString()
         assertM(res)(containsString("foo"))
+      } +
+      testM("Response Content length for Unicode Charset bytes") {
+        val app           = Http.text("åäö")
```
tusharmath: Use a Gen to generate Unicode values.

sumawa: Done. Using Gen.anyString  which internally use anyUnicodeChar

```
  def anyString: Gen[Random with Sized, String] =
    Gen.string(Gen.anyUnicodeChar)
```

zio-http/src/main/scala/zhttp/http/HttpData.scala
sumawa: Done. I assumed we are renaming BinaryByteBuf to Complete everywhere.

zio-http/src/main/scala/zhttp/http/HttpData.scala
sumawa: I have kept HttpData.Text in the end 
```
  private[zhttp] def unsafeSize: Long = self match {
    case HttpData.Empty               => 0L
    case HttpData.Complete(data)      => data.readableBytes().toLong
    case HttpData.BinaryStream(_)     => -1L
    case HttpData.Text(text, charset) => text.getBytes(charset).size.toLong
    case _                            => -1L
```
As there is one test failing in EncodeClientParamsSpec
```
      testM("content-length") {
        check(clientParamWithFiniteData) { params =>
          val req = encodeClientParams(HttpVersion.HTTP_1_1, params)
          assert(req.headers().getInt(HttpHeaderNames.CONTENT_LENGTH).toLong)(equalTo(params.data.unsafeSize))
        }
      }
```
Looking into it.

sumawa: Done.

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -118,7 +117,7 @@ object HttpData {
 
   private[zhttp] final case class Text(text: String, charset: Charset) extends HttpData[Any, Nothing] with Cached
```
tusharmath: ~Delete `Text` `BinaryChunk`~

tusharmath: We can change the approach a bit — Instead of adding any default headers via the `Response` constructors we could add them via the handler directly.

**Advantages**
1. You can be a bit more intelligent about what headers should be set.
2. You get the final response, ie. after all the middlewares etc. So you can make better assumptions about content-length and transfer-encoding etc.
3. For sending files (later) we can use the same technique, ie create a chunked writer instance and use that to write a file directly. 
4. We can do better memoization this way, specially for files.

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -54,9 +54,9 @@ sealed trait HttpData[-R, +E] { self =>
    */
   private[zhttp] def unsafeSize: Long = self match {
     case HttpData.Empty               => 0L
-    case HttpData.Text(text, _)       => text.length.toLong
+    case HttpData.Text(text, charset) => text.getBytes(charset).size.toLong
     case HttpData.BinaryChunk(data)   => data.size.toLong
```
sumawa, line 58: Restored Text and BinaryChunk. As per test cases only one code path uses this (EncodeClientParamsSpec.scala) . 
The regular Request => Response flow doesn't compute text bytes using this. It is handled in post response processing in Handler.

zio-http/src/main/scala/zhttp/http/HttpData.scala
sumawa: Done.
Also refactored Handler a bit, as there was code duplication.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -106,87 +107,84 @@ private[zhttp] final case class Handler[R](
...
-              )
-          }
-        }
+        processResponse(jReq, res)
+      case HExit.Failure(e)   =>
+        processErrorResponse(jReq, Some(e))
+      case HExit.Empty        =>
+        processErrorResponse(jReq, None)
+    }
+
```
sumawa: Had to refactor as there was the same (success / error) response  flow with just one difference of effect and non effect computation.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -106,87 +107,84 @@ private[zhttp] final case class Handler[R](
...
+      self.upgradeToWebSocket(ctx, jReq, res)
+    } else {
+      val responseWithHeadersAndByteBuf = attachHeadersWithComputedBuf(res)
+      unsafeWriteAnyResponse(responseWithHeadersAndByteBuf)
+
+      responseWithHeadersAndByteBuf.data match {
+        case HttpData.Empty =>
+          unsafeWriteAndFlushLastEmptyContent()
+
+        case data @ HttpData.Text(_, _) =>
```
sumawa: This path ```data @ HttpData.Text(_, _)``` along with ```case data @ HttpData.BinaryChunk(_) ``` will probably not be reached, so I will remove it.

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -54,9 +54,9 @@ sealed trait HttpData[-R, +E] { self =>
...
-    case HttpData.BinaryByteBuf(data) => data.readableBytes().toLong
```
tusharmath, line 59: Can revert this to BinaryByteBuf

zio-http/src/main/scala/zhttp/service/Handler.scala
tusharmath: Don't allocate additional `Some` along the way. Slowly such allocations start affecting performance quite badly.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -106,87 +107,84 @@ private[zhttp] final case class Handler[R](
...
+    val size         = modifiedData.unsafeSize
+    val isChunked    = modifiedData.isChunked
+    Response(
+      res.status,
+      res.headers ++
+        Headers(Name.ContentLength -> size.toString).when(size >= 0) ++
+        Headers(Name.TransferEncoding -> Value.Chunked).when(isChunked),
+      modifiedData,
+      res.attribute,
+    )
```
tusharmath: Again a lot more allocations happening here unnecessarily. I think this too will affect the performance of the system. I would recommend in this PR keep the implementation exactly as it was, except the content-length problem. 

Also I observed, there aren't enough tests asserting for `content-length` .

sumawa: Sure, will add asserts as well.

zio-http/src/main/scala/zhttp/http/HttpData.scala
sumawa: Reverted

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -67,14 +67,9 @@ object Response {
...
   ): Response[R, E] = {
-    val size      = data.unsafeSize
-    val isChunked = data.isChunked
-
     Response(
       status,
-      headers ++
-        Headers(Name.ContentLength -> size.toString).when(size >= 0) ++
-        Headers(Name.TransferEncoding -> Value.Chunked).when(isChunked),
+      headers,
```
sumawa: All headers are being attached in the Handler

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -119,29 +120,70 @@ private[zhttp] final case class Handler[R](
...
                     case HttpData.Empty =>
                       UIO {
+                        val size = res.data.unsafeSize
+                        unsafeWriteAnyResponse(
+                          res.updateHeaders(
+                            _ ++
+                              Headers(Name.ContentLength -> size.toString).when(size >= 0)
+                              ++ Headers(Name.TransferEncoding -> Value.Chunked).when(res.data.isChunked),
+                          ),
+                        )
```
sumawa: For each case we are adding headers individually.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -119,29 +120,70 @@ private[zhttp] final case class Handler[R](
...
+                        val byteBuf = data.encodeAndCache(res.attribute.memoize)
+                        val size    = byteBuf.readableBytes().toLong
+                        unsafeWriteAnyResponse(
+                          res.updateHeaders(
+                            _ ++
+                              Headers(Name.ContentLength -> size.toString).when(size >= 0)
+                              ++ Headers(Name.TransferEncoding -> Value.Chunked).when(res.data.isChunked),
+                          ),
+                        )
+                        unsafeWriteLastContent(byteBuf)
```
sumawa: Only for Text and BinaryChunk headers are computed via ByteBuf returned by encodeAndCache

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -119,29 +120,70 @@ private[zhttp] final case class Handler[R](
...
```
sumawa: Only for Text and BinaryChunk headers are computed via ByteBuf returned by encodeAndCache

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -154,26 +196,65 @@ private[zhttp] final case class Handler[R](
...
-
           res.data match {
             case HttpData.Empty =>
+              val size = res.data.unsafeSize
+              unsafeWriteAnyResponse(
+                res.updateHeaders(
+                  _ ++
+                    Headers(Name.ContentLength -> size.toString).when(size >= 0)
+                    ++ Headers(Name.TransferEncoding -> Value.Chunked).when(res.data.isChunked),
+                ),
```
amitksingh1490: Code duplication

sumawa: Done, added another method for update response headers

sumawa:
Fixed by #711

PR #680, opened 2021-12-23
`content-length` header should be ignored in `HttpData.fromStream`

### LINKED ISSUE #679: Transfer-encoding is set to chunked even if content-length is known
The following code:

```scala
import zio._
import zhttp.http._
import zhttp.service.Server
import zio.stream.ZStream

object Test extends App {
  val app = Http.collect[Request] { case _ =>
    Response(
      status = Status.OK,
      headers = List(Header.contentLength(4)),
      data = HttpData.fromStream(ZStream.fromIterable("test".getBytes))
    )
  }

    Server.start(8090, app).exitCode
}
```

returns the following response:

```
$ curl --raw -v http://localhost:8090

< HTTP/1.1 200 OK
< content-length: 4
< transfer-encoding: chunked
<
4
test
0
```

Which is incorrect as the content is longer (due to the chunks delimiters). If the length is known, I think a normal response should be returned.

I'm using version `1.0.0.0-RC19`.

amitksingh1490:
to resolve #679


amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 433694.14**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 434933.81**

zio-http/src/main/scala/zhttp/http/Headers.scala
```diff
@@ -21,6 +21,8 @@ final case class Headers(toChunk: Chunk[Header]) extends HeaderExtension[Headers
 
   override def updateHeaders(f: Headers => Headers): Headers = f(self)
 
+  def filter(f: Header => Boolean): Headers = Headers(self.toChunk.filter(f(_)))
```
tusharmath, line 24: - We can add this to `HeadersModifiers`
- For this PR we don't need this function it. You could have use `removeHeader(Name.ContentLength)`

zio-http/src/test/scala/zhttp/http/ResponseSpec.scala
```diff
@@ -0,0 +1,30 @@
+package zhttp.http
+import zhttp.http.Headers.Literals.Name
+import zhttp.internal.HttpGen
+import zio.test.Assertion._
+import zio.test._
+object ResponseSpec extends DefaultRunnableSpec {
```
tusharmath: can keep these headers inside of `ServerSpec`

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -67,12 +67,12 @@ object Response {
     headers: Headers = Headers.empty,
     data: HttpData[R, E] = HttpData.Empty,
   ): Response[R, E] = {
-    val size      = data.unsafeSize
-    val isChunked = data.isChunked
-
+    val size         = data.unsafeSize
+    val isChunked    = data.isChunked
+    val smartHeaders = if (isChunked) headers.filter(h => h._1 != Name.ContentLength) else headers
```
tusharmath: Actually `TransferEncoding: Chunked` doesn't work this way. We should use `ChunkedWriter` to support this capability. The protocol is quite different than what we have been doing.

PR #682, opened 2021-12-23
Feature: Return `Start` on server start

### LINKED ISSUE #633: Return the server start on starting the server
When `0` is provided as the port value, the server will bind to any available free port. There is no way to get what port the server is using at the moment.

I would like it if the `start` method on the server could return an `Int` representing the port the server is using.

There is no alternative way to get the port info.

**Additional context**
ServerBootstrap has an API to extract the port information.

girdharshubham:
Closes #633


amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 436423.14**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 426759.21**

amitksingh1490:
## 🚀🚀🚀 Benchmark Results 
 **Concurrency: 256** 
 **Requests/sec: 436891.79**

zio-http/src/main/scala/zhttp/service/ServerStart.scala
```diff
@@ -0,0 +1,31 @@
+package zhttp.service
+
+sealed trait ServerStart { self =>
```
tusharmath: This isn't a domain, its data. You can simply use a product type here. ie. just have a case class called `Start(port: Int)`. That should do the job for now. Later we can see, if there is something else that we would like to add.

tusharmath: Add scala doc for this class.

zio-http/src/test/scala/zhttp/internal/AppPort.scala
```diff
@@ -0,0 +1,24 @@
...
+    .toLayer
+
+  case class Live(ref: Ref[Int]) extends Service {
+    override def set(n: Int): IO[Nothing, Unit] = ref.set(n)
+
+    override def get: IO[Nothing, Int] = ref.get
+  }
+
+  def set(n: Int): ZIO[AppPort, Nothing, Unit] = ZIO.accessM(_.get.set(n))
+  def get: ZIO[AppPort, Nothing, Int]          = ZIO.accessM[AppPort](_.get.get)
```
tusharmath: This is dangerous. The call to `get` might return `0` if not already set. I would prefer using a promise instead of a Ref over here.

Also - U can move this logic to `HttpAppCollection` class. We can rename that class to something else.

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -124,10 +127,13 @@ object Server {
...
       serverBootstrap = new ServerBootstrap().channelFactory(channelFactory).group(eventLoopGroup)
-      _ <- ChannelFuture.asManaged(serverBootstrap.childHandler(init).bind(settings.address))
-
+      chf  <- ZManaged.effect(serverBootstrap.childHandler(init).bind(settings.address))
+      _    <- ChannelFuture.asManaged(chf)
+      port <- ZManaged.effect(chf.channel().localAddress().asInstanceOf[InetSocketAddress].getPort)
+      serverSettings = ServerStart.port(port)
     } yield {
       ResourceLeakDetector.setLevel(settings.leakDetectionLevel.jResourceLeakDetectionLevel)
+      serverSettings()
```
tusharmath:
```suggestion
      Start(port = port)
```

zio-http/src/main/scala/zhttp/service/ServerStart.scala
girdharshubham: Got it. I thought of it as a sum type, which currently has only port but could also have other types.
Ref: https://en.m.wikipedia.org/wiki/Tagged_union

zio-http/src/test/scala/zhttp/internal/AppPort.scala
girdharshubham: But won't ref ensure an atomic operation in the over-all effect? What I mean is, the first operation effectively defines a server start up(sets up the value) and then subsequent operations get the value.
Won't set take precedence in this case everytime because of how ref is designed?

zio-http/src/main/scala/zhttp/service/ServerStart.scala
girdharshubham: If we are sticking with a product type for now, we should definitely define it within Server.scala. I'll move Start there and add the doc.

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -92,11 +99,12 @@ object Server {
   def start[R <: Has[_]](
     port: Int,
     http: HttpApp[R, Throwable],
-  ): ZIO[R, Throwable, Nothing] =
+  ): ZIO[R, Throwable, Nothing] = {
```
tusharmath, line 100:
```suggestion
  ): ZIO[R, Throwable, Start] = {
```

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -56,6 +58,11 @@ object Server {
     flowControl: Boolean = false,
   )
 
+  /**
+   * Holds server bootstrap information.
+   */
+  private[zhttp] final case class Start(port: Int = 0)
```
tusharmath:
```suggestion
  final case class Start(port: Int = 0)
```

zio-http/src/test/scala/zhttp/internal/AppCollection.scala
```diff
@@ -30,9 +30,13 @@ object AppCollection {
...
 
+  def setPort(n: Int): ZIO[HttpAppCollection, Nothing, Boolean] = ZIO.accessM[HttpAppCollection](_.get.setPort(n))
+  def getPort: ZIO[HttpAppCollection, Nothing, Int]             = ZIO.accessM[HttpAppCollection](_.get.getPort)
+
   def live: ZLayer[Any, Nothing, HttpAppCollection] = Ref
     .make(Map.empty[Id, HttpApp[HttpEnv, Throwable]])
-    .map(ref => new Live(ref))
+    .zip(Promise.make[Nothing, Int])
+    .map(a => new Live(a._1, a._2))
     .toLayer
```
tusharmath: Use a `for` expression for better readability.

zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -17,41 +17,54 @@ import zio.{Chunk, Has, Task, ZIO, ZManaged}
...
     path: Path = !!,
     headers: Headers = Headers.empty,
-  ): ZIO[SttpClient, Throwable, SResponse[Either[String, WebSocket[Task]]]] = {
+  ): ZIO[SttpClient with HttpAppCollection, Throwable, SResponse[Either[String, WebSocket[Task]]]] = {
     // todo: uri should be created by using URL().asString but currently support for ws Scheme is missing
-    val url                       = s"ws://localhost:$port${path.asString}"
-    val headerConv: List[SHeader] = headers.toList.map(h => SHeader(h._1.toString(), h._2.toString()))
-    send(basicRequest.get(uri"$url").copy(headers = headerConv).response(asWebSocketUnsafe))
+    for {
+      port <- ZIO.accessM[HttpAppCollection](_.get.getPort)
```
tusharmath:
```suggestion
      port <- HttpAppCollection.port
```

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -31,7 +31,9 @@ sealed trait Server[-R, +E] { self =>
     case ConsolidateFlush     => s.copy(consolidateFlush = true)
   }
 
-  def make(implicit ev: E <:< Throwable): ZManaged[R with EventLoopGroup with ServerChannelFactory, Throwable, Unit] =
+  def make(implicit
+    ev: E <:< Throwable,
+  ): ZManaged[R with EventLoopGroup with ServerChannelFactory, Throwable, Start] =
```
tusharmath, line 36: Needs unit tests :P

zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
girdharshubham: Used AppCollection.getPort

zio-http/src/main/scala/zhttp/service/Server.scala
girdharshubham: @tusharmath it's already supported in a way in the make method.

example/src/main/scala/example/HelloWorldAdvanced.scala
```diff
@@ -32,9 +32,9 @@ object HelloWorldAdvanced extends App {
 
     // Create a new server
     server.make
-      .use(_ =>
+      .use(bootstrap =>
```
girdharshubham: @tusharmath @amitksingh1490  added this just as an example of how we are supporting this feature.

original comment: https://github.com/dream11/zio-http/pull/682/files#r776072140

tusharmath:
```suggestion
      .use(start =>
```

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -56,6 +58,11 @@ object Server {
     flowControl: Boolean = false,
   )
 
+  /**
+   * Holds server bootstrap information.
+   */
+  final case class Bootstrap(port: Int = 0)
```
tusharmath: Bootstrap is confusing with ServerBootstrap, no? What was wrong with `Start`?

girdharshubham: ~I like `Start` as well but, to me, it doesn't convey the whole intent - if it makes sense.
Maybe we could make it `BootstrapConfig`?~

Start, it is. 💯

zio-http/src/main/scala/zhttp/service/Server.scala
girdharshubham: @tusharmath I ran an scoverage report. We are already covering make using HttpRunnableSpec. I think what we should be doing instead is writing unit test cases for different server settings.

zio-http/src/test/scala/zhttp/internal/DynamicServer.scala
```diff
@@ -1,16 +1,16 @@
...
 import zhttp.http._
+import zio._
 import zio.blocking.Blocking
 import zio.console.Console
-import zio.{Ref, UIO, ZIO, ZLayer}
 
 import java.util.UUID
 
-object AppCollection {
+object DynamicServer {
```
tusharmath, line 11:
```suggestion
object ReusableServer {
```

An alternative name to consider (Optional)

zio-http/src/test/scala/zhttp/internal/DynamicServer.scala
```diff
@@ -27,29 +27,38 @@ object AppCollection {
...
 
   sealed trait Service {
     def add(app: HttpApp[HttpEnv, Throwable]): UIO[Id]
     def get(id: Id): UIO[Option[HttpApp[HttpEnv, Throwable]]]
+    def setPort(n: Int): UIO[Boolean]
+    def getPort: IO[Nothing, Int]
   }
 
-  final class Live(ref: Ref[Map[Id, HttpApp[HttpEnv, Throwable]]]) extends Service {
+  final class Live(ref: Ref[Map[Id, HttpApp[HttpEnv, Throwable]]], pr: Promise[Nothing, Int]) extends Service {
```
tusharmath:
```suggestion
  final class Live(ref: Ref[Map[Id, HttpApp[HttpEnv, Throwable]]], pr: Promise[Nothing, Start]) extends Service {
```

zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -18,47 +18,59 @@ import zio.{Chunk, Has, Task, ZIO, ZManaged}
...
  * actual Http server and makes requests.
  */
-abstract class HttpRunnableSpec(port: Int) extends DefaultRunnableSpec { self =>
+abstract class HttpRunnableSpec extends DefaultRunnableSpec { self =>
+  def serve[R <: Has[_]](
+    app: HttpApp[R, Throwable],
+  ): ZManaged[R with EventLoopGroup with ServerChannelFactory with DynamicServer, Nothing, Unit] =
+    for {
+      start <- Server.make(Server.app(app) ++ Server.port(0) ++ Server.paranoidLeakDetection).orDie
+      _     <- ZIO.accessM[DynamicServer](_.get.setPort(start.port)).toManaged_
```
tusharmath:
```suggestion
      _     <- DynamicServer.setPort(start.port).toManaged_
```

girdharshubham: Removed setPort and added setStart.

zio-http/src/test/scala/zhttp/internal/DynamicServer.scala
girdharshubham: I think DynamicServer is better. 😅

girdharshubham:
Ran it by @tusharmath, we need to cover tests for server start as well. It will be covered in another PR

zio-http/src/main/scala/zhttp/service/Server.scala
girdharshubham: Done.

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -204,10 +208,21 @@ object ServerSpec extends HttpRunnableSpec(8088) {
...
     }.provideCustomLayerShared(env) @@ timeout(30 seconds)
+
+  def serverStartSpec = suite("ServerStartSpec") {
+    testM("make") {
+      (Server.port(0) ++ Server.app(Http.ok)).make.use { start =>
+        val actual = Client
+          .request(s"http://localhost:${start.port}")
+          .map(_.status)
+
+        assertM(actual)(equalTo(Status.OK))
```
tusharmath: - [x] Test should assert if the port is set. 
- [x] Test should assert that if no port is provided, a random port is assigned and returned.

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -204,10 +208,23 @@ object ServerSpec extends HttpRunnableSpec(8088) {
...
 
-  override def spec = {
+  override def spec =
     suiteM("Server") {
-      app.as(List(staticAppSpec, dynamicAppSpec, responseSpec, requestSpec)).useNow
+      app.as(List(serverStartSpec, staticAppSpec, dynamicAppSpec, responseSpec, requestSpec)).useNow
     }.provideCustomLayerShared(env) @@ timeout(30 seconds)
+
+  def serverStartSpec = suite("ServerStartSpec") {
+    testM("make: port set") {
```
girdharshubham:
```suggestion
testM("make set port")
````
Remove colon

File: zio-http/src/test/scala/zhttp/service/ServerSpec.scala [hunk 16]
```diff
@@ -204,10 +208,23 @@ object ServerSpec extends HttpRunnableSpec(8088) {
...
+      val port = 8088
+      (Server.port(port) ++ Server.app(Http.empty)).make.use { start =>
+        assertM(ZIO.effect(start.port))(equalTo(port))
+      }
+    } +
+      testM("make: random port") {
```
tusharmath:
```suggestion
      testM("available port") {
```

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
tusharmath:
```suggestion
    testM("desired port") {
```

tusharmath (approved):

PR #711, opened 2021-12-28
Fix: Content-Length issues

### LINKED ISSUE #641: Encoding problem when using HttpData.fromText and multi byte characters

tusharmath:
fixes #641 #679

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/Handler.scala
+++ b/zio-http/src/main/scala/zhttp/service/Handler.scala
@@ -49,16 +49,26 @@ private[zhttp] final case class Handler[R](
    * response. It will also set the server time if requested by the client.
    */
   private def encodeResponse(res: Response[_, _]): HttpResponse = {
+
     val jResponse = res.attribute.encoded match {
+
+      // Check if the encoded response exists and/or was modified.
       case Some((oRes, jResponse)) if oRes == res =>
         jResponse match {
-          case response: FullHttpResponse => response.retainedDuplicate()
+
+          // Duplicate the response without allocating much memory
+          case response: FullHttpResponse =>
+            // TODO: use response.retainedDuplicate()
+
+            response.content().resetReaderIndex()
+            response.retain()
           case response                   => response
         }
 
       case _ => res.unsafeEncode()
     }
 
+    // Identify if the server time should be set and update if required.
     if (res.attribute.serverTime) jResponse.headers().set(HttpHeaderNames.DATE, serverTime.refreshAndGet())
     jResponse
   }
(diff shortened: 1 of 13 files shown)


zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -157,13 +159,25 @@ object ServerSpec extends HttpRunnableSpec(8088) {
...
+            }
+        }
       }
   }
 
   override def spec = {
     suiteM("Server") {
       app.as(List(staticAppSpec, dynamicAppSpec, responseSpec, requestSpec)).useNow
-    }.provideCustomLayerShared(env) @@ timeout(30 seconds) @@ sequential
+    }.provideCustomLayerShared(env) @@ timeout(3000 seconds) @@ sequential
```
amitksingh1490: Do we need this change?

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -154,29 +138,10 @@ private[zhttp] final case class Handler[R](
...
-
-            case data @ HttpData.BinaryChunk(_) =>
-              unsafeWriteLastContent(data.encodeAndCache(res.attribute.memoize))
-              releaseRequest(jReq)
-
-            case HttpData.BinaryStream(stream) =>
-              unsafeRunZIO(
-                writeStreamContent(stream) *> Task(releaseRequest(jReq)),
-              )
+            case HttpData.BinaryStream(stream) => unsafeRunZIO(writeStreamContent(stream) *> Task(releaseRequest(jReq)))
```
amitksingh1490, line 143: This might introduce a memory leak when writeStreamContent fails `releaseRequest(jReq)` might not get called. But this is still there in main, We can ignore this for now, but can we  add a todo here.

tusharmath: Yes, you are right. I think this issue is there even in the current main.

amitksingh1490 (approved):

zio-http/src/main/scala/zhttp/service/Handler.scala
amitksingh1490: I will raise an separate issue for this

amitksingh1490 (approved):

PR #716, opened 2021-12-29
chore: ignore ClientHttpsSpec

dinojohn:
Ignoring the flaky ClientHttpsSpec.


tusharmath (approved):

lachezar:
I have tried having a unit test in a project of my own that makes a request to an https url and then I get strange Netty errors (NoRouteToHostException). Is this test ignored for the same reason... is there a way to have a client making an https request in a ZIO unit test?

PR #731, opened 2021-12-29
Performance: Improve benchmarking code

tusharmath:


zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
```diff
@@ -30,11 +30,24 @@ final case class ServerChannelInitializer[R](
...
     // ServerCodec
     // Always add ServerCodec
-    pipeline.addLast(HTTP_SERVER_CODEC, new HttpServerCodec()) // TODO: See if server codec is really required
+    pipeline.addLast(
+      "decoder",
+      new HttpRequestDecoder(
+        DEFAULT_MAX_INITIAL_LINE_LENGTH,
+        DEFAULT_MAX_HEADER_SIZE,
+        DEFAULT_MAX_CHUNK_SIZE,
+        false,
```
amitksingh1490: should be taken from server config, defaulting to given these values

amitksingh1490:
try not adding HttpObjectAggregator in benchmarks.

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
tusharmath: for now I think it's alright.

amitksingh1490 (approved):

PR #751, opened 2022-01-03
Feature: Signed Cookie

### LINKED ISSUE #419: Allow creation of signed cookie

ShrutiVerma97:
fixes #419


example/src/main/scala/example/SignCookies.scala
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Example to make app using signed-cookies
+ */
+object SignCookies extends App {
+
+  // Setting cookies with an expiry of 5 days
+  private val cookie = Cookie("key", "hello") @@ maxAge(5 days)
+
+  private val app = Http.collect[Request] { case Method.GET -> !! / "cookie" =>
+    Response.ok.addCookie(cookie @@ path(!! / "cookie") @@ httpOnly @@ sign("secret") withContent "world")
```
ShrutiVerma97: https://github.com/dream11/zio-http/pull/533/files#r773066949

How about this approach?

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -31,7 +36,10 @@ final case class Cookie(
   /**
    * Sets content in cookie
    */
-  def withContent(v: String): Cookie = copy(content = v)
+  def withContent(v: String): Cookie = self.secret match {
```
tusharmath: - drop the `@@` feature for cookies.

```scala
val cookie = Cookie("foo", "bar").withContent("bar0")
val cookieSigned = cookie.withSecret("secret")

// Problem 1
println(cookieSigned.content) // Actual: "bar.01010101" Expected: "bar"

// Problem 2
cookie.withSecret("secret") != cookie.withSecret("secret").withSecret("secret")
- `withSecret` should be idempotent.

```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -73,6 +78,13 @@ final case class Cookie(
    */
   def withSameSite(v: Cookie.SameSite): Cookie = copy(sameSite = Some(v))
 
+  /**
+   * Adds secret in the cookie
+   */
+  def withSecret(key: String): Cookie = {
```
```suggestion
  def sign(secret: String): Cookie = {
```

Similarly we should also have `unsign`

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -125,6 +142,34 @@ final case class Cookie(
     cookie.flatten.mkString("; ")
   }
 
+  /**
+   * Signs cookie content with a secret
+   */
+  def sign(secret: String, content: String): String = {
+    try {
```
tusharmath: remove try

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -125,6 +142,34 @@ final case class Cookie(
     cookie.flatten.mkString("; ")
   }
 
+  /**
+   * Signs cookie content with a secret
+   */
+  def sign(secret: String, content: String): String = {
```
tusharmath: keep encode/decode methods private

zio-http/src/main/scala/zhttp/http/Cookie.scala
ShrutiVerma97: `getInstance` and `doFinal` throws exceptions such as `NoSuchAlgorithmException` or `IllegalStateException`

tusharmath: Is there a way we can avoid it? Like guarantee that the algorithm exists?
May be there is something like — `availableAlgorithms` ?

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -120,6 +140,34 @@ final case class Cookie(
     cookie.flatten.mkString("; ")
   }
 
+  /**
+   * Signs cookie content with a secret
+   */
+  private def sign(secret: String, content: String): String = {
```
```suggestion
  private def signContent(secret: String, content: String): String = {
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -68,6 +73,16 @@ final case class Cookie(
    */
   def withSameSite(v: Cookie.SameSite): Cookie = copy(sameSite = Some(v))
 
+  /**
+   * Adds secret in the cookie
```
```suggestion
   * Signs the cookie at the time of encoding using the provided secret.
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -120,6 +140,34 @@ final case class Cookie(
...
+  private def sign(secret: String, content: String): String = {
+    try {
+      val sha256    = Mac.getInstance("HmacSHA256")
+      val secretKey = new SecretKeySpec(secret.getBytes(), "RSA")
+      sha256.init(secretKey)
+      val signed    = sha256.doFinal(content.getBytes())
+      val mda       = MessageDigest.getInstance("SHA-512")
+      content + '.' + getEncoder.encodeToString(mda.digest(signed))
+    } catch {
+      case _: Exception => content
```
tusharmath: Returning the same content is quite dangerous and difficult to debug. I would rather have a failure on the server this way.

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -237,6 +238,22 @@ object Middleware {
...
+   */
+  def signCookies[R, E](secret: String): Middleware[R, E] =
+    patch((_, _) =>
+      Patch.updateHeaders(h => {
+        Headers(
+          h.toList.collect { case h if contentEqualsIgnoreCase(h._1, HeaderNames.setCookie) => h._2.toString }
+            .map(Cookie.decodeResponseCookie)
+            .collect { case Some(cookie) =>
+              (HeaderNames.setCookie, cookie.sign(secret).encode)
+            },
```
tusharmath: can we do all of this in one `collect`?

zio-http/src/main/scala/zhttp/http/Patch.scala
```diff
@@ -29,11 +30,13 @@ object Patch {
...
   final case class Combine(left: Patch, right: Patch)   extends Patch
+  final case class UpdateHeaders(f: Headers => Headers) extends Patch
 
   def empty: Patch                                              = Empty
   def addHeader(headers: Headers): Patch                        = AddHeaders(headers)
   def addHeader(headers: Header): Patch                         = AddHeaders(Headers(headers))
   def addHeader(name: CharSequence, value: CharSequence): Patch = AddHeaders(Headers(name, value))
   def removeHeaders(headers: List[String]): Patch               = RemoveHeaders(headers)
   def setStatus(status: Status): Patch                          = SetStatus(status)
+  def updateHeaders(f: Headers => Headers): Patch               = UpdateHeaders(f)
```
tusharmath: Can simply pull these changes in a separate PR. It will also get automatically documented in the release notes.

zio-http/src/main/scala/zhttp/http/headers/HeaderGetters.scala
```diff
@@ -277,10 +277,15 @@ trait HeaderGetters[+A] { self =>
     getHeaderValue(HeaderNames.setCookie)
 
   final def getSetCookiesDecoded: List[Cookie] =
-    getHeaderValues(HeaderNames.cookie)
+    getHeaderValues(HeaderNames.setCookie)
```
tusharmath, line 280: This is a bug? Probably a separate PR?

zio-http/src/main/scala/zhttp/http/Cookie.scala
tusharmath: Also why is `content` an argument here?

zio-http/src/main/scala/zhttp/http/Cookie.scala
ShrutiVerma97: But if we are not able to sign the cookie then it should return cookie(without sign)?

zio-http/src/main/scala/zhttp/http/Patch.scala
ShrutiVerma97: Merged in main

zio-http/src/main/scala/zhttp/http/headers/HeaderGetters.scala
ShrutiVerma97: Merged in main

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -120,6 +140,30 @@ final case class Cookie(
     cookie.flatten.mkString("; ")
   }
 
+  /**
+   * Signs cookie content with a secret
+   */
+  private def signContent(secret: String): String = {
```
tusharmath, line 146: `sign` should only return the signature not combine it with the original content.

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -120,6 +140,30 @@ final case class Cookie(
...
+    sha256.init(secretKey)
+    val signed    = sha256.doFinal(self.content.getBytes())
+    val mda       = MessageDigest.getInstance("SHA-512")
+    self.content + '.' + getEncoder.encodeToString(mda.digest(signed))
+  }
+
+  /**
+   * Unsigns cookie content with a secret
+   */
+  private def unSignContent(secret: String): Option[Cookie] = {
```
```suggestion
  private def verify(content: String, secret: String): Option[Cookie] = {
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -120,6 +140,30 @@ final case class Cookie(
...
+    val str             = self.content.slice(0, content.lastIndexOf('.'))
+    val unSignedCookie  = self.withContent(str)
+    val encryptedCookie = unSignedCookie.signContent(secret)
+
+    if (encryptedCookie == self.content)
+      Some(unSignedCookie.sign(secret))
+    else None
+  }
```
tusharmath: private method/functions should either return null or throw an error.  In this case we should throw a error.

example/src/main/scala/example/SignCookies.scala
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Example to make app using signed-cookies
+ */
+object SignCookies extends App {
+
+  // Setting cookies with an expiry of 5 days
+  private val cookie = Cookie("key", "hello").withMaxAge(5 days)
+
+  private val app = Http.collect[Request] { case Method.GET -> !! / "cookie" =>
+    Response.ok.addCookie(cookie.sign("secret").withContent("world"))
```
tusharmath:
```suggestion
    Response.ok.addCookie(cookie.sign("secret"))
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -120,6 +140,30 @@ final case class Cookie(
...
+    sha256.init(secretKey)
+    val signed    = sha256.doFinal(self.content.getBytes())
+    val mda       = MessageDigest.getInstance("SHA-512")
+    getEncoder.encodeToString(mda.digest(signed))
+  }
+
+  /**
+   * Unsigns cookie content with a secret
+   */
+  private def verify(content: String, secret: String): Cookie = {
```
```suggestion
  private def verify(content: String, signature: String, secret: String): Boolean = {
```

tusharmath: Verify should only return a True or a False.

zio-http/src/test/scala/zhttp/internal/HttpGen.scala
```diff
@@ -38,7 +38,8 @@ object HttpGen {
     httpOnly <- Gen.boolean
     maxAge   <- Gen.option(Gen.anyLong)
     sameSite <- Gen.option(Gen.fromIterable(List(Cookie.SameSite.Strict, Cookie.SameSite.Lax)))
-  } yield Cookie(name, content, expires, domain, path, secure, httpOnly, maxAge, sameSite)
+    secret   <- Gen.option(Gen.anyString)
+  } yield (Cookie(name, content, expires, domain, path, secure, httpOnly, maxAge, sameSite, secret), secret)
```
tusharmath: do we need to output the secret?

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -144,6 +182,22 @@ object Cookie {
     case object None   extends SameSite { def asString = "None"   }
   }
 
+  def decodeResponseSignedCookie(headerValue: String, secret: String): Option[Cookie] = {
```
tusharmath: I think we can do better here. Instead of adding a new function to decode, it would be easier, to use it like this - `decodeResponseCookie(message, Some("secret"))`.

```scala
def decodeResponseCookie(message: String, secret: Option[String] = None): Option[Cookie] = ...
```

The logic for this code should also move to `unsafeDecodeResponseCookie`. That we have all the decoding logic at one place.

```scala
def unsafeDecodeResponseCookie(headerValue: String, secret: String = null): Cookie = ???
```

Basic idea is to use `null` in unsafe code. This helps with performance.

zio-http/src/test/scala/zhttp/http/CookieSpec.scala
```diff
@@ -7,11 +7,24 @@ import zio.test._
...
+              )(
+                isSome(equalTo(cookieString)),
+              )
+            }
+            case None         => {
+              val cookieString = cookie.encode
+              assert(Cookie.decodeResponseCookie(cookieString))(isSome(equalTo(cookie))) &&
+              assert(Cookie.decodeResponseCookie(cookieString).map(_.encode))(isSome(equalTo(cookieString)))
+            }
+          }
```
tusharmath: I think, we do this https://github.com/dream11/zio-http/pull/751/files#r782728917

We won't need to modify this test much. Only pass the secret as an additional operation.

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -237,6 +238,22 @@ object Middleware {
...
+      Patch.updateHeaders(h => {
+        Headers(
+          h.toList.collect {
+            case h if contentEqualsIgnoreCase(h._1, HeaderNames.setCookie) => {
+              (HeaderNames.setCookie, Cookie.decodeResponseCookie(h._2.toString).get.sign(secret).encode)
+            }
+          },
+        )
+      }),
+    )
```
tusharmath: we can add a new operator inside of `Middleware` called `modifyHeaders`

```scala
def modifyHeaders(f: PartialFunction[Header, Header]): Middleware[Any, Nothing]
```

Use this constructor to implement signCookies.

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -237,6 +238,22 @@ object Middleware {
         } yield Patch.empty
     }
 
+  /**
+   * Creates a middleware for signing cookies
+   */
+  def signCookies[R, E](secret: String): Middleware[R, E] =
```
tusharmath:
```suggestion
  def signCookies(secret: String): Middleware[Any, Nothing] =
```

zio-http/src/test/scala/zhttp/middleware/MiddlewareSpec.scala
```diff
@@ -210,6 +210,20 @@ object MiddlewareSpec extends DefaultRunnableSpec with HttpAppTestExtensions {
...
+      } +
+      suite("signCookie") {
+        testM("should sign cookies") {
+          val cookieRes = Headers.setCookie(Cookie("key", "value").withHttpOnly)
+          val app       =
+            Http.ok.addHeaders(cookieRes) @@ signCookies("secret") getHeader "set-cookie"
+          assertM(app(Request()))(
+            isSome(
+              equalTo(
+                "key=value.fm67q+j8zQjFXnLi22ckhgRC5qaQ9srgU3/Fli94OOWmtuo68xcm5LkXbcODb9taM/B48j6kws3eZ0MYDIeWTA==; HttpOnly",
```
tusharmath: Just use `cookie.encode`

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -147,10 +185,10 @@ object Cookie {
   /**
    * Decodes from Set-Cookie header value inside of Response into a cookie
    */
-  def decodeResponseCookie(headerValue: String): Option[Cookie] =
-    Try(unsafeDecodeResponseCookie(headerValue)).toOption
+  def decodeResponseCookie(headerValue: String, secret: Option[String] = None): Option[Cookie] =
+    Try(unsafeDecodeResponseCookie(headerValue, secret.orNull)).toOption
 
-  private[zhttp] def unsafeDecodeResponseCookie(headerValue: String): Cookie = {
+  private[zhttp] def unsafeDecodeResponseCookie(headerValue: String, secret: String = null): Cookie = {
```
```suggestion
  private[zhttp] def unsafeDecodeResponseCookie(headerValue: String, secret: Option[String] = None): Cookie = {
```

zio-http/src/test/scala/zhttp/middleware/MiddlewareSpec.scala
```diff
@@ -210,6 +210,15 @@ object MiddlewareSpec extends DefaultRunnableSpec with HttpAppTestExtensions {
               res <- r.get
             } yield assert(res)(equalTo(false))
           }
+      } +
+      suite("signCookie") {
+        testM("should sign cookies") {
+          val cookie       = Cookie("key", "value").withHttpOnly
+          val cookieRes    = Headers.setCookie(cookie)
+          val cookieString = cookie.sign("secret").encode
+          val app          = Http.ok.addHeaders(cookieRes) @@ signCookies("secret") getHeader "set-cookie"
```
tusharmath:
```suggestion
          val app          = Http.ok.addCookie(cookie) @@ signCookies("secret") getHeader "set-cookie"
```

zio-http/src/test/scala/zhttp/middleware/MiddlewareSpec.scala
```diff
@@ -210,6 +210,15 @@ object MiddlewareSpec extends DefaultRunnableSpec with HttpAppTestExtensions {
               res <- r.get
             } yield assert(res)(equalTo(false))
           }
+      } +
+      suite("signCookie") {
```
tusharmath:
```suggestion
      suite("signCookies") {
```

zio-http/src/main/scala/zhttp/http/headers/HeaderGetters.scala
```diff
@@ -278,7 +278,12 @@ trait HeaderGetters[+A] { self =>
 
   final def getSetCookiesDecoded: List[Cookie] =
     getHeaderValues(HeaderNames.setCookie)
-      .map(Cookie.decodeResponseCookie)
+      .map(Cookie.decodeResponseCookie(_))
+      .collect { case Some(cookie) => cookie }
+
+  final def getSignedSetCookiesDecoded(secret: String): List[Cookie] =
```
tusharmath: we don't need this method. We can use `getSetCookiesDecoded(secret: Option[String])` instead.

tusharmath (approved):

PR #761, opened 2022-01-04
Feature: CSRF Middleware

### LINKED ISSUE #420: CSRF Middleware | Double Cookie Pattern | Stateless

smehta91:
fixes #420

### DIFF
--- a/example/src/main/scala/example/CSRF.scala
+++ b/example/src/main/scala/example/CSRF.scala
@@ -1,11 +1,12 @@
 package example
 
+import zhttp.http.Middleware.{addCookie, csrf}
 import zhttp.http._
-import zhttp.http.middleware.HttpMiddleware.{addCookie, csrf}
-import zio.UIO
+import zhttp.service.Server
+import zio.{App, ExitCode, UIO, URIO}
 
 object CSRF extends App {
-  val app: HttpApp[Any, Nothing] = HttpApp.collectM {
+  val app: HttpApp[Any, Nothing] = Http.collectM {
     case Method.GET -> !! / "safeEndpoint"    =>
       UIO(Response.ok)
     case Method.POST -> !! / "unsafeEndpoint" => UIO(Response.ok)
@@ -13,4 +14,6 @@ object CSRF extends App {
   app @@ csrf("x-token", "token").when((method, _, _) => method == Method.POST) @@
     addCookie(Cookie("token", "randomToken")).when((method, _, _) => method == Method.GET)
 
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
+    Server.start(8091, app).exitCode
 }
--- a/example/src/main/scala/example/CSRF.scala
+++ b/example/src/main/scala/example/CSRF.scala
@@ -1,20 +1,16 @@
 package example
 
-import zhttp.http.Middleware.{addCookie, csrf}
+import zhttp.http.Middleware.csrf
 import zhttp.http._
 import zhttp.service.Server
 import zio.{App, ExitCode, URIO}
 
 object CSRF extends App {
-  val app: HttpApp[Any, Nothing] =
-    Http.collect[Request] {
-      case Method.GET -> !! / "safeEndpoint"    => Response.text("Hello World!")
-      case Method.POST -> !! / "unSafeEndpoint" => Response.jsonString("""{"greetings": "Hello World!"}""")
-    } @@ addCookie(Cookie("token", "randomToken")).when((method, _, _) => method == Method.GET) @@ csrf(
-      "x-token",
-      "token",
-    ).when((method, _, _) => method == Method.POST)
+  // CSRF middleware
+  // To prevent Cross-site request forgery attacks. This middleware is modeled after the double submit cookie pattern.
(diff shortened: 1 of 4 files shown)


example/src/main/scala/example/CSRF.scala
```diff
@@ -0,0 +1,21 @@
...
+import zio.random.Random
+
+object CSRF extends App {
+  val privateApp: HttpApp[Any, Nothing] = Http.collect[Request] { case Method.GET -> !! / "unsafeEndpoint" =>
+    Response.text("secure info")
+  } @@ csrfValidate() // Check for matching csrf header and cookie
+
+  val publicApp: HttpApp[Any, Nothing] = Http.collect[Request] { case Method.GET -> !! / "safeEndpoint" =>
+    Response.text("hello")
+  } @@ csrfGenerate(random.nextString(5).provideLayer(Random.live)) // set x-csrf token cookie
```
tusharmath:
```suggestion
  } @@ csrfGenerate() // set x-csrf token cookie
```

tusharmath: We din't need to provide the env over here.

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -113,6 +113,30 @@ object Middleware {
   def basicAuth(u: String, p: String): Middleware[Any, Nothing] =
     basicAuth { case (user, password) => (user == u) && (password == p) }
 
+  def addCookieM[R, E](cookie: ZIO[R, E, Cookie]): Middleware[R, E] =
+    patchM((_, _) => cookie.mapBoth(Option(_), c => Patch.addHeader(Headers.setCookie(c))))
+
+  def csrfGenerate[R, E](tokenGen: ZIO[R, E, String], tokenName: String = "x-csrf-token"): Middleware[R, E] =
```
tusharmath:
```suggestion
  def csrfGenerate[R, E](tokenGen: ZIO[R, E, String] = random.nextString(6), tokenName: String = "x-csrf-token"): Middleware[R, E] =
```

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -113,6 +113,38 @@ object Middleware {
...
+   * @method
+   *   csrfGenerate - Sets cookie with CSRF token
+   * @method
+   *   csrfValidate - Validate token value in request headers against value in cookies
+   * @see
+   *   https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie
+   */
+
+  def csrfGenerate[R, E](
+    tokenGen: ZIO[R, Nothing, String] = random.nextUUID.map(_.toString),
```
tusharmath:
```suggestion
    tokenGen: ZIO[R, Nothing, String] = UIO(java.util.UUID.randomUUID.toString),
```

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -113,6 +113,38 @@ object Middleware {
...
+    tokenName: String = "x-csrf-token",
+  ): Middleware[R, E] =
```
tusharmath, line 136: I  would also flip the arguments, ie have `token` before the generator.

zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -113,6 +113,38 @@ object Middleware {
   def basicAuth(u: String, p: String): Middleware[Any, Nothing] =
     basicAuth { case (user, password) => (user == u) && (password == p) }
 
+  def addCookieM[R, E](cookie: ZIO[R, E, Cookie]): Middleware[R, E] =
+    patchM((_, _) => cookie.mapBoth(Option(_), c => Patch.addHeader(Headers.setCookie(c))))
```
amitksingh1490:
```suggestion
    patchZIO((_, _) => cookie.mapBoth(Option(_), c => Patch.addHeader(Headers.setCookie(c))))
```

zio-http/src/main/scala/zhttp/http/Middleware.scala
smehta91: We won't be able to test this if we make this change.

amitksingh1490 (approved):

PR #771, opened 2022-01-06
Scala doc generation CI step

### LINKED ISSUE #736: Include a step in ci to also check scala Doc generation
A step could be added in the workflow that also checks doc generation.

**Why is it important?**
Build failed during doc generation on `main`, we would like to eliminate the chances of it happening again in the future.

girdharshubham:
Fixes doc generation 
fixes #736

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,14 +4,14 @@ import Dependencies._
 val releaseDrafterVersion = "5"
 
 // CI Configuration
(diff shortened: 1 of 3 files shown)


amitksingh1490 (approved):

build.sbt
```diff
@@ -66,6 +66,14 @@ ThisBuild / githubWorkflowBuildPreamble :=
     scalas = List(Scala213),
   ).steps
 
+ThisBuild / githubWorkflowBuildPostamble := Seq(
```
amitksingh1490: should run only on PR also on one scala version

amitksingh1490 (approved):

PR #782, opened 2022-01-07
`Http.fromFile` should send an Internal Server Error response when the file does not exist.

ashprakasan:
This PR has 2 changes -

1. Internal Server Error when there is an Exception on `Http.fromFile`
2. `CONTENT_ENCODING : Chunked` is not present on `Http.fromFile` response


ashprakasan:
This will be addressed as a part of [#850](https://github.com/dream11/zio-http/pull/850)

PR #786, opened 2022-01-08
Fix cors middleware failing test

amitksingh1490:


zio-http/src/main/scala/zhttp/http/Middleware.scala
```diff
@@ -106,7 +106,10 @@ object Middleware {
       case Codec(decoder, encoder)       => http.contramapZIO(decoder(_)).mapZIO(encoder(_))
       case Identity                      => http.asInstanceOf[Http[R, E, AOut, BOut]]
       case Constant(http)                => http
-      case OrElse(self, other)           => self.execute(http).orElse(other.execute(http))
+      case OrElse(self, other)           =>
+        Http.fromOptionFunction { a =>
+          self.execute(http)(a) orElse other.execute(http)(a)
+        }
```
tusharmath, line 112: This should mean the same imo. Why is the behaviour different?

amitksingh1490: Due to option type in Error we want it to go to next if Its error but only want to execute other if error is some. fromOptionFunction handles this behaviour.

PR #789, opened 2022-01-08
fix: lazily evaluate default in foldExit

brendo-m:
Noticed when debugging the other day that the default route is always evaluated, even if the main route evaluates


tusharmath:
This is by design. Primary reason would be performance. Why would you need this to be lazily evaluated?

brendo-m:
> This is by design. Primary reason would be performance. Why would you need this to be lazily evaluated?

Interesting, wouldn't it be more performant not to do this? It's doing extra work on each request for potentially no reason

tusharmath:
It depends how you use it. In this context, calling by name has a bit of memory overhead and using it in hot paths can turn out to be counter productive.

brendo-m:
Makes sense. Happy to close

PR #800, opened 2022-01-11
Test/keepalive httpversion

### LINKED ISSUE #791: Enable Server KeepAlive by default while creating server and add test cases.
Keep Alive disabling may not disable TCP connection re-use internally.  So it is an ambiguous behaviour. Keep alive can be enabled / disabled per request using connection headers

Also with tools like apache workbench does not work with zio-http server if keep alive is disabled. 
So it is necessary to use Server.keepAlive while creating Server, otherwise we get this error.

```
ab -n 100 -c 10  http://localhost:61859/foo

Benchmarking localhost (be patient)...apr_pollset_poll: The timeout specified has expired (70007)
```
The reason being Apache Bench uses HTTP 1.0

Rejecting Http 1.0 requests with connection headers may break clients coming through proxy servers.

	* Use keepAlive always and remove Server.keepAlive
	* HTTP 1.1 requests can use "Connection: close" in the request header to control re-use mechanism
	* HTTP 1.0 requests can use "Connection: keep-alive" in the request header to control re-use mechanism
	* Also may be we can explore keepAliveTimeout mechanism rather, if there is a substantial use case.

The "proof" that persistent connection is still working at the transport level, even if we don't use Server.keepAlive 

Run HelloWorldAdvanced in examples test (with Server.keepAlive disabled), and fire these curl requests from command line 
```
 curl -Iv http://localhost:8090/foo --next http://localhost:8090/foo 2>&1 | grep -i '#0'
```
We see the message regarding Re-use 
```
* Connected to localhost (::1) port 8090 (#0)
* Connection #0 to host localhost left intact
* Re-using existing connection! (#0) with host localhost
* Connected to localhost (::1) port 8090 (#0)
* Connection #0 to host localhost left intact	
```

**Additional context**

As per [RFC 2616] <https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.10>  every connection is by default assumed to be persistent if not otherwise indicated by an explicit Connection: close header.

```
HTTP/1.1 defines the "close" connection option for the sender to signal that the connection will be closed after completion of the response. For example,

       Connection: close

in either the request or the response header fields indicates that the connection SHOULD NOT be considered `persistent' (section 8.1) after the current request/response is complete.
HTTP/1.1 applications that do not support persistent connections MUST include the "close" connection option in every message.
```
##### For Http 1.0 (WITH Connection: keep-alive in request header)
```
curl  --http1.0 -Iv -H "Connection: keep-alive"  http://localhost:8090/foo
```
We don't get connection close in response header and we get "left intact" message indicating a re-use with Http 1.0
```
< 
* Connection #0 to host localhost left intact
* Closing connection 0
```

##### For Http 1.0 (WITHOUT Connectin: keep-alive in request header)
```
curl  --http1.0 -Iv http://localhost:8090/foo  
```
We get connection: close in response header (correct behaviour) indicating no re-use
```
	< connection: close
	* Closing connection 0	
```

##### For Http 1.1 (Without Connection: close in request header)

We get "left intact" message indicating  re-use (correct behaviour)
```
curl  --http1.1 -Iv http://localhost:8090/foo
	< 
	* Connection #0 to host localhost left intact
	* Closing connection 0
```

##### For Http 1.1 (WITH Connection: close in request header)
```
curl  --http1.1 -Iv -H "Connection: Close" http://localhost:8090/foo
```
We get Connection: close in response header (and no "left intact" message) indicating no re-use (correct behaviour)
```
	< connection: close
	* Closing connection 0
```

sumawa:


zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -19,15 +20,19 @@ import zio.{Chunk, Has, Task, ZIO, ZManaged}
  * actual Http server and makes requests.
  */
 abstract class HttpRunnableSpec extends DefaultRunnableSpec { self =>
+  val http10 = HttpVersion.HTTP_1_0
```
amitksingh1490, line 23:
```suggestion
  val http1_0 = HttpVersion.HTTP_1_0
```

zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -19,15 +20,19 @@ import zio.{Chunk, Has, Task, ZIO, ZManaged}
...
+  val http11 = HttpVersion.HTTP_1_1
```
amitksingh1490: not needed import `io.netty.handler.codec.http.HttpVersion._` instead

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -123,7 +123,7 @@ object ServerSpec extends HttpRunnableSpec {
     } +
       testM("POST Request.getBody") {
         val app = Http.collectZIO[Request] { case req => req.getBody.as(Response.ok) }
-        val res = app.requestStatus(!!, Method.POST, "some text")
+        val res = app.requestStatus(method = Method.POST, content = "some text")
```
amitksingh1490, line 126:
```suggestion
        val res = app.requestStatus(path= !!, method = Method.POST, content = "some text")
```

amitksingh1490 (approved):

amitksingh1490 (approved):

PR #806, opened 2022-01-12
Logging Middleware

gciuloaica:


zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -45,6 +46,58 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
         } yield Patch.empty
     }
 
+  /**
+   * Provides a logging middleware
+   */
+  def log[R](
+    logger: (String, LogLevel) => ZIO[R, Throwable, Unit],
```
tusharmath, line 53:
```suggestion
    logger: LogChunk => ZIO[R, Throwable, Unit],
```

```scala
case class LogChunk {
  def request: Request
  def response: Response

  // May be should be part of response
  def isError: Boolean
  def isBadRequest: Boolean
  def isAuthorized: Boolean

  // ...end


  def duration: Long

  // Actual string that can be logged.
  def encode(fmt: LogFmt): String
}

```

Need a better name.

zio-http/src/main/scala/zhttp/http/LOG.scala
```diff
@@ -0,0 +1,107 @@
...
+  private def execute(
+    isRequest: Boolean,
+    logFmt: LogFmt,
+    method: Option[Method],
+    url: Option[URL],
+    headers: Headers,
+    body: Task[String],
+    status: Option[Status],
+    duration: Option[Long],
+  ): Task[String] = {
```
tusharmath, line 57:
```suggestion
  private def execute(logChunk: LogChunk): String = {
```

zio-http/src/main/scala/zhttp/http/LOG.scala
```diff
@@ -0,0 +1,107 @@
...
+      case LogFmt.Status if status.isDefined     => Task.succeed(s" Status: ${status.get},")
+      case LogFmt.Duration if duration.isDefined => Task.succeed(s" Duration: ${duration.get}ms")
+      case LogFmt.Body(limit)                    => body.map(b => s" Body: ${b.take(limit)}, ")
+      case LogFmt.Request(fmt) if isRequest   => execute(isRequest, fmt, method, url, headers, body, status, duration)
+      case LogFmt.Response(fmt) if !isRequest => execute(isRequest, fmt, method, url, headers, body, status, duration)
+      case LogFmt.Headers(filter)             =>
+        Task.succeed(s" Headers: ${LogFmt.stringifyHeaders(filter(headers)).mkString}")
+      case _                                  => Task.succeed("")
+    }
+  }
```
tusharmath, line 75: Can we add more unit tests for every scenario.

tusharmath: This is going to be significantly slow, because we are allocating a lot of Strings in the process.

tusharmath: **Use case 1 — Column Spacing**
```
GET /foo     30ms  200
GET /foo-bar 120ms 401
```
What if I want data to maintain columns.
We need ability to pad and trim or space out columns.

tusharmath: - Formatting one particular header differently eg: `Content-Length` should be shown in Bold.
- How do I specify common headers eg: `content-type`?
- How do I log content as JSON?

tusharmath:
Closing because of no activity.

PR #794, opened 2022-01-10
Websocket client support

girdharshubham:
Closes #650 

Depends on #894 #898 #900 #901 #902 #903

PR #809, opened 2022-01-12
ZIO 2 support

jdegoes:
Web socket tests are temporarily disabled as sttp does not support ZIO 2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .idea
 .bloop/
 .metals/
+null
 out
 .cache/
 .history/
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.ipr
 *.iws
 .idea
+.bloop/
+.metals/
 out
 .cache/
 .history/
@@ -28,3 +30,6 @@ node_modules/
 .bsp/
 .DS_Store
 web/.DS_Store
+
+.vscode
+project/metals.sbt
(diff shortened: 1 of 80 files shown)


jdegoes:
I think the benchmarks are expected to not pass, but please let me know if I'm missing something there (i.e. a generalization that would let them pass with a non-main branch).

amitksingh1490:
> I think the benchmarks are expected to not pass, but please let me know if I'm missing something there (i.e. a generalization that would let them pass with a non-main branch).

Yes benchmarks fail on forked PR please ignore https://github.com/dream11/zio-http/issues/619

tusharmath:
@jdegoes Thanks!

We can merge this PR and make sure it's updated with every change in the `main` branch. The websocket client work is being developed in #794 and should be out soon so we won't need to have sttp dependency.

Would you require a maven release also? The challenge that I see could be with versioning since ZIO http is still in RC.

jdegoes:
@tusharmath Yes, I need a Maven release, and it will help the ecosystem as well to have something available for testing.

This is my first PR (based on @kitlangton and @adamfraser's work), but expect many more to come!

jdegoes:
I would suggest for versioning we do: ZIO HTTP 2.0-RC<X>. So there will exist the ZIO HTTP 1.x line, and the ZIO HTTP 2.x line. Obviously folks will continue using 1.x for around 6 months but the future will be 2.x

amitksingh1490 (approved):

amitksingh1490 (approved):

PR #824, opened 2022-01-13
Feature: API to modify headers

### LINKED ISSUE #823: API to modify headers
Right now, there is only one operator in `Headers` to update `Headers` which takes  param```(update: Headers => Headers) ```
There should be one operator which modifies `Header` for eg: ```(f: Header => Header)``` 

```def modifyHeaders(f: Header => Header): Headers = ???```

### PR DESCRIPTION
**Pull request by ShrutiVerma97**
fixes #823

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Headers.scala
+++ b/zio-http/src/main/scala/zhttp/http/Headers.scala
@@ -26,6 +26,8 @@ final case class Headers(toChunk: Chunk[Header]) extends HeaderExtension[Headers
 
   def toList: List[(String, String)] = toChunk.map { case (name, value) => (name.toString, value.toString) }.toList
 
+  def modifyHeaders(f: Header => Header): Headers = Headers(toChunk.map(f(_)))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: zio-http/src/main/scala/zhttp/http/Headers.scala [hunk 1]
```diff
@@ -26,6 +26,8 @@ final case class Headers(toChunk: Chunk[Header]) extends HeaderExtension[Headers
 
   def toList: List[(String, String)] = toChunk.map { case (name, value) => (name.toString, value.toString) }.toList
 
+  def modifyHeaders(f: Header => Header): Headers = Headers(toChunk.map(f(_)))
```
tusharmath:
```suggestion
  def modify(f: Header => Header): Headers = Headers(toChunk.map(f(_)))
```

Let's keep it as `modify` because we know it's applicable only on headers.

tusharmath (approved):

tusharmath (approved):

PR #846, opened 2022-01-17
URL Refactoring

### LINKED ISSUE #501: Improve URL Parser performance.
Optimise URL.fromString function
Write jmh benchmarks and try micro optimisations

ShrutiVerma97:
fixes #501 #846 
Benchmark results
```
[info] Benchmark                                Mode  Cnt     Score     Error  Units
[info] URLParserBenchmark.benchmarkURLParser   thrpt   10  1795.675 ? 142.343  ops/s
[info] URLParserBenchmark.benchmarkURLParser2  thrpt   10  2131.620 ?  71.715  ops/s
[info] URLParserBenchmark.benchmarkURLParser3  thrpt   10  1930.228 ?  83.573  ops/s
```


zio-http/src/main/scala/zhttp/http/Scheme.scala
```diff
@@ -26,4 +26,11 @@ object Scheme       {
       case "HTTP"  => Option(HTTP)
       case _       => None
     }
+
+  def fromString2(scheme: String): Scheme =
+    scheme.toUpperCase match {
+      case "HTTPS" => HTTPS
+      case "HTTP"  => HTTP
+      case _       => null
```
amitksingh1490: Try adding a subtype empty.

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -67,6 +67,30 @@ object URL {
...
+  private def fromAbsoluteURI2(uri: URI): URL = {
+
+    def portFromScheme(scheme: Scheme): Int = scheme match {
+      case Scheme.HTTP  => 80
+      case Scheme.HTTPS => 443
+      case null         => -1
+    }
+
+    val scheme = Scheme.fromString2(uri.getScheme)
+    val port   = if (uri.getPort != -1) uri.getPort else portFromScheme(scheme)
```
amitksingh1490: try moving uri.getPort to val

amitksingh1490: This code is also not safe since portFromScheme can alo return -1

ShrutiVerma97: We are handling that in the next statement

tusharmath:
One thing we can do @ShrutiVerma97 here is to convert URL to a sum type. 

```scala
sealed trait URL

final case class FromString(string: String) extends URL
final case class Cons(
  path: Path,
  kind: URL.Location = URL.Location.Relative,
  queryParams: Map[String, List[String]] = Map.empty,
  fragment: Option[Fragment] = None,
) extends URL
```

So this way we will keep an unparsed version as string, and parse it only when its truly needed.

ShrutiVerma97:
Benchark results URL vs URL2 
```
[info] Benchmark                                Mode  Cnt     Score     Error  Units
[info] URLParserBenchmark.benchmarkURL2Parser  thrpt    5  2122.770 ? 302.008  ops/s
[info] URLParserBenchmark.benchmarkURLParser   thrpt    5  2033.823 ?  51.423  ops/s
[info] URLParserBenchmark.benchmarkURL2Parser2  thrpt    5  2315.495 ?  53.017  ops/s
[info] URLParserBenchmark.benchmarkURL2Parser3  thrpt    5  2060.433 ?  68.606  ops/s
```

zio-http/src/main/scala/zhttp/http/PathModule.scala
```diff
@@ -63,8 +63,8 @@ private[zhttp] trait PathModule { module =>
   }
 
   object Path {
-    def apply(): Path                   = End
-    def apply(string: String): Path     = if (string.trim.isEmpty) End else Path(string.split("/").toList)
+    def apply(): Path               = End
+    def apply(string: String): Path = if (string.trim.isEmpty || string == null) End else Path(string.split("/").toList)
```
tusharmath: this change should not be required

zio-http/src/main/scala/zhttp/http/Scheme.scala
tusharmath: Can be optimized further.

zio-http/src/main/scala/zhttp/http/URL5.scala
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+}
+object URL5 {
+
+  def apply(path: Path): URL5                                = Cons(path = path)
+  def apply(str: String): Either[HttpError.BadRequest, URL5] = Decode(string = str).decode
+
+  final case class Decode(string: String) extends URL5 {
```
tusharmath:
```suggestion
  final case class Raw(string: String) extends URL5 {
```

ShrutiVerma97:
Updated Benchmarks:
```
[info] Benchmark                                  Mode  Cnt    Score     Error  Units
[info] URLParserBenchmark.benchmarkURLOldParser  thrpt    3  423.353 ? 414.729  ops/s
[info] URLParserBenchmark.benchmarkURLParser     thrpt    3  456.072 ?  95.470  ops/s`

PR #868, opened 2022-01-20
Fix: EncodeClientParams

### LINKED ISSUE #867: HTTP requests contain malformed path
**Describe the bug**
On 1.0.0.0-RC22, we noticed that HTTPS requests succeed, and that HTTP requests (non-secure) fail. By digging deeper, we found that the HTTP commands are malformed during a normal HTTP request.

**To Reproduce**
1. Run netcat `nc -l 9000`
2. Run 

```scala
import zhttp.service.{ChannelFactory, Client, EventLoopGroup}
import zio.{App, ExitCode, URIO, console}

object SimpleClient extends App {

  val program = for {
    res <- Client.request("http://localhost:9000/foo")
    data <- res.getBodyAsString
    _ <- console.putStrLn(data)
  } yield ()

    program.exitCode.provideCustomLayer(ChannelFactory.auto ++ EventLoopGroup.auto())
}
```
3. Go back to netcat and see that it received `GET http://localhost:9000/foo HTTP/1.1`
4. Confirm that we see `http://localhost:9000/foo` (observed) for a path instead of `/foo` (expected)

**Expected behavior**
That a HTTP request does `GET /foo HTTP/1.1` instead of `GET http://localhost:9000/foo HTTP/1.1`

 - OS: MacOS 11.6

ShrutiVerma97:
fixes #867

### DIFF
--- a/zio-http/src/test/scala/zhttp/http/EncodeClientRequestSpec.scala
+++ b/zio-http/src/test/scala/zhttp/http/EncodeClientRequestSpec.scala
@@ -1,8 +1,10 @@
 package zhttp.http
 
-import io.netty.handler.codec.http.{HttpHeaderNames, HttpVersion}
+import io.netty.handler.codec.http.{FullHttpRequest, HttpHeaderNames, HttpVersion}
 import zhttp.internal.HttpGen
+import zhttp.service.Client.ClientRequest
 import zhttp.service.{Client, EncodeClientParams}
+import zio.ZIO
 import zio.random.Random
 import zio.test.Assertion._
 import zio.test._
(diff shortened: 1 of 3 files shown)


zio-http/src/test/scala/zhttp/http/EncodeClientRequestSpec.scala
```diff
@@ -38,13 +38,13 @@ object EncodeClientRequestSpec extends DefaultRunnableSpec with EncodeClientPara
...
           val req = encodeClientParams(HttpVersion.HTTP_1_1, params)
-          assert(req.uri())(equalTo(params.url.asString))
+          assert(req.uri())(equalTo(params.url.relative.asString))
         }
       } +
       testM("uri on HttpData.File") {
         check(HttpGen.clientParamsForFileHttpData()) { params =>
           val req = encodeClientParams(HttpVersion.HTTP_1_1, params)
-          assert(req.uri())(equalTo(params.url.asString))
+          assert(req.uri())(equalTo(params.url.relative.asString))
```
tusharmath: Let's tests with actual and expected urls. Here we are testing implementation details.

tusharmath (approved):

PR #880, opened 2022-01-22
Fix/skip benchmark documentation

kaushik143:
- Skip benchmark for PR with documentation label
- Trigger workflow when PR is labeled or unlabeled

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ name: Continuous Integration
 on:
   pull_request:
     branches: ['**']
-    types: [opened, synchronize, reopened, edited]
+    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
   push:
     branches: ['**']
     tags: [v*]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ name: Continuous Integration
 on:
   pull_request:
     branches: ['**']
-    types: [opened, synchronize, reopened, edited]
+    types: [opened, synchronize, reopened, edited, labeled, unlabeled]
```
amitksingh1490, line 13: this will skip the label doc but it will rerun the Benchmarks whenever any label is added or removed

amitksingh1490 (changes requested):

tusharmath:
Should we only run benchmarks when there is a `benchmark` label attached?

.github/workflows/ci.yml
kaushik-barodiya: I added this so that if we remove the documentation label then again CI runs and run workflow accordingly

amitksingh1490:
closing this due to inactivity  please reopen when you work on it again

PR #899, opened 2022-01-26
Feature: `Content-Type` should be set using a custom MimeDB

### LINKED ISSUE #883: `CONTENT_TYPE` header gets incorrect values because of multiple bugs with `Files.probleFileContent` used in its implementation.
**Describe the bug**
FileType detections using `Files.probeFileContent` is platform-specific and largely unreliable.
`CONTENT_TYPE` header is `null` (when serving files with extension '.mp4' ) on openJDK because Files.probleFileContent returns `null`.

File type detection is largely platform-specific and was not considered reliable.
**To Reproduce**
1. Run the example `FileStreaming.scala` on `Linux` os with `Open JDK11`
2. Request 'http://www.localhost:8090/video' throws an exception

**Expected behavior**
We should see a proper response with `CONTENT_TYPE` header set as 'video/mp4`

**Screenshots**

**Platform (please complete the following information):**
Ubuntu 11.04 + OpenJDK

**Additional context**
A few bugs being tracked with `Files.probeFileContent`

https://bugs.openjdk.java.net/browse/JDK-7142921
https://bugs.openjdk.java.net/browse/JDK-8186071
https://bugs.openjdk.java.net/browse/JDK-8129632

ashprakasan:
Fixes #883 

`MimeDb.scala` should ideally be auto-generated from https://www.iana.org/assignments/media-types/media-types.xhtml. As of now, the list of `MediaType` is hard-coded.


tusharmath:
Take a pull from `dream11:main`

ashprakasan:
As I understand, there are a few different options for auto-generating `MimeDB.scala`

1. Generating code using string templates ([Twirl](https://github.com/playframework/twirl)).

     **Pros** : This seems to have the easiest learning curve of all the approaches. 
                Twirl works very well with scala. The syntax is `scala-ish`
                Code is easy to read and understand.

     **Cons** : Can produce code that doesn't compile.

2. Generating code from an abstract syntax tree ([treehugger](http://eed3si9n.com/treehugger/)).

      **Pros** : Uncompilable code cannot be generated.
                        Uses scala syntax.

      **Cons** : Steep learning curve. Will take me some time to do this.
                        Not much IDE support.
                        Code looks complex, reader might take some time to understand what is happening.
                         
3. Building an abstract syntax tree and sending it directly to the compiler ([Scala macros](https://scalamacros.org/)).
      
      **Pros** : Directly passes bytecode of AST to next phase of compilation without generating `.scala` file
                       Syntactically invalid code will not be generated.

      **Cons** : Steep learning curve. Will take me some time to complete this.
                        Not much IDE support.
                        Code can be complex. Reader needs to know macros to understand what is happening.

What do you think will be a good approach here? @tusharmath, @amitksingh1490

ashprakasan:
[#934](https://github.com/dream11/zio-http/issues/934) was created to separately address the auto-generation of `MimeDB.scala`

zio-http/src/main/scala/zhttp/http/MediaType.scala
```diff
@@ -0,0 +1,21 @@
...
+  val fileExtensions: List[String] = Nil,
+  extensions: Map[String, String] = Map.empty,
+)
+
+object MediaType extends MimeDB {
+  def forExtention(ext: String): Option[MediaType] = extensionMap.get(ext.toLowerCase)
+
+  val extensionMap: Map[String, MediaType] = allMediaTypes.flatMap(m => m.fileExtensions.map(_ -> m)).toMap
+
+  def probeContentType(name: String): Option[String] = name.lastIndexOf(".") match {
```
tusharmath: Let's have an unsafe implementation which is package private also. That returns a `String` or `null`.

Purely for perf reasons.

zio-http/src/main/scala/zhttp/http/MediaType.scala
```diff
@@ -0,0 +1,21 @@
+package zhttp.http
+
+case class MediaType(
+  mainType: String,
+  val subType: String,
+  val compressible: Boolean = false,
+  val binary: Boolean = false,
+  val fileExtensions: List[String] = Nil,
```
tusharmath: We don't need `val`. It's a `case class`

zio-http/src/test/scala/zhttp/http/ContentTypeSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+            .map(_.getOrElse("Content type header not found."))
+        assertM(res)(equalTo("application/javascript"))
+      } +
+      testM("no extension") {
+        val file = new File(getClass.getResource("/TestFile4").getPath)
+        val res  =
+          Http
+            .fromFile(file)
+            .deploy
+            .getHeaderValue(HttpHeaderNames.CONTENT_TYPE)
```
tusharmath:
```suggestion
            .getContentType
```
Everywhere.

zio-http/src/test/scala/zhttp/http/ContentTypeSpec.scala
```diff
@@ -0,0 +1,92 @@
...
+        val file = new File(getClass.getResource("/truststore.jks").getPath)
+        val res  =
+          Http
+            .fromFile(file)
+            .deploy
+            .getHeaderValue(HttpHeaderNames.CONTENT_TYPE)
+            .run()
+            .map(_.getOrElse("Content type header not found."))
+        assertM(res)(equalTo("Content type header not found."))
+      }
```
tusharmath, line 78: I think all the tests are doing the same thing. We can maybe reduce the boilerplate.
It's a nice idea to test for multiple extensions.

ashprakasan: okay.

zio-http/src/test/scala/zhttp/http/ContentTypeSpec.scala
ashprakasan: done.

zio-http/src/main/scala/zhttp/http/MediaType.scala
ashprakasan: Done.

zio-http/src/main/scala/zhttp/http/MediaType.scala
ashprakasan: I've memoised the result for now, will do other perf enhancements in the later PRs.

zio-http/src/main/scala/zhttp/http/MediaType.scala
```diff
@@ -0,0 +1,34 @@
+package zhttp.http
+
+import java.util
+
+case class MediaType(
```
tusharmath:
```suggestion
final class MediaType private[zhttp](
```

The MediaType should only be constructed in `MimeDB`

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -203,6 +204,13 @@ sealed trait Http[-R, +E, -A, +B] extends (A => ZIO[R, Option[E], B]) { self =>
   final def headerValue(name: CharSequence)(implicit eb: IsResponse[B]): Http[R, E, A, Option[CharSequence]] =
     headers.map(_.headerValue(name))
 
+  /**
+   * Extracts the value of ContentType header
+   */
+  final def contentType(implicit eb: IsResponse[B]): Http[R, E, A, Option[CharSequence]] = headerValue(
+    HttpHeaderNames.CONTENT_TYPE,
+  )
```
tusharmath, line 212: This is a nice addition!

zio-http/src/main/scala/zhttp/http/MimeDB.scala
```diff
@@ -0,0 +1,7787 @@
+package zhttp.http
+
+private[zhttp] trait MimeDB {
```
tusharmath, line 3: We should be able to set and get content type via MimeDB in `Headers`, `Request` and `Response`

zio-http/src/main/scala/zhttp/http/Http.scala
tusharmath: Should this return a `MediaType`?

zio-http/src/main/scala/zhttp/http/MediaType.scala
ashprakasan: This can be used to add new `MediaType`

tusharmath (approved):

PR #900, opened 2022-01-26
Feature: add `echo` operator to `Socket`

tusharmath:

PR #901, opened 2022-01-26
feature: add `Socket.empty`

tusharmath:
depends on #900


zio-http/src/main/scala/zhttp/socket/Socket.scala
```diff
@@ -59,6 +60,16 @@ sealed trait Socket[-R, +E, -A, +B] { self =>
...
 
+  /**
+   * Simply echos the incoming message back
+   */
+  def echo[A]: Socket[Any, Nothing, A, A] = Socket.collect[A] { case a => ZStream.succeed(a) }
+
+  /**
+   * Creates a socket that doesn't do anything.
+   */
+  def empty: Socket[Any, Nothing, Any, Nothing] = Socket.Empty
```
amitksingh1490, line 72: unit test

girdharshubham: done

girdharshubham (approved):

PR #903, opened 2022-01-26
Feature: add `toHttp` to Response

tusharmath:


zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -61,6 +61,11 @@ final case class Response private (
    */
   def withServerTime: Response = self.copy(attribute = self.attribute.withServerTime)
 
+  /**
+   * Wraps the current response as a Http
+   */
+  def wrapHttp: Http[Any, Nothing, Any, Response] = Http.succeed(self)
```
amitksingh1490: Unit test missing

jdegoes: Could we call it `toHttp`, like `ZIO#toLayer`, `ZIO#toManaged`, etc.?

zio-http/src/test/scala/zhttp/http/ResponseSpec.scala
```diff
@@ -1,15 +1,16 @@
 package zhttp.http
 
+import zio.test.Assertion._
 import zio.test._
 
-object ResponseHelpersSpec extends DefaultRunnableSpec {
-  val redirectSpec = {
-    val location = "www.google.com"
-    suite("redirectSpec")(
+object ResponseSpec extends DefaultRunnableSpec {
```
girdharshubham, line 6: Test case organization made similar to how test cases are written for HeaderSpec, HttpSpec, etc.

```scala
suite("flatMap")(
      test("should flatten") {...}
    )
```

amitksingh1490 (approved):

PR #909, opened 2022-01-27
feature: Add `collectManaged` to Http

beem812:
Figured I would take a stab at this, I wasn't sure if `useNow` was the right way to go about implementing this but It let me lean on the existing code to handle a ZIO. This will also be my first contribution to anything so if I've done this PR incorrectly I'm ready to learn.


tusharmath (approved):

tusharmath:
Perfect! Thanks!

PR #894, opened 2022-01-26
Refactor: Merge client and server `Request`

tusharmath:
Merging `Client.ClientRequest` and `Request` into one type. This was unnecessarily duplicated.

PR #913, opened 2022-01-27
Fix: add host header and port in Request automatically

### LINKED ISSUE #912: `Host` header not automatically set from the `URL`
**Describe the bug**
`Host` header should be automatically set when a url is passed to create a Request.

**To Reproduce**
```scala
val req = URL.fromString("http://www.foo.com").map(url => Request(url = url)).get

req.getHost // None
```

`req.getHost` returns a `None`.

**Expected behavior**
`req.getHost` should return `Some("www.foo.com")`

**Additional context**

tusharmath:
fixes #912 
depends on #894

### DIFF
--- a/zio-http/src/test/scala/zhttp/http/RequestSpec.scala
+++ b/zio-http/src/test/scala/zhttp/http/RequestSpec.scala
@@ -1,0 +1,45 @@
+package zhttp.http
+
+import zio.test.Assertion.{equalTo, isSome}
+import zio.test.{DefaultRunnableSpec, _}
+
+object RequestSpec extends DefaultRunnableSpec {
+  def spec = suite("Request") {
+    suite("host header") {
+      suite("without port") {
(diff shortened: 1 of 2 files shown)


zio-http/src/test/scala/zhttp/http/RequestSpec.scala
```diff
@@ -0,0 +1,45 @@
...
+            } +
+            test("ssl with non-default port") {
+              val url = URL.fromString("https://example.com:8090").getOrElse(null)
+              val req = Request(url = url)
+              assert(req.getHost)(isSome(equalTo("example.com:8090")))
+            } +
+            test("ssl with default port") {
+              val url = URL.fromString("https://example.com:443").getOrElse(null)
+              val req = Request(url = url)
+              assert(req.getHost)(isSome(equalTo("example.com")))
```
tusharmath, line 40: PORT should not be added to host header.

PR #914, opened 2022-01-27
 Response.json method change

yisraelU:
First of all , awesome library . thank you
changed the method to take in CharSequence instead of String.
provides better performance,flexibility plus compatibility with zio-json which encodes to charsequence
Please let me know if this implementation is acceptable , any feedback is desired


zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -180,9 +180,9 @@ object Response {
   /**
    * Creates a response with content-type set to application/json
    */
-  def json(data: String): Response =
+  def json(data: CharSequence): Response =
     Response(
-      data = HttpData.fromChunk(Chunk.fromArray(data.getBytes(HTTP_CHARSET))),
+      data = HttpData.fromString(data.toString),
```
jdegoes: Would be good to push this change into `Http.fromString`, all the way down as low as possible.

tusharmath: We can add a new primitive for `HttpData.fromCharSequence` and `Http.fromString` can also use it since `String <: CharSequence`

yisraelU: @jdegoes , @tusharmath is there a need for 2 primitives? perhaps just change HttpData.Text to hold a CharSequence

jdegoes: I think that's correct, since `CharSequence` is always more efficient than `String`.

yisraelU: I pushed change for review . I can add a new method fromCharSequence to replace the fromString method and mark fromString deprecated as it will be subsumed under the method fromCharSequence.

jdegoes: I guess the main problem is people will look for `fromString`, expecting it to be there. Not sure.

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -80,14 +80,14 @@ object HttpData {
...
-  def fromString(text: String, charset: Charset = HTTP_CHARSET): HttpData = Text(text, charset)
+  def fromString(text: CharSequence, charset: Charset = HTTP_CHARSET): HttpData = Text(text, charset)
 
   /**
    * Helper to create HttpData from contents of a file
    */
   def fromFile(file: java.io.File): HttpData = File(file)
 
-  private[zhttp] final case class Text(text: String, charset: Charset)                   extends HttpData
+  private[zhttp] final case class Text(text: CharSequence, charset: Charset)             extends HttpData
```
tusharmath, line 95: This is good.

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -80,14 +80,14 @@ object HttpData {
   /**
    * Helper to create HttpData from String
    */
-  def fromString(text: String, charset: Charset = HTTP_CHARSET): HttpData = Text(text, charset)
+  def fromString(text: CharSequence, charset: Charset = HTTP_CHARSET): HttpData = Text(text, charset)
```
tusharmath: Keep this, and let's add a `fromCharSequence` also.

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -86,7 +81,7 @@ final case class Response private (
 
     val jHeaders = self.getHeaders.encode
     val jContent = self.data match {
-      case HttpData.Text(text, charset) => Unpooled.wrappedBuffer(text.getBytes(charset))
+      case HttpData.Text(text, charset) => Unpooled.copiedBuffer(text, charset)
```
tusharmath, line 82: Is this change required?

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -185,9 +180,9 @@ object Response {
   /**
    * Creates a response with content-type set to application/json
    */
-  def json(data: String): Response =
+  def json(data: CharSequence): Response =
```
tusharmath, line 181: We can do this for `Response.text("abc")` also.

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -156,7 +156,7 @@ object Client {
   ) extends HeaderExtension[ClientRequest] { self =>
 
     def getBodyAsString: Option[String] = data match {
```
tusharmath, line 179: We can add `getBodyAsCharSequence` also. It's alright if you want to do it in a different PR :)

zio-http/src/main/scala/zhttp/http/Response.scala
yisraelU: @tusharmath 
wrappedbuffer doesn't take in CharSequence .
we can do text.toString.getBytes instead .
Is that preferred?
String.getBytes performs a copy either way

zio-http/src/main/scala/zhttp/http/HttpData.scala
yisraelU: Ok great

zio-http/src/main/scala/zhttp/http/Response.scala
yisraelU: I'll add

zio-http/src/main/scala/zhttp/service/Client.scala
yisraelU: done

zio-http/src/main/scala/zhttp/http/Response.scala
tusharmath: Got it. May be we can do something like this.

```scala
case HttpData.Text(ascii: AsciiString, _) => Unpooled.wrappedBuffer(ascii)
case HttpData.Text(text, charset)         => Unpooled.copiedBuffer(text, charset)
```

The overall idea is to reduce the copies we create as much as possible. This makes a big difference in performance when we send large payloads to the clients.

tusharmath:
Can you take a pull from main.

zio-http/src/main/scala/zhttp/http/Response.scala
yisraelU: ok , ill look into it

yisraelU:
> Can you take a pull from main.

@tusharmath done.

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,6 +175,22 @@ object Client {
   ) extends HeaderExtension[ClientRequest] {
     self =>
 
+    def bodyAsString: Task[String] = getBodyAsByteBuf.map(_.toString(headers.charset))
+    def getBodyAsString: Option[String] = data match {
```
tusharmath, line 179:
```suggestion
    def getBodyAsString: Task[String] = data match {
```

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -175,6 +175,22 @@ object Client {
...
+      case HttpData.Text(text, _)       => Some(text.toString)
+      case HttpData.BinaryChunk(data)   => Some(new String(data.toArray, HTTP_CHARSET))
+      case HttpData.BinaryByteBuf(data) => Some(data.toString(HTTP_CHARSET))
+      case _                            => Option.empty
+    }
+
+    def getBodyAsCharSequence: Option[CharSequence] = data match {
```
tusharmath, line 186:
```suggestion
    def getBodyAsCharSequence: Task[CharSequence] = data match {
```

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -77,6 +77,11 @@ object HttpData {
   def fromStream(stream: ZStream[Any, Throwable, String], charset: Charset = HTTP_CHARSET): HttpData =
     HttpData.BinaryStream(stream.map(str => Unpooled.copiedBuffer(str, charset)))
 
+  /**
+   * Helper to create HttpData from a CharSequence
+   */
+  def fromCharSequence(text: CharSequence, charset: Charset = HTTP_CHARSET): HttpData = Text(text, charset)
```
tusharmath, line 83:
```suggestion
  def fromCharSequence(text: CharSequence): HttpData = Text(text, charset)
```

I think `fromCharSequence` doesn't need `Charset`

zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -87,7 +92,7 @@ object HttpData {
    */
   def fromFile(file: java.io.File): HttpData = File(file)
 
-  private[zhttp] final case class Text(text: String, charset: Charset)                   extends HttpData
+  private[zhttp] final case class Text(text: CharSequence, charset: Charset)             extends HttpData
```
tusharmath, line 95:
```suggestion
  private[zhttp] final case class Text(text: CharSequence)             extends HttpData
```

tusharmath:
Expecting many more conflicts from main. So closing this in favour of #1142

PR #923, opened 2022-01-30
[middleware] Serve precompressed assets

ex0ns:
This is currently a proof of concept, it might not belongs in the main repo and it might even not be the role for a middleware.
The idea behind this is to be able to serve pre-compressed static assets based on the accept-encoding headers of the client. This is often a feature of the reverse proxy but could be integrated inside zio-http if needed.

The PR currently does the following:

- [X] Parse AcceptEncoding header (with weight) and sort them
- [X] Match the compression format from the middleware
- [X] Alter the request to add the pre-defined prefix (can be tuned by the user, I've added a default for GZIP and BR)
- [X] Try multiple content encoding in order 
- [X] Correctly fallback on the original request if none of the compressed one worked

I have two implementation in the code, and I can't decide which one would be better (this is mostly a readability tradeoff).
Any review/advise welcome, but feel free to close/reject if this does not belongs to zio-http repository or don't have time to review.

The `example/src/main/scala/example/CompressedAssets.scala` example will change when the middleware is functional enough, but was used to demonstrate the purpose of this PR.


ex0ns:
Are you still interested by that ? should I spend some time to update it with the lastest version ? And maybe try it with the newly merge file server ?

tusharmath:
Thanks @ex0ns! This seems really interesting. I also think this is a very specific use-case and doesn't really belong here. Closing it for now.

PR #951, opened 2022-02-03
Url macro

gciuloaica:
URL interpolator that fails compilation when provided string is not a valid URL.


gciuloaica:
based on today agreement, the PR is closed.

PR #953, opened 2022-02-04
Doc: for WebSocketFrame

girdharshubham:
Adds in documentation for `WebSocketFrame`

### DIFF
--- a/docs/website/docs/v1.x/dsl/socket/websocketframe.md
+++ b/docs/website/docs/v1.x/dsl/socket/websocketframe.md
@@ -4,7 +4,7 @@ sidebar_label: "WebSocketFrame"
 ---
 In the [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) protocol, communication happens using frames. ZIO
 HTTP's [WebSocketFrame](https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala)
-is its representation of those frames. The protocol defines the following types of frames:
+is its representation of those frames. The domain defines the following types of frames:
 
 * Text
 * Binary
--- a/docs/website/docs/v1.x/dsl/socket/websocketframe.md
+++ b/docs/website/docs/v1.x/dsl/socket/websocketframe.md
@@ -1,0 +1,103 @@
+---
+title: "WebSocketFrame"
+sidebar_label: "WebSocketFrame"
+---
+In the [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) protocol, communication happens using frames. ZIO
+HTTP's [WebSocketFrame](https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala)
+is its representation of those frames. The protocol defines the following types of frames:
+
+* Text
+* Binary
+* Continuation
+* Close
+* Ping
+* Pong
+
+### Text
+
+Text models textual data in the WebSocket protocol. To create a Text frame, you can use the `text` constructor.
+
+```scala
+val text = WebSocketFrame.text("Hello from ZIO-HTTP")
+```
+
+### Binary
+
+Binary models raw binary data. To create a Binary frame, you can use the `binary` constructor.
+
+```scala
+import io.netty.buffer.Unpooled.copiedBuffer
+import io.netty.util.CharsetUtil.UTF_16
+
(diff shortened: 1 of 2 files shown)


docs/website/docs/v1.x/dsl/socket/websocketframe.md
```diff
@@ -0,0 +1,111 @@
+---
+title: "WebSocketFrame"
+sidebar_label: "WebSocketFrame"
+---
+In the [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) protocol, communication happens using frames. ZIO
+HTTP's [WebSocketFrame](https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala)
+is its representation of those frames. The protocol defines the following types of frames, which are generally divided
+into two categories - **Non-Control** and **Control** Frames:
```
devsprint: I think that will be enough to specify: 

`The protocol defines the following types of frames: ` instead of
`The protocol defines the following types of frames, which are generally divided
into two categories - **Non-Control** and **Control** Frames:`

girdharshubham: Sure thing. Thanks for the review, @devsprint

docs/website/docs/v1.x/dsl/socket/websocketframe.md
```diff
@@ -0,0 +1,111 @@
...
+sent back.
+
+```scala
+val ping = WebSocketFrame.ping
+```
+
+## Pattern Matching on WebSocketFrame
+
+ZIO HTTP envisions the WebSocketFrame as a [Sum](https://en.wikipedia.org/wiki/Tagged_union) type, which allows
+exhaustive pattern matching to be performed on it, meaning that the compiler can leverage this knowledge to warn us if
```
ShrutiVerma97: I think we can remove  `meaning that the compiler can leverage this knowledge to warn us if
we write code that does not handle all the cases.`
as we are covering the term exhaustive pattern matching and this is related to scala, not websocket.

girdharshubham: This is done. Thanks, @ShrutiVerma97

docs/website/docs/v1.x/dsl/socket/websocketframe.md
```diff
@@ -0,0 +1,103 @@
+---
+title: "WebSocketFrame"
+sidebar_label: "WebSocketFrame"
+---
+In the [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) protocol, communication happens using frames. ZIO
+HTTP's [WebSocketFrame](https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala)
+is its representation of those frames. The protocol defines the following types of frames:
```
amitksingh1490: check grammar 
```suggestion
is representation of those frames. The Domain defines the following types of frames:
```

girdharshubham: @amitksingh1490 The grammar for the sentence seems correct. However, using `domain` instead of `protocol` makes more sense.

docs/website/docs/v1.x/dsl/socket/websocketframe.md
```diff
@@ -0,0 +1,103 @@
+---
+title: "WebSocketFrame"
+sidebar_label: "WebSocketFrame"
+---
+In the [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) protocol, communication happens using frames. ZIO
+HTTP's [WebSocketFrame](https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala)
+is its representation of those frames. The domain defines the following types of frames:
```
amitksingh1490:
```suggestion
In the [WebSocket](https://datatracker.ietf.org/doc/html/rfc6455) protocol, communication happens using frames. ZIO
HTTP's [WebSocketFrame](https://github.com/dream11/zio-http/blob/main/zio-http/src/main/scala/zhttp/socket/WebSocketFrame.scala)
is the representation of those frames. The domain defines the following type of frames:
```

ShrutiVerma97 (approved):

docs/website/docs/v1.x/dsl/socket/websocketframe.md
girdharshubham: This is done. Thanks, @amitksingh1490.

amitksingh1490 (approved):

PR #965, opened 2022-02-07
Feature: add `Version`

### LINKED ISSUE #954: Create `Version` type to encode `HttpVersion`
We should be able to specify `Version` via `ClientRequest`. This is currently supported only via `HttpVersion`

A sum type should be created for `Version` with `Http_1_1` and `Http_1_2` and `Http_2` as it's specialized types.

tusharmath:
Fixes #954

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Method.scala
+++ b/zio-http/src/main/scala/zhttp/http/Method.scala
@@ -3,8 +3,7 @@ package zhttp.http
 import io.netty.handler.codec.http.HttpMethod
 
 sealed trait Method { self =>
-  lazy val asHttpMethod: HttpMethod = Method.asHttpMethod(self)
-
+  lazy val toJava: HttpMethod     = Method.asHttpMethod(self)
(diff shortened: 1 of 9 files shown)


amitksingh1490 (approved):

zio-http/src/main/scala/zhttp/http/Method.scala
```diff
@@ -3,8 +3,7 @@ package zhttp.http
 import io.netty.handler.codec.http.HttpMethod
 
 sealed trait Method { self =>
-  lazy val asHttpMethod: HttpMethod = Method.asHttpMethod(self)
```
lemony312, line 6: What was the motivation for renaming this field?
Tapir depends on this method for its zio-server interpreter and it is now incompatible with the latest tapir.

lemony312: My mistake, I was using an outdated version of tapir.
Still curious why the name was changed `toJava`
thanks.

PR #972, opened 2022-02-07
Refactor: ServerSSLHandler

dinojohn:
This PR enables further customisations for SSlContext. This is also a supporting PR for http2.

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/server/ServerSSLBuilder.scala
+++ b/zio-http/src/main/scala/zhttp/service/server/ServerSSLBuilder.scala
@@ -1,8 +1,12 @@
 package zhttp.service.server
 
 import io.netty.handler.codec.http2.Http2SecurityUtil
-import io.netty.handler.ssl.ApplicationProtocolConfig.{Protocol, SelectedListenerFailureBehavior, SelectorFailureBehavior}
-import io.netty.handler.ssl.{ApplicationProtocolConfig, ApplicationProtocolNames, SslContext, SslContextBuilder, SslProvider, SupportedCipherSuiteFilter}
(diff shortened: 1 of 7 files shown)


zio-http/src/main/scala/zhttp/service/server/ServerSSLBuilder.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  def ctxFromCert(certInputStream: InputStream, keyInputStream: InputStream): SslContextBuilder = {
+    SslContextBuilder
+      .forServer(certInputStream, keyInputStream)
+      .sslProvider(SslProvider.JDK)
+  }
+
+  def build(sslContextBuilder: SslContextBuilder, enableHttp2: Boolean = false): SslContext = {
+    if (enableHttp2) {
```
smehta91, line 55: Test is missing for this condition.

zio-http/src/main/scala/zhttp/service/server/ServerSSLBuilder.scala
```diff
@@ -0,0 +1,82 @@
...
+      sslContextBuilder
+        .ciphers(Http2SecurityUtil.CIPHERS, SupportedCipherSuiteFilter.INSTANCE)
```
smehta91, line 57: I think we should use `builder pattern` to build the arguments (eg. ciphers, ApplicationProtocolConfig) for`SSLContextBuilder` and that should have `build` method which actually calls `sslContextBuilder`

smehta91 (changes requested):

dinojohn:
on hold for now

PR #976, opened 2022-02-07
first functional version of logging-middleware for ZIO-Http 2.x

gciuloaica:


gciuloaica:
not need it anymore. a different approach will be implemented.

PR #993, opened 2022-02-09
PlainTextBenchmark refactor

ShrutiVerma97:


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -720,6 +726,8 @@ object Http {
     def apply[R, E, B](f: A => HExit[R, E, B]): Http[R, E, A, B] = FromFunctionHExit(f)
   }
 
+  private final case class When[R, E, A, B](f: A => Boolean, other: Http[R, E, A, B]) extends Http[R, E, A, B]
```
amitksingh1490: We can define `IfThenElse` instead and can derive when from it. Let's discuss it.
cc: @tusharmath

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -77,6 +77,11 @@ sealed trait Http[-R, +E, -A, +B] extends (A => ZIO[R, Option[E], B]) { self =>
    */
   final def apply(a: A): ZIO[R, Option[E], B] = execute(a).toZIO
 
+  /**
+   * Applies Http based only if the condition function evaluates to true
+   */
+  final def when[R1 <: R, E1 >: E, A2 <: A, B1 >: B](f: A2 => Boolean): Http[R, E, A2, B] = Http.When(f, self)
+
```
amitksingh1490, line 90: Add HttpApp specific operators like `whenPath`, `whenHeaders` and `whenRequest`

example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -21,15 +21,17 @@ object Main extends App {
     .withServer(STATIC_SERVER_NAME)
     .freeze
 
+  val path = "/plaintext"
+
+  def app(res: Response) = Http.succeed(res).when((a: Request) => a.url.path.toString.equals(path))
```
amitksingh1490: use `whenPath` instead

amitksingh1490 (approved):

example/src/main/scala/example/PlainTextBenchmarkServer.scala
amitksingh1490: as per below comment

amitksingh1490 (changes requested):

zio-http/src/main/scala/zhttp/http/Http.scala
tusharmath: We aren't using all the type params R1, E1 etc., why have them?

tusharmath: We should have two overloads for `whenPath` one should take a `Path` and one should take a `String`.

zio-http/src/main/scala/zhttp/http/Http.scala
tusharmath: We can use `Http.fromFunction[A]` instead I think.

PR #1002, opened 2022-02-10
Refactor: Remove Collect and FromFunctionZIO

### LINKED ISSUE #1000: Remove FromFunctionZIO
it can be implemented using FromFunctionHExit

use FromFunctionHExit

**Why is it important?**
performance improvement

dinojohn:
closes #1000


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -392,9 +392,7 @@ sealed trait Http[-R, +E, -A, +B] extends (A => ZIO[R, Option[E], B]) { self =>
       case Http.Identity        => HExit.succeed(a.asInstanceOf[B])
       case Succeed(b)           => HExit.succeed(b)
       case Fail(e)              => HExit.fail(e)
```
tusharmath, line 394: Can drop these also.

tusharmath (approved):

PR #1012, opened 2022-02-10
Perf: URL refactor

ShrutiVerma97:


example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -28,7 +28,9 @@ object Main extends App {
       .exitCode
   }
 
-  private def app(response: Response) = Http.response(response)
+  private def app(response: Response) = Http.collect[Request] { case _ =>
```
amitksingh1490: not in the scope of this PR

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -1,167 +1,143 @@
...
-  fragment: Option[Fragment] = None,
-) { self =>
+sealed trait URL { self =>
 
-  def encode: String = URL.encode(self)
+  def getHost: Option[String]   = self.toAbsolute.host
+  def getPort: Option[Int]      = self.toAbsolute.port
+  def getScheme: Option[Scheme] = self.toAbsolute.scheme
+  def getRelative: Relative     = self.toAbsolute.relative
+  def getPath: Path             = getRelative.path
```
amitksingh1490: drop get from the function names

zio-http/src/main/scala/zhttp/http/Scheme.scala
```diff
@@ -48,6 +48,13 @@ object Scheme       {
     case _       => None
   }
 
+  def decode2(scheme: String): Scheme = scheme.length match {
```
amitksingh1490: this implementation can introduce bugs

amitksingh1490 (changes requested):

ShrutiVerma97:
Final benchmarks:
```
[info] Benchmark                                  Mode  Cnt       Score        Error  Units
[info] URLParserBenchmark.benchmarkURLOldParser  thrpt    3  374112.291 ? 298723.960  ops/s
[info] URLParserBenchmark.benchmarkURLParser     thrpt    3  437285.905 ? 278341.194  ops/s
```
Improvement: 16%

zio-http/src/main/scala/zhttp/http/Request.scala
```diff
@@ -78,7 +79,7 @@ trait Request extends HeaderExtension[Request] { self =>
   /**
    * Overwrites the path in the request
    */
-  def setPath(path: Path): Request = self.copy(url = self.url.copy(path = path))
+  def setPath(path: Path): Request = self.copy(url = self.url.toAbsolute.copy(urlRelative = Relative(path)))
```
tusharmath:
```suggestion
  def setPath(path: Path): Request = self.copy(url = self.url.setPath(path))
```

zio-http/src/main/scala/zhttp/http/Scheme.scala
tusharmath: - Should be called as unsafeDecode
- Should be package private

zio-http/src/main/scala/zhttp/http/Scheme.scala
```diff
@@ -48,6 +48,13 @@ object Scheme       {
...
+    case 5 => HTTPS
+    case 3 => WSS
+    case 2 => WS
+    case _ => HTTP
```
tusharmath:
```suggestion
    case _ => null
```

PR #1020, opened 2022-02-11
Feat: add support for trailing slash

### LINKED ISSUE #919: No way to determine if a request path ends with a trailing slash
**Describe the bug**
The trailing slash seems to be removed. The application can't then determine if the request URL ended with a trailing slash. If there's a redirect adding a trailing slash (so that relative references in an html page work correctly), this causes an infinite lop

**To Reproduce**
```scala
import zio._
import zhttp.http._
import zhttp.service.Server

object Test2 extends App {
  val app = Http.collect[Request] { case r => Response.text(r.url.toString) }

    Server.start(8090, app).exitCode
}

/*
$ curl http://localhost:8090/a/b/c
URL(/a/b/c,Relative,Map(),None)

$ curl http://localhost:8090/a/b/c/
URL(/a/b/c,Relative,Map(),None)
*/
```

**Expected behavior**
I'd expect to be able to determine if there was a trailing slash in the second request.

1.0.0-RC23

ShrutiVerma97:
fixes #919

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/PathModule.scala
+++ b/zio-http/src/main/scala/zhttp/http/PathModule.scala
@@ -70,9 +70,7 @@ private[zhttp] trait PathModule { module =>
     def apply(string: String): Path     = if (string.trim.isEmpty) End
     else {
       val path = Path(string.split("/").toList)
-      if (string.endsWith("/")) {
-        path.setTrail
-      } else path
+      if (string.endsWith("/")) path.setTrail else path
     }
     def apply(seqString: String*): Path = Path(seqString.toList)
     def apply(list: List[String]): Path = list.foldRight[Path](End)((s, a) => a.append(s))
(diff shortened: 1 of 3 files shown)


tusharmath:
What would happen if we created a path like this — 

```scala
val a = Cons("a", Root, true)
val b = Cons("b", a, true)
b.encode
```

My guess is that there will multiple slashes in the final output.
Correct me if I am wrong.

ShrutiVerma97:
> What would happen if we created a path like this —
> 
> ```scala
> val a = Cons("a", Root, true)
> val b = Cons("b", a, true)
> b.encode
> ```
> 
> My guess is that there will multiple slashes in the final output. Correct me if I am wrong.

Output is `/b/a`
We are making use of `trailingSlash` info in URL encode only.

ShrutiVerma97:
We need to treat 
"/a/" and "/a" as two different routes.

PR #1033, opened 2022-02-15
Remove tuple from ResponseHandler

amitksingh1490:


zio-http/src/main/scala/zhttp/service/server/content/handlers/ServerResponseHandler.scala
```diff
@@ -2,56 +2,39 @@ package zhttp.service.server.content.handlers
...
 
 import java.io.File
 
 @Sharable
-private[zhttp] case class ServerResponseHandler[R](
-  runtime: HttpRuntime[R],
-  config: Server.Config[R, Throwable],
-  serverTime: ServerTimeGenerator,
-) extends SimpleChannelInboundHandler[(Response, FullHttpRequest)](false) {
+private[zhttp] trait ServerResponseHandler[R] {
```
tusharmath, line 16: convert to abstract class.

PR #1036, opened 2022-02-15
doc: socket

girdharshubham:
Adds in website documentation for Socket

### DIFF
--- a/docs/website/docs/v1.x/dsl/socket/socket.md
+++ b/docs/website/docs/v1.x/dsl/socket/socket.md
@@ -87,7 +87,7 @@ val socket = s1 merge s2
 
 ### `map` over a Socket's output channel
 
-Socket is a monad, so you can use `map` to transform the output of a Socket from type `Socket[R, E, A, B]` to type `Socket[R, E, A, C]`, it takes a function from `B => Socket[R, E, A, C]`.
+Socket is a domain, so you can use `map` to transform the output of a Socket from type `Socket[R, E, A, B]` to type `Socket[R, E, A, C]`, it takes a function from `B => Socket[R, E, A, C]`.
 
 ```scala
 val sc     = Socket.succeed("Hello, from ZIO-HTTP")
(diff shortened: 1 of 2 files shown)


docs/website/docs/v1.x/dsl/socket/socket.md
```diff
@@ -0,0 +1,166 @@
...
+val s1 = Socket.succeed(WebSocketFrame.text("Hello, from ZIO-HTTP"))
+val s2 = Socket.succeed(WebSocketFrame.text("Welcome to the party"))
+val socket = s1 merge s2
+```
+
+## Transforming Sockets
+
+### `map` over a Socket's output channel
+
+Socket is a monad, so you can use `map` to transform the output of a Socket from type `Socket[R,E,A,B]` to type `Socket[R,E,A,C]`, it takes a function from `B => Socket[R,E,A,C]`.
```
girdharshubham:
```suggestion
Socket is a functor, so you can use `map` to transform the output of a Socket from type `Socket[R,E,A,B]` to type `Socket[R,E,A,C]`, it takes a function from `B => Socket[R,E,A,C]`.
```
Socket by definition looks more like a Functor. Need feedback.

Ref: https://medium.com/beingprofessional/understanding-functor-and-monad-with-a-bag-of-peanuts-8fa702b3f69e

ShrutiVerma97: Use Socket is a `domain` for now to maintain consistency in the docs.

ShrutiVerma97 (approved):

docs/website/docs/v1.x/dsl/socket/socket.md
```diff
@@ -0,0 +1,166 @@
+---
+title: "Socket"
+sidebar_label: "Socket"
+---
+
+`Socket[-R, +E, -A, +B]` models a function from `A` to `ZStream[R, E, B]`. When a value of type `A` is evaluated against a `Socket[R, E, A, B]`, it can either succeed with a `B`, or fail with an `E`, and it could have its requirement on `R`.
```
amitksingh1490:
```suggestion
`Socket[-R, +E, -A, +B]` models a function from `A` to `ZStream[R, E, B]`. When a value of type `A` is evaluated against a `Socket[R, E, A, B]`, it can either succeed with a stream of values of type `B`, or fail with an `E`, and it could have its requirement on `R`.
```

amitksingh1490 (approved):

PR #1039, opened 2022-02-16
Fix: Handle HttpClient Interruption on connection close

gciuloaica:
…ssage in case the netty chanel is getting closed due to a unhandled exception


zio-http/src/main/scala/zhttp/service/client/ClientInboundHandler.scala
```diff
@@ -28,15 +28,15 @@ final class ClientInboundHandler[R](
...
+    zExec.unsafeRun(ctx)(promise.succeed(ClientResponse.unsafeFromJResponse(msg)).uninterruptible)
     if (isWebSocket) {
       ctx.fireChannelRead(msg.retain())
       ctx.pipeline().remove(ctx.name()): Unit
     }
   }
 
   override def exceptionCaught(ctx: ChannelHandlerContext, error: Throwable): Unit = {
-    zExec.unsafeRun(ctx)(promise.fail(error))
+    zExec.unsafeRun(ctx)(promise.fail(error).uninterruptible)
```
tusharmath, line 41: Is it possible to write a test case for this?

tusharmath: If it's hard to test then we should write a note about what is the issue and why are we making it interruptible.

gciuloaica: Added note as the test quite hard to be written.

tusharmath (approved):

PR #1042, opened 2022-02-16
enhanced client  Part 1 Introduced Types for DefaultClient / Manager / State

sumawa:
## Objective 
Introduce types for client enhancement
Introduce Client.make which returns a default implementation of the new client

### Client (Existing)
Added only one new “make” to existing Client that returns concrete DefaultClient

```
def make(client: ClientSettings = ClientSettings.defaultSetting): Task[DefaultClient] = {
 val settings = client.settings()
 for {
   connectionManager <- ClientConnectionManager(settings)
   clientImpl = DefaultClient(connectionManager)
 } yield clientImpl
}
```

### ClientSettings
Various configuration settings to be used for implementing connection pool and various timeouts.

### ClientConnectionState
Defines data structure and case classes for maintaining connection pool
~~Or build a new channel and attach new request / promise and return response using Inbound Handler~~

### DefaultClient
Depends on ConnectionManager to obtain next available connection from the pool
Exposes variation of APIs to be used by end users like 
```
def run(url: URL): Task[Response] = ???
/** Submits a request and returns the response status */
def status(req: Request): Task[Status] = ???
def stream(req: Request): ZStream[Any, Throwable, ByteBuf] = ???
```

Maintains backward compatibility with existing ClientRequest and ClientResponse

### ClientConnectionManager
Set of functions that uses underlying connection pool to retrieve existing idle channel 
Or build a new channel and attach new request / promise and return response

### TransportFactory
Transport Factory : Factory for various types of transports like "kQueue", "epoll", etc. used to retrieve ServerChannel, ClientChannel and EventLoopGroup combinations


sumawa:
Replaced by #1119

PR #1043, opened 2022-02-16
enhanced client Part 2 : one connection per request and NewHttpRunnableSpec

sumawa:
- Added one request per connection implementation using attachHandler in ClientConnectionManager
- Added NewHttpRunnableSpec and EnhancedClientSpec


sumawa:
Not needed.

PR #1050, opened 2022-02-17
Jmh Benchmark comparison

ShrutiVerma97:
Main without optimisation
```
[info] Benchmark                                   Mode  Cnt          Score          Error  Units
[info] HttpExecuteBenchmark.benchmarkHttpProgram  thrpt    3  246985859.821 ? 75279200.251  ops/s
```


ShrutiVerma97:
First optimisation: added `fromHExit` primitive in `Http`
```
[info] Benchmark                                   Mode  Cnt          Score          Error  Units
[info] HttpExecuteBenchmark.benchmarkHttpProgram  thrpt    3  300162057.758 ? 59037422.487  ops/s
```

PR #1052, opened 2022-02-17
feat: http server response compression

girdharshubham:
Adds in support for Http Response Compression.

Depends on: #1053

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/server/content/compression/CompressionOptions.scala
+++ b/zio-http/src/main/scala/zhttp/service/server/content/compression/CompressionOptions.scala
@@ -1,0 +1,29 @@
+package zhttp.service.server.content.compression
+
+import io.netty.handler.codec.compression.{CompressionOptions => JCompressionOptions, StandardCompressionOptions}
+
+sealed trait CompressionOptions { self =>
+  def toJava: JCompressionOptions = self match {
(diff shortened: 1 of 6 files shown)


zio-http/src/main/scala/zhttp/service/server/content/compression/CompressionOptions.scala
```diff
@@ -0,0 +1,29 @@
...
+sealed trait CompressionOptions { self =>
+  def toJava: JCompressionOptions = self match {
+    case CompressionOptions.GZip(cl, wb, ml)    => StandardCompressionOptions.gzip(cl, wb, ml)
+    case CompressionOptions.Deflate(cl, wb, ml) => StandardCompressionOptions.deflate(cl, wb, ml)
+  }
+}
+
+object CompressionOptions {
+  private final case class GZip(compressionLevel: Int, windowBits: Int, memLevel: Int)    extends CompressionOptions
+  private final case class Deflate(compressionLevel: Int, windowBits: Int, memLevel: Int) extends CompressionOptions
```
tusharmath, line 14:
```scala
case class CompressionOptions(level: Int, bits: Int, mem: Int, kind: Kind) {
  def toJava: JCompressionOptions = ???
}
sealed trait Kind
object GZIp extends Kind
object Deflate extends Kind


```

PR #1058, opened 2022-02-18
Feature: Added helper checkers to `Status` class

rleibman:


tusharmath:
Can you rebase from `main` we now have a `.code` operator so you don't have to do `asJava.code` anymore.

zio-http/src/main/scala/zhttp/http/Status.scala
```diff
@@ -4,6 +4,13 @@ import io.netty.handler.codec.http.HttpResponseStatus
 
 sealed trait Status extends Product with Serializable { self =>
 
+  lazy val isInformational: Boolean = asJava.code() >= 100 && asJava.code() < 200
+  lazy val isSuccess: Boolean = asJava.code() >= 200 && asJava.code() < 300
+  lazy val isRedirection: Boolean = asJava.code() >= 300 && asJava.code() < 400
+  lazy val isClientError: Boolean = asJava.code() >= 400 && asJava.code() < 500
+  lazy val isServerError: Boolean = asJava.code() >= 500 && asJava.code() < 600
+  lazy val isError: Boolean = isClientError | isServerError
```
tusharmath: I think we should use `def` here `lazy val` will add additional complexity around thread safety using `synchronized` blocks. Since these are primitive int operations that can be computed very fast let's use `def`.


Do you mind also adding some unit tests and scala doc :)

tusharmath:
Can you also run `fmt` command from sbt. The CI is complaining about formatting.

rleibman:
How about now?

tusharmath (approved):

tusharmath:
Thanks @rleibman !

PR #850, opened 2022-01-17
Feature : Implement a static file server

ashprakasan:
Fixes #830

I am thinking to complete this feature in multiple PRs, as it will be more productive to merge things as and when they're ready.

The following features are implemented here -

- [x] METHOD_NOT_ALLOWED response if `Http` method is not GET
- [x] NOT-FOUND if the file does not exist
- [x] FORBIDDEN if there are permission issues on the file
- [x] INTERNAL SERVER ERROR in case some other exception is thrown
- [x] Show a listing of all files for a directory path

Some features that will be done in upcoming PRs -

Support for `config` or `options` - [#862](https://github.com/dream11/zio-http/issues/862)
Use templating library for listing files in directory in `listFilesHtml` - [#863](https://github.com/dream11/zio-http/issues/863)

PR #1061, opened 2022-02-19
Feature: Static Server

### LINKED ISSUE #830: Enhancement : Build a static file server

`Http.fromFile(file: java.io.File)` is currently incomplete. For anybody to be able to successfully use it out of the box, the following features need to be supported -

1. `BAD_REQUEST` response in case of invalid request
2. `METHOD_NOT_ALLOWED` response if `Http method` is not `GET`
3.  `NOT-FOUND` if the file does not exist
4. `FORBIDDEN` if there are permission issues on the file
5.  `INTERNAL SERVER ERROR` in case some other exception is thrown
6. Cache validation and correct handling of `IF_MODIFIED_SINCE` header.
  
The following should be appropriately handled - 

1. `BAD_REQUEST` response in case of invalid request (This will apply to all requests, not just `Http.fromFile`)
2. `METHOD_NOT_ALLOWED` response if `Http method` is not `GET`
3.  `NOT-FOUND` if the file does not exist
4. `FORBIDDEN` if there are permission issues on the file
5.  `INTERNAL SERVER ERROR` in case some other exception is thrown
6. Cache validation and correct handling of `IF_MODIFIED_SINCE` header on request

The developer using `zio-http` will have to handle these scenarios explicitly, which does not provide great devx.

**Additional context**
Implementation of Static File Server on netty - https://github.com/netty/netty/tree/4.1/example/src/main/java/io/netty/example/http/file

tusharmath:
Continuation of #1024 #850 
Closes #830


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -574,11 +610,27 @@ object Http {
...
+
   /**
    * Creates an Http app from a resource path
    */
-  def fromResource(path: String): HttpApp[Any, Throwable] =
-    Http.fromFile(new File(getClass.getResource(path).getPath))
+  def fromResource(path: String): HttpApp[Any, Throwable] = {
+    for {
+      path <- Http
+        .attempt(getClass.getResource(path) match {
```
guizmaii: In my implementation (inspired by yours), `getClass.getResource(path)` was returning `null`, I had to use `getClass.getClassLoader.getResource(path)`

amitksingh1490 (approved):

amitksingh1490 (approved):

PR #1071, opened 2022-02-20
Feature: add `withMediaType` operator in `HeaderModifier`

tusharmath:


zio-http/src/main/scala/zhttp/http/headers/HeaderModifier.scala
```diff
@@ -177,6 +177,8 @@ trait HeaderModifier[+A] { self =>
   final def withMaxForwards(value: CharSequence): A =
     addHeaders(Headers.maxForwards(value))
 
+  def withMediaType(mediaType: MediaType): A = self.withContentType(mediaType.fullType)
```
amitksingh1490, line 180: unit test, use gen if possible.

amitksingh1490 (approved):

PR #1080, opened 2022-02-21
Query ergonomics

gciuloaica:
Implementation that supports:
- up to 22 parameters in Case Class
- unlimited cases in a sum type - but provides support only for `case object`
- primitives could be wrapped in `Option` or `List`


zio-http/src/main/scala/zhttp/http/QueryParameters.scala
```diff
@@ -0,0 +1,15 @@
+package zhttp.http
+
+final case class QueryParameters(raw: Map[String, List[String]]) { self =>
```
smehta91, line 3: This should use `DynamicValue` of ZSchema

zio-http/src/main/scala/zhttp/http/QueryParameters.scala
```diff
@@ -0,0 +1,15 @@
...
+  def ++(other: QueryParameters): QueryParameters = self.combine(other)
+
+  def combine(other: QueryParameters): QueryParameters = QueryParameters(self.raw ++ other.raw)
+
+  def queryParams: QueryParameters = self
+}
```
smehta91, line 9:
```suggestion
final case class QueryParameters(raw: ListMap[String, DynamicValue]) { self =>
  def ++(other: QueryParameters): QueryParameters = self.combine(other)

  def combine(other: QueryParameters): QueryParameters = QueryParameters(self.raw ++ other.raw)

  def queryParams: QueryParameters = self
}

object QueryParameters {

  val empty: QueryParameters = QueryParameters(ListMap.empty[String, DynamicValue])

}
```

zio-http-query-params/src/main/scala/zhttp/http/query/QueryParams.scala
```diff
@@ -0,0 +1,1669 @@
...
+
+import java.time._
+import java.util.UUID
+import scala.util.Try
+
+object QueryParams {
+
+  implicit class QueryParametersWrapper(queryParams: QueryParameters) {
+    def decode[A](implicit schema: Schema[A]): Either[String, A] =
+      CaseClassDecoder.decode(schema, queryParams.raw)
```
smehta91, line 15: `schema.fromDynamic(DynamicValue.Record(queryParams.raw))`

smehta91 (changes requested):
We can build `DynamicValue` using QueryParameter domain and while decoding we can decode typed value if we have schema for the type

gciuloaica:
new API in progress https://github.com/dream11/zio-http/pull/1209

PR #1081, opened 2022-02-22
Fix: Close connection when thrown inside handler

### LINKED ISSUE #1037: Connection not closed when exception thrown directly in request handler
With an http handler:

```scala
    val app = Http.collectZIO[Request] {
      case Method.GET -> !! / "fail" =>
        throw new Exception("GET Failed!")
    }
```

Call the `/fail` route (say via) curl. An exception is written to the console of the server but the connection is never closed and th client waits until the server is shutdown.

**Expected behaviour**
The expected behaviour would be to return some sort of 500 error to the client. Instead the connection remains open which is confusing.

**Setup**
 - scala 2.13.8
 - zhttp 2.0.0-RC3

dinojohn:
closes #1037

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/Handler.scala
+++ b/zio-http/src/main/scala/zhttp/service/Handler.scala
@@ -51,7 +51,7 @@ private[zhttp] final case class Handler[R](
         writeResponse(
           Response
             .fromHttpError(HttpError.InternalServerError(cause = Some(throwable)))
-            .addHeader((HeaderNames.connection, HeaderValues.close)),
+            .withConnection(HeaderValues.close),
           jReq,
         ): Unit
     }
--- a/zio-http/src/main/scala/zhttp/service/Handler.scala
+++ b/zio-http/src/main/scala/zhttp/service/Handler.scala
@@ -48,8 +48,12 @@ private[zhttp] final case class Handler[R](
       )
     catch {
       case throwable: Throwable =>
-        config.error.fold(())(f => runtime.unsafeRun(ctx)(f(throwable)))
-        writeResponse(Response.fromHttpError(HttpError.InternalServerError(cause = Some(throwable))), jReq): Unit
+        writeResponse(
+          Response
+            .fromHttpError(HttpError.InternalServerError(cause = Some(throwable)))
+            .addHeader((HeaderNames.connection, HeaderValues.close)),
+          jReq,
+        ): Unit
     }
   }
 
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -23,26 +23,33 @@ private[zhttp] final case class Handler[R](
...
+            }
+
+            override def data: HttpData = HttpData.fromByteBuf(jReq.content())
+          },
+        ),
+      )
+    catch {
+      case throwable: Throwable =>
+        writeResponse(Response.fromHttpError(HttpError.InternalServerError(cause = Some(throwable))), jReq): Unit
+    }
```
tusharmath, line 59: - We should call the `Config.onError` handler if available
- If handler not available we should call `ctx.fireExceptionCaught`
- Connection is still not getting closed here?

dinojohn: If Config.error is not present, Connection is still not getting closed, Unless we write a 500 response or explicitly close the connection by `ctx.close()` .

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -23,26 +23,34 @@ private[zhttp] final case class Handler[R](
...
+              }
+            }
+
+            override def data: HttpData = HttpData.fromByteBuf(jReq.content())
+          },
+        ),
+      )
+    catch {
+      case throwable: Throwable =>
+        config.error.fold(())(f => runtime.unsafeRun(ctx)(f(throwable)))
```
tusharmath:
```suggestion
```

zio-http/src/main/scala/zhttp/service/Handler.scala
tusharmath: Add a `connection:close` header. This will be understood automatically by the `Keep Alive` handle added to the channel pipeline.

zio-http/src/test/scala/zhttp/service/ServerSpec.scala
```diff
@@ -29,13 +29,17 @@ object ServerSpec extends HttpRunnableSpec {
     case Method.GET -> !! / "get%2Fsuccess" => ZIO.succeed(Response.ok)
   }
 
+  private val throwableApp: Http[Any, Nothing, Request, Nothing] = Http.collectZIO[Request] {
```
tusharmath: Define the Http inside the test. Deploy and run then assert the status code.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -23,26 +23,38 @@ private[zhttp] final case class Handler[R](
...
+            override def data: HttpData = HttpData.fromByteBuf(jReq.content())
+          },
+        ),
+      )
+    catch {
+      case throwable: Throwable =>
+        writeResponse(
+          Response
+            .fromHttpError(HttpError.InternalServerError(cause = Some(throwable)))
+            .addHeader((HeaderNames.connection, HeaderValues.close)),
```
tusharmath:
```suggestion
            .withConnection(HeaderValues.close),
```

tusharmath (approved):

dinojohn:
closing in favour of #1099

PR #1087, opened 2022-02-23
Query ergonomics macro

gciuloaica:
Added macro implementation to not expose zio-schema into users api.


gciuloaica:
new API in Progress https://github.com/dream11/zio-http/pull/1209

PR #1091, opened 2022-02-24
Added json response in benchmark

ShrutiVerma97:
Benchmark code:
```scala
package zhttp.benchmarks

import org.openjdk.jmh.annotations._
import zhttp.http._

import java.util.concurrent.TimeUnit

@State(Scope.Thread)
@BenchmarkMode(Array(Mode.Throughput))
@OutputTimeUnit(TimeUnit.SECONDS)
class HttpRouteTextPerf {

  private val res          = Response.text("HELLO WORLD")
  private val app          = Http.fromHExit(HExit.succeed(res)).whenPathEq("/text") ++ Http
    .fromHExit(HExit.succeed(res))
    .whenPathEq("/plain")
  private val req: Request = Request(Method.GET, URL(!! / "text"))

  @Benchmark
  def benchmarkHttpProgram(): Unit = {
    val _ = app.execute(req)
    ()
  }

}
```

jmh benchmark results
```
[info] Benchmark                                Mode  Cnt        Score         Error  Units
[info] HttpRouteTextPerf.benchmarkHttpProgram  thrpt    3  4052918.714 ? 1315668.386  ops/s
```

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -209,7 +209,7 @@ sealed trait Http[-R, +E, -A, +B] extends (A => ZIO[R, Option[E], B]) { self =>
     dd: Http[R1, E1, A1, B1],
   ): Http[R1, E1, A1, B1] = Http.FoldHttp(self, ee, bb, dd)
 
-  final def composeHttp[R1 <: R, A1 <: A, E1, B1](other: Http[R1, E1, A1, B1]): Http[R1, E1, A1, B1] =
+  final def composeHttp[R1 <: R, A1 <: A, E1 >: E, B1 >: B](other: Http[R1, E1, A1, B1]): Http[R1, E1, A1, B1] =
     Http.Compose(self, other)
 
   /**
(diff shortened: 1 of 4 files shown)


tusharmath:
Can we improve performance by using a special `Combine` operator that isn't built on `fold`?

ShrutiVerma97:
> Can we improve performance by using a special `Combine` operator that isn't built on `fold`?

After adding composeHttp:
```
[info] Benchmark                                Mode  Cnt        Score         Error  Units
[info] HttpRouteTextPerf.benchmarkHttpProgram  thrpt    3  7615488.786 ? 3227764.384  ops/s
```

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -843,6 +865,11 @@ object Http {
 
   private case class Attempt[A](a: () => A) extends Http[Any, Nothing, Any, A]
 
+  private final case class Compose[R, E, EE, A, B, BB](
```
```suggestion
  private final case class Combine[R, E, EE, A, B, BB](
```

Also this is faster than our current `++`.

ShrutiVerma97:
Benchmark with HttpCombineEval(++):
```
[info] Benchmark                           Mode  Cnt      Score       Error  Units
[info] HttpCombineEval.benchmarkNotFound  thrpt    3  65230.423 ? 10185.351  ops/s
[info] HttpCombineEval.benchmarkOk        thrpt    3  42583.211 ?  7217.063  ops/s
```

Benchmark with HttpComposeHttpEval(composeHttp):

```
[info] Benchmark                           Mode  Cnt       Score       Error  Units
[info] HttpCombineEval.benchmarkNotFound  thrpt    3   77365.635 ? 56534.469  ops/s
[info] HttpCombineEval.benchmarkOk        thrpt    3  106470.738 ? 51292.490  ops/s
```

PR #1101, opened 2022-02-27
setCookiesDecode uses wrong date format

dhyanpatel:
https://github.com/dream11/zio-http/issues/1093

### DIFF
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
@@ -3,8 +3,7 @@ package zhttp.benchmarks
 import org.openjdk.jmh.annotations._
 import zhttp.http._
 
-import java.time.format.DateTimeFormatter
-import java.time.{Instant, ZoneOffset}
+import java.time.Instant
 import java.util.concurrent.TimeUnit
 
 @State(Scope.Thread)
@@ -21,7 +20,7 @@ class CookieDecodeBenchmark {
   private val cookie    = Cookie(
     name,
     value,
-    Some(DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC).format(Instant.now)),
+    Some(Instant.now()),
     Some(domain),
     Some(path),
     true,
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
@@ -3,7 +3,8 @@ package zhttp.benchmarks
 import org.openjdk.jmh.annotations._
 import zhttp.http._
 
-import java.time.Instant
+import java.time.format.DateTimeFormatter
+import java.time.{Instant, ZoneOffset}
 import java.util.concurrent.TimeUnit
 
 @State(Scope.Thread)
@@ -20,7 +21,7 @@ class CookieDecodeBenchmark {
   private val cookie    = Cookie(
     name,
     value,
-    Some(Instant.now()),
+    Some(DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC).format(Instant.now)),
     Some(domain),
     Some(path),
     true,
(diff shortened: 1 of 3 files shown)


dhyanpatel:
Hey! This is my first time contributing to open source, but was hoping to help with the project and learn about open source contributions. Let me know if there's anything I need to change!

I'm still working on getting the unit test for response cookies to work.

You can also reach me on discord at Aethen#8828 (I'm on the ZIO discord server)

zio-http-benchmarks/src/main/scala/zhttp.benchmarks/CookieDecodeBenchmark.scala
```diff
@@ -20,7 +21,7 @@ class CookieDecodeBenchmark {
   private val cookie    = Cookie(
     name,
     value,
-    Some(Instant.now()),
+    Some(DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC).format(Instant.now)),
```
tusharmath: Keep the formatter inside `Cookie` companion object, something like this —

```scala
object Cookie {
  val formater = DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC)
}
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -12,7 +13,7 @@ import scala.util.Try
 final case class Cookie(
   name: String,
   content: String,
-  expires: Option[Instant] = None,
+  expires: Option[String] = None,
```
tusharmath: We should keep time as an `Instant` only. The memory usage is much lower and modification is easier and doesn't require us to parse it multiple times.

Finally when we encode the cookie as a `String` that's where we should use the formatter.

dhyanpatel: Sounds good! Took another shot at this. Still not able to figure out this one unit test issue though.

dhyanpatel:
Not sure why this error is happening:
```
java.lang.IllegalArgumentException: Empty key
  	at java.base/javax.crypto.spec.SecretKeySpec.<init>(SecretKeySpec.java:110)
```
An empty key is being passed to SecretKeySpec. In the debugger I found it's because an empty string is being set as the `secret` of the Cookie for one of the cookies generated by `HttpGen.cookies`.

However, I'm not sure why this is happening, as I haven't touched the `secret` parameter in this PR. 

Potential solution:
```scala
  def encode: String = {
    val c = secret match {
      case Some(sec) if sec.nonEmpty => content + "." + signContent(sec)
      case None | Some(_)    => content
    }
```

zio-http/src/main/scala/zhttp/http/Cookie.scala
```diff
@@ -233,7 +238,7 @@ object Cookie {
           domain = headerValue.substring(curr + 7, next)
         } else if (headerValue.regionMatches(true, curr, fieldPath, 0, fieldPath.length)) {
           val v = headerValue.substring(curr + 5, next)
-          if (!v.isEmpty) {
+          if (v.nonEmpty) {
```
dhyanpatel, line 241: Changed these since I was already working in the file, and v.nonEmpty is more readable than !v.isEmpty

kaushik143:
> Not sure why this error is happening:
> 
> ```
> java.lang.IllegalArgumentException: Empty key
>   	at java.base/javax.crypto.spec.SecretKeySpec.<init>(SecretKeySpec.java:110)
> ```
> 
> An empty key is being passed to SecretKeySpec. In the debugger I found it's because an empty string is being set as the `secret` of the Cookie for one of the cookies generated by `HttpGen.cookies`.
> 
> However, I'm not sure why this is happening, as I haven't touched the `secret` parameter in this PR.
> 
> Potential solution:
> 
> ```scala
>   def encode: String = {
>     val c = secret match {
>       case Some(sec) if sec.nonEmpty => content + "." + signContent(sec)
>       case None | Some(_)    => content
>     }
> ```

I have created a separate issue for this https://github.com/dream11/zio-http/issues/1114. Would you like to give it a shot?

kaushik143:
Also on this PR you have formatting issue , please run `fmt` and push the changes

dhyanpatel:
Note: Blocked by #1114 
Will continue after that is resolved

gciuloaica (approved):
Looks good to me.

gciuloaica:
new PR open based on solution from this PR on top of latest Cookie implementation changes. #1288

PR #1104, opened 2022-02-28
Netty layer logging

gciuloaica:


build.sbt
```diff
@@ -102,6 +102,7 @@ lazy val zhttp = (project in file("zio-http"))
       `zio-test-sbt`,
       `netty-incubator`,
       `scala-compact-collection`,
+      logback,
```
tusharmath, line 105: - Can we use `netty`'s Logging capabilities?
- When we add internal logging, is there an additional overhead of String allocations if it's disabled?

gciuloaica: We are using netty's logging capabilities, but netty's logging capabilities is just a facade over slf4j, log4j or jdk logging, so it still needs a front-end logger like logback.
If logging is disabled there is no overhead.

gciuloaica:
Closing this PR as a new implementation will added.

PR #1106, opened 2022-02-28
Enhancement: Added combine operator

ShrutiVerma97:
Currently, ++ is implemented using `FoldHttp` primitive, I've added `Combine` primitive which improves performance(as mentioned below)
HttpCombineEval Benchmark results:

Main branch:
```
[info] Benchmark                           Mode  Cnt      Score       Error  Units
[info] HttpCombineEval.benchmarkNotFound  thrpt    3  70211.931 ? 18985.369  ops/s
[info] HttpCombineEval.benchmarkOk        thrpt    3  39266.153 ? 96066.575  ops/s
```

Current Branch:
```
[info] Benchmark                           Mode  Cnt      Score       Error  Units
[info] HttpCombineEval.benchmarkNotFound  thrpt    3  81206.263 ? 11267.440  ops/s
[info] HttpCombineEval.benchmarkOk        thrpt    3  91960.203 ? 43519.961  ops/s
```

### DIFF
--- a/example/src/main/scala/example/PlainTextBenchmarkServer.scala
+++ b/example/src/main/scala/example/PlainTextBenchmarkServer.scala
@@ -12,8 +12,18 @@ import zio.{App, ExitCode, UIO, URIO}
 object Main extends App {
 
   private val message: String = "Hello, World!"
+  private val json: String    = """{"greetings": "Hello World!"}"""
+
+  val path1 = "/plaintext"
+  val path2 = "/json"
 
   private val STATIC_SERVER_NAME = AsciiString.cached("zio-http")
+
+  private val frozenJsonResponse = Response
+    .json(json)
+    .withServerTime
+    .withServer(STATIC_SERVER_NAME)
+    .freeze
 
   private val frozenResponse = Response
     .text(message)
@@ -22,16 +32,20 @@ object Main extends App {
     .freeze
 
   override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] = {
-    frozenResponse
-      .flatMap(server(_).make.useForever)
-      .provideCustomLayer(ServerChannelFactory.auto ++ EventLoopGroup.auto(8))
-      .exitCode
+    val s = for {
+      res1 <- frozenResponse
+      res2 <- frozenJsonResponse
+    } yield server(res1, res2)
+    s.flatMap(_.make.useForever.provideCustomLayer(ServerChannelFactory.auto ++ EventLoopGroup.auto(8))).exitCode
   }
-  private val path                                               = "/plaintext"
-  private def app(response: Response) = Http.fromHExit(HExit.succeed(response)).whenPathEq(path)
 
-  private def server(response: Response) =
-    Server.app(app(response)) ++
+  private def app(response: Response, json: Response) =
+    Http.fromHExit(HExit.succeed(response)).whenPathEq(path1) ++ Http
+      .fromHExit(HExit.succeed(json))
+      .whenPathEq(path2)
+
+  private def server(response: Response, json: Response) =
+    Server.app(app(response, json)) ++
       Server.port(8080) ++
       Server.error(_ => UIO.unit) ++
       Server.disableLeakDetection ++
--- a/example/src/main/scala/example/PlainTextBenchmarkServer.scala
+++ b/example/src/main/scala/example/PlainTextBenchmarkServer.scala
@@ -1,7 +1,7 @@
 package example
 
 import io.netty.util.AsciiString
-import zhttp.http._
+import zhttp.http.{Http, _}
 import zhttp.service.server.ServerChannelFactory
 import zhttp.service.{EventLoopGroup, Server}
 import zio.{App, ExitCode, UIO, URIO}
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -421,6 +422,24 @@ sealed trait Http[-R, +E, -A, +B] extends (A => ZIO[R, Option[E], B]) { self =>
       case RunMiddleware(app, mid) => mid(app).execute(a)
 
       case When(f, other) => if (f(a)) other.execute(a) else HExit.empty
+
+      case Combine(self, other) => {
```
amitksingh1490, line 600: Add tests

example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -12,26 +12,40 @@ import zio.{App, ExitCode, UIO, URIO}
...
     .withServerTime
     .withServer(STATIC_SERVER_NAME)
     .freeze
 
   override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] = {
-    frozenResponse
-      .flatMap(server(_).make.useForever)
-      .provideCustomLayer(ServerChannelFactory.auto ++ EventLoopGroup.auto(8))
-      .exitCode
+    val s = for {
```
amitksingh1490: clean this up

example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -12,26 +12,40 @@ import zio.{App, ExitCode, UIO, URIO}
...
 
-  private def server(response: Response) =
-    Server.app(app(response)) ++
+  private def app(response: Response, json: Response) =
+    Http.fromHExit(HExit.succeed(response)).whenPathEq(path1) ++ Http
+      .fromHExit(HExit.succeed(json))
+      .whenPathEq(path2)
+
+  private def server(response: Response, json: Response) =
+    Server.app(app(response, json)) ++
```
amitksingh1490: compose apps here instead of App function

example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -12,26 +12,40 @@ import zio.{App, ExitCode, UIO, URIO}
...
+    } yield server(res1, res2)
+    s.flatMap(_.make.useForever.provideCustomLayer(ServerChannelFactory.auto ++ EventLoopGroup.auto(8))).exitCode
   }
-  private val path                                               = "/plaintext"
-  private def app(response: Response) = Http.fromHExit(HExit.succeed(response)).whenPathEq(path)
 
-  private def server(response: Response) =
-    Server.app(app(response)) ++
+  private def app(response: Response, json: Response) =
+    Http.fromHExit(HExit.succeed(response)).whenPathEq(path1) ++ Http
```
amitksingh1490: path1 and path2 should have better names, e.g plainText, json

gciuloaica:
LGTM, nothing to add on top of @amitksingh1490 comments.

sumawa:
Changes and test cases looks good. Nothing to add.

sumawa (approved):

gciuloaica (approved):

zio-http/src/test/scala/zhttp/http/HttpSpec.scala
```diff
@@ -136,11 +136,40 @@ object HttpSpec extends DefaultRunnableSpec with HExitAssertion {
...
+            val c      = Http.collect[Int] { case 3 => "C" }
+            val actual = (a ++ b ++ c).execute(3)
+            assert(actual)(isSuccess(equalTo("C")))
+          } +
+          test("should resolve third") {
+            val a      = Http.collectZIO[Int] { case 1 => UIO("A") }
+            val b      = Http.collectZIO[Int] { case 2 => UIO("B") }
+            val c      = Http.collectZIO[Int] { case 3 => UIO("C") }
+            val actual = (a ++ b ++ c).execute(3)
+            assert(actual)(isEffect)
```
girdharshubham: The above test would only test if the result is an effect or not. I think the intent is to test if the output is `C` or not. The above test would pass even for `(a ++ b ++ c).execute(2)`
```suggestion
            assert(actual)(isEffect(equalTo("C")))
```

girdharshubham (changes requested):

girdharshubham (approved):
Looks good to me.

zio-http/src/test/scala/zhttp/http/HttpSpec.scala
amitksingh1490: This needs to be resolved. May be we should run the effect and assert on value.

zio-http/src/test/scala/zhttp/http/HttpSpec.scala
```diff
@@ -176,6 +205,13 @@ object HttpSpec extends DefaultRunnableSpec with HExitAssertion {
             val b      = Http.succeed("B")
             val actual = (a ++ b).execute(2)
             assert(actual)(isSuccess(equalTo("B")))
+          } +
+          test("should resolve third") {
+            val a      = Http.empty.flatten
+            val b      = Http.empty.flatten
```
amitksingh1490: flatten ??
flatten works on Http[Http[]]

zio-http/src/test/scala/zhttp/http/HttpSpec.scala
ShrutiVerma97: This test intends to check if after combining apps, it's reaching the third app that returns an effect. I discussed it with @girdharshubham  and this works

ShrutiVerma97: I can add this test to assert on effect value:
```scala
     testM("should resolve third") {
            val a      = Http.collectZIO[Int] { case 1 => UIO("A") }
            val b      = Http.collectZIO[Int] { case 2 => UIO("B") }
            val c      = Http.collectZIO[Int] { case 3 => UIO("C") }
            val actual = (a ++ b ++ c).execute(3).toZIO
            assertM(actual)(equalTo("C"))
          } 
```

amitksingh1490: Not needed, marking it as resolved

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -843,6 +862,11 @@ object Http {
 
   private case class Attempt[A](a: () => A) extends Http[Any, Nothing, Any, A]
 
+  private final case class Combine[R, E, EE, A, B, BB](
```
girdharshubham, line 865: A better signature for combine could be
```scala
final case class Combine[R, R1 <: R, E, EE >: E, A, B, BB >: B](
  self: Http[R, E, A, B],
  other: Http[R1, EE, A, BB],
) extends Http[R1, EE, A, BB]
```

this could be better represented as:
```scala
val a = Http.collectZIO[Int] {
  case 1 => ZIO.fail(new InterruptedException())
  case 2 => ZIO.environment[Int]
}

val b = Http.collectZIO[Int] {
  case 1 => ZIO.fail(new RuntimeException())
  case 4 => ZIO.environment[String]
}
val r: Combine[Int, String with Int, InterruptedException, Exception, Int, Int, Any] = Combine(a, b)
```

girdharshubham, line 865: A better type signature for `Combine` could be:
```scala
private final case class Combine[R, R1 <: R, E, EE >: E, A, B, BB >: B](
  self: Http[R, E, A, B],
  other: Http[R1, EE, A, BB],
) extends Http[R1, EE, A, BB]
```

girdharshubham, line 865: A better type signature for `Combine` could be:
```scala
private final case class Combine[R, R1 <: R, E, EE >: E, A, B, BB >: B](
  self: Http[R, E, A, B],
  other: Http[R1, EE, A, BB],
) extends Http[R1, EE, A, BB]
```

girdharshubham (changes requested):

zio-http/src/test/scala/zhttp/http/HttpSpec.scala
```diff
@@ -136,11 +136,40 @@ object HttpSpec extends DefaultRunnableSpec with HExitAssertion {
...
             assert(actual)(isSuccess(equalTo("B")))
           } +
+          test("should resolve third") {
+            val a      = Http.collect[Int] { case 1 => "A" }
+            val b      = Http.collect[Int] { case 2 => "B" }
+            val c      = Http.collect[Int] { case 3 => "C" }
+            val actual = (a ++ b ++ c).execute(3)
+            assert(actual)(isSuccess(equalTo("C")))
+          } +
+          test("should resolve third") {
```
amitksingh1490: add 1 more test with collectHttp also

amitksingh1490 (approved):

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -421,6 +422,24 @@ sealed trait Http[-R, +E, -A, +B] extends (A => ZIO[R, Option[E], B]) { self =>
...
+            Effect(
+              zio.foldM(
+                {
+                  case Some(error) => ZIO.fail(Option(error.asInstanceOf[E]))
+                  case None        => other.execute(a).toZIO
+                },
+                b => ZIO.succeed(b.asInstanceOf[B]),
+              ),
+            )
+          }
```
tusharmath: Why can't we do something like 

```scala
self.execute(a) match {
  case HExit.Empty => other.execute(a)
  case hExit => hExit
}
```

zio-http/src/test/scala/zhttp/http/HttpSpec.scala
```diff
@@ -136,11 +136,47 @@ object HttpSpec extends DefaultRunnableSpec with HExitAssertion {
             val actual = (a ++ b).execute(2)
             assert(actual)(isSuccess(equalTo("B")))
           } +
+          test("should resolve third") {
```
tusharmath: - Move tests for `++` into the `combine` suite.
- For testing the `++` use `succeed` and `empty` to build Https, instead of `collect`

zio-http/src/main/scala/zhttp/http/Http.scala
ShrutiVerma97: Currently, in the case of None, the other app executes
IF we return the effect as it is then we won't be handling the failure correctly? And In no case, the other app would execute.

tusharmath (approved):

PR #1109, opened 2022-03-01
Removed flatten from test

ShrutiVerma97:
flatten is not needed to create an empty Http app.


girdharshubham (approved):
Looks good to me.

amitksingh1490 (approved):

PR #1111, opened 2022-03-02
Enhancement: Merge server and client Responses

amitksingh1490:


gciuloaica (approved):
The only change that I find somehow confusing is the `IsResponse` but it is not blocking approval.

amitksingh1490:
> The only change that I find somehow confusing is the `IsResponse` but it is not blocking approval.

I Think you are correct we could get rid of isResponse completely since we have only one type of response now. Making that change

gciuloaica (approved):
Nice!

tusharmath (approved):

PR #1113, opened 2022-03-03
Refactor: Make `Http.getResource` consistent with `ZStream.fromResource`

guizmaii:
Code based on https://github.com/zio/zio/blob/v1.0.13/streams/jvm/src/main/scala/zio/stream/platform.scala#L385-L398

I debugged a request trying to access a `index.html` file present in the `resources` folder:

Screen Shot 2022-03-03 at 9 08 35 pm

As you can see in this debug evaludation window, when I execute `getClass.getResource(path)`, it returns me `null`:
Screen Shot 2022-03-03 at 9 08 47 pm
While when I execute `getClass.getClassLoader.getResource(path)` I get the file:
Screen Shot 2022-03-03 at 9 08 59 pm

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -8,7 +8,7 @@ import zhttp.http.headers.HeaderModifier
 import zhttp.service.server.ServerTime
 import zhttp.service.{Handler, HttpRuntime, Server}
 import zio._
-import zio.blocking.{Blocking, effectBlockingIO}
+import zio.blocking.Blocking
 import zio.clock.Clock
 import zio.duration.Duration
 import zio.stream.ZStream
@@ -859,7 +859,7 @@ object Http {
   /**
    * Creates an Http app from a resource path
    */
-  def fromResource(path: String): HttpApp[Blocking, Throwable] =
+  def fromResource(path: String): HttpApp[Any, Throwable] =
     Http.getResource(path).flatMap(url => Http.fromFile(new File(url.getPath)))
 
   /**
@@ -886,13 +886,13 @@ object Http {
    */
   def getResource(path: String): Http[Any, Throwable, Any, net.URL] =
     Http
-      .fromZIO(effectBlockingIO(getClass.getClassLoader.getResource(path)))
+      .fromZIO(Blocking.Service.live.effectBlockingIO(getClass.getClassLoader.getResource(path)))
       .flatMap { resource => if (resource == null) Http.empty else Http.succeed(resource) }
 
   /**
    * Attempts to retrieve files from the classpath.
    */
-  def getResourceAsFile(path: String): Http[Blocking, Throwable, Any, File] =
+  def getResourceAsFile(path: String): Http[Any, Throwable, Any, File] =
     Http.getResource(path).map(url => new File(url.getPath))
 
   /**
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -860,7 +860,7 @@ object Http {
    * Creates an Http app from a resource path
    */
   def fromResource(path: String): HttpApp[Blocking, Throwable] =
-    Http.getResourceAsFile(path) >>= { Http.fromFile(_) }
+    Http.getResource(path).flatMap(url => Http.fromFile(new File(url.getPath)))
 
   /**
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -669,18 +670,15 @@ object Http {
...
    */
-  def getResource(path: String): Http[Any, Throwable, Any, net.URL] =
+  def getResource(path: String): Http[Blocking, Throwable, Any, net.URL] =
     Http
-      .attempt(Option(getClass.getResource(path)) match {
-        case Some(path) => Http.succeed(path)
-        case None       => Http.empty
-      })
-      .flatten
+      .fromZIO(effectBlockingIO(getClass.getClassLoader.getResource(path)))
```
guizmaii: I'm not sure we need the `effectBlockingIO`, I just copied the code of ZIO

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -644,8 +645,8 @@ object Http {
   /**
    * Creates an Http app from a resource path
    */
-  def fromResource(path: String): HttpApp[Any, Throwable] =
-    Http.getResourceAsFile(path) >>= { Http.fromFile(_) }
+  def fromResource(path: String): HttpApp[Blocking, Throwable] =
+    Http.getResource(path).flatMap(url => Http.fromFile(new File(url.getPath)))
```
guizmaii, line 863: This change is not related to the fix of `Http.getResource`.
It just helps to delay the creation of the `File` instance as much as possible.

guizmaii:
Ok I understand what was my mistake in my code. I need to write `Http.getResource("/index.html")` instead of `Http.getResource("index.html")`

Do we want to keep the actual behaviour (requiring the path to start with a `/`) or do we want to align with ZIO and not have to pass this `/` (it doesn't find the resource if the path starts with `/`)?

amitksingh1490:
Could you please describe why are these changes required?

guizmaii:
@amitksingh1490, well I tried to explain the problem in my different previous comments. 
Here's a summary:

AFAIK, in ZIO (as we can see [here](https://github.com/zio/zio/blob/v1.0.13/streams/jvm/src/main/scala/zio/stream/platform.scala#L385-L398)) when accessing a resource, it's `getClass.getClassLoader.getResource(path)` that is used and not `getClass.getResource(path)`, as used in the current code of zio-http.

The difference for the user is that with the "ZIO solution", `getClass.getClassLoader.getResource`, the user will have to pass a `path` that **does not start with a `/`** while with the current solution in zio-http, `getClass.getResource`, to get the same file the user **has pass a `path` starting with a `/`**.

To acces the same resource, with ZIO you'll have to write:
```scala
ZStream.fromResource("static/index.html")
```
while with zio-http, you'll have to write:
```scala
Http.getResource("/static/index.html")
```

To me, being used to zio, this inconsistency between the two projects makes the behaviour of zio-http surprising:
returning `null` when we forget to add the initial `/` in the `path` we pass to the `Http.getResource`

Also, being used to the Scala `scala.io.Source.fromResource(...)`, which is using the "ZIO solution", (`getClass.getClassLoader.getResource`, see [here](https://github.com/scala/scala/blob/2.13.x/src/library/scala/io/Source.scala#L180-L181)), I think it'd more globally consistent to use the "ZIO solution" in zio-http too.

My PR is changing the zio-http behaviour to adopt ZIO and `scala.io.Source.fromResource` one.

WDYT?

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -669,18 +670,15 @@ object Http {
   /**
    * Attempts to retrieve files from the classpath.
    */
-  def getResource(path: String): Http[Any, Throwable, Any, net.URL] =
+  def getResource(path: String): Http[Blocking, Throwable, Any, net.URL] =
```
```suggestion
  def getResource(path: String): Http[Any, Throwable, Any, net.URL] =
```
I don't think we need this.

tusharmath:
Can you fix the build issue.

guizmaii:
@tusharmath 

> Can you fix the build issue.

Done :)

guizmaii:
I rebased my PR on `main`

guizmaii:
I don't know why the Benchmarks CI step is always failing on my PR. 
Is it normal? 
Do I have to do something to fix it?

🤔

tusharmath (approved):

PR #1115, opened 2022-03-04
Maintenance: Add internal Logging capabilities using `zhttp.log`

### LINKED ISSUE #1222: Consider removing a wild `println` from `Server.start`
One of the `Server.start` methods have what I guess is some local development logging that reached productive code:

https://github.com/dream11/zio-http/blob/a50397fc11743b32ed95ec350178437da8405ce1/zio-http/src/main/scala/zhttp/service/Server.scala#L218

As one peer suggested, it has the following issues:

- Having that log only for a very specific overloaded method of Server.start
- Using `println`
- Not considering that `println` is an effect, and could eventually throw `=> Task`

Please, consider removing it 🙏🏽

gciuloaica:
fixes #1222 

Added zhttp-logging library that could be used to debug internals of zio-http. 
It is outputting results in console and provides a simple DLS to construct a log line.

API is similar with any other logging library.


zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -133,12 +134,21 @@ sealed trait Server[-R, +E] { self =>
   def withRequestDecompression(enabled: Boolean, strict: Boolean): Server[R, E] =
     Concat(self, RequestDecompression(enabled, strict))
 
+  /**
+   * <<<<<<< HEAD Creates a new server with netty's LoggingHandler to log http
```
sumawa: we can delete HEAD part

docs/website/docs/v1.x/dsl/server.md
```diff
@@ -110,3 +110,5 @@ object HelloWorldAdvanced extends App {
 | `Server.simpleLeakDetection`   | Simplistic leak detection comes with small over head. Refer netty's [ResourceLeakDetector](https://netty.io/4.0/api/io/netty/util/ResourceLeakDetector.Level.html)                               |
 | `Server.paranoidLeakDetection` | Comes with highest possible overhead (for testing purposes only). Refer netty's [ResourceLeakDetector](https://netty.io/4.0/api/io/netty/util/ResourceLeakDetector.Level.html)                              |
 | `Server.consolidateFlush`      | Flushing content is done in batches. Can potentially improve performance.                               |
+| `Server.lowLevelLogging`       | Enable internal netty logger. It is decreasing the performance for which reason it should be used only for troubleshooting. |
```
sumawa: We can say something like:
WARNING: Enabling this will degrade the performance. It should be used only for troubleshooting.

docs/website/docs/v1.x/dsl/server.md
```diff
@@ -110,3 +110,5 @@ object HelloWorldAdvanced extends App {
...
+| `Server.enableLogging(logLevel)` | Enable internal zhttp logger. It is decreasing the performance for which reason it should be used only for troubleshooting. |
```
sumawa: Warning message same as suggested above.

example/src/main/scala/example/HelloWorldAdvanced.scala
```diff
@@ -24,7 +24,8 @@ object HelloWorldAdvanced extends App {
   private val server =
     Server.port(PORT) ++              // Setup port
       Server.paranoidLeakDetection ++ // Paranoid leak detection (affects performance)
-      Server.app(fooBar ++ app)       // Setup the Http app
+      Server.app(fooBar ++ app) ++    // Setup the Http app
+      Server.lowLevelLogging          // Setup logging
```
sumawa: Maybe we can add performance warning comment here as well. Although not 100% sure, but sometimes the new users of a library refer to only examples rather than the documentation. 
```suggestion
      Server.lowLevelLogging          // Setup logging, degrades performance (only for low level troubleshooting)
```

build.sbt
```diff
@@ -118,6 +121,20 @@ lazy val zhttpTest = (project in file("zio-http-test"))
...
 
+lazy val zhttpLogging = (project in file("zio-http-logging"))
+  .settings(stdSettings("zhttp-logging"))
+  .settings(publishSetting(false))
+  .settings(
+    libraryDependencies ++= {
+      if (isScala3(scalaVersion.value)) Seq.empty
+      else Seq(reflect.value % Provided)
+    },
+  )
```
tusharmath, line 134: These settings required?

project/Dependencies.scala
```diff
@@ -19,4 +20,6 @@ object Dependencies {
   val `zio-streams`  = "dev.zio" %% "zio-streams"  % ZioVersion
   val `zio-test`     = "dev.zio" %% "zio-test"     % ZioVersion % "test"
   val `zio-test-sbt` = "dev.zio" %% "zio-test-sbt" % ZioVersion % "test"
+
+  val reflect = Def.map(scalaVersion)("org.scala-lang" % "scala-reflect" % _)
```
tusharmath, line 32: Why do we need this?

zio-http-logging/src/test/scala/zhttp/logging/LoggerSpec.scala
```diff
@@ -0,0 +1,62 @@
+package zhttp.logging
+import zio.test._
+
+import java.nio.file.{Files, Paths}
+
+object LoggerSpec extends DefaultRunnableSpec {
+  private val logFile             = Paths.get("target/file.log")
+  private val consoleLogTransport = LoggerTransport.console
+  private val fileLogTransport    = LoggerTransport.file(logFile)
```
tusharmath: Should use an in-memory logger for unit tests.

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -294,6 +311,10 @@ object Server {
 
   private final case class RequestDecompression(enabled: Boolean, strict: Boolean) extends UServer
 
+  private final case class LowLevelLogging(logLevel: LogLevel) extends UServer
+
+  private final case class CustomLogger(logger: Logger) extends UServer
+
```
tusharmath: Logging at the moment should only be used internally by contributors while they are debugging a problem. 
It's too early to share a public API for users that allows them to configure it. This increases our API's surface areas and will come in the way of producing a stable release.

tusharmath: Thinking more about this it feels like this might be a usecase. For eg: We could implement #1222 by changing the default log level to `info`. Which can be turned completely off in production.

build.sbt
gciuloaica: Yes, it is adding a compile time dependency on `scala-reflect` only for cala 2.x

project/Dependencies.scala
gciuloaica: we need it for scala 2.x macro implementation

zio-http-logging/src/test/scala/zhttp/logging/LoggerSpec.scala
gciuloaica: My goal was to implement integration testing.

zio-http/src/main/scala/zhttp/service/Server.scala
gciuloaica: done.

docs/website/docs/v1.x/dsl/server.md
tusharmath: We have removed this setting now.

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
```diff
@@ -70,6 +75,12 @@ final case class ServerChannelInitializer[R](
     // Flushing content is done in batches. Can potentially improve performance.
     if (cfg.consolidateFlush) pipeline.addLast(HTTP_SERVER_FLUSH_CONSOLIDATION, new FlushConsolidationHandler)
 
+    if (Log.isEnabled) {
```
tusharmath: we should be using `log` not `Log` over here.

zio-http-logging/src/main/scala-2/zhttp/logging/macros/LoggerMacroImpl.scala
```diff
@@ -0,0 +1,59 @@
...
+    val sourceLocation: Tree =
+      if (logLevel == LogLevel.Trace)
+        q"Some(_root_.zhttp.logging.Logger.SourcePos($cname, $lno))"
+      else
+        q"None"
+    val isEnabled: Tree      = q"${c.prefix.tree}.isEnabled"
+    val level: Tree          = q"_root_.zhttp.logging.LogLevel.${TermName(logLevel.name.toLowerCase.capitalize)}"
+
+    q"""
+      if($isEnabled) {
```
tusharmath, line 37: This isn't the most optimized implementation. Consider for eg:
- The level is set to `Info` and aimed at printing the server port when the server has successfully started.
- Now technically the logging is enabled and `isEnabled` will true. 
- This will allow the logger to perform dispatches to all the transports even for `debug` statements.
- We should rather check for `isDebugEnabled` (like we did before)

example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -53,5 +53,4 @@ object Main extends App {
       Server.disableLeakDetection ++
       Server.consolidateFlush ++
       Server.disableFlowControl
-
```
tusharmath: Let's revert changes that aren't part of this PR.

zio-http-logging/src/test/scala/zhttp/logging/LoggerSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+  private def imMemoryTransport(ref: mutable.ListBuffer[CharSequence]): Transport = unsafeSync { line =>
+    ref += line
+  }
+
+  private def inMemoryLogTransport(ref: mutable.ListBuffer[CharSequence]): LoggerTransport =
+    LoggerTransport(
+      format = LogFormat.min,
+      level = LogLevel.Info,
+      transport = imMemoryTransport(ref),
+    )
```
tusharmath: I think we should reify the in memory implementation

```scala
final class MemoryLogger() extends Transport {
  val buffer: ListBuffer[String] = ListBuffer.empty[String]
  override def run(msg: CharSequence): Unit = buffer += msg.toString
  def stdout: String = buffer.mkString("\n")
}
```

zio-http-logging/src/test/scala/zhttp/logging/LoggerSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  private val logFile             = Paths.get("target/file.log")
+  private val consoleLogTransport = LoggerTransport.console
```
tusharmath: We don't need these two transports any more. All the tests should be based on the in memory Logger.

docs/website/docs/v1.x/dsl/server.md
gciuloaica: right, forgot about them.

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
gciuloaica: Done

zio-http-logging/src/main/scala-2/zhttp/logging/macros/LoggerMacroImpl.scala
gciuloaica: I have  updated the code to check if it is enabled for a certain level.

zio-http-logging/src/test/scala/zhttp/logging/LoggerSpec.scala
gciuloaica: done

zio-http-logging/src/main/scala-2/zhttp/logging/macros/LoggerMacroImpl.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    val cname: Tree          = q"${c.internal.enclosingOwner.owner.fullName}"
+    val lno: Tree            = q"${c.enclosingPosition.line}"
+    val sourceLocation: Tree =
+      if (logLevel == LogLevel.Trace)
+        q"Some(_root_.zhttp.logging.Logger.SourcePos($cname, $lno))"
+      else
+        q"None"
+    val level: Tree          = q"_root_.zhttp.logging.LogLevel.${TermName(logLevel.name.toLowerCase.capitalize)}"
+    val isEnabled: Tree      = q"${c.prefix.tree}.isEnabled(${level})"
```
tusharmath:
```suggestion
    val isEnabled: Tree      = q"${c.prefix.tree}.is${level}Enabled"
```

I think this will perform a bit better.

zio-http/src/main/scala/zhttp/service/server/ServerChannelInitializer.scala
```diff
@@ -75,7 +75,7 @@ final case class ServerChannelInitializer[R](
     // Flushing content is done in batches. Can potentially improve performance.
     if (cfg.consolidateFlush) pipeline.addLast(HTTP_SERVER_FLUSH_CONSOLIDATION, new FlushConsolidationHandler)
 
-    if (log.isEnabled()) {
+    if (log.transports.exists(_.level != LogLevel.Disable)) {
```
tusharmath: `isEnabled` should still exist for this usecase.

amitksingh1490:
@gciuloaica We need to disable logging in plainTextBenchMarks

zio-http-logging/src/test/scala/zhttp/logging/LogLevelSpec.scala
```diff
@@ -15,5 +15,10 @@ object LogLevelSpec extends DefaultRunnableSpec {
         assertTrue(LogLevel.fromString(level.toString) == level)
       }
     },
+    testM("any value with the exception of defined values for LogLevel should be set to Error log level.") {
+      checkAll(Gen.fromIterable(List("not defined", "unknown", "disable"))) { level =>
```
tusharmath, line 19:
```scala
Gen.anyString
```

This would be better.

gciuloaica: But it might fail if Gen will produce Info, Error or any defined log levels. Probability is small but still it might fail.

tusharmath (approved):

PR #1121, opened 2022-03-07
Feature: Support Custom statuses code

### LINKED ISSUE #1117: Support custom `Status`es for `HttpError`
I would like to create an `HttpError` with a custom status [451 Unavailable For Legal Reasons](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/451).

A way to create a custom `Status` instance for my `HttpError`.

gciuloaica:
fixes #1117


zio-http/src/main/scala/zhttp/http/HttpError.scala
```diff
@@ -121,4 +122,6 @@ object HttpError {
 
   final case class BadGateway(msg: String = "Bad Gateway") extends HttpError(Status.BAD_GATEWAY, msg)
 
+  final case class CustomResponseStatus(code: Int, reason: String) extends HttpError(CUSTOM_STATUS(code), reason)
```
tusharmath:
```suggestion
  final case class Custom(code: Int, reason: String) extends HttpError(Status.Custom(code), reason)
```

zio-http/src/main/scala/zhttp/http/Status.scala
```diff
@@ -90,62 +91,63 @@ sealed trait Status extends Product with Serializable { self =>
 }
 
 object Status {
```
tusharmath: All Statuses should be in `CamelCase`. May be a separate PR for this?

zio-http/src/main/scala/zhttp/http/HttpError.scala
gciuloaica: done

zio-http/src/main/scala/zhttp/http/HttpError.scala
```diff
@@ -121,4 +121,6 @@ object HttpError {
 
   final case class BadGateway(msg: String = "Bad Gateway") extends HttpError(Status.BAD_GATEWAY, msg)
 
+  final case class CustomResponseStatus(code: Int, reason: String) extends HttpError(Status.Custom(code), reason)
```
tusharmath:
```suggestion
  final case class Custom(code: Int, reason: String) extends HttpError(Status.Custom(code), reason)
```

tusharmath (approved):

PR #1123, opened 2022-03-07
Enhance/transport changes 2 - Remove ServerChannelFactory

sumawa:
## Objective 

- Remove the need to pass EventLoop / Server Channel as dependency to Server
- Instead make it available via a newly introduced "Transport" type
- Remove EventLoopGroup / ServerChannelFactory / ChannelFactory classes 
- Replace them with one TransportFactory defining types (Epoll/Auto/Uring/Nio etc)

### TransportFactory
Transport Factory : Factory for various types of transports like "kQueue", "epoll", etc. used to retrieve ServerChannel, ClientChannel and EventLoopGroup combinations

Advantages:
- No need to dependency inject Channel and EventLoopGroup while creating Server (and in future enhanced client will use this)
- Transport mechanism and thread pool configurable via Server Config (and in future Enhanced client will use ClientSettings to configure transport)

### Server side usage change
Specify transport type (default Auto) and event loop threads like this
```
private val server =
    Server.port(PORT) ++              // Setup port
      Server.paranoidLeakDetection ++ // Paranoid leak detection (affects performance)
      Server.app(fooBar ++ app) ++    // Setup the Http app
      Server.auto                     // Specifying transport type
      Server.threads(4)           // Specifying event loop threads
```
### Client side change
Instead of composing environment layer like (old way)
```
```
Just specify environment type like this (Auto / Epoll / Uring / Nio etc )

``` val env = Auto.live() ```  for transport type "Auto" and default number of threads

or 

``` val env = Nio.live(4) ``` for transport type "Nio" and thread count 4.

### CODE REVIEW

File: example/src/main/scala/example/HelloWorldAdvanced.scala [hunk 1]
```diff
@@ -24,22 +23,24 @@ object HelloWorldAdvanced extends App {
   private val server =
     Server.port(PORT) ++              // Setup port
       Server.paranoidLeakDetection ++ // Paranoid leak detection (affects performance)
-      Server.app(fooBar ++ app)       // Setup the Http app
+      Server.app(fooBar ++ app) ++    // Setup the Http app
+      Server.auto                     // Specifying a Transport type (by default Auto)
```
gciuloaica, line 27: instead of `auto` we might be using `default` as we will be consistent across project

example/src/main/scala/example/HttpsClient.scala
```diff
@@ -2,16 +2,17 @@ package example
...
+import zhttp.service.transport.Transport.Auto
 import zio.{App, ExitCode, URIO, console}
 
 import java.io.InputStream
 import java.security.KeyStore
 import javax.net.ssl.TrustManagerFactory
 
 object HttpsClient extends App {
-  val env     = ChannelFactory.auto ++ EventLoopGroup.auto()
+  val env     = Auto.live()
```
gciuloaica, line 15: it is quite confusing, maybe it should be `Transport.Default`

sumawa:
TransportFactory not needed. To focus on approach enabling sharing of EventLoop b/w client and server.

PR #1128, opened 2022-03-09
Refactor: Http.Status names to Camel Case

gciuloaica:


zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -142,7 +142,7 @@ private[zhttp] final case class Handler[R](
                 }
               case None        =>
                 UIO {
-                  writeResponse(Response.status(Status.NOT_FOUND), jReq)
+                  writeResponse(Response.status(Status.Not_Found), jReq)
```
tusharmath:
```suggestion
                  writeResponse(Response.status(Status.NotFound), jReq)
```

Underscores should be dropped.

gciuloaica: I have removed the underscores from all Status names.

amitksingh1490 (approved):

PR #1140, opened 2022-03-17
fix: Handle content encoding in `hasContentType`

frekw:
I suppose we could be even more correct at the cost of some performance, e.g split at `;` and only check the head, but I don't think it's worth it.


zio-http/src/main/scala/zhttp/http/headers/HeaderChecks.scala
```diff
@@ -11,8 +11,12 @@ import zhttp.http.HeaderValues
...
  */
 trait HeaderChecks[+A] { self: HeaderExtension[A] with A =>
-  final def hasContentType(value: CharSequence): Boolean =
-    contentType.exists(contentEqualsIgnoreCase(value, _))
+  final def hasContentType(value: CharSequence): Boolean = {
+    contentType.exists(h => {
+      val max = Math.min(value.length, h.length)
+      contentEqualsIgnoreCase(h.subSequence(0, max), value)
+    })
+  }
```
tusharmath: We can use `contentStartsWithIgnoreCase` which will be more performant.

tusharmath: Can one not use `mediaType` field on Headers? Something like -

```scala
headers.mediaType match {
  case Some(MediaType.application.json) => true
  case None                             => false
}
```
We can wrap this into a function `hasMediaType` and add it to HeaderChecks also.

frekw: > We can use contentStartsWithIgnoreCase which will be more performant.

I can't seem to find that method, but I can absolutely implement it if we want to!

> Can one not use mediaType field on Headers? Something like -

Not right now, because `mediaType` suffers from the same problem:

```
  final def mediaType: Option[MediaType] =
    contentType.flatMap(ct => MediaType.forContentType(ct.toString))
```

zio-http/src/main/scala/zhttp/http/headers/HeaderChecks.scala
```diff
@@ -11,8 +11,12 @@ import zhttp.http.HeaderValues
  * true or false based on if the conditions are met or not.
  */
 trait HeaderChecks[+A] { self: HeaderExtension[A] with A =>
-  final def hasContentType(value: CharSequence): Boolean =
-    contentType.exists(contentEqualsIgnoreCase(value, _))
+  final def hasContentType(value: CharSequence): Boolean = {
+    contentType.exists(h => {
+      val max = Math.min(value.length, h.length)
+      contentEqualsIgnoreCase(h.subSequence(0, max), value)
```
tusharmath:
```suggestion
      contentEqualsIgnoreCase(h.subSequence(0, max, false), value)
```

zio-http/src/main/scala/zhttp/http/headers/HeaderChecks.scala
tusharmath: There is a `startsWith` in `AsciiString` may be we can use that?

tusharmath:
@frekw are you still on this?

zio-http/src/main/scala/zhttp/http/headers/HeaderChecks.scala
frekw: I don't think a 3-arity version exists: https://docs.oracle.com/javase/8/docs/api/java/lang/CharSequence.html#subSequence-int-int- 🤔

zio-http/src/main/scala/zhttp/http/headers/HeaderChecks.scala
frekw: > There is a startsWith in AsciiString may be we can use that?

It will be used under the hood by `contentEqualsIgnoreCase` is `a` or `b` is an AsciiString 👍

frekw:
@tusharmath yeah, sorry, I've just been strapped on time. I rewrote it using `HttpUtil` which should be optimized enough (and also handle more edge cases from the spec).

zio-http/src/main/scala/zhttp/http/headers/HeaderChecks.scala
```diff
@@ -12,10 +13,11 @@ import zhttp.http.HeaderValues
  */
 trait HeaderChecks[+A] { self: HeaderExtension[A] with A =>
   final def hasContentType(value: CharSequence): Boolean = {
-    contentType.exists(h => {
-      val max = Math.min(value.length, h.length)
-      contentEqualsIgnoreCase(h.subSequence(0, max), value)
-    })
+    contentType
+      .flatMap(ct => Option(HttpUtil.getMimeType(ct)))
```
frekw: This can throw is `ct == Null` – do you think we should handle that @tusharmath ?

vigoo (approved):

PR #1149, opened 2022-03-21
ZIO-2.0.0-RC3

vladimir-lu:
New RC of ZIO is out and unfortunately has some source-breaking changes that mean you can't use it together with libraries that use `ZManaged` from previous versions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -101,6 +101,7 @@ lazy val zhttp = (project in file("zio-http"))
     libraryDependencies ++= Seq(
       netty,
       `zio`,
+      `zio-managed`,
       `zio-streams`,
       `zio-test`,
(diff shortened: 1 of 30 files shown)


ghost:
So I was also thinking about picking this up. However, instead of using zio.managed the API should be migrated to use the new idea of scopes. Just an opinion.

vladimir-lu:
Definitely not against that - I think the `ZManaged` removal is a nice simplification. In this PR I just wanted to get something out the door quickly as it's blocking zio-telemetry to be published for RC3. There's another already approved PR that brings bigger changes too #1136 that might be useful to merge in before changing all the places where `ZManaged` is used.

tusharmath:
This is fixed in #1152 :)

PR #1155, opened 2022-03-23
fix: for websocket onClose

### LINKED ISSUE #1147: WebsocketApp.onClose race condition in effect evaluation and interruption
(using 1.0.0.0-RC25 in combination with ZIO 1.0.13)

Websocket server:

```scala
SocketApp
.onClose(_ => {
    println(s"onClose called");
    ZIO.effect(println(s"onClose called ZIO"));
})....
```

based on the signature:
```scala
def onClose[R1 <: R](close: Connection => ZIO[R1, Nothing, Any]): SocketApp[R1] =
```

`onClose` expects a callback function which returns an effect, which is evaluated when the websocket connection is closed.

**Symptom**: "onClose called" is seen, but "onClose called ZIO" is not.

In `WebSocketAppHandler.scala` we have:
```scala
    app.close match {
        case Some(v) => zExec.unsafeRun(ctx)(v(ctx.channel().remoteAddress()).uninterruptible)
```

which leads to `HttpRuntime.scala':

```scala
 13 /**                                                                                                                        
 14  * Provides basic ZIO based utilities for any ZIO based program to execute in a                                            
 15  * channel's context. It will automatically cancel the execution when the                                                  
 16  * channel closes.                                                                                                       
 17  */
```
which `forks` the provided effect and indeed sets up an interruption mechanism in case the channel is closed:
```scala
 fiber <- program.fork
 close <- ...
```

However, despite setting the `onClose` effect `uninterruptible`, this does prevent (for my app) the effect from being evaluated, as it is interrupted immediately.

If I remove `fiber.interrupt` my effect gets evaluated and everything works as expected.

My first guess would be that the `fiber` is interrupted before ZIO has scheduled it, i.e. before it has even started and therefore still "before" the uninterruptible marker/block.

I don't know if I am the only one running into this (it's a big app with its own execution pools so not sure if there is something else fishy), and I would like to ask for thoughts on the best fix, if this is defined behaviour in ZIO (1.0), etc.

girdharshubham:
TODO: unit tests:
- find a way to provide an instance of ChannelHandlerContext 
- refactor functions to return something else than Unit as in current implementation are not testable.

fixes #1147

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/HttpRuntime.scala
+++ b/zio-http/src/main/scala/zhttp/service/HttpRuntime.scala
@@ -15,8 +15,8 @@ import scala.jdk.CollectionConverters._
  * channel closes.
  */
 final class HttpRuntime[+R](strategy: HttpRuntime.Strategy[R]) {
-  def unsafeRun(ctx: ChannelHandlerContext)(program: ZIO[R, Throwable, Any]): Unit = {
 
+  def unsafeRun(ctx: ChannelHandlerContext)(program: ZIO[R, Throwable, Any]): Unit = {
     val rtm = strategy.runtime(ctx)
 
     // Close the connection if the program fails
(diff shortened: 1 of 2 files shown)


gciuloaica:
It looks ok to me.

tusharmath:
I think we can add a unit test for this.

tusharmath:
Closing in favour of #1156

PR #1166, opened 2022-03-29
add client benchmark ci

sumawa:


sumawa:
Will be replaced with a new draft PR

PR #1169, opened 2022-03-31
Feat/ci client benchmark

sumawa:
Client benchmarking


sumawa:
replaced by another cleaner PR

PR #1189, opened 2022-04-06
Merge main into zio-series/2.x

kitlangton:


kitlangton:
I think that the default behavior of squashing these changes made it very difficult to merge main back into series/2.x. I needed to go through nearly every changed file. Does that sound like a possible culprit? Would it be possible to merge without squashing?

amitksingh1490:
> I think that the default behavior of squashing these changes made it very difficult to merge main back into series/2.x. I needed to go through nearly every changed file. Does that sound like a possible culprit? Would it be possible to merge without squashing?

Do you recommend rebase and merge?

kitlangton:
> Do you recommend rebase and merge?

Good question! I'm not sure though. Maybe main was just not merged for a while and a squash would be okay.

amitksingh1490 (approved):

PR #1193, opened 2022-04-08
API DSL

kitlangton:
Closing #1190 in favor of this (against ZIO 1.0).
Here's a variant against the current main branch @tusharmath :)

## Todos
- [ ] Customize Response headers, status code, etc. in handler
- [ ] Refine api
- [ ] Flesh out Doc type
- [ ] Add Error type
- [ ] Unify Input & Params into single type
- [ ] Improve Client
- [ ] Extract model into shared project so we can have a Scala.js client
- [ ] Improve OpenAPI support
- [ ] Expose OpenAPI from server

## Performance Explorations
- [ ] Tuple Allocation Optimizations (during Zips)
- [ ] OrElse Optimizations (create Trie-like request parsing structure)


kitlangton:
`zio-schema` needs to be updated for Scala 3. But I believe that's in the works, at least according to discord :)

zio-http/src/main/scala/zhttp/api/API.scala
```diff
@@ -0,0 +1,87 @@
...
+import zhttp.http.{Headers => _, Path => _}
+import zio.schema.Schema
+
+/**
+ *   - Input and Output as Schemas.
+ *   - Support wide range of Codecs including Json
+ *     - Dynamically decide response format based upon Request Header
+ */
+final case class API[Params, Input, Output](
+  method: HttpMethod,
```
tusharmath:
```suggestion
  method: Method,
```

zio-http/src/main/scala/zhttp/api/API.scala
```diff
@@ -0,0 +1,87 @@
...
+  requestParser: RequestParser[Params], // Path / QueryParams / Headers
+  doc: Doc,
+  inputSchema: Schema[Input],           // Generate any sort of codec, generate OpenAPI docs.
+  outputSchema: Schema[Output],
```
tusharmath, line 16:
```suggestion
  decoder: BodyDecoder[Input],           // Generate any sort of codec, generate OpenAPI docs.
  encoder: BodyEncoder[Output],
```

Should we avoid `Schema` here? To support Scala 3 and reduce onboarding cost for new devs?

zio-http/src/main/scala/zhttp/api/API.scala
kitlangton: One potential reason not to do that is so we can break the model into a _shared_ project so that the API DSL can be interpreted into Scala.js fetch requests. Though perhaps the `Method` sealed trait could be moved there as well, so long as it isn't part of `netty`.

I was going to give this a shot today, as well as cleaning some other stuff up.

zio-http/src/main/scala/zhttp/api/RequestParser.scala
```diff
@@ -0,0 +1,333 @@
...
+}
+
+/**
+ * A DSL for describe Paths
+ *   - ex: /users
+ *   - ex: /users/:id/friends
+ *   - ex: /users/:id/friends/:friendId
+ *   - ex: /posts/:id/comments/:commentId
+ */
+sealed trait Path[A] extends RequestParser[A] { self =>
```
tusharmath:
```suggestion
sealed trait Route[A] extends RequestParser[A] { self =>
```

`Path` already exists in ZIO Http. Consider an alternative name — `Route` ?

zio-http/src/main/scala/zhttp/api/Handler.scala
```diff
@@ -0,0 +1,27 @@
+package zhttp.api
+
+import zhttp.http.HttpApp
+import zio.ZIO
+
+final case class Handler[-R, +E, Params, Input, Output](
+  api: API[Params, Input, Output],
+  handle: ((Params, Input)) => ZIO[R, E, Output],
```
tusharmath, line 8:
```suggestion
  toHttp: Http[R, E, (Params, Input), Output],
```

Basically we are saying here that — A handler is a description of the Http endpoint AND a way to execute it.

tusharmath, line 8:
```suggestion
  handle: Http[R, E, (Params, Input), Output],
```

`Http` will allow us to gain some performance, control over dynamic routing and also middlewares.

zio-http/src/main/scala/zhttp/api/Handler.scala
```diff
@@ -0,0 +1,27 @@
...
+) { self =>
+
+  def ++[R1 <: R, E1 >: E](that: Handler[R1, E1, _, _, _]): Handlers[R1, E1] =
+    Handlers(this) ++ that
+
+  def toHttp: HttpApp[R, E] =
+    ServerInterpreter.handlerToHttpApp(self)
```
tusharmath, line 15:
```suggestion
```

This isn't required anymore.

zio-http/src/main/scala/zhttp/api/Handler.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import zhttp.http.HttpApp
+import zio.ZIO
+
+final case class Handler[-R, +E, Params, Input, Output](
+  api: API[Params, Input, Output],
+  handle: ((Params, Input)) => ZIO[R, E, Output],
+) { self =>
+
+  def ++[R1 <: R, E1 >: E](that: Handler[R1, E1, _, _, _]): Handlers[R1, E1] =
```
tusharmath, line 11:
```suggestion
  def ++[R1 <: R, E1 >: E](that: Handler[R1, E1, _, _, _]): Handler[R1, E1] =
```
I think combining two handler(s) should produce a handler. We can drop `Handlers` type completely this way.

tusharmath, line 11:
```suggestion
  def ++[R1 <: R, E1 >: E](that: Handler[R1, E1, _, _, _]): Handler[R1, E1] =
```
I think combining two handler(s) should produce a handler. We can drop `Handlers` type completely this way.

zio-http/src/main/scala/zhttp/api/Handler.scala
tusharmath, line 15:
```suggestion
  def toHttp: HttpApp[R, E] = self match {
    case Concat(left, right) => left.handle ++ right.handle
    case Single(_, handle)   => http
  }
```
- This way we won't need a `ServerInterpreter`
- We can also mount the `Handler` on any route as an HttpApp

tusharmath, line 15:
```suggestion
  def toHttp: HttpApp[R, E] = self match {
    case Concat(left, right) => left.handle ++ right.handle
    case Single(_, handle)   => http
  
  }
```
This way we won't need a ServerInterpreter

zio-http/src/main/scala/zhttp/api/Handler.scala
```diff
@@ -0,0 +1,27 @@
...
+    Handlers(this) ++ that
+
+  def toHttp: HttpApp[R, E] =
+    ServerInterpreter.handlerToHttpApp(self)
+
```
tusharmath, line 16:
```suggestion
  def openAPI: UIO[String] = ???
```

May be it's worth adding a special method for `openAPI`

zio-http/src/main/scala/zhttp/api/Handler.scala
```diff
@@ -0,0 +1,27 @@
+package zhttp.api
+
+import zhttp.http.HttpApp
+import zio.ZIO
+
+final case class Handler[-R, +E, Params, Input, Output](
```
tusharmath, line 6: This should be a sum type like `Handlers` and then we won't need `Handlers`.

May be we can rename `Handler` because it's not just the handle any more. It contains info about the endpoint also.

- `Handler` to `API`
- `API` to `Endpoint`

zio-http/src/main/scala/zhttp/api/Handlers.scala
```diff
@@ -0,0 +1,28 @@
+package zhttp.api
+
+import scala.language.implicitConversions
+
+sealed trait Handlers[-R, +E] { self =>
```
tusharmath: We can drop this 😇

tusharmath: We can drop this 😇

zio-http/src/main/scala/zhttp/api/HttpMethod.scala
```diff
@@ -0,0 +1,20 @@
...
+    }
+}
+
+object HttpMethod {
+  case object GET    extends HttpMethod
+  case object POST   extends HttpMethod
+  case object PATCH  extends HttpMethod
+  case object PUT    extends HttpMethod
+  case object DELETE extends HttpMethod
+}
```
tusharmath: `API` being a native to ZIO Http, having `Method` and `HttpMethod` can be a bit confusing.

tusharmath: `API` being a native to ZIO Http, having `Method` and `HttpMethod` can be a bit confusing.

zio-http/src/main/scala/zhttp/api/RequestParser.scala
```diff
@@ -0,0 +1,333 @@
...
+        Some(route)
+      case _: Query[_]                         =>
+        None
+      case _: Path[_]                          =>
+        None
+    }
+
+  private[api] def parseRequest(request: Request): Option[A] = Option(parseRequestImpl(request))
+
+  private[api] def parseRequestImpl(request: Request): A
```
tusharmath, line 77:
```suggestion
  private[api] def parseRequestImpl(request: HttpRequest): A
```

We can improve performance by using the netty request. That will further improve performance.

zio-http/src/main/scala/zhttp/api/Server.scala
```diff
@@ -0,0 +1,33 @@
+package zhttp.api
+
+import zio.{Has, ZIO}
+
+final case class Server[R <: Has[_], E <: Throwable](apis: APIs[_], handlers: Handlers[R, E]) {
```
tusharmath: Can drop this file.

zio-http/src/main/scala/zhttp/api/ServerInterpreter.scala
```diff
@@ -0,0 +1,47 @@
+package zhttp.api
+
+import zhttp.http.{Http, HttpApp, Request, Response}
+import zio.schema.Schema
+import zio.schema.codec.JsonCodec
+import zio.{Chunk, UIO, ZIO}
+
+private[api] object ServerInterpreter {
```
tusharmath, line 8: And drop this file also.

zio-http/src/main/scala/zhttp/api/API.scala
```diff
@@ -0,0 +1,87 @@
...
+    copy(requestParser = requestParser ++ headers)
+
+  def input[Input2](implicit schema: Schema[Input2]): API[Params, Input2, Output] =
+    copy(inputSchema = schema)
+
+  def output[Output2](implicit schema: Schema[Output2]): API[Params, Input, Output2] =
+    copy(outputSchema = schema)
+
+  def ++(that: API[_, _, _]): APIs[Id with that.Id] =
+    APIs(self) ++ that
```
tusharmath: The type param to API will have an intersection of multiple tuples when we use `++`. Not something that's very readable or usable IMHO. Using the `++` in `Handler` makes a lot more sense.

tusharmath (changes requested):

zio-http/src/main/scala/zhttp/api/API.scala
kitlangton: We need something Schema in order to support OpenAPI generation, as well as the ability to encode/decode to different formats based on headers. But perhaps for a first release, until Schema supports Scala 3.

zio-http/src/main/scala/zhttp/api/RequestParser.scala
kitlangton: 💯

zio-http/src/main/scala/zhttp/api/Handler.scala
kitlangton: How would we turn the API description into an request parser?

zio-http/src/main/scala/zhttp/api/Handler.scala
kitlangton: 👍 I think that will work. Yeah, I want to get rid of the Handlers/APIs types if possible.

zio-http/src/main/scala/zhttp/api/Handler.scala
kitlangton: We need to, somewhere, interpret the API DSL into something that can parse requests. But I may not understand fully the intended change.

zio-http/src/main/scala/zhttp/api/Handler.scala
kitlangton: 👍

zio-http/src/main/scala/zhttp/api/RequestParser.scala
kitlangton: Not sure how to get the netty HttpRequest to where I need it. Might we need to add a variant to Http that allows direct access to the netty Request?

zio-http/src/main/scala/zhttp/api/API.scala
kitlangton: Yeah. It was for a fancy feature where you could make sure the types of the handlers align with the API. Not worth the complexity.

kitlangton:
@tusharmath Made a couple of the changes. Maybe we could link up again on Zoom and discuss some of the bigger changes. Might help to sync up :)

PR #1204, opened 2022-04-13
Feature: Add implicit conversion from `Unit` to `Html`

nartamonov:
The conversion is useful in cases when we don't want to construct any `Html` but want to leave it empty. It is is expected by users of `scalatags` or similar libraries.


tusharmath (approved):

tusharmath:
thanks @nartamonov !

PR #1209, opened 2022-04-16
API DSL

tusharmath:
## Todos
- [x] Customize Response headers, status code, etc. in handler
- [x] Add Error type
- [x] Refine api
- [x] Flesh out Doc type
- [x] Unify Input & Params into single type
- [x] Improve Client
- [ ] Extract model into shared project so we can have a Scala.js client
- [ ] Improve OpenAPI support
- [ ] Expose OpenAPI from server

## Performance Explorations
- [ ] Tuple Allocation Optimizations (during Zips)
- [ ] OrElse Optimizations (create Trie-like request parsing structure)


kitlangton:
@tusharmath I added some (very basic) support for customizing headers/status codes when handling responses. There are a few ways of doing this, but I think an overload where the user simply specifies what to return in the handler is the simplest.

Next, I'm going to get zio-schema published for Scala 3 and fix the 2.12 compilation errors.

jdegoes (approved):

jdegoes (approved):

PR #1214, opened 2022-04-18
Refactor: Catch SSLHandshakeException

dinojohn:
Currently the server throws an exception if the client ssl certificates are invalid. This PR will ignore SSLHandshakeException.


tusharmath:
I don't think we should ignore the handshake exception. Ideally the user should have control over this. The should be able to act on the exception or ignore it completely.

tusharmath:
I don't think we should just ignore the exception. Closing it for now.

Issue #101, opened 2021-03-31
Add to Client API proxy server settings


declspec-cdecl:
It's good idea to add possibility to specify proxy server settings for Client

heksesang:
Any progress on this?

ShrutiVerma97:
> Any progress on this?

We have not picked it yet, do you want to give it a try?

tusharmath:
This is fixed in #1226

PR #1226, opened 2022-04-24
Feature: Add proxy handler for client API

pizzaeueu:
Hi there!
My attempt to provide a solution for #101 

## Why PR to zio-http 2.x ?
Actually, I don't know the right flow. Please correct me if I need to do it in a different way. I personally need it in 2.x, so that was a motivation 😌 

## How did you test it?
1) Run locally any free forward-proxy, e.g. https://github.com/clue/docker-polipo
2) Run the following code with updated zio-http version

```
import zhttp.http.{Proxy, Request, URL}
import zhttp.service.Client.Config
import zhttp.service.{ChannelFactory, Client, EventLoopGroup}
import zio.ZIO

object Main extends zio.ZIOAppDefault {
  override def run = {
    val req = Request(
      url = URL
        .fromString("https://gorest.co.in/public/v2/posts")
        .toOption
        .getOrElse(URL.empty)
    )
    val proxy = Proxy(
      URL.fromString("http://127.0.0.1:8123").toOption.get
    )
    Client
      .request(
        req,
        Config().withProxy(proxy),
      )
      .provide(
        ChannelFactory.auto,
        EventLoopGroup.auto(),
      )
      .flatMap(response => ZIO.log(response.status.toString))
  }
}
```

**Result with Proxy**
```
timestamp=2022-04-24T07:28:44.302564Z level=INFO thread=#zio-fiber-2 message="Ok" location=test.Main.run file=Main.scala line=26
```

**Result without Proxy**
```
timestamp=2022-04-24T07:30:29.919041Z level=INFO thread=#zio-fiber-2 message="Ok" location=test.Main.run file=Main.scala line=26
```

**Results with wrong proxy address**
```
... java.nio.channels.UnresolvedAddressException: java.nio.channels.UnresolvedAddressException ...
```
```
... Connection refused: /127.0.0.1:1111
```

etc


zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -167,9 +180,15 @@ object Client {
     } yield res
   }
 
-  case class Config(socketApp: Option[SocketApp[Any]] = None, ssl: Option[ClientSSLOptions] = None) { self =>
+  case class Config(
+    socketApp: Option[SocketApp[Any]] = None,
+    ssl: Option[ClientSSLOptions] = None,
+    proxyAddress: Option[InetSocketAddress] = None,
```
tusharmath: May be we want to add a new type `Proxy`. Where we can pass, `username` `password` and `headers`. They could be optional. Then we can add build a new `HttpProxyHandler` using `Proxy` class.

zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -71,13 +72,25 @@ final case class Client[R](rtm: HttpRuntime[R], cf: JChannelFactory[Channel], el
...
+          // Adding proxy handler
+          if (isProxy)
+            pipeline.addLast(
+              PROXY_HANDLER,
+              new HttpProxyHandler(
+                clientConfig.proxyAddress.getOrElse(
+                  new InetSocketAddress(host, port)
+                )
+              ),
+            )
```
tusharmath, line 93: Can we also add some tests? You can refer to `ClientSpec.scala`.

pizzaeueu: Hi @tusharmath 

For some reason, `ClientSpec.scala` is just stucking when I'm trying to run it locally. As well as other tests that use `provideLayerShared` method.

I experimented a bit and found that versions before [Updating to ZIO 2.0.0-RC4](https://github.com/dream11/zio-http/pull/1178) works fine

Sorry for disturbing you with that, but maybe you have any CI tool to check if the issue is reproduced only on my local machine? I didn't find it on my own unfortunatelly

pizzaeueu: Hi @tusharmath 

For some reason, `ClientSpec.scala` is just stucking when I'm trying to run it locally. As well as other tests that use `provideLayerShared` method.

I experimented a bit and found that versions before [Updating to ZIO 2.0.0-RC4](https://github.com/dream11/zio-http/pull/1178) works fine

Sorry for disturbing you with that, but maybe you have any CI tool to check if the issue is reproduced only on my local machine? I didn't find it on my own unfortunatelly

zio-http/src/main/scala/zhttp/service/Client.scala
pizzaeueu: new `Proxy` class was introduced

zio-http/src/main/scala/zhttp/service/Client.scala
pizzaeueu: Added test to ClientSpec + covered new Proxy class

zio-http/src/main/scala/zhttp/http/Proxy.scala
```diff
@@ -0,0 +1,37 @@
...
+import io.netty.handler.codec.http.EmptyHttpHeaders
+import io.netty.handler.proxy.HttpProxyHandler
+import zhttp.http.middleware.Auth.Credentials
+
+import java.net.InetSocketAddress
+
+final case class Proxy(
+  url: URL,
+  credentials: Option[Credentials] = None,
+  headers: Option[Headers] = None,
```
tusharmath:
```suggestion
  headers: Headers = Headers.empty,
```

zio-http/src/main/scala/zhttp/http/Proxy.scala
```diff
@@ -0,0 +1,37 @@
+package zhttp.http
+import io.netty.handler.codec.http.EmptyHttpHeaders
+import io.netty.handler.proxy.HttpProxyHandler
+import zhttp.http.middleware.Auth.Credentials
+
+import java.net.InetSocketAddress
+
+final case class Proxy(
```
tusharmath, line 18: May be we can add a bit of scala doc for the users to understand the usage of `Proxy` a little better.

zio-http/src/main/scala/zhttp/http/Proxy.scala
```diff
@@ -0,0 +1,37 @@
...
+    encodedHeaders = headers.map(_.encode).getOrElse(EmptyHttpHeaders.INSTANCE)
+  } yield {
+    new HttpProxyHandler(proxyAddress, encodedHeaders)
+  }
+
+  private def buildProxyAddress: Option[InetSocketAddress] = for {
+    proxyHost <- url.host
+    proxyPort <- url.port
+  } yield new InetSocketAddress(proxyHost, proxyPort)
+}
```
tusharmath, line 51: We can also add some helper methods and use builder pattern to build proxy in parts. eg :

```scala
def withCredentials(cred: Credential): Proxy = ???
def withHeaders(headers: Headers): Proxy = ???
```

zio-http/src/test/scala/zhttp/service/ClientSpec.scala
```diff
@@ -43,6 +43,19 @@ object ClientSpec extends HttpRunnableSpec {
...
+          for {
+            validServerPort <- ZIO.service[DynamicServer].flatMap(_.port)
+            serverUrl       <- ZIO.fromEither(URL.fromString(s"http://localhost:$validServerPort"))
+            proxyUrl        <- ZIO.fromEither(URL.fromString("http://localhost:0001"))
+            out             <- Client.request(
+              Request(url = serverUrl),
+              Config().withProxy(Proxy(proxyUrl)),
+            )
+          } yield out
+        assertM(res.either)(isLeft(isSubtype[ConnectException](anything)))
```
tusharmath, line 60: Can we add a test for success also?

pizzaeueu: Hi!
Thanks for the feedback
I've fixed all comments above 😌

zio-http/src/test/scala/zhttp/http/ProxySpec.scala
```diff
@@ -0,0 +1,40 @@
...
+    } + suite("Unauthenticated proxy") {
+      test("successfully encode valid proxy") {
+        val proxy   = Proxy(validUrl)
+        val encoded = proxy.encode
+
+        assert(encoded)(isSome)
+        assert(encoded.map(_.username()))(isSome(isNull))
+        assert(encoded.map(_.password()))(isSome(isNull))
+        assert(encoded.map(_.authScheme()))(isSome(equalTo("none")))
+      }
```
tusharmath, line 37: I don't think we can write `assert` expressions like this. You will have to compose them using `&&`. Right now only the last `assert` is going to be executed.

pizzaeueu: Yeah, that's a nice catch
Thank you

Fixed

tusharmath:
@pizzaeueu Can you make the PR wrt `main`. We make a ZIO 2.0 branch only at the time of release.

pizzaeueu:
Hi @tusharmath 
Thanks for the feedback

Target branch was changed to main
All conflicts were fixed

pizzaeueu:
Hi @tusharmath 
Sorry for disturbing you, but are there any chances to merge this MR?
Maybe I should apply any additional fixes or smth?

I really need such feature 😌

tusharmath (approved):

PR #1236, opened 2022-05-02
fix HttpData unsafe cast

senia-psm:

PR #1244, opened 2022-05-07
Fix: Echo `HttpData` directly from `Request`

### LINKED ISSUE #1229: Exception on unsafe cast in writeResponse
A simple test that starts echo `Server` and performs a single `Client` call runs forever.

This looks like regression: runs forever on `2.0.0-RC7`, works as expected on `2.0.0-RC4` (with `ZManaged` instead of `Scope`).

**To Reproduce**
See this project with a single test: https://github.com/senia-psm/zhttp-bug

This test runs forever:
```scala
test("Server should be able to start on an unoccupied port") {

  val dummyRoute = Http.collect[Request] { case req @ Method.POST -> !! / "echo" =>
    Response(status = Status.Ok, data = req.data)
  }

  val layer: ZLayer[ServerChannelFactory & EventLoopGroup, Nothing, Server.Start] = ZLayer.scoped {
    Server(dummyRoute).withBinding("localhost", 0).make.orDie
  }
  val env                                                                         = ZLayer
    .make[Server.Start & EventLoopGroup & ChannelFactory](
      layer,
      EventLoopGroup.auto(0),
      ServerChannelFactory.auto,
      ChannelFactory.auto,
    )

  val action = for {
    port           <- ZIO.serviceWith[Server.Start](_.port)
    resp           <- Client.request(
                        s"http://localhost:$port/echo",
                        method = Method.POST,
                        content = HttpData.fromString("echo string"),
                      )
    responseString <- resp.bodyAsString
  } yield assertTrue(port > 0, responseString == "echo string")

  action.provideLayer(env)
},
```

**Expected behaviour**
Test above should complete in finite amount of time (< 1 second).

**Desktop:**
 - OS: Linux
 - Version: 5.4.179-gentoo

**Additional context**
Might be related to https://github.com/zio/interop-cats/issues/533

tusharmath:
fixes #1229

related to #1236

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/HttpData.scala
+++ b/zio-http/src/main/scala/zhttp/http/HttpData.scala
@@ -126,18 +126,10 @@ object HttpData {
     }
   }
 
-  private[zhttp] final class UnsafeContent(private val httpContent: HttpContent) extends AnyVal {
-    def content: ByteBuf = httpContent.content()
-
-    def isLast: Boolean = httpContent.isInstanceOf[LastHttpContent]
-  }
-
-  private[zhttp] final class UnsafeChannel(private val ctx: ChannelHandlerContext) extends AnyVal {
-    def read(): Unit = ctx.read(): Unit
-  }
-
-  private[zhttp] final case class UnsafeAsync(unsafeRun: (UnsafeChannel => UnsafeContent => Unit) => Unit)
+  private[zhttp] final case class UnsafeAsync(unsafeRun: (ChannelHandlerContext => HttpContent => Unit) => Unit)
       extends HttpData {
+
+    private def isLast(msg: HttpContent): Boolean = msg.isInstanceOf[LastHttpContent]
 
     /**
      * Encodes the HttpData into a ByteBuf.
@@ -148,7 +140,7 @@ object HttpData {
           val buffer = Unpooled.compositeBuffer()
           msg => {
             buffer.addComponent(true, msg.content)
-            if (msg.isLast) cb(UIO(buffer)) else ch.read()
+            if (isLast(msg)) cb(UIO(buffer)) else ch.read(): Unit
           }
         }),
       )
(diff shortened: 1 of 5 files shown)


zio-http/src/main/scala/zhttp/http/HttpData.scala
```diff
@@ -126,19 +126,11 @@ object HttpData {
...
-    def content: ByteBuf = httpContent.content()
-
-    def isLast: Boolean = httpContent.isInstanceOf[LastHttpContent]
-  }
-
-  private[zhttp] final class UnsafeChannel(private val ctx: ChannelHandlerContext) extends AnyVal {
-    def read(): Unit = ctx.read(): Unit
-  }
-
-  private[zhttp] final case class UnsafeAsync(unsafeRun: (UnsafeChannel => UnsafeContent => Unit) => Unit)
```
tusharmath, line 139: Removing unnecessary wrapping from HttpContent and ChannelHandlerContext. Since we are being explicit about it being an "unsafe" API, the wrapping seems a bit unnecessary.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -63,7 +64,11 @@ private[zhttp] final case class Handler[R](
                 HttpData.UnsafeAsync(callback =>
                   ctx
                     .pipeline()
-                    .addAfter(HTTP_REQUEST_HANDLER, HTTP_CONTENT_HANDLER, new RequestBodyHandler(callback)): Unit,
+                    .addAfter(
+                      HTTP_REQUEST_HANDLER,
+                      HTTP_CONTENT_HANDLER,
+                      new RequestBodyHandler(callback(ctx)),
+                    ): Unit,
```
tusharmath, line 71: Initializing the callback here to avoid race conditions inside of RequestBodyHandler

zio-http/src/main/scala/zhttp/service/ServerResponseWriter.scala
```diff
@@ -103,6 +103,12 @@ private[zhttp] final class ServerResponseWriter[R](
       case HttpData.JavaFile(unsafeGet) =>
         unsafeWriteFileContent(unsafeGet())
         releaseAndRead(jReq)
+
+      case HttpData.UnsafeAsync(unsafeRun) =>
+        unsafeRun { _ => msg =>
+          ctx.writeAndFlush(msg)
+          if (!msg.isInstanceOf[LastHttpContent]) ctx.read(): Unit
+        }
```
tusharmath, line 111: This will allow responses to use the HttpData received in the request directly. This is also going significantly more performant than echo using streams.

PR #1251, opened 2022-05-11
Feature: Add `empty` and `allow` constructors to `Middleware`

ex0ns:
As discussed previously, I think that having an `empty` middleware could be useful.
It was also noted that an `allow` one could be as useful, it could be used as a fail-fast middleware to allow/disallow a request to continue based on a condition.

I'm not sure about the location of those constructors and where I should place them instead, maybe you can point me to the correct location if you want me to change them.

Note: I've renamed and added a test for the `identity` middleware (one for Http.empty and one for a Http.succeed)


tusharmath (approved):

tusharmath:
Perfect!

PR #1261, opened 2022-05-18
feature: Provide access to `ChannelHandlerContext` using `Http`.

tusharmath:
Summary of changes produced in this PR

1. Adding support to access `ChannelHandlerContext` via `Http.context`

  ```scala
  val context: Http[Any, Nothing, Request, ChannelHandlerContext] = Http.context 
  ```

2. Adding to wrap an `HttpApp` using the `Http.usingContext` api

  ```scala
  val app = Http.usingContext { 
    ctx => Http.collect[Request] {
      case GET -> !! / "address" => Response.text(s"You connected via ${ctx.channel.remoteAddress}")
    }
  }
  ```
3. Deprecation of `request.remoteAddress`. This was an awkward API and made sense only in server context. Hence the public API is now moved into `Http.context`.

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -1,7 +1,7 @@
 package zhttp.http
 
 import io.netty.buffer.{ByteBuf, ByteBufUtil}
-import io.netty.channel.ChannelHandler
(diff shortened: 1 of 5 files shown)


tusharmath:
closing in favour of #1263

PR #1267, opened 2022-05-23
Fix `Http.fromResource` implementation

guizmaii:
See https://howtodoinjava.com/java/io/read-file-from-resources-folder/

The issue, reported by [strokyl](https://github.com/strokyl), a developer of my team, is that when the lib is packaged in a jar then `getResource` returns something such as `jar-filename.jar/!filename.txt` which makes the `new File(url.getPath)` not working.

@strokyl, am I explaining the issue correctly?

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Http.scala
+++ b/zio-http/src/main/scala/zhttp/http/Http.scala
@@ -891,8 +891,36 @@ object Http {
   /**
    * Creates an Http app from a resource path
    */
-  def fromResource(path: String): HttpApp[Blocking, Throwable] =
-    Http.getResource(path).flatMap(url => Http.fromFile(new File(url.getPath)))
+  def fromResource(path: String): HttpApp[Any, Throwable] = {
+    @inline def header[A](v: A)(f: A => Headers): Headers = if (v == null) Headers.empty else f(v)
+
+    Http
+      .getResource(path)
+      .flatMap { url =>
+        val response =
+          effectBlocking {
+            val connection = url.openConnection()
+            val mediaType  =
+              path.lastIndexOf(".") match {
+                case -1 => None
+                case i  => MediaType.forFileExtension(path.substring(i + 1)).map(_.fullType)
+              }
+
+            val contentLengthHeader   = header(connection.getContentLengthLong)(Headers.contentLength)
+            val contentTypeHeader     = header(mediaType.orNull)(Headers.contentType)
+            val contentEncodingHeader = header(connection.getContentEncoding)(Headers.contentEncoding)
+            val data                  = ZStream.fromInputStream(connection.getInputStream).provideLayer(Blocking.live)
+
+            Response(
+              headers = contentLengthHeader ++ contentTypeHeader ++ contentEncodingHeader,
+              data = HttpData.fromStream(data),
+            )
+          }
+
+        Http.fromZIO(response)
+      }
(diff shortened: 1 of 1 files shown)


strokyl:
> See https://howtodoinjava.com/java/io/read-file-from-resources-folder/
> 
> The issue, reported by [strokyl](https://github.com/strokyl), a developer of my team, is that when the lib is packaged in a jar then `getResource` returns something such as `jar-filename.jar/!filename.txt` which makes the `new File(url.getPath)` not working.
> 
> @strokyl, am I explaining the issue correctly?

Yes exactly, I did open a more complete PR with some explanation about limitations with this new solution here: https://github.com/dream11/zio-http/pull/1268

strokyl:
Ok nice catch using `.getConnection` to get size and content length information of the resource!
We still have to figure out the problem with directories, I explain it here: https://github.com/dream11/zio-http/pull/1268#issue-1245280356. (unit test here: https://github.com/dream11/zio-http/pull/1268/files#diff-f107f2cb5f3bf791a7b494cf7c94236b94aeb8e377ca72562686a91ae9b3278dR27-R50)
@guizmaii we should propose a single PR instead of two going to close mine.

strokyl:
For the directory issue, the only solution I have in mind is too;

- open a stream on it
- get the first line
- if the resource is a directory first line is a file inside the resource directory, we can check it by opening a resource
- concat resource path and this first line content to get potential subresource path
- try opening a resource on it...

Note: if a resource is an empty directory opening a resource on it returns null. 
I don't know what happens for a directory containing a single empty directory, and in this case, I am afraid my solution does not work

tusharmath:
Thanks @strokyl @guizmaii!
This is indeed an  interesting find. Few thoughts though —

1. Can we add a failing test for this case? May be we need a separate module with a little bit of mocking.
2. Using `Blocking` Stream is going to have a heavy penalty on performance. I think if we can convert it to `fromFile` API, then the file can be served in a non-blocking way.

guizmaii:
Hi @tusharmath,

> 1. Can we add a failing test for this case? May be we need a separate module with a little bit of mocking.

Sure, we will

> 2. Using Blocking Stream is going to have a heavy penalty on performance.

I'm sorry, I'm not sure to understand your comment. 
The initial `Http.getResource`, which was already used in the previous version of the code, is already a `Blocking` operation.

🤔

>  I think if we can convert it to fromFile API, then the file can be served in a non-blocking way.

I didn't find how to transform an `InputStream` into a `File` without consuming it.
What I understood, and I can be wrong, is that if we transform it into a `File` we'll have to load the entire `InputStream` content in memory and the code will "consume" this data twice: once to create the `File` and once to serve it to the HTTP client. 
Am I wrong? 🤔

tusharmath:
> I'm sorry, I'm not sure to understand your comment.
The initial Http.getResource, which was already used in the previous version of the code, is already a Blocking operation.

You right, but there is a difference. The previous version uses the blocking thread pool only to get the resource path, rest of the file is served asynchronously. 
In this PR, we are reading the file in a blocking thread pool, that's going to dramatically affect performance.

guizmaii:
@strokyl @tusharmath I'm sorry I have absolutely no time to finish this PR. If one of you wants to take this subject and finish it that'd be very nice of you. I can help if needed

tusharmath:
Sure. I'll take this up.

PR #1275, opened 2022-05-29
Fix: Drop dependency on `scala-collection-compat` when using Scala >= 2.13

jpablo:
This is only an issue for Scala 3 projects that depend on Scala 2 projects using `scala-collection-compat_2.13` (such as `scalameta`).

In such cases there's an sbt conflict:

```
[error] Modules were resolved with conflicting cross-version suffixes ...
[error]    org.scala-lang.modules:scala-collection-compat _3, _2.13
```

As `scalameta` is not available yet for Scala 3, there's no easy solution for now.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -105,8 +105,13 @@ lazy val zhttp = (project in file("zio-http"))
       `zio-test`,
       `zio-test-sbt`,
       `netty-incubator`,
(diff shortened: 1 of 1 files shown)


tusharmath (approved):

tusharmath:
Can u create a PR wrt main also?

PR #1279, opened 2022-05-30
Backport #1259 into zio2

gurghet:
Backport #1259 like I said in https://github.com/dream11/zio-http/issues/1250


gurghet:
Currently does not compile in scala 3.1.2 but don't understand why

zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -76,7 +76,7 @@ abstract class HttpRunnableSpec extends ZIOSpecDefault { self =>
         id       <- Http.fromZIO(DynamicServer.deploy(app))
         url      <- Http.fromZIO(DynamicServer.wsURL)
         response <- Http.fromFunctionZIO[SocketApp[HttpEnv]] { app =>
-          ZIO.scoped {
+          ZIO.scoped[HttpEnv] {
```
gurghet, line 79: Scala 3 makes a mess here without this.
I haven't investigated why. If someone has time please find the root cause of this as there might be something wrong with the implementation of `scoped`.

amitksingh1490:
covered in https://github.com/dream11/zio-http/pull/1280

PR #1288, opened 2022-06-02
Fix: Cookie expires format

### LINKED ISSUE #1093: setCookiesDecode uses wrong date format
Ran into this when trying to use `Response.setCookiesDecode`.  It always returned `None`.

Turns out currently it expects a date on this format for the `expires` value :
`2022-02-24T17:10:10.682151Z`
The Set-Cookie I was trying decode had an expires value on this format:
`Fri, 25 Feb 2022 04:49:46 GMT`
This also seems to be the format that the [HTTP specification](https://httpwg.org/specs/rfc7231.html#header.date) calls for. I belive zhttp is not following the specification here

gciuloaica:
Fix #1093

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Cookie.scala
+++ b/zio-http/src/main/scala/zhttp/http/Cookie.scala
@@ -9,7 +9,7 @@ import java.util.Base64.getEncoder
 import java.util.Locale
 import javax.crypto.Mac
 import javax.crypto.spec.SecretKeySpec
-import scala.util.{Failure, Success, Try}
+import scala.util.Try
 
 final case class Cookie(
   name: String,
@@ -178,7 +178,7 @@ object Cookie {
   private val sameSiteStrict = "strict"
   private val sameSiteNone   = "none"
 
-  private val dateTimeFormatter: DateTimeFormatter =
+  val dateTimeFormatter: DateTimeFormatter =
     DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss O", Locale.ENGLISH).withZone(ZoneOffset.UTC)
 
   sealed trait SameSite {
@@ -194,12 +194,7 @@ object Cookie {
    * Decodes from Set-Cookie header value inside of Response into a cookie
    */
   def decodeResponseCookie(headerValue: String, secret: Option[String] = None): Option[Cookie] =
-    Try(unsafeDecodeResponseCookie(headerValue, secret)) match {
-      case Failure(exception) =>
-        println(exception)
-        None
-      case Success(value)     => Some(value)
-    }
+    Try(unsafeDecodeResponseCookie(headerValue, secret)).toOption
 
   private[zhttp] def unsafeDecodeResponseCookie(headerValue: String, secret: Option[String] = None): Cookie = {
     var name: String              = null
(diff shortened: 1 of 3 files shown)


Dennis4b:
Please see the last comment in #1093, specifically:

* `DateTimeFormatter.RFC_1123_DATE_TIME` is problematic, because days-of-the-month `1-9` are not prefixed by a `0` in RFC 1123 but they are in the HTTP specification (more discussion of this at `https://stackoverflow.com/questions/7707555/getting-date-in-http-format-in-java`)

* The users default locale might not be English and output a month like 'Mai' instead of 'May'.

Both would be solved with

`DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss O", Locale.ENGLISH).withZone(ZoneOffset.UTC)`

gciuloaica:
> DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss O", Locale.ENGLISH).withZone(ZoneOffset.UTC)

Thanks @Dennis4b, somehow I missed that part

gciuloaica:
We are hitting a bug in JDK 8 - https://bugs.openjdk.org/browse/JDK-8154050
The issue is fixed in JDK 9 and all the other newer versions.

gciuloaica:
@Dennis4b the proposed format is having issues in JDK8 that is the minimum version that we support. See the link in the previous comment.

Dennis4b:
I see. I need/use it to set a proper expiration date on cookies (i.e. one that browsers understand :-), haven't needed to parse such a timestamp yet so haven't run into any issues.

What are the options/possibilities for a different approach?

gciuloaica:
> I see. I need/use it to set a proper expiration date on cookies (i.e. one that browsers understand :-), haven't needed to parse such a timestamp yet so haven't run into any issues.
> 
> What are the options/possibilities for a different approach?

I think the fastest solution is to drop support for JDK8 but as far as I'm aware there are still a lot of companies stuck on JDK8  for different reasons.

Dennis4b:
Sooo reading RFC 2616 closer there are actually 3 variants:
```
       HTTP-date    = rfc1123-date | rfc850-date | asctime-date
       rfc1123-date = wkday "," SP date1 SP time SP "GMT"
       rfc850-date  = weekday "," SP date2 SP time SP "GMT"
       asctime-date = wkday SP date3 SP time SP 4DIGIT
       date1        = 2DIGIT SP month SP 4DIGIT
                      ; day month year (e.g., 02 Jun 1982)
       date2        = 2DIGIT "-" month "-" 2DIGIT
                      ; day-month-year (e.g., 02-Jun-82)
       date3        = month SP ( 2DIGIT | ( SP 1DIGIT ))
                      ; month day (e.g., Jun  2)
       time         = 2DIGIT ":" 2DIGIT ":" 2DIGIT
                      ; 00:00:00 - 23:59:59
       wkday        = "Mon" | "Tue" | "Wed"
                    | "Thu" | "Fri" | "Sat" | "Sun"
       weekday      = "Monday" | "Tuesday" | "Wednesday"
                    | "Thursday" | "Friday" | "Saturday" | "Sunday"
       month        = "Jan" | "Feb" | "Mar" | "Apr"
                    | "May" | "Jun" | "Jul" | "Aug"
                    | "Sep" | "Oct" | "Nov" | "Dec"
```

For output the RFC1123+plus+leading+zero format can be used, but if we want to parse dates as well then all cases should be handled (so regardless of JDK8 it's not a complete solution).

The best options are probably:
* Find appropriately licensed code which already solves this, or
* Write a proper parser for the above grammar, or
* Determine which variant is used via a simple string check, drop the trailing " GMT" (only GMT is allowed anyway), and use the appropriate (out of 3) java dateformatter without any `O` flags to avoid the JDK8 bug.

tusharmath (approved):

PR #1291, opened 2022-06-07
Feature: Middleware for TrailingSlash

### LINKED ISSUE #1290: Unable to match on root (1.0.0.0-RC29)
**Describe the bug**
Following an upgrade to version 1.0.0.0-RC29 I noticed that the endpoint responsible for dealing with the root of my application was not behaving correctly.

I used to have this
```scala
val app = Http.collect[Request] { case Method.GET -> !! => Response.text("ok") }
```
But as `!!` is equals to `Path.empty` it does not match the incoming request  (that has a default trailing slash when looking at the path in the debugger).

**To Reproduce**
```scala
val app = Http.collect[Request] { case Method.GET -> !! => Response.text("ok") }
val app2 = Http.collect[Request] { case req @ Method.GET -> path if path == Path.decode("/")  => Response.text("ok") }
```

**Expected behaviour**
I would expect `app` and `app2` to behave the same in this part, or if `!!` is really supposed to be `Path.empty` then there should be an easy way to pattern match on the root (right now by using the unapply method of `/:` and `/` it is not possible).

**Screenshots**

**Additional context**

The following spec can be added to `PathSpec.scala` to reproduce the issue, however I have no clue on how to fix this bug/behavior (if this is a bug):

```scala
val path = collect { case !! => true }
assert(path(Path.decode("/")))(isSome(equalTo(true)))
```

gciuloaica:
fixes #1290 
Introduced `Middleware.trailingSlashDrop` and `Middleware.trailingSlashRedirect` to drop or redirect to client in case of  trailing slash presence, as the trailing slash is not heaving any meaning in a `HttpApp`.


tusharmath:
- [x] Update PR description with a more meaningful change

tusharmath:
- [x] Update PR description with a more meaningful change

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +173,31 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
       _ => middleware,
       _ => Middleware.identity,
     )
+
+  /**
+   * Removes the trailing slash from the path.
+   */
+  final def trailingSlashDrop[R, E]: HttpMiddleware[R, E] =
+    Middleware.identity[Request, Response].contramap[Request](removeTrailingSlash)
```
tusharmath:
```suggestion
    Middleware.identity[Request, Response].contramap[Request](_.dropTrailingSlash)
```

- Will need to add `dropTrailingSlash` in `Request` and `Path`
- Also add `addTrailingSlash` in `Request` and `Path`

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +173,31 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
       _ => middleware,
       _ => Middleware.identity,
     )
+
+  /**
+   * Removes the trailing slash from the path.
+   */
+  final def trailingSlashDrop[R, E]: HttpMiddleware[R, E] =
```
tusharmath:
```suggestion
  final def dropTrailingSlash: HttpMiddleware[Any, Nothing] =
```

We don't need `R, E`

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +173,31 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
...
+    Middleware.identity[Request, Response].contramap[Request](removeTrailingSlash)
+
+  /**
+   * Permanent redirect if the trailing slash is present in the request URL.
+   */
+  final def trailingSlashRedirect[R, E]: HttpMiddleware[R, E] =
```
tusharmath:
```suggestion
  final def redirectTrailingSlashRedirect(permanent: Boolean): HttpMiddleware[Any, Nothing] =
```

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +173,31 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
...
+    Middleware.whenRequest(_.url.path.trailingSlash)(
+      trailingSlashDrop ++ Middleware.intercept[Request, Response](identity)((_, req) =>
+        Response(
+          status = Status.PermanentRedirect,
+          headers = Headers(HttpHeaderNames.LOCATION, encode(removeTrailingSlash(req).url)),
+        ),
```
tusharmath:
```suggestion
        Response.redirect(req.dropTrailingSlash, isPermanent)
```

zio-http/src/main/scala/zhttp/http/Path.scala
```diff
@@ -36,6 +36,10 @@ final case class Path(segments: Vector[String], trailingSlash: Boolean) { self =
 
   def toList: List[String] = segments.toList
 
+  def dropTrailingSlash: Path = self.copy(trailingSlash = false)
+
+  def addTrailingSlash: Path = self.copy(trailingSlash = true)
+
```
tusharmath: Need this API in `Url` also as mentioned here — https://github.com/dream11/zio-http/pull/1291#pullrequestreview-1002306902

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +172,22 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
...
+
+  /**
+   * Permanent redirect if the trailing slash is present in the request URL.
+   */
+  final def redirectTrailingSlash(permanent: Boolean): HttpMiddleware[Any, Nothing] =
+    Middleware.whenRequest(_.url.path.trailingSlash)(
+      dropTrailingSlash ++ Middleware.intercept[Request, Response](identity)((_, req) =>
+        Response.redirect(encode(req.dropTrailingSlash.url), isPermanent = permanent),
+      ),
+    )
```
tusharmath: - Trailing slash should only be removed if there are no query params as per the RFC
- I think this middleware will still execute the HttpApp but ignore it's result.

zio-http/src/test/scala/zhttp/http/middleware/WebSpec.scala
```diff
@@ -178,4 +211,12 @@ object WebSpec extends DefaultRunnableSpec with HttpAppTestExtensions {
       res <- fib.join
     } yield res
   }
+
+  private def runTrailingSlash[R, E](app: HttpApp[R, E]): ZIO[TestClock with R, Option[E], Response] = {
+    for {
+      fib <- app { Request(url = URL.fromString("/").toOption.getOrElse(URL(!!))) }.fork
+      _   <- TestClock.adjust(10 seconds)
+      res <- fib.join
+    } yield res
```
tusharmath: why do we need `fork` here? There is not clock usage anywhere.

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
gciuloaica: you are right, it is executing the  HttpApp.

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +172,24 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
       _ => middleware,
       _ => Middleware.identity,
     )
+
+  /**
+   * Removes the trailing slash from the path.
+   */
+  final def dropTrailingSlash: HttpMiddleware[Any, Nothing] =
+    Middleware.identity[Request, Response].contramap[Request](_.dropTrailingSlash).when(_.url.queryParams.isEmpty)
```
tusharmath: This is nice! Finally an actual usecase for `when`

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +172,24 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
...
+
+  /**
+   * Permanent redirect if the trailing slash is present in the request URL.
+   */
+  final def redirectTrailingSlash(permanent: Boolean): HttpMiddleware[Any, Nothing] =
+    Middleware.ifThenElse[Request](_.url.path.trailingSlash)(
+      req =>
+        dropTrailingSlash ++ Middleware.fromHttp(
+          Http.response(Response.redirect(encode(req.dropTrailingSlash.url), isPermanent = permanent)),
+        ),
```
tusharmath:
```suggestion
        dropTrailingSlash ++ Middleware.redirect(req.dropTrailingSlash.url, permanent),
```
We can add `Middleware.redirect` as another Middleware. I think it's useful.

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
```diff
@@ -171,6 +172,24 @@ private[zhttp] trait Web extends Cors with Csrf with Auth with HeaderModifier[Ht
...
+  final def dropTrailingSlash: HttpMiddleware[Any, Nothing] =
+    Middleware.identity[Request, Response].contramap[Request](_.dropTrailingSlash).when(_.url.queryParams.isEmpty)
+
+  /**
+   * Permanent redirect if the trailing slash is present in the request URL.
+   */
+  final def redirectTrailingSlash(permanent: Boolean): HttpMiddleware[Any, Nothing] =
+    Middleware.ifThenElse[Request](_.url.path.trailingSlash)(
+      req =>
+        dropTrailingSlash ++ Middleware.fromHttp(
```
tusharmath: Why do we need to combine it with `dropTrailingSlash`?

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
gciuloaica: Done

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
gciuloaica: added

zio-http/src/main/scala/zhttp/http/middleware/Web.scala
gciuloaica: by mistake, no point to combine it.

tusharmath (approved):

PR #1307, opened 2022-06-17
Removed html bodies from default error responses

### LINKED ISSUE #1282: Cannot override the default error pages which contain sensitive info
The default error pages cannot be overridden and they contain sensitive info such as the internal error message and stack trace.
This is a security concern.

* The default error pages should either have no body or shouldn't show any dangerous info such as internal error messages or stack traces in production.
* The error pages should be fully customizable. They are currently hard-coded in a few places and cannot be replaced.
* Provide an error handler to allow the user to customize the response in certain failure scenarios.

Overriding the error pages or mapping the `Response` object to remove the body of non 2xx pages. Neither are currently supported by `zio-http`. 

**Additional context**
Screenshot 2022-06-01 at 17 09 11

andcea:
fixes https://github.com/dream11/zio-http/issues/1282

This change removes the html bodies of the default error pages, and only returns empty status codes as requested in the issue above.

### DIFF
--- a/project/.bloop/zio-http-build.json
+++ b/project/.bloop/zio-http-build.json
@@ -1,0 +1,1994 @@
+{
+    "version": "1.4.0",
+    "project": {
+        "name": "zio-http-build",
+        "directory": "/Users/andcea/workspace/zio-http/project",
+        "workspaceDir": "/Users/andcea/workspace/zio-http/project",
+        "sources": [
+            "/Users/andcea/workspace/zio-http/project/src/main/scala",
+            "/Users/andcea/workspace/zio-http/project/src/main/scala-2.12",
+            "/Users/andcea/workspace/zio-http/project/src/main/scala-2",
+            "/Users/andcea/workspace/zio-http/project/src/main/java",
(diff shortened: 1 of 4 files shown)


987Nabil:
I think there are ppl that do like the error pages for dev mode. I'd recommend to not just delete it, but as @tusharmath suggested in the issue, offer a middleware that gives you the option to "beautify" the errors. This should most likely only be moving some code instead of deleting it. I think it is reasonable to do this in one PR.

guizmaii:
IMO, having a different XP in dev and prod mode is a bad idea

tusharmath:
I think step one should be to move the beautification option to a middleware. That way people who want it can use it and others can keep it simple, just status codes.

I believe when we make requests from curl and there is an error, I think it's worth while to send the stacktrace to the user.

andcea:
This PR is trying to address the immediate security risk that stack traces introduce to a service when exposed publicly.
Exposing them only when using curl is as dangerous IMO.

I'm not sure I'll have time to contribute a middleware for pretty errors, but I think this fix should be merged in regardless to mitigate the issue.

andcea:
Closing in favour of https://github.com/dream11/zio-http/pull/1334

PR #1318, opened 2022-06-27
Feature: Add `seeOther` to `Response`

rleibman:


zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -219,6 +219,12 @@ object Response {
     Response(status, Headers.location(location))
   }
 
+  /**
+   * Creates an empty response with status 303
+   */
+  def seeOther(location: String): Response =
```
tusharmath: Take url as input insted of string

rleibman: Yes, probably.... but then shouldn't redirect do the same? It currently takes a CharSequence

tusharmath: Hmm, in that case, let's keep it as `CharSequence`. Atleast it's consistent that way.

tusharmath (approved):

PR #1320, opened 2022-06-28
Fix: Multiread on HttpData get's hung

### LINKED ISSUE #1207: IllegalReferenceCountException
**Describe the bug**
When we upgraded from RC25 to RC27 our very simple server started throwing this exception

**To Reproduce**
Sorry, not clear, it's a simple server, the request is a POST with json (circe) data, responding also with json data (though I don't think it gets that far)

**Expected behaviour**
We shouldn't get an exception, and if we do, there should be clear logs.

 - Device: reproduced locally on IOS or ubuntu cloud

**Additional context**
Hey... after we upgraded from RC25 to RC27, we started getting the following error:

```
Uncaught failure: io.netty.util.IllegalReferenceCountException: refCnt: 0

<init>:30, IllegalReferenceCountException (io.netty.util)
ensureAccessible:1454, AbstractByteBuf (io.netty.buffer)
checkIndex:1383, AbstractByteBuf (io.netty.buffer)
checkDstIndex:1409, AbstractByteBuf (io.netty.buffer)
getBytes:1062, CompositeByteBuf (io.netty.buffer)
getBytes:2031, CompositeByteBuf (io.netty.buffer)
getBytes:49, CompositeByteBuf (io.netty.buffer)
getBytes:1336, ByteBufUtil (io.netty.buffer)
getBytes:1309, ByteBufUtil (io.netty.buffer)
getBytes:1301, ByteBufUtil (io.netty.buffer)
$anonfun$bodyAsByteArray$2:21, HttpDataExtension (zhttp.http)
apply:-1, HttpDataExtension$$Lambda$2824/0x000000080121c0c8 (zhttp.http)
evaluateNow:496, FiberContext (zio.internal)
$anonfun$fork$15:781, FiberContext (zio.internal)
run:-1, FiberContext$$Lambda$220/0x0000000800e0eba0 (zio.internal)
safeExecute$$$capture:164, AbstractEventExecutor (io.netty.util.concurrent)
safeExecute:-1, AbstractEventExecutor (io.netty.util.concurrent)
```

BTW.... I couldn't detect any logging on this issue coming out of zhttp or netty, so I had to go into the debugger, put a breakpoint in the exception's constructor and capture the stack trace from there.

### LINKED ISSUE #1309: bodyAsString can only be called once
**Describe the bug**

**To Reproduce**
2. Apply the middleware twice
3. Send a request

**Expected behaviour**

gciuloaica:
Fixes #1309


example/src/main/scala/example/MultipleCallsOfRequestBodyAsString.scala
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+  private val server =
+    Server.port(8090) ++ Server.app(app @@ logMiddleware) ++ Server.enableObjectAggregator(Int.MaxValue)
+
+  override def run(args: List[String]): URIO[zio.ZEnv, ExitCode] =
+    server.make.useForever
+      .provideCustomLayer(ServerChannelFactory.auto ++ EventLoopGroup.auto(0))
+      .exitCode
+}
```
tusharmath: We should delete it before merging.

ithinkicancode:
Hi @gciuloaica, does this PR actually fix the issue reported ("bodyAsString can only be called once") or does this PR merely report the error ("The body content could not be read multiple times in case of streaming. Use `Server.enableObjectAggregator`")?

I skimped through the changes - I might be wrong - it doesn't seem this PR allows multiple read from the request body. If this PR indeed fixes the issue and allows multiple reads, could you add a unit test that calls `bodyAsString` multiple times?

I ran into a similar/same issue with `request.data.toByteBuf`, which exhibits the same issue that @ghostdogpr encountered: the second call hangs forever. In fact, `bodyAsString` indirectly calls `toByteBuf` therefore I believe the issue is related to `toByteBuf`.

Please kindly let me know the intent of this PR: is it a fix or is it just improvement over reporting "multiple read isn't possible"?

Thank you very much.

gciuloaica:
> Hi @gciuloaica, does this PR actually fix the issue reported ("bodyAsString can only be called once") or does this PR merely report the error ("The body content could not be read multiple times in case of streaming. Use `Server.enableObjectAggregator`")?
> 
> I skimped through the changes - I might be wrong - it doesn't seem this PR allows multiple read from the request body. If this PR indeed fixes the issue and allows multiple reads, could you add a unit test that calls `bodyAsString` multiple times?
> 
> I ran into a similar/same issue with `request.data.toByteBuf`, which exhibits the same issue that @ghostdogpr encountered: the second call hangs forever. In fact, `bodyAsString` indirectly calls `toByteBuf` therefore I believe the issue is related to `toByteBuf`.
> 
> Please kindly let me know the intent of this PR: is it a fix or is it just improvement over reporting "multiple read isn't possible"?
> 
> Thank you very much.

@ithinkicancode  `Server.objectAggregator` flag is dictating the behaviour of how the request body get processed, in chunked or not checked mode. For the situation when this flag is enabled the multi-read of the body is fixed. But if this flag is not enabled, so for chunked mode, there will be an exception telling the fact that multi read is not  possible.

ithinkicancode:
Thanks for the explanation @gciuloaica !

It'd be great if you can include two unit tests that makes multiple calls to `bodyAsString`:

1. one calls Server.enableObjectAggregator, and the multiple calls to `bodyAsString` should succeed.
2. one  *doesn't call* Server.enableObjectAggregator, and the second call to `bodyAsString` should expect the error.

Thanks again.

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -53,23 +54,31 @@ private[zhttp] final case class Handler[R](
         val hasBody = canHaveBody(jReq)
         log.debug(s"HasBody: [${hasBody}]")
         log.debug(s"HttpRequest: [${jReq.method} ${jReq.uri()}]")
-        if (hasBody) ctx.channel().config().setAutoRead(false): Unit
+        if (hasBody) {
+          ctx.channel().attr(bodyReadFlag).set(false)
```
tusharmath: This will introduce a bug when the same channel is being used for multiple requests.

gciuloaica: resolved.

gciuloaica:
Closing this PR as it is split in 2 different   PRs:
- fix for the situation when object aggregator is in the pipeline. (#1350 )
- fix for the chunked request. (TBD)

PR #1334, opened 2022-07-06
added implementation to prettify error message as middleware

### LINKED ISSUE #1282: Cannot override the default error pages which contain sensitive info

gciuloaica:


tusharmath:
There are some places in the `Handler` also where we catch exceptions and create a 500 exception. Ideally, those too should handled by this middleware, otherwise we will have a bit of inconsistent way to display error messages.

gciuloaica:
> There are some places in the `Handler` also where we catch exceptions and create a 500 exception. Ideally, those too should handled by this middleware, otherwise we will have a bit of inconsistent way to display error messages.

@tusharmath the error handling in `Handler` are happening after the `HttpApp` has been evaluated, so the middleware execution has been completed (unsafeRun completed). One way to unify both parts will be to migrate the formatting code to a different place (outside middleware implementation) and re-use it in Handler. To implement this it will require to know in the Handler if the middleware is in the pipeline or not.

zio-http/src/main/scala/zhttp/http/middleware/WebWithError.scala
```diff
@@ -0,0 +1,3 @@
+package zhttp.http.middleware
+
+class WebWithError {}
```
jdegoes, line 3: Dead code, let's delete

zio-http/src/main/scala/zhttp/http/middleware/package.scala
```diff
@@ -3,4 +3,6 @@ package zhttp.http
 package object middleware {
   type HttpMiddleware[-R, +E]       = MonoMiddleware[R, E, Request, Response]
   type MonoMiddleware[-R, +E, A, B] = Middleware[R, E, A, B, A, B]
+
+  type HttpMiddlewareWithError[-R, +E] = MonoMiddleware[R, E, Option[Request], Option[Response]]
```
jdegoes, line 7: Dead alias, let's delete

zio-http/src/main/scala/zhttp/service/Handler.scala
```diff
@@ -25,70 +25,62 @@ private[zhttp] final case class Handler[R](
...
-
-              override def unsafeContext: Ctx = ctx
-            },
-          )
-        catch {
-          case throwable: Throwable => resWriter.write(throwable, jReq)
-        }
-
-      case msg: HttpContent =>
+        unsafeRun(
```
jdegoes, line 54: These changes seem unnecessary for the scope of this PR. We should aim to do this as a "pure middleware" that does not touch anything else in the library.

jdegoes (changes requested):
Would love to get this in!

Please address comments, fix conflicts, and rebase. Thank you!

devsprint:
Closing this in favor of #1492

PR #1336, opened 2022-07-06
Fix HttpRuntime 

gciuloaica:


zio-http/src/test/scala/zhttp/service/TestSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+import zhttp.http.{Http, HttpData, Request, Response}
+import zhttp.internal.{DynamicServer, HttpRunnableSpec}
+import zhttp.service.server.ServerChannelFactory
+import zio.stream.ZStream
+import zio.test.Assertion.{containsString, equalTo}
+import zio.test.TestAspect.{sequential, timeout}
+import zio.test.{TestEnvironment, assertZIO}
+import zio.{ZIO, durationInt}
+
+object TestSpec extends HttpRunnableSpec {
```
amitksingh1490: rename this spec properly

gciuloaica: @amitksingh1490  do we really need this test? It is already part of the ServerSpec

amitksingh1490: delete it

PR #1350, opened 2022-07-14
Fix HttpData multiread when  aggregator is in the pipeline

### LINKED ISSUE #1309: bodyAsString can only be called once

gciuloaica:
Fixes #1309


tusharmath:
This is fixed in #1351

PR #1351, opened 2022-07-15
Refactor: Re-write `HttpData` to `Body`

### LINKED ISSUE #1309: bodyAsString can only be called once

tusharmath:
fixes #1309


zio-http/src/main/scala/zhttp/http/Body.scala
```diff
@@ -0,0 +1,248 @@
...
+                        ctx.read(): Unit
+                      }
+                    },
+                  ): Unit
+              },
+              1,
+            )
+          }
+        }
+      }
```
tusharmath: Too much of unsafe pipeline logic that can be moved to the backend.

zio-http/src/main/scala/zhttp/http/Response.scala
```diff
@@ -31,40 +29,30 @@ final case class Response private (
...
-      case HttpData.UnsafeAsync(_) => null
-      case data: HttpData.Complete =>
-        data match {
-          case HttpData.FromAsciiString(text) => Unpooled.wrappedBuffer(text.array())
-          case HttpData.BinaryChunk(data)     => Unpooled.wrappedBuffer(data.toArray)
-          case HttpData.BinaryByteBuf(data)   => data
-          case HttpData.BinaryStream(_)       => null
-          case HttpData.Empty                 => Unpooled.EMPTY_BUFFER
-          case HttpData.JavaFile(_)           => null
+  private[zhttp] def unsafeEncode(): Task[HttpResponse] = for {
```
tusharmath: - [ ] rename to `encode` 
- [ ] Can be moved into ResponseWriter

PR #1353, opened 2022-07-16
Feature: Add `toJavaURI` and `toJavaURL` to `URL`

### LINKED ISSUE #1337: Add method to convert `zhttp.http.URL` to `java.net.URI`
Maybe also `java.net.URL`

jgoday:
fixes #1337

It adds a conversion method to create a java.net.URI/URL from a zhttp.http.URL.

Uses HttpGen.genAbsoluteURL to test it against random urls. To avoid an encode exception, any relative path within an absolute location must be converted to an absolute one.

```
java.net.URISyntaxException: Relative path in absolute URI: https://S:321971123a/b
  at java.base/java.net.URI.checkPath(URI.java:1965)
```

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -87,6 +87,35 @@ final case class URL(
 
     copy(kind = location)
   }
+
+  /**
+   * Returns a new java.net.URI representing this URL.
+   */
+  def toJavaURI: Option[java.net.URI] = self.kind match {
+    case URL.Location.Relative                     => None
+    case URL.Location.Absolute(scheme, host, port) =>
+      val queryStr    = queryParams.filter { case (k, _) => k.nonEmpty }.map { case (k, v) =>
+        v.map(v => s"$k=$v").mkString("&")
+      }.filter(_.nonEmpty).mkString("&")
+      val fragmentStr = fragment.map(_.raw).getOrElse(null)
+
+      Some(
+        new java.net.URI(
+          scheme.encode,
+          null,
+          host,
+          port,
+          if (path.leadingSlash || path.isEmpty) path.encode else (Path.root ++ path).encode,
+          if (queryStr.isEmpty) null else queryStr,
+          fragmentStr,
+        ),
+      )
+  }
+
+  /**
+   * Returns a new java.net.URL representing this URL.
+   */
+  def toJavaURL: Option[java.net.URL] = toJavaURI.map(_.toURL)
 }
 object URL {
   private def fromAbsoluteURI(uri: URI): Option[URL] = {
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -94,12 +94,10 @@ final case class URL(
   def toJavaURI: java.net.URI = new URI(encode)
 
   /**
-   * Returns a new java.net.URL representing this URL.
+   * Returns a new java.net.URL only if this URL represents an absolute
+   * location.
    */
-  def toJavaURL: Option[java.net.URL] = self.kind match {
-    case URL.Location.Relative => None
-    case _                     => Some(toJavaURI.toURL)
-  }
+  def toJavaURL: Option[java.net.URL] = if (self.kind == URL.Location.Relative) None else Some(toJavaURI.toURL)
 }
 object URL {
   private def fromAbsoluteURI(uri: URI): Option[URL] = {
--- a/zio-http/src/main/scala/zhttp/http/URL.scala
+++ b/zio-http/src/main/scala/zhttp/http/URL.scala
@@ -91,31 +91,15 @@ final case class URL(
   /**
    * Returns a new java.net.URI representing this URL.
    */
-  def toJavaURI: Option[java.net.URI] = self.kind match {
-    case URL.Location.Relative                     => None
-    case URL.Location.Absolute(scheme, host, port) =>
-      val queryStr    = queryParams.filter { case (k, _) => k.nonEmpty }.map { case (k, v) =>
-        v.map(v => s"$k=$v").mkString("&")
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -87,6 +87,35 @@ final case class URL(
 
     copy(kind = location)
   }
+
+  /**
+   * Returns a new java.net.URI representing this URL.
+   */
+  def toJavaURI: Option[java.net.URI] = self.kind match {
```
tusharmath: I think this should not return an Option. Relative urls can also be encoded to URI.

zio-http/src/test/scala/zhttp/http/URLSpec.scala
```diff
@@ -82,5 +82,30 @@ object URLSpec extends ZIOSpecDefault {
...
+                autoUrl
+            val url         = URL.fromString(absoluteUrl.encode).getOrElse(absoluteUrl)
+            val uri         = url.toJavaURI
+            val reversedUrl = uri.flatMap(u => URL.fromString(u.toString).toOption)
+
+            assert(reversedUrl)(
+              equalTo(Some(url)),
+            )
+          }
+        },
```
tusharmath, line 97: I think we are doing too much over here. I think the test should assert for symmetry. Something like 

```scala
val gen: Gen[URL] = ???

check(gen) { url =>
  val httpURLString = url.encode
  val javaURLString = url.toJava.toString
  assertTrue(httpURLString == javaURLString)
}
```

tusharmath (changes requested):

zio-http/src/main/scala/zhttp/http/URL.scala
jgoday: you're right.
however, java.net.URL does not support relative paths,
should we keep the Option return signature in toJavaURL ?

zio-http/src/test/scala/zhttp/http/URLSpec.scala
jgoday: That's exactly the reason about this workaround,
right now, URL, does not seems to be symmetric

```scala
val original: String = "https://localhost/?b=a&H=c&b=b"
val target: String = zhttp.http.URL.fromString(original).toOption.get.encode

// val original: String = https://localhost/?b=a&H=c&b=b
// val target: String = https://localhost/?b=a&b=b&H=c

```

Some other problem is relative paths over an absolute location,
encode generates an invalid url:
```scala
val c = zhttp.http.URL(
    zhttp.http.Path.decode("a/b"),
    zhttp.http.URL.Location.Absolute(zhttp.http.Scheme.HTTPS, "localhost", 433)).encode
// val c: String = https://localhost:433a/b

zhttp.http.URL.fromString(c)
// val res0: Either[Exception,zhttp.http.URL] = Left(java.net.MalformedURLException: Invalid URL: "https://localhost:433a/b")
```	
and some automatically generated urls has this very shape.

Should we consider this ?:
 * and encode problem
 * An incosistent combination of path and location
 * return 'None' or 'Either.left' in toJavaURI method

zio-http/src/main/scala/zhttp/http/URL.scala
tusharmath: I think we can have both.

zio-http/src/test/scala/zhttp/http/URLSpec.scala
tusharmath: In that case, we can create a `Gen` from an iterable and manually specify the input cases.

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -87,6 +87,35 @@ final case class URL(
...
+          if (queryStr.isEmpty) null else queryStr,
+          fragmentStr,
+        ),
+      )
+  }
+
+  /**
+   * Returns a new java.net.URL representing this URL.
+   */
+  def toJavaURL: Option[java.net.URL] = toJavaURI.map(_.toURL)
```
987Nabil:
```suggestion
  def toJavaURL: Option[java.net.URL] = 
  if(self.kind == URL.Location.Relative) None else Some(toJavaURI.toURL)
```

987Nabil: Then you can just always convert in `toJavaURI`

zio-http/src/main/scala/zhttp/http/URL.scala
jgoday: Ok, changed in the latest commit

zio-http/src/test/scala/zhttp/http/URLSpec.scala
jgoday: Seems that java.net.URI inner parser takes into account all these cases, updated too.

zio-http/src/main/scala/zhttp/http/URL.scala
jgoday: Ok, just updated and fixed test case too

zio-http/src/main/scala/zhttp/http/URL.scala
```diff
@@ -91,31 +91,15 @@ final case class URL(
...
-        ),
-      )
-  }
+  def toJavaURI: java.net.URI = new URI(encode)
 
   /**
    * Returns a new java.net.URL representing this URL.
    */
-  def toJavaURL: Option[java.net.URL] = toJavaURI.map(_.toURL)
+  def toJavaURL: Option[java.net.URL] = self.kind match {
```
987Nabil: I think it is preferable to use `if` in simple cases. Especially since this is http lib code and might have performance implications.

jgoday: Of course, updated in the latest commit

tusharmath:
Thanks @jgoday !

tusharmath (approved):

PR #1373, opened 2022-07-28
Fix: Host header must include port

### LINKED ISSUE #1372: `Host` header doesn't include port
**Describe the bug**

According to the HTTP 1.1 spec (https://datatracker.ietf.org/doc/html/rfc2616#section-14.23) `Host` header must include both host and port if they're specified in the requested URL. `zhttp.service.EncodeRequest` only includes the host. 

This makes zio-http client unsuitable for calling some security-related services - e.g. `userinfo` request in Keycloak (https://www.keycloak.org/docs/latest/securing_apps/index.html#userinfo-endpoint) will fail if the service runs on non-standard port

**To Reproduce**
1. Pick some HTTP server running on non-standard port (say, 1234)
1. Run a traffic interceptor (such as Wireshark) for this port
2. Send a request with zio-http client - e.g. `Client.request("http://localhost:1234/some/api", Method.GET)`
3. See HTTP GET request having `Host` header set to `localhost` instead of `localhost:1234`

**Expected behaviour**
HTTP GET request having `Host` header contains both host and port - e.g. `localhost:1234`

paul-lysak:
Fixes #1372 - according to HTTP 1.1 spec (https://datatracker.ietf.org/doc/html/rfc2616#section-14.23) Host header must include port if the request is sent to the custom port

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/Client.scala
+++ b/zio-http/src/main/scala/zhttp/service/Client.scala
@@ -63,12 +63,8 @@ final case class Client[R](rtm: HttpRuntime[R], cf: JChannelFactory[JChannel], e
   ): JChannelFuture = {
 
     try {
-      val (host, port) = (req.url.host, req.url.port) match {
-        case (Some(host), Some(port)) => (host, port)
-        case (Some(host), None) => (host, 80)
-        case _ => assert(false, "Host name is required")
-          ("", -1)
-      }
+      val host = req.url.host.getOrElse { assert(false, "Host name is required"); "" }
+      val port = req.url.port.getOrElse(80)
 
       val isWebSocket = req.url.scheme.exists(_.isWebSocket)
       val isSSL       = req.url.scheme.exists(_.isSecure)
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zhttp/service/Client.scala
```diff
@@ -63,12 +63,12 @@ final case class Client[R](rtm: HttpRuntime[R], cf: JChannelFactory[JChannel], e
...
-
-      assert(host != null, "Host name is required")
-
-      val port = req.url.port.getOrElse(80)
+      val (host, port) = (req.url.host, req.url.port) match {
+        case (Some(host), Some(port)) => (host, port)
+        case (Some(host), None) => (host, 80)
+        case _ => assert(false, "Host name is required")
+          ("", -1)
+      }
```
tusharmath: Let's avoid tuple allocations as much as possible.

tusharmath:
Run `scalafmt` once on your code.

zio-http/src/main/scala/zhttp/service/Client.scala
paul-lysak: Is this for performance reasons or for style preferences?
I've changed the approach a bit, please take a look.

paul-lysak:
@tusharmath @amitksingh1490 can you please run the CI? It's been 2 weeks since I've committed the latest fixes

paul-lysak:
`Build and Test (ubuntu-latest, 2.12.15, temurin@8) Expected — Waiting for status to be reported` - @tusharmath - should I rebase to the fresh code from the `main` branch to have test environment versions updated?

tusharmath (approved):

PR #1379, opened 2022-07-30
Fix: Propagate `FiberRef` To `HttpRuntime`

### LINKED ISSUE #1352: Logger doesn't propagate to `Http`
**Describe the bug**
When using an SLF4J logger with ZIO HTTP, the logger logs in the correct format only outside an `Http`, but inside the `Http` the logger format is the default one.

**To Reproduce**
Run this program:

```scala
  val app = Http.collectZIO[Request] { case Method.GET -> !! =>
    ZIO.logInfo("this is a message") *> // this is logged in the default format
    ZIO.succeed(Response.ok)
  }

  def run = {
    val logger = Runtime.removeDefaultLoggers >>> SLF4J.slf4j(LogLevel.All, LogFormat.line, _ => "zio-rest-api")
    val program = for {
      _ <- ZIO.logInfo("Starting app") // this logs fine
      app <- Server.start(8090, app)
    } yield app
    program.provide(logger)
  }
}
```

**Expected behaviour**
Both "Starting app" and "this is a message" should be logged in the SLF4J format.

**Additional context**
I'm using Scala 3.1.3, ZIO HTTP 2.0.0-RC10, ZIO 2.0.0, and ZIO Logging 2.0.0.

Many thanks 🙂

adamgfraser:
Resolves #1352.

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/HttpRuntime.scala
+++ b/zio-http/src/main/scala/zhttp/service/HttpRuntime.scala
@@ -1,10 +1,9 @@
 package zhttp.service
 
 import io.netty.channel.{ChannelHandlerContext, EventLoopGroup => JEventLoopGroup}
-import io.netty.util.concurrent.{EventExecutor, Future, GenericFutureListener}
+import io.netty.util.concurrent.{Future, GenericFutureListener}
 import zio._
 
-import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 /**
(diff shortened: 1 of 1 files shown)


gciuloaica:
@adamgfraser I have tested the change and it is working fine. The proper logger is used. Thank you.
However, it seems that are some issues with the tests.

tusharmath (approved):

tusharmath:
Thank you Adam! 🙏

PR #1385, opened 2022-08-08
Fix: Validate cookie signature on request

### LINKED ISSUE #1383: No way to verify cookie signature
Good evening dear team,

As far as I understand, there's no clear way to verify the signature of a request cookie (as of v1.0.0.0-RC29)
There's the `verify` method in `Cookie`, but it's private.

What's the right way of verifying incoming cookies then?
Ideally, I'd expect the server to verify signatures during decoding (like here for example https://github.com/pillarjs/cookies)

Thank you.

gciuloaica:
Fixes #1383


zio-http/src/main/scala/zhttp/http/headers/HeaderGetters.scala
```diff
@@ -166,6 +166,11 @@ trait HeaderGetters[+A] { self =>
       Cookie.decodeRequestCookie(header)
     }
 
+  final def signedCookiesDecoded(secret: String): List[Cookie] =
+    headerValues(HeaderNames.cookie).flatMap { header =>
+      Cookie.decodeRequestCookie(header, Some(secret))
+    }
+
```
tusharmath, line 173: I think we can overload `cookiesDecoded` with a `secret` for consistency.

gciuloaica: updated.

tusharmath (approved):

PR #1387, opened 2022-08-10
Remove event loop

gciuloaica:


gciuloaica:
Different approach was suggested

PR #1393, opened 2022-08-15
Feature: Add the ability to get resources from JAR archives

nartamonov:
Currently `Http.fromResource` is able to provide file resources only. However, it is often necessary to provide a resource that is inside a JAR archive. For example, if you run an application through `sbt run`, the application classes and resources will be packaged into a temporary JAR archive, and the resource URLs will have a `jar:file` scheme. Now such resources cannot be obtained with `Http.fromResource`. This PR adds this feature. Based on [the implementation](https://github.com/akka/akka-http/blob/main/akka-http/src/main/scala/akka/http/scaladsl/server/directives/FileAndResourceDirectives.scala#L263-L291) of `ResourceFile` from `akka-http`.


zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -869,7 +863,57 @@ object Http {
...
+                    .withContentLength(contentLength)
+              } yield mediaType.fold(app)(t => app.withMediaType(t))).onError(_ => closeJar)
+            }
+
+        Http.fromZIO(appZIO).flatten.catchAll {
+          case Some(e) => Http.fail(e)
+          case None    => Http.empty
+        }
+      case proto  =>
+        Http.die(new IllegalArgumentException(s"Unsupported protocol: $proto"))
```
```suggestion
        Http.fail(new IllegalArgumentException(s"Unsupported protocol: $proto"))
```

zio-http/src/test/scala/zhttp/service/StaticFileServerSpec.scala
```diff
@@ -64,6 +64,36 @@ object StaticFileServerSpec extends HttpRunnableSpec {
...
+          test("should have content-type") {
+            val res = resourceOk.run().map(_.mediaType)
+            assertZIO(res)(isSome(equalTo(MediaType.text.plain)))
+          } +
+          test("should respond with empty") {
+            val res = resourceNotFound.run().map(_.status)
+            assertZIO(res)(equalTo(Status.NotFound))
+          }
+      },
+    ),
```
tusharmath, line 97: Can you please use `,` instead of `+` to combine the tests.

zio-http/src/main/scala/zhttp/http/Http.scala
```diff
@@ -869,7 +863,57 @@ object Http {
...
+    }
+  }
+
+  private def determineMediaType(filePath: String): Option[MediaType] = {
+    // Extract file extension
+    val ext = filePath.lastIndexOf(".") match {
+      case -1 => None
+      case i  => Some(filePath.substring(i + 1))
+    }
+    ext.flatMap(MediaType.forFileExtension)
```
tusharmath: instead of doing a flatmap we can move this operation above. That will save us a lambda and an additional Option.

zio-http/src/test/scala/zhttp/service/StaticFileServerSpec.scala
nartamonov: Fixed.

zio-http/src/main/scala/zhttp/http/Http.scala
nartamonov: Thanks, fixed.

tusharmath (approved):

PR #1394, opened 2022-08-18
feat: add metrics middleware

frekw:
Since ZIO 2.0 ships with integrated metrics support, I think it makes sense to include middleware for some standard metrics into zio-http.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -99,7 +99,7 @@ lazy val root = (project in file("."))
     zioHttpBenchmarks,
     zioHttpLogging,
     zioHttpExample,
-    zioHttpTestkit
+    zioHttpTestkit,
   )
 
 lazy val zioHttp = (project in file("zio-http"))
--- a/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
+++ b/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
@@ -1,7 +1,7 @@
 package zio.http.middleware
 
-import zio.http.{Http, Middleware, Request, Response}
 import zio._
+import zio.http.{Http, Middleware, Request, Response}
 import zio.metrics.{Metric, MetricKeyType, MetricLabel}
 
 private[zio] trait Metrics {
--- a/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
+++ b/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
@@ -11,6 +11,7 @@ private[zio] trait Metrics {
     totalRequestsName: String = "http_requests_total",
     requestDurationName: String = "http_request_duration_seconds",
     requestDurationBoundaries: MetricKeyType.Histogram.Boundaries = Metrics.defaultBoundaries,
+    extraLabels: Set[MetricLabel] = Set.empty,
   ): HttpMiddleware[Any, Nothing] = {
     new Middleware[Any, Nothing, Request, Response, Request, Response] {
       val requestsTotal      = Metric.counterInt(totalRequestsName)
@@ -24,7 +25,7 @@ private[zio] trait Metrics {
         Set(
           MetricLabel("method", req.method.toString),
           MetricLabel("path", pathLabelMapper.lift(req).getOrElse(req.path.toString())),
-        )
+        ) ++ extraLabels
 
       def labelsForResponse(res: Response): Set[MetricLabel] =
         Set(
--- a/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
+++ b/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
@@ -54,7 +54,7 @@ private[zio] trait Metrics {
 
       def apply[R1 <: Any, E1 >: Nothing](
         http: Http[R1, E1, Request, Response],
-      ): Http[R1, E1, Request, Response] =
+      )(implicit trace: Trace): Http[R1, E1, Request, Response] =
         Http.fromOptionFunction[Request] { req =>
           val requestLabels = labelsForRequest(req)
 
--- a/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
+++ b/zio-http/src/main/scala/zio/http/middleware/Metrics.scala
@@ -5,98 +5,72 @@ import zio._
 import zio.metrics.{Metric, MetricKeyType, MetricLabel}
 
 private[zio] trait Metrics {
-  def metrics[R, E](
-    mapper: Metrics.Mapper = Map.empty,
-    boundaries: MetricKeyType.Histogram.Boundaries = Metrics.defaultBoundaries,
-  ): HttpMiddleware[R with Metrics.InFlightRequests, E] =
-    new Middleware[R with Metrics.InFlightRequests, E, Request, Response, Request, Response] {
-      private def metricLabelsForRequest(req: Request): Set[MetricLabel] =
+  def metrics(
+    pathLabelMapper: PartialFunction[Request, String] = Map.empty,
+    concurrentRequestsName: String = "http_concurrent_requests_total",
+    totalRequestsName: String = "http_requests_total",
+    requestDurationName: String = "http_request_duration_seconds",
+    requestDurationBoundaries: MetricKeyType.Histogram.Boundaries = Metrics.defaultBoundaries,
+  ): HttpMiddleware[Any, Nothing] = {
+    new Middleware[Any, Nothing, Request, Response, Request, Response] {
+      val requestsTotal      = Metric.counterInt(totalRequestsName)
+      val concurrentRequests = Metric.gauge(concurrentRequestsName)
+      val requestDuration    = Metric.histogram(requestDurationName, requestDurationBoundaries)
+      val status404          = Set(MetricLabel("status", "404"))
+      val status500          = Set(MetricLabel("status", "500"))
+      val nanosToSeconds     = 1e9d
+
+      def labelsForRequest(req: Request): Set[MetricLabel] =
         Set(
           MetricLabel("method", req.method.toString),
-          MetricLabel("path", mapper.lift(req).getOrElse(req.path.toString())),
+          MetricLabel("path", pathLabelMapper.lift(req).getOrElse(req.path.toString())),
         )
 
-      private def withRequestMeta[Type, In, Out](metric: Metric[Type, In, Out])(req: Request) =
-        metric.tagged(metricLabelsForRequest(req))
+      def labelsForResponse(res: Response): Set[MetricLabel] =
+        Set(
+          MetricLabel("status", res.status.code.toString),
+        )
 
-      private def requestsTotal(req: Request) =
-        withRequestMeta(Metric.counterInt("http_requests_total"))(req)
+      def observeRequest(start: Long, labels: Set[MetricLabel]) =
+        for {
+          _   <- requestsTotal.tagged(labels).increment
+          end <- Clock.nanoTime
+          took = end - start
+          _ <- requestDuration.tagged(labels).update(took / nanosToSeconds)
+        } yield ()
 
-      private val errorCount = Metric.counterInt("http_errors_total")
-
-      private def errorCountWithRequest(req: Request) =
-        withRequestMeta(errorCount)(req)
-
-      private def errorCountWithResponse(req: Request, res: Response) =
-        errorCountWithStatus(req, res.status.code.toString)
-
-      private def errorCountWithStatus(req: Request, status: String) =
-        errorCountWithRequest(req).tagged("status_code", status)
-
-      private def requestDuration(req: Request) =
-        withRequestMeta(
-          Metric.histogram(
-            "http_request_duration_seconds",
-            boundaries,
-          ),
-        )(req).trackDurationWith(toSeconds)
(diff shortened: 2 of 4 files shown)


zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,102 @@
...
+                  err.fold(errorCountWithStatus(req, "404").increment)(_ => errorCountWithStatus(req, "500").increment),
+                res => ZIO.when(res.status.code > 399)(errorCountWithResponse(req, res).increment),
+              )
+              .tapDefect(_ => errorCountWithStatus(req, "500").increment)) @@ requestDuration(req)
+        }
+    }
+}
+
+object Metrics {
+  type Mapper = PartialFunction[Request, String]
```
frekw: Can be used to override mappings (i.e map certain paths like `/users/1` to `/users/:id`) to avoid having Prometheus label count explode. It would be nice if we could deal with placeholders automatically. Perhaps via the new API work @kitlangton is doing?

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  // Prometheus defaults
```
frekw: I think defaulting to Prometheus' default buckets make sense since I suspect most people would be using it, but can be overriden in the constructor.

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,102 @@
...
+    ),
+  )
+
+  trait InFlightRequests {
+    def increment(labels: Set[MetricLabel]): UIO[Int]
+    def decrement(labels: Set[MetricLabel]): UIO[Int]
+    def get: UIO[Int]
+  }
+
+  object InFlightRequests {
```
frekw: This is needed because gauges can't be incremented or decremented so we have to keep track of the current number of in flight requests ourselves :(

We have to make a tradeoff here: either we use a `Ref` to keep track of the current value and base the Gauge's value off of that. This won't give us readings under concurrency (since the ref and the updating of the gauge aren't atomic).

The alternative is to use `RefM` and update the state and the gauge from within `RefM#update`, but that means that _every request_ needs to acquire two locks for updating the `RefM` which just seems worse to me.

tusharmath:
@frekw This would be a great addition to the ZIO Http ecosystem.
I think it would better to maintain it as separate project, in the same repo.

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,102 @@
...
+    def increment(labels: Set[MetricLabel]): URIO[InFlightRequests, Any] = ZIO.serviceWithZIO(_.increment(labels))
+    def decrement(labels: Set[MetricLabel]): URIO[InFlightRequests, Any] = ZIO.serviceWithZIO(_.decrement(labels))
+    val get: URIO[InFlightRequests, Int]                                 = ZIO.serviceWithZIO(_.get)
+
+    val live = ZLayer.fromZIO {
+      for {
+        ref <- Ref.make(0)
```
tusharmath: Though I am ok to let people just play around with `zio-http-metrics` initially, 
to gather feedback and focus on performance later. Using a Ref will increase contention between threads for every request. 
What if we use `FiberRef` instead?

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,102 @@
...
+import zhttp.http.{Http, Middleware, Request, Response}
+import zio._
+import zio.metrics.{Metric, MetricKeyType, MetricLabel}
+
+private[zhttp] trait Metrics {
+  def metrics[R, E](
+    mapper: Metrics.Mapper = Map.empty,
+    boundaries: MetricKeyType.Histogram.Boundaries = Metrics.defaultBoundaries,
+  ): HttpMiddleware[R with Metrics.InFlightRequests, E] =
+    new Middleware[R with Metrics.InFlightRequests, E, Request, Response, Request, Response] {
```
tusharmath: I think there is quite a lot of logic that's got nothing to do with a middleware. 
How about we create a `Metrics` service that does all this stuff we test that directly instead of doing it via the middleware API.

That way we will later get flexibility to also benchmark the service and see how it performs under stress.

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  trait InFlightRequests {
+    def increment(labels: Set[MetricLabel]): UIO[Int]
+    def decrement(labels: Set[MetricLabel]): UIO[Int]
+    def get: UIO[Int]
+  }
+
+  object InFlightRequests {
+    def increment(labels: Set[MetricLabel]): URIO[InFlightRequests, Any] = ZIO.serviceWithZIO(_.increment(labels))
+    def decrement(labels: Set[MetricLabel]): URIO[InFlightRequests, Any] = ZIO.serviceWithZIO(_.decrement(labels))
```
tusharmath:
```suggestion
    def increment(labels: Set[MetricLabel]): URIO[InFlightRequests, Int] = ZIO.serviceWithZIO(_.increment(labels))
    def decrement(labels: Set[MetricLabel]): URIO[InFlightRequests, Int] = ZIO.serviceWithZIO(_.decrement(labels))
```
Shouldn't this be int?

frekw:
> @frekw This would be a great addition to the ZIO Http ecosystem.
I think it would better to maintain it as separate project, in the same repo.

Definitely! What do you think is a good project structure in this case? Should we add an `extras` folder or something similar or should we just add `metrics` to the top level?

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
frekw: I don't really see what that service would do – all helpers essentially extract different metadata about the request or the response to appropriately tag the metrics with. We can definitely move that to its own module if you think it helps readability. 👍

zio-http/src/main/scala/zhttp/http/middleware/Metrics.scala
frekw: Yes, I have [an open ticket](https://github.com/zio/zio/issues/7217) in ZIO to add `Gauge#increment` and `Gauge#decrement` to be able to eliminate the ref entirely. Until it's been decided how that should progress I think we should hold off merging because the performance impact will be too big.

>  What if we use FiberRef instead?

I don't think that'll work unfortunately because the active requests gauge needs to be global for all requests. :/

frekw: This will be fixed by https://github.com/zio/zio/pull/7284 🥳🥳🥳

987Nabil:
@frekw I would highly recommend to make metric names configurable. Orgs might have naming schemas that a team has to stick to. We have this and I know others as well. And I think making names configurable should not be too complicated.

jdegoes:
@frekw  If you want this to become #1459, I can help shepherd this through to merging. The functionality is most-welcome!

frekw:
@jdegoes now that Gauge.increment/decrement is in core I'd love to take another stab at it 👍

frekw:
@jdegoes I've pushed an updated version that uses the new Gauge.increment/decrement. Temporarily changed the dependency to the snapshot version of ZIO to get access to them but should hold off merging until there's a proper release. It gets a bit repetitive to handle all cases (sucess / error / defect) but nothing too bad.

What I'm struggling a bit with is writing tests that aren't dependent on global state (and execution order) – but I guess that's an issue with testing any logic around metrics, or is there a way to create a scoped, fresh metrics registry solely for testing?

zio-http/src/main/scala/zio/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,76 @@
+package zio.http.middleware
+
+import zio.http.{Http, Middleware, Request, Response}
+import zio._
+import zio.metrics.{Metric, MetricKeyType, MetricLabel}
+
+private[zio] trait Metrics {
+  def metrics(
+    pathLabelMapper: PartialFunction[Request, String] = Map.empty,
```
frekw, line 29: Can be used to override mappings (i.e map certain paths like /users/1 to /users/:id) to avoid having Prometheus label count explode (and thus crippling its performance).

adamgfraser:
@frekw I think you can do this with tags.

zio-http/src/main/scala/zio/http/middleware/Metrics.scala
```diff
@@ -0,0 +1,76 @@
...
+                  observeRequest(start, requestLabels ++ err.fold(status404)(_ => status500))
+                },
+                response => {
+                  observeRequest(start, requestLabels ++ labelsForResponse(response))
+                },
+              )
+              .tapDefect(_ => {
+                observeRequest(start, requestLabels ++ status500)
+              })
+          } yield res).ensuring(concurrentRequests.tagged(requestLabels).decrement)
```
frekw: Not sure about the performance implications of `.ensuring` here – would it be better to just have `observeRequest` accept the `requestLabels` as well and decrement the gauge?

project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   val NettyVersion                  = "4.1.82.Final"
   val NettyIncubatorVersion         = "0.0.15.Final"
   val ScalaCompactCollectionVersion = "2.8.1"
-  val ZioVersion                    = "2.0.2"
+  val ZioVersion                    = "2.0.2+47-e1206a44-SNAPSHOT"
```
frekw: Only for dev, make sure we have 2.0.3 before merge.

build.sbt
```diff
@@ -82,6 +82,7 @@ ThisBuild / githubWorkflowBuildPostamble :=
     ),
     scalas = List(Scala213),
   ).steps
+ThisBuild / resolvers ++= Resolver.sonatypeOssRepos("snapshots")
```
frekw: Only for dev, remove before merge.

zio-http/src/main/scala/zio/http/middleware/Metrics.scala
adamgfraser: You could replace the `tapBoth`, `tapDefect` and `ensuring` with `onExit`.

frekw:
@adamgfraser ah, by accepting something like `extraTags: Set[MetricLabel] = Set.empty`? Clever! 

I'll look at that in the morning, need to jump to bed now! 😴

zio-http/src/main/scala/zio/http/middleware/Metrics.scala
frekw: Ah, TIL! 👍

frekw:
Ok, I think this is good to go @adamgfraser @jdegoes

zio-http/src/test/scala/zio/http/middleware/MetricsSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+        .flatten @@ metrics(extraLabels = Set(MetricLabel("test", "http_requests_total & http_errors_total")))
+
+      val total   = Metric.counterInt("http_requests_total").tagged("test", "http_requests_total & http_errors_total")
+      val totalOk = total.tagged("path", "/ok").tagged("method", "GET").tagged("status", "200")
+      val totalErrors   = total.tagged("path", "/error").tagged("method", "GET").tagged("status", "500")
+      val totalDefects  = total.tagged("path", "/defect").tagged("method", "GET").tagged("status", "500")
+      val totalNotFound = total.tagged("path", "/not-found").tagged("method", "GET").tagged("status", "404")
+
+      for {
+        _ <- app(Request(method = Method.GET, url = URL(!! / "ok")))
```
swoogles: Very minor - `Request.method` defaults to `GET`, so that bit can be removed from lines 30-32, 58, and 77.
However, you might prefer the more explicit approach here, since we're tagging things by the method.

swoogles:
Great work here! 👍

jdegoes:
@swoogles Since you also did some work, can you work with @frekw to get this one through?

swoogles:
> @swoogles Since you also did some work, can you work with @frekw to get this one through?

Yep, I've already submitted a PR against his branch with a few small additions. Otherwise I think he nailed it.

frekw:
Sorry @swoogles I didn't have notifications turned on for my fork so I had missed it. Merged now ✅

frekw:
Ok, I've rebased based on the latest changes to `main` and updated the test accordingly. CI should pass now. 👍

frekw:
@jdegoes I noticed master is now using a ZIO snapshot so I've rebased. This should be good to merge now.

frekw:
Whoops, I spoke too early, looks like the Middleware API has changed. I'll adjust.

jdegoes (approved):

jdegoes:
Thank you!

PR #1397, opened 2022-08-19
Feat: add support for server response compression

girdharshubham:
# Description
This PR adds in support for Http Response Compression as a Server config. The server needs to be built with the appropriate config `Server.responseCompression` by setting `contentSizeThreshold` and an `IndexedSeq` of `CompressionOptions`, based on this config, the internal netty handler will be added to the pipeline with the set `CompressionOptions`.

## Changelog
* Server Config for HttpResponseCompression
* Add HttpContentCompressor in ServerChannelInitializer
* New Type for CompressionOptions
* Test Cases

### DIFF
--- a/zio-http/src/main/scala/zhttp/service/Server.scala
+++ b/zio-http/src/main/scala/zhttp/service/Server.scala
@@ -5,7 +5,7 @@ import io.netty.channel.ChannelPipeline
 import io.netty.util.ResourceLeakDetector
 import zhttp.http.{Http, HttpApp}
 import zhttp.service.server.ServerSSLHandler._
-import zhttp.service.server.{CompressionOptions, _}
+import zhttp.service.server._
 import zio._
 
 import java.net.{InetAddress, InetSocketAddress}
--- a/zio-http/src/main/scala/zhttp/service/Server.scala
+++ b/zio-http/src/main/scala/zhttp/service/Server.scala
@@ -5,7 +5,7 @@ import io.netty.channel.ChannelPipeline
 import io.netty.util.ResourceLeakDetector
 import zhttp.http.{Http, HttpApp}
 import zhttp.service.server.ServerSSLHandler._
-import zhttp.service.server._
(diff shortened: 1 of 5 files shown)


girdharshubham:
@jdegoes @adamw @kitlangton @tusharmath could you guys review this?

jdegoes:
@girdharshubham Love it! Please update to latest, address feedback, and we can get this merged.

zio-http/src/main/scala/zhttp/service/Server.scala
```diff
@@ -207,6 +217,9 @@ object Server {
 
   def requestDecompression(strict: Boolean): UServer = Server.RequestDecompression(enabled = true, strict = strict)
 
+  def responseCompression(contentSizeThreshold: Int, options: IndexedSeq[CompressionOptions]): UServer =
```
jdegoes, line 220: Add `ServerConfig.ResponseCompression` and add `responseCompression: Option[ResponseCompression]` to `ServerConfig` case class, possibly choosing a reasonable default.

zio-http/src/main/scala/zhttp/service/server/CompressionOptions.scala
```diff
@@ -0,0 +1,34 @@
+package zhttp.service.server
+
+import io.netty.handler.codec.compression.{CompressionOptions => JCompressionOptions, StandardCompressionOptions}
+
+final case class CompressionOptions(level: Int, bits: Int, mem: Int, kind: CompressionOptions.CompressionType) { self =>
```
jdegoes, line 5: I think we should move this to `ServerConfig` companion object, if that config is the only one that needs it.

zio-http/src/main/scala/zhttp/service/Server.scala
girdharshubham: Thanks for the review @jdegoes
Will add them next.

girdharshubham:
@jdegoes closing this in favour of #1486 
Don't have access to push straight to this org

PR #1401, opened 2022-08-23
Remove space after Expires in cookie

onsoup42:
Parsing cookies with and expires fails because the parser expects a space at the end. Also encoding cookie will have an extra space after the expires date.
Expected:
```Expires=Tue, 23 Aug 2022 21:07:21 GMT;```
actual:
```Expires=Tue, 23 Aug 2022 21:07:21 GMT ;```

### DIFF
--- a/zio-http/src/main/scala/zhttp/http/Cookie.scala
+++ b/zio-http/src/main/scala/zhttp/http/Cookie.scala
@@ -179,7 +179,7 @@ object Cookie {
   private val sameSiteNone   = "none"
 
   val dateTimeFormatter: DateTimeFormatter =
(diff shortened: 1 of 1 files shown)


girdharshubham (approved):

onsoup42:
I see now that the space was added as a workaround for JDK 8 in #1288.  We can probably close this PR unless we are ok dropping java 8 support. For now I will work around this by not using `Cookie.decodeResponseCookie` or the `setCookiesDecoded` on a Response.

girdharshubham:
Thanks for raising this PR @onsoup42. I don't think dropping support for jdk 8 is in the works right now.

PR #1402, opened 2022-08-23
Add convenient request method to Client class

palanga:
Thumbs up ?


tusharmath (approved):

cowfoodbagman:
Your return type should be `Task[Response]` I believe.

PR #1409, opened 2022-08-30
Fix: CORS Headers implementation

### LINKED ISSUE #1408: CORS middleware header `access-control-allow-headers` is broken
With the current zio-http code from `main` branch (v2.0.0-RC10 doesn't have this bug) browsers can't parse `access-control-allow-headers` and fail to process cross-origin requests. 
Chrome fails with an error: "Cannot parse Access-Control-Allow-Headers response header field in preflight response."

It happens because the header value is quoted and browsers consider everything in quotes as a header name. 
E.g.  `access-control-allow-headers: "content-type,authorization,*"` - `content-type,authorization,*` is considered a header name, which fails to parse.

Pull request https://github.com/dream11/zio-http/pull/1355 added ability to specify multiple headers with the same name
(by switching from `headers.set` to `headers.add`) which is good, but it also switched from `DefaultHttpHeaders` to `CombinedHttpHeaders` 
which tries to combine multiple headers into one. This has 2 consequences:

- developers that use zio-http aren't able to specify that they really want 2 headers sent unless this is a special case (such as Set-Cookie), `CombinedHttpHeaders` will forcibly compress them into a single
header with multiple values
- values containing comas will be quoted even if there's just one header with such name - so, during the encoding the `access-control-allow-headers: content-type,authorization,*` header 
will become `access-control-allow-headers: "content-type,authorization,*"` which can't be parsed by the browser.

### LINKED ISSUE #1421: Illegal header formatting in RC11
Since RC11, headers seem to be incorrectly formatted - there are some problems with adding `"`. Specifically, when the `WWW-Authenticate` header is set to `Basic realm="realm"`, the client receives `"Basic realm=""realm"""` (with many `"` added).

This used to work in RC10 and works for all other server implementations in tapir.

This breaks clients, e.g. the clients that come bundled with java:

```scala
import zhttp.http._
import zhttp.service.Server
import zio._

import java.net.URI
import java.net.http.HttpResponse.BodyHandlers
import java.net.http.{HttpClient, HttpRequest}
import scala.jdk.CollectionConverters.{CollectionHasAsScala, MapHasAsScala}

object Demo extends ZIOAppDefault {
  val app: HttpApp[Any, Nothing] = Http.collect[Request] { case _ =>
    Response.text("").setStatus(Status.Unauthorized).addHeader(("WWW-Authenticate", "Basic realm=\"realm\""))
  }

  override val run = for {
    _ <- Server.start(8090, app).fork
    _ <- ZIO.sleep(Duration.fromSeconds(1))
    _ <- ZIO.succeed {
      println("Using java.net.URL:")
      val client1 = new java.net.URL("http://localhost:8090/")
      val conn = client1.openConnection()
      val headers = conn.getHeaderFields.asScala.toVector
        .filter(_._1 != null)
        .flatMap { case (k, vv) => vv.asScala.map(k -> _) }
      println("Headers:")
      headers.foreach(println(_))

      // Fails on Java 11 (but works on 17) 
      println("Using java.net.http.HttpClient:")
      val client = HttpClient.newHttpClient()
      client.send(HttpRequest.newBuilder().uri(URI.create("http://localhost:8090/")).build(), BodyHandlers.ofString())
    }
  } yield ()
}
```

gciuloaica:
Fixed CORS headers implementation to provide them to encoder as individual headers instead of merging headers with same name into one. This will avoid the encode method to consider that the header value need to be escaped.

Fixes: #1408 
Fixes: #1421


zio-http/src/test/scala/zhttp/internal/HttpRunnableSpec.scala
```diff
@@ -114,4 +114,22 @@ abstract class HttpRunnableSpec extends ZIOSpecDefault { self =>
...
+      headers <- Client
+        .request(
+          "http://localhost:%d/%s".format(port, path),
+          method,
+          ssl = ClientSSLOptions.DefaultSSL,
+          headers = headers,
+        )
+        .map(_.headers)
+    } yield headers
+  }
```
tusharmath: Do we need a integration test for this? Will just unit test not suffice?

gciuloaica: Unit test might be sufficient for CORS specific issues but the integration tests will help also in future to catch issues in http encoder/decoder code, or other netty internal changes related to headers

zio-http/src/main/scala/zhttp/http/Headers.scala
```diff
@@ -37,11 +37,18 @@ final case class Headers(toChunk: Chunk[Header]) extends HeaderExtension[Headers
...
    */
-  private[zhttp] def encode: HttpHeaders =
-    self.toList
-      .foldLeft[HttpHeaders](new CombinedHttpHeaders(true)) { case (headers, entry) =>
+  private[zhttp] def encode: HttpHeaders = {
+    val (exceptions, regularHeaders) = self.toList.span(h => h._1.contains(HeaderNames.setCookie))
+    val combinedHeaders              = regularHeaders
+      .groupBy(_._1)
+      .map { case (key, tuples) =>
+        key -> tuples.map(_._2).map(value => if (value.contains(",")) s"""\"$value\"""" else value).mkString(",")
```
adwells-ds: Since you are already on it, please use default Netty methods, that supports `Iterable` with proper escape mechanism, thanks!
https://github.com/netty/netty/blob/3efd9642ca17b9470f38eea8d100e4cec9c41291/codec-http/src/main/java/io/netty/handler/codec/http/CombinedHttpHeaders.java#L168-L170

Example
https://scastie.scala-lang.org/8lW5DwGYSh6HcJnyXCgNUA

gciuloaica: Just went away from `CombinedHttpHeaders` as it has multiple issues.

jdegoes (approved):

PR #1423, opened 2022-09-08
Default environment 

gciuloaica:
Improved the Client and Server API by providing traits that could be mixed-in client and server app.


jdegoes:
Closed in favor of #1460

PR #1434, opened 2022-09-11
Fix resource leak in `Http.fromResource`

987Nabil:
Hey, I am pretty sure, the current impl. is leaking resources.
`attemptBlockingIO(new java.util.zip.ZipFile(filePath))` is open the resource, but it is only **ensured** that is is closed once the stream is created. Wich for example does not happen, if the requested path leads to a directory `ZIO.when(entry.isDirectory)(ZIO.fail(None))`. There is this `onError(_ => closeJar)`, but this will not ensure that the resource is closed in case of an defect/interrupt.

I ended up with opening the resource twice. I tried it first with an `aquireReleaseWith` and creating the stream inside based on the jar, but this closes the resource before the stream finishes.

@adamgfraser am I right that this is leaking? Maybe you have an idea how to open it only once?

### DIFF
--- a/zio-http/src/main/scala/zio/http/Http.scala
+++ b/zio-http/src/main/scala/zio/http/Http.scala
@@ -9,11 +9,12 @@ import zio.http.html._
 import zio.http.socket.{SocketApp, WebSocketChannelEvent}
 import zio.stream.ZStream
 
-import java.io.{File, IOException}
+import java.io.{File, FileNotFoundException, IOException}
 import java.net
 import java.net.{InetAddress, InetSocketAddress}
 import java.nio.charset.Charset
 import java.nio.file.Paths
+import java.util.zip.ZipFile
 import scala.annotation.unused
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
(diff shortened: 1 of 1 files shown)


tusharmath:
You can actually write a unit test for this. If there is a resource leak, there will be an error thrown.

987Nabil:
@tusharmath how? The leak would only happen in case of a defect or in interrupt at the right time. I could maybe write similar code and a test for it to proof I am right, but I don't see how I can provoke this with the real impl.

jdegoes (approved):

jdegoes:
@987Nabil This looks like a legit resource leak, can you fix conflicts?

987Nabil:
@jdegoes is this solution good, or is there a way to open the file only once?

987Nabil:
I actually wonder, if there should not be better support for reading jar resources in ZIO core (ZIO/ZStream) on which this could be build. Or is there and I just don't know?

vigoo:
I think there is no simpler way right now. As we already have support for compression in streams in ZIO streams it could make sense to have support for reading zip entries too - could you open an issue for that in the ZIO repo?

Issue #1338, opened 2022-07-06
Add a method to `zhttp.http.URL` to retrieve the query params as a String 


987Nabil:
Just a simple String representation of all query params `?k1=v1&k2=v2...`
Maybe add a proxy to `Request`

adrianfilip:
Picked it up.

adrianfilip:
#1443

PR #1443, opened 2022-09-12
Feature: add QueryParams

adrianfilip:


adrianfilip:
#1338

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -98,6 +98,9 @@ final case class URL(
    * location.
    */
   def toJavaURL: Option[java.net.URL] = if (self.kind == URL.Location.Relative) None else Some(toJavaURI.toURL)
+
+  def queryParamsAsString: String =
+    self.queryParams.toList.map { case (k, v) => s"$k=${v.mkString(",")}" }.mkString("&")
```
jdegoes: Each key and value has to be URL encoded, because they may have spaces, commas, `&`, and other characters inside them, which should not be interpreted as part of the structure of the query params.

I'd add a test to ensure they are URL encoded properly.

This also needs to be symmetric with the parsing of a URL. That is, if we parse a URL into `URL`, and then generate a string, that string needs to be equal to the original URL.

Perhaps you could use the `QueryStringEncoder` for this method that is used in `URL.encode` in order to ensure integrity between these different ways of producing the query param string.

adrianfilip: Changed the implementation to use code that has decoding built in. Changed the test to cover more cases.

zio-http/src/test/scala/zio/http/URLSpec.scala
```diff
@@ -96,5 +97,40 @@ object URLSpec extends ZIOSpecDefault {
...
+              "http://abc.com/users",
+              "http://abc.com/users/?u=1&u=2&ord=ASC&txt=scala%3Fis%2Bawesome%21",
+              "http://abc.com/users#the%20hash",
+              "ws://abc.com/subscriptions",
+              "ws://abc.com/subscriptions?",
+              "wss://abc.com/subscriptions",
+            ),
+          )
+
+          checkAll(urls) { genUrl =>
```
adrianfilip: I think this covers the added method query. 
There is another test for URL symmetry - https://github.com/zio/zio-http/blob/d9d18bb5332b3452f9e395da704e7cf0b824e077/zio-http/src/test/scala/zio/http/URLSpec.scala#L10

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -98,6 +98,10 @@ final case class URL(
    * location.
    */
   def toJavaURL: Option[java.net.URL] = if (self.kind == URL.Location.Relative) None else Some(toJavaURI.toURL)
+
+  def queryParamsAsString: String =
+    Option(self.toJavaURI.getQuery).getOrElse("")
+
```
tusharmath, line 109: I am not really convinced if this is the right way to expose query params.
We should ideally have a type such as - `QueryParam` that has methods such as `add`, `remove`, `encode` etc.

The above API seems a bit to specific for a particular use-case where the default value is empty string and will introduce an unnecessary breaking change later when something more ergonomic is available.

I think for now we can do this - 
```scala
url.toJavaURL.map(_.getQuery).getOrElse("")
```

zio-http/src/test/scala/zio/http/URLSpec.scala
tusharmath: Yes, also the internal implementation uses built in codecs so we are testing a library function here.
Not worth it IMO :)

zio-http/src/main/scala/zio/http/URL.scala
adrianfilip: @tusharmath 
I pushed an alternative that I think satisfies your intention.

I think that `queryParamsAsString` supporting only java urls might be confusing. Someone may make a URL from a ws endpoint and this new impl would throw because ws endpoints are a malformed java urls.

I kept my initial implementation and renamed it `uriQueryParamsAsString`.
I added a `urlQueryParamsAsString` where I use your suggestion. The only tweaks are to make it return the query decoded. I also refactored `toJavaURL` to make it pure (let me know if this should be reverted)

jdegoes: I do think it would be a good idea to have `def queryParams: QueryParams`, with methods like add/remove/etc.

Perhaps the optimal way to solve the problem would be `url.queryParams.encode`. 

Would you sign off on that API? Only a bit more typing for someone looking for a "fast" way to get the query parameters as an encoded string.

adrianfilip: Like this?
```
  trait QueryParams{
    def add(key: String, value: String): QueryParams
    def remove(key: String): QueryParams
    //this returns URI query params
    def queryParamsMap: Map[String, List[String]]
    //these return URI query params - or we can have a URLQueryParams and URIQuery params
    def decoded: String
    def encoded: String
  }
```

Also do we work on the URI level or want to make a distinction between URI and URL?

adrianfilip:
Added QueryParams but still have more tests to add and to make QueryParams extend Map. Moving it to WIP in the meantime.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -184,4 +178,79 @@ object URL {
...
+      val encoder = new QueryStringEncoder(s"")
+      map.foreach { case (key, values) =>
+        if (key != "") values.foreach { value => encoder.addParam(key, value) }
+      }
+
+      encoder.toString
+
+    }
+
+    def queryParamsMap: Map[String, List[String]] = map
```
tusharmath:
```suggestion
    def toMap: Map[String, List[String]] = map
```

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -184,4 +178,79 @@ object URL {
...
+
+    def apply(tuples: Tuple2[String, String]*): QueryParams =
+      QueryParams(map = tuples.groupBy(_._1).map { case (key, values) =>
+        key -> values.map(_._2).toList
+      })
+
+    def decode(queryStringFragment: String): QueryParams = queryParams(queryStringFragment)
+    val empty: QueryParams                               = QueryParams(Map.empty[String, List[String]])
+
+    private[http] def queryParams(query: String): QueryParams = {
```
tusharmath: Can be inlined at call-site.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -184,4 +178,79 @@ object URL {
     } yield Fragment(raw, decoded)
   }
 
+  final case class QueryParams private[http] (map: Map[String, List[String]])
```
tusharmath: I think `QueryParams` should be available at `zio.http` like `Path`.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -73,11 +75,11 @@ final case class URL(
     copy(kind = location)
   }
 
-  def setQueryParams(queryParams: Map[String, List[String]]): URL =
+  def setQueryParams(queryParams: QueryParams): URL =
```
tusharmath, line 78: I think we should keep the older API also, and add this as an overload.
We can even add  `(String, List[String]) *` as an overload.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -184,4 +178,79 @@ object URL {
...
+
+    override def get(key: String): Option[List[String]] = map.get(key)
+
+    override def iterator: Iterator[(String, List[String])] = map.iterator
+
+  }
+
+  object QueryParams {
+
+    def apply(tuples: Tuple2[String, String]*): QueryParams =
```
tusharmath: Shouldn't this be `Tuple2(String, List[String])` ?

zio-http/src/main/scala/zio/http/URL.scala
adrianfilip: Done

zio-http/src/main/scala/zio/http/URL.scala
adrianfilip: Done

zio-http/src/main/scala/zio/http/URL.scala
adrianfilip: Done

zio-http/src/main/scala/zio/http/URL.scala
adrianfilip: Done

zio-http/src/main/scala/zio/http/URL.scala
adrianfilip: Yes that would be a better fit. Updated it.

adrianfilip:
@tusharmath @jdegoes Can you give it another look please? I think it covers all feedback.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -41,11 +31,22 @@ final case class URL(
 
   def isRelative: Boolean = !isAbsolute
 
+  private[zio] def normalize: URL = {
```
adrianfilip, line 35: Moved it in alphabetical position.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -41,11 +31,22 @@ final case class URL(
...
+      self.queryParams.toMap.toList.filter(i => i._1.nonEmpty && i._2.nonEmpty).sortBy(_._1).toMap
+    self.copy(queryParams = QueryParams(queryParamsMap))
+  }
+
   def port: Option[Int] = kind match {
     case URL.Location.Relative      => None
     case abs: URL.Location.Absolute => Option(abs.port)
   }
 
+  private[zio] def relative: URL = self.kind match {
```
adrianfilip, line 46: Moved it in alphabetical position.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -0,0 +1,74 @@
+package zio.http
+
+import io.netty.handler.codec.http.{QueryStringDecoder, QueryStringEncoder}
+
+import scala.jdk.CollectionConverters._
+
+final case class QueryParams private[http] (map: Map[String, List[String]])
+    extends scala.collection.Map[String, List[String]] {
```
tusharmath: I am not sure if we should extend Map. Its possible to change the implementation to a declarative one where we might end up introducing a breaking change @jdegoes thoughts?

zio-http/src/test/scala/zio/http/QueryParamsSpec.scala
```diff
@@ -0,0 +1,220 @@
...
+import zio.test.Assertion.equalTo
+import zio.test.{ZIOSpecDefault, _}
+
+object QueryParamsSpec extends ZIOSpecDefault {
+
+  def spec =
+    suite("QueryParams")(
+      suite("-")(
+        test("removes query param") {
+          val gens = Gen.fromIterable(
```
tusharmath, line 13: We can use a Gen with random alpha numerics and check for symmetry between encoding and decoding

zio-http/src/main/scala/zio/http/QueryParams.scala
jdegoes: I think no matter the encoding we should be able to implement `Map` for it (by implementing an iterator). In some designs, such an iterator may not be performant, but even here, I regard it as a convenience (to be able to have collection methods on it, like `foldLeft`).

jdegoes:
@adrianfilip Let us know when it's ready for final review!

adrianfilip:
@jdegoes @tusharmath Applied changes. Can you review again please?

jdegoes (approved):

PR #1444, opened 2022-09-12
Refactor: Rename `Server.error` to `Server.onError`

### LINKED ISSUE #1032: Error recovery behavior is not clear
**Describe the bug**
1. It is not clear what `Server.error` is supposed to do. Does it recover from errors, or is it a `.tap` on them? 
2. It does not work either way...
3. By default, not recovering from an error results in fiber death

**To Reproduce**
```scala
  val app =
    Http.collectZIO[Request] {
      case Method.GET -> !! / "foo" => ZIO.fail(new Throwable("boom"))
    }

Server.app(app).withPort(9090).withError(e => ZIO.debug(s"an error happened: $e")).make <*> AsyncHttpClientZioBackend.managed() use { case (_, backend) =>
            basicRequest.get(uri"http://localhost:9090/foo").send(backend).debug
        }

// result: a 500 response with fiber trace
```

```
Response(Left(<html><head></head><body><h1>Internal Server Error</h1><pre><div>java.lang.Throwable: boom
	at <REDACTED>)
	at zio.ZIO$.$anonfun$fail$1(ZIO.scala:2898)
	at zio.internal.FiberContext.evaluateNow(FiberContext.scala:413)
	at zio.internal.FiberContext.$anonfun$fork$17(FiberContext.scala:781)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)</div></pre></body></html>),500,Internal Server Error,List(content-length: 700),List(),RequestMetadata(GET,http://localhost:9090/foo,Vector(Accept-Encoding: gzip, deflate)))
```

**Expected behaviour**
1. The error handler should have printed its content
2. By default, the error handler should be more graceful, e.g. return error 500 with `Internal Server Error` message

```scala
val recover = Http.error("Internal Server Error")
Server.app(app orElse recover)...
```

**Additional context**
Scala version: **2.13.7**
zio-http version: **1.0.0-RC23**
zio version: **1.0.13**

adrianfilip:
fixes #1032

### DIFF
--- a/zio-http/src/main/scala/zio/http/Server.scala
+++ b/zio-http/src/main/scala/zio/http/Server.scala
@@ -81,7 +81,7 @@ sealed trait Server[-R, +E] { self =>
   /**
    * Creates a new server with the errorHandler provided.
    */
-  def withError[R1](errorHandler: Throwable => ZIO[R1, Nothing, Unit]): Server[R with R1, E] =
+  def onError[R1](errorHandler: Throwable => ZIO[R1, Nothing, Unit]): Server[R with R1, E] =
     Concat(self, Server.OnError(errorHandler))
 
   /**
--- a/zio-http/src/main/scala/zio/http/Server.scala
+++ b/zio-http/src/main/scala/zio/http/Server.scala
@@ -175,7 +175,7 @@ object Server {
 
   def enableObjectAggregator(maxRequestSize: Int = Int.MaxValue): UServer = ObjectAggregator(maxRequestSize)
 
-  def error[R](errorHandler: Throwable => ZIO[R, Nothing, Unit]): Server[R, Nothing] = Server.Error(errorHandler)
+  def onError[R](errorHandler: Throwable => ZIO[R, Nothing, Unit]): Server[R, Nothing] = Server.Error(errorHandler)
 
   def make[R](
     server: Server[R, Throwable],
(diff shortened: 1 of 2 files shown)


zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -175,7 +175,7 @@ object Server {
 
   def enableObjectAggregator(maxRequestSize: Int = Int.MaxValue): UServer = ObjectAggregator(maxRequestSize)
 
-  def error[R](errorHandler: Throwable => ZIO[R, Nothing, Unit]): Server[R, Nothing] = Server.Error(errorHandler)
+  def onError[R](errorHandler: Throwable => ZIO[R, Nothing, Unit]): Server[R, Nothing] = Server.Error(errorHandler)
```
tusharmath: We should change it here also -
https://github.com/zio/zio-http/pull/1444/files#diff-5acd6bc78f56104e72f88d80f3126f9f7bb88f1264839086d0fc72160775d927R84

adrianfilip: Not sure what you mean. That link takes me to my change. :)

jdegoes: Maybe change the case class from `Error` to `OnError`:

<https://github.com/zio/zio-http/blob/aa01acb433e82b87b178d9a471122ac8886e5b36/zio-http/src/main/scala/zio/http/Server.scala#L277>

adrianfilip: Makes sense. Pushed.

tusharmath: There is a method inside of `Server::withError` which should also be renamed to `onError`.
Just search in the file `withError`, you will find it :)

adrianfilip: Thanks. Updated that also.

jdegoes (approved):

tusharmath (approved):

PR #1452, opened 2022-09-12
Unsafe API

### LINKED ISSUE #1448: Adopt ZIO 2.0 pattern for unsafe methods

Unsafe methods use the ZIO 1.x convention of prefixing with `unsafe`.

The ZIO 2.0 pattern for unsafe methods is to place them into a separate `.unsafe` namespace, and to not add any prefix to the method name, but rather, to accept an implicit `Unsafe` in the implicit parameter list.

**Why is it important?**

Consistency with ZIO 2.0 idioms.

vigoo:
Resolves #1448

### DIFF
--- a/zio-http/src/main/scala/zio/http/Cookie.scala
+++ b/zio-http/src/main/scala/zio/http/Cookie.scala
@@ -1,9 +1,9 @@
 package zio.http
 
-import zio.{Duration, Unsafe}
 import zio.http.Cookie.{SameSite, Type}
 import zio.http.CookieDecoder.log
 import zio.http.service.Log
+import zio.{Duration, Unsafe}
 
 import java.security.MessageDigest
 import java.util.Base64.getEncoder
--- a/zio-http/src/main/scala/zio/http/Cookie.scala
+++ b/zio-http/src/main/scala/zio/http/Cookie.scala
@@ -1,6 +1,6 @@
 package zio.http
 
-import zio.Duration
+import zio.{Duration, Unsafe}
 import zio.http.Cookie.{SameSite, Type}
 import zio.http.CookieDecoder.log
 import zio.http.service.Log
@@ -31,7 +31,9 @@ final case class Cookie[T](name: String, content: String, target: Cookie.Type[T]
    */
   def encode(validate: Boolean)(implicit ev: CookieEncoder[T]): Either[Exception, String] =
     try {
-      Right(ev.unsafeEncode(self, validate))
+      Unsafe.unsafe { implicit u =>
+        Right(ev.unsafe.encode(self, validate))
+      }
     } catch {
       case e: Exception =>
         log.error("Cookie encoding failure", e)
@@ -213,7 +215,9 @@ object Cookie {
    */
   def decode[S](string: String, validate: Boolean = false)(implicit ev: CookieDecoder[S]): Either[Exception, ev.Out] = {
     try {
-      Right(ev.unsafeDecode(string, validate))
+      Unsafe.unsafe { implicit u =>
+        Right(ev.unsafe.decode(string, validate))
+      }
     } catch {
       case e: Exception =>
         log.error("Cookie decoding failure", e)
--- a/zio-http/src/main/scala/zio/http/CookieDecoder.scala
+++ b/zio-http/src/main/scala/zio/http/CookieDecoder.scala
@@ -9,8 +9,8 @@ import scala.jdk.CollectionConverters._
 sealed trait CookieDecoder[A] {
   type Out
 
-  final def apply(cookie: String): Out =
-    Unsafe.unsafe { implicit u => unsafe.decode(cookie, validate = false) }
+  final def apply(cookie: String)(implicit unsafe: Unsafe): Out =
+    this.unsafe.decode(cookie, validate = false)
 
   trait UnsafeAPI {
     def decode(header: String, validate: Boolean)(implicit unsafe: Unsafe): Out
--- a/zio-http/src/main/scala/zio/http/CookieDecoder.scala
+++ b/zio-http/src/main/scala/zio/http/CookieDecoder.scala
@@ -1,6 +1,7 @@
 package zio.http
 
 import io.netty.handler.codec.http.{cookie => jCookie}
+import zio.Unsafe
 import zio.http.Cookie.SameSite
 import zio.http.service.Log
 
@@ -8,8 +9,14 @@ import scala.jdk.CollectionConverters._
 sealed trait CookieDecoder[A] {
   type Out
 
-  final def apply(cookie: String): Out = unsafeDecode(cookie, validate = false)
-  def unsafeDecode(header: String, validate: Boolean): Out
+  final def apply(cookie: String): Out =
+    Unsafe.unsafe { implicit u => unsafe.decode(cookie, validate = false) }
+
+  trait UnsafeAPI {
+    def decode(header: String, validate: Boolean)(implicit unsafe: Unsafe): Out
+  }
+
+  val unsafe: UnsafeAPI
 }
 
 object CookieDecoder {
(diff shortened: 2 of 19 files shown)


zio-http/src/main/scala/zio/http/CookieDecoder.scala
```diff
@@ -1,15 +1,22 @@
...
 import zio.http.service.Log
 
 import scala.jdk.CollectionConverters._
 sealed trait CookieDecoder[A] {
   type Out
 
-  final def apply(cookie: String): Out = unsafeDecode(cookie, validate = false)
-  def unsafeDecode(header: String, validate: Boolean): Out
+  final def apply(cookie: String): Out =
+    Unsafe.unsafe { implicit u => unsafe.decode(cookie, validate = false) }
```
vigoo: The unsafety of the decoder is not handled here at all

zio-http/src/main/scala/zio/http/CookieEncoder.scala
```diff
@@ -1,41 +1,52 @@
...
 
 import io.netty.handler.codec.http.{cookie => jCookie}
+import zio.Unsafe
 import zio.http.Cookie.SameSite
 
 sealed trait CookieEncoder[A] {
-  final def apply(a: Cookie[A]): String = unsafeEncode(a, false)
-  def unsafeEncode(a: Cookie[A], validate: Boolean): String
+  final def apply(a: Cookie[A]): String =
+    Unsafe.unsafe { implicit u => unsafe.encode(a, validate = false) }
```
vigoo: The unsafety of the encoder is not handled here

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -672,7 +672,8 @@ object Http {
     /**
      * Applies Http based on the path as string
      */
-    def whenPathEq(p: String): HttpApp[R, E] = http.when(_.unsafeEncode.uri().contentEquals(p))
+    def whenPathEq(p: String): HttpApp[R, E] =
+      http.when(request => Unsafe.unsafe { implicit u => request.unsafe.encode.uri().contentEquals(p) })
```
vigoo: Unsafety not handled

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -721,7 +722,8 @@ object Http {
   /**
    * Provides access to the request's ChannelHandlerContext
    */
-  def context: Http[Any, Nothing, Request, ChannelHandlerContext] = Http.fromFunction[Request](_.unsafeContext)
+  def context: Http[Any, Nothing, Request, ChannelHandlerContext] =
+    Http.fromFunction[Request](request => Unsafe.unsafe { implicit u => request.unsafe.context })
```
vigoo: This should probably be `fromFunctionZIO` to handle the unsafety

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -133,43 +141,63 @@ object Request {
...
-      override def method: Method            = Method.fromHttpMethod(jReq.method())
-      override def url: URL                  = URL.fromString(jReq.uri()).getOrElse(URL.empty)
-      override def version: Version          = Version.unsafeFromJava(jReq.protocolVersion())
-      override def unsafeEncode: HttpRequest = jReq
-      override def unsafeContext: Ctx        = ctx
-      override def body: Body = Body.fromAsync { async => ServerInboundHandler.Unsafe.addAsyncBodyHandler(async) }
+      override final def headers: Headers = Headers.make(jReq.headers())
+      override final def method: Method   = Method.fromHttpMethod(jReq.method())
+      override final def url: URL         = URL.fromString(jReq.uri()).getOrElse(URL.empty)
+      override final def version: Version =
```
vigoo: Unsafety not handled in `version` and `body`

zio-http/src/main/scala/zio/http/service/ServerInboundHandler.scala
```diff
@@ -29,22 +29,28 @@ private[zio] final case class ServerInboundHandler[R](
         val req  = Request.fromFullHttpRequest(jReq)
         val exit = http.execute(req)
 
-        if (self.attemptFastWrite(exit)) {
-          Unsafe.releaseRequest(jReq)
-        } else
-          runtime.unsafeRun {
-            self.attemptFullWrite(exit, jReq) ensuring ZIO.succeed { Unsafe.releaseRequest(jReq) }
-          }
+        Unsafe.unsafe { implicit u =>
```
vigoo: Maybe it would be better to have an implicit unsafe for the whole class as it's implementing a Netty interface

zio-http/src/main/scala/zio/http/CookieEncoder.scala
jdegoes: I think we should require `implicit unsafe: Unsafe` here, then.

zio-http/src/main/scala/zio/http/Http.scala
jdegoes: Ditto.

zio-http/src/main/scala/zio/http/Http.scala
jdegoes: Yes although it will have the same effect (die) unless we want to expose an error here.

zio-http/src/main/scala/zio/http/Request.scala
jdegoes: 😬 Not sure we really want `Unsafe` here. Ideally we would have "safe by construction" rather than "safe by deconstruction", to push the problem onto constructor.

zio-http/src/main/scala/zio/http/service/ServerInboundHandler.scala
jdegoes: Agreed. You can also cheat and `import zio.Unsafe.unsafe` since it's in the `zio` package. 🤫

zio-http/src/main/scala/zio/http/Http.scala
tusharmath: This will result in performance degradation. 
The older implementation was able to match path with Zero allocations.

zio-http/src/main/scala/zio/http/service/ServerFastResponseWriter.scala
```diff
@@ -9,22 +10,24 @@ import zio.http.{HExit, Response}
...
-      case HExit.Success(response) =>
-        response.attribute.encoded match {
-          case Some((oResponse, jResponse: FullHttpResponse)) if Unsafe.hasChanged(response, oResponse) =>
-            val djResponse = jResponse.retainedDuplicate()
-            Unsafe.setServerTime(time, response, djResponse)
-            ctx.writeAndFlush(djResponse, ctx.voidPromise()): Unit
-            log.debug("Fast write performed")
-            true
+  def attemptFastWrite(exit: HExit[R, Throwable, Response])(implicit ctx: Ctx): Boolean =
+    Unsafe.unsafe { implicit u =>
```
tusharmath: Will cause performance degradation. This interferes with fast path execution.

zio-http/src/main/scala/zio/http/Http.scala
vigoo: @tusharmath I think the user can opt out of this by providing an implicit `Unsafe` once (by getting and storing one initialization time because the singleton value is `private[zio]`) and then there is no extra allocations compared to the original version

vigoo: @tusharmath I think with @jdegoes's recommended change the user can opt out of this by providing an implicit `Unsafe` once (by getting and storing one initialization time because the singleton value is `private[zio]`) and then there is no extra allocations compared to the original version. (and even without that it is not per request but at `HttpApp` definition time)

zio-http/src/main/scala/zio/http/service/ServerFastResponseWriter.scala
vigoo: fixed

zio-http/src/main/scala/zio/http/Http.scala
jdegoes: @tusharmath Yes, there would have been performance degradation with `Unsafe.unsafe { ... }`, but if we reflect that in the API with (implicit unsafe) rather than trying to "escape" the unsafeness, then there will be no impact on performance, since ZIO HTTP will itself simply import the unsafe implicit into scope.

zio-http/src/main/scala/zio/http/Cookie.scala
```diff
@@ -31,7 +31,9 @@ final case class Cookie[T](name: String, content: String, target: Cookie.Type[T]
    */
   def encode(validate: Boolean)(implicit ev: CookieEncoder[T]): Either[Exception, String] =
     try {
-      Right(ev.unsafeEncode(self, validate))
+      Unsafe.unsafe { implicit u =>
```
jdegoes, line 34: I think when we are inside this library, we should always pass `Unsafe.unsafe` explicitly or import it at the top of the file. We should aim to have net zero performance impact.

zio-http/src/main/scala/zio/http/Cookie.scala
```diff
@@ -213,7 +215,9 @@ object Cookie {
    */
   def decode[S](string: String, validate: Boolean = false)(implicit ev: CookieDecoder[S]): Either[Exception, ev.Out] = {
     try {
-      Right(ev.unsafeDecode(string, validate))
+      Unsafe.unsafe { implicit u =>
```
jdegoes, line 218: Same here.

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -721,7 +725,8 @@ object Http {
   /**
    * Provides access to the request's ChannelHandlerContext
    */
-  def context: Http[Any, Nothing, Request, ChannelHandlerContext] = Http.fromFunction[Request](_.unsafeContext)
+  def context: Http[Any, Nothing, Request, ChannelHandlerContext] =
+    Http.fromFunctionZIO[Request](request => ZIO.succeedUnsafe { implicit u => request.unsafe.context })
```
jdegoes, line 729: To avoid the extra allocation, maybe just `Http.fromFunction { request => request.unsafe.context(Unsafe.unsafe) }`

tusharmath:
@vigoo Can you create a PR from `zio:unsafe` instead of `vigoo:unsafe`. 
The performance workflow isn't able to comment on this PR. 

I can observe a 20% drop in performance thru the logs.

zio-http/src/main/scala/zio/http/Http.scala
tusharmath: There is still an additional `Function` that's being allocated. Which is degrading the performance.

Earlier we used to be at around 1.01M rps now its around 0.80 rps.

vigoo:
> @vigoo Can you create a PR from `zio:unsafe` instead of `vigoo:unsafe`. The performance workflow isn't able to comment on this PR.
> 
> I can observe a 20% drop in performance thru the logs.

Yeah I will reopen it. Hopefully with @jdegoes's suggestion to just use the unsafe object internally it won't hurt the performance. Will try out after a few meetings :)

Issue #1451, opened 2022-09-12
Add Middleware that does response logging using ZIO 2 Logging


jdegoes:
**Is your feature request related to a problem? Please describe.**

Response logging is a standard feature of some web servers. It is possible to add response logging now into any ZIO HTTP server, but it requires writing custom code, particularly if the end-goal is to pipe that response logging through the ZIO 2 logging front-end, to benefit from the work done in supporting different logging backends in the ZIO Logging project.

**Describe the solution you'd like**

Middleware that does industry-standard response logging using ZIO 2 integrated logging, including request timings.

**Describe alternatives you've considered**

Writing custom code to do the same.

swoogles:
I'll look into this.
Do we have a clear spec for "industry-standard response logging" ?

jdegoes:
No but some inspiration can be taken from [httpd](https://httpd.apache.org/docs/current/mod/mod_log_config.html#page-header) or Nginx.

swoogles:
Looking at those 2, I lean towards Nginx, which uses full words:
```
$remote_addr
$remote_user
```

Httpd is more cryptic/terse with single letter options. I don't think stuff like this is user-friendly:
```
%u       | Remote user if the request was authenticated. 
%U       | The URL path requested, not including any query string.
```

jdegoes:
Oh I agree and I am not even suggesting user-configurability in these tickets. Just choosing some reasonable defaults for request/response logging.

The docs can be taken as a reference to what sort of things these servers are capable of embedding into the access logs.

swoogles:
`I am not even suggesting user-configurability`
Ah, good to know, I was starting down that rabbit hole 🙂

swoogles:
Looks like I can't assign the ticket to myself in this repo for some reason, but I'll claim this one.
You can assign it to me when convenient.

PR #1456, opened 2022-09-12
Create default Response logging middleware

swoogles:
I'm sure we'll want to modify this, but it's just a starting point for discussion around #1451

Sample logs:
```
timestamp=2022-09-12T20:28:38.340849Z level=INFO thread=#zio-fiber-102 message="404 GET /health 0ms" location=zio.http.middleware.Web.logResponseDefault file=Web.scala line=46
timestamp=2022-09-12T20:28:38.340685Z level=INFO thread=#zio-fiber-103 message="200 GET /health 1000ms" location=zio.http.middleware.Web.logResponseDefault file=Web.scala line=46
```

### DIFF
--- a/zio-http/src/main/scala/zio/http/middleware/Web.scala
+++ b/zio-http/src/main/scala/zio/http/middleware/Web.scala
@@ -37,17 +37,13 @@ private[zio] trait Web extends Cors with Csrf with Auth with HeaderModifier[Http
         } yield Patch.empty
     }
 
-  /*
-   * Questions:
-   *   What log level? Info?
-   */
-  final def logDefault: HttpMiddleware[Any, Nothing] = // TODO Confirm Nothing error.
+  final def logResponseDefault: HttpMiddleware[Any, Nothing] =
     interceptZIOPatch(req => Clock.nanoTime.map(start => (req.method, req.url, start))) {
       case (response, (method, url, start)) =>
         for {
           end <- Clock.nanoTime
           duration = (end - start) / 1000000
-          _   <- ZIO.log(s"${response.status.asJava.code()} ${method} ${url.encode} ${duration}ms")
+          _ <- ZIO.log(s"${response.status.asJava.code()} ${method} ${url.encode} ${duration}ms")
         } yield Patch.empty
     }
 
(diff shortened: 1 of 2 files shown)


zio-http/src/main/scala/zio/http/middleware/Web.scala
```diff
@@ -37,6 +37,16 @@ private[zio] trait Web extends Cors with Csrf with Auth with HeaderModifier[Http
...
 
+  final def logResponseDefault: HttpMiddleware[Any, Nothing] =
+    interceptZIOPatch(req => Clock.nanoTime.map(start => (req.method, req.url, start))) {
+      case (response, (method, url, start)) =>
+        for {
+          end <- Clock.nanoTime
+          duration = (end - start) / 1000000
+          _ <- ZIO.log(s"${response.status.asJava.code()} ${method} ${url.encode} ${duration}ms")
+        } yield Patch.empty
+    }
```
tusharmath, line 48: `debug` does this already isn't it.

jdegoes: If this is a duplicate of `debug`, what do you think about using a new name, so as not to imply that the middleware only be used for debugging??

swoogles: > `debug` does this already isn't it.

It's very similar, but it strictly uses the Console, rather than the logging machinery. In addition to the rename that John proposed, we would also want to change the `debug` implementation to this version.

jdegoes (approved):

zio-http/src/main/scala/zio/http/middleware/Web.scala
tusharmath: Agree. Let's do that.

zio-http/src/test/scala/zio/http/middleware/WebSpec.scala
```diff
@@ -47,6 +47,22 @@ object WebSpec extends ZIOSpecDefault with HttpAppTestExtensions { self =>
...
+    suite("logging")(
+      test("log status method url and time") {
+        val program = runApp(app @@ logResponseDefault) *> ZTestLogger.logOutput
+        for {
+          res <- program
+          logMessages = res.map(_.message())
+        } yield assertTrue(logMessages == Chunk("200 GET /health 1000ms"))
+      },
+      test("log 404 status method url and time") {
+        val program = runApp(Http.empty ++ Http.notFound @@ logResponseDefault) *> ZTestLogger.logOutput
```
tusharmath, line 59:
```suggestion
        val program = runApp(Http.empty @@ logResponseDefault) *> ZTestLogger.logOutput
```
We should have to compose it with a `notFound`. Logger should work automatically for not-found statuses.

jdegoes:
Closed in favor of #1467

PR #1463, opened 2022-09-13
1007 convert headers to sealed trait

adrianfilip:


zio-http/src/main/scala/zio/http/Headers.scala
```diff
@@ -1,81 +1,151 @@
...
+
+    override private[http] def encode: HttpHeaders = Headers.encode(self.toList)
+
+    def flatten: Iterable[Header] = Iterable(self)
+
+    override def iterator: Iterator[Header] =
+      Iterator.single(self)
+
+    // Unless we implement it headers.toString results in StackOverflow
+    override def toString(): String = (key, value).toString()
```
adrianfilip, line 64: Looking into if this can be avoided. If it's acceptable as a solution I am ok to leave it like this.

zio-http/src/main/scala/zio/http/Headers.scala
```diff
@@ -1,81 +1,151 @@
...
+    override def iterator: Iterator[Header] =
+      Iterator.empty
+
+  }
+
   val BasicSchemeName  = "Basic"
   val BearerSchemeName = "Bearer"
 
-  def apply(name: CharSequence, value: CharSequence): Headers = Headers(Chunk((name, value)))
+  def apply(name: CharSequence, value: CharSequence): Headers = Headers.Header(name.toString, value.toString)
```
adrianfilip, line 120: The CorsSpec tests fail without the toString because the headers from HeadersConstructors have quoted values and retain them when passed as keys without the toString.  The alternative was to apply a toString on the headers in HeaderConstructors.

jdegoes:
@adrianfilip Let us know when it's ready for final review (and conflicts fixed, PR rebased, etc.)!

adrianfilip:
@jdegoes @tusharmath It's ready for review :).

jdegoes (approved):

PR #1467, opened 2022-09-13
 ZIO logger based logger transport and request logging middleware 

### LINKED ISSUE #1450: Add Middleware that does request logging using ZIO 2 Logging

Request logging is a standard feature of Nginx, httpd, and other web servers. ZIO 2 users who have configured ZIO 2's integrated logging do not have an easy way to pipe request logging through to their configured logging backend.

A Middleware that both enriches log annotations with specified (default?) headers, but also produces industry-standard request logging details, such as HTTP method, path, IP address, etc.

**Why is it important?**

Request logging is valuable to some users; integration of such logging into ZIO 2 logging (with access to all supported logging backends, and logging annotations, etc.), is similarly valuable to some users.

### LINKED ISSUE #1451: Add Middleware that does response logging using ZIO 2 Logging

vigoo:
- Adds a `LoggerTransport` that forwards all zio-http logs to ZIO core loggers
- Defines a `Middleware` that logs request/response details directly to ZIO core, taking advantage of its annotation support.

Notes:
- Instead of directly logging through ZIO core we could extend the zio-http specfific logger interface to support annotations and then go through that
- One thing I wanted to add (and I did with akka-http multiple times in the past) but does not seem to be possible at the moment is to wrap streaming response bodies in a way to measure its completion (when the stream is fully consumed) and optionally capture and log the whole streamed response body (or optionally just its length)

Resolves #1451 and #1450

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -133,7 +133,7 @@ lazy val zioHttpLogging = (project in file("zio-http-logging"))
   )
   .settings(
     testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework"),
-    libraryDependencies ++= Seq(`zio-test`, `zio-test-sbt`),
+    libraryDependencies ++= Seq(`zio`, `zio-test`, `zio-test-sbt`),
   )
 
 lazy val zioHttpExample = (project in file("zio-http-example"))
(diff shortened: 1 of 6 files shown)


jdegoes (approved):

jdegoes:
Good to merge as soon as CI passing!

PR #1468, opened 2022-09-13
Convert some ignored SSLSpec cases to flaky. Ignore only 1.

swoogles:
Doesn't fix anything, but uses the more appropriate `flaky` annotation, rather than disabling all of cases in this Spec.
Should document the current situation a bit better through code.

### DIFF
--- a/zio-http/src/test/scala/zio/http/service/SSLSpec.scala
+++ b/zio-http/src/test/scala/zio/http/service/SSLSpec.scala
@@ -6,7 +6,7 @@ import zio.http._
 import zio.http.service.ClientSSLHandler.ClientSSLOptions
 import zio.http.service.ServerSSLHandler.{ServerSSLOptions, ctxFromCert}
 import zio.test.Assertion.equalTo
-import zio.test.TestAspect.{ignore, timeout}
+import zio.test.TestAspect.{flaky, ignore, timeout}
 import zio.test.{Gen, TestEnvironment, ZIOSpecDefault, assertZIO, check}
 import zio.{Scope, ZIO, durationInt}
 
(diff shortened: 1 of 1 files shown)


jdegoes:
Great idea! Looks like benchmarks failed, can you please take a look?

swoogles:
> Great idea! Looks like benchmarks failed, can you please take a look?

Yeah, still trying to figure out what is going wrong with this build.
That stage copies some files into the Benchmark project that eventually cause this compilation error:
```
zio-http: [error] /zio-http/src/main/scala/Main.scala:53:8: value onError is not a member of zio.http.Server[Any,Nothing]
zio-http: [error] possible cause: maybe a semicolon is missing before `value onError`?
zio-http: [error]       .onError(_ => ZIO.unit)
```
That method is now called `withError`, so it makes sense that it wouldn't compile.
But when I go to the file that's being copied, it correctly uses `withError`:
Screen Shot 2022-09-14 at 8 44 35 AM

It seems like it's caching and/or referencing an old version of the file.

swoogles:
So we're copying files from a different repo, `dream11/FrameworkBenchmarks`, and _it_ has the offending file.
However, this was recently/deliberately changed to `onError` -

https://github.com/dream11/FrameworkBenchmarks/commit/449aff1819a70bc6f0b968f106367e0cce5e1d2d

@tusharmath Can you help me understand why this was changed `onError` when the code I'm looking at indicates that it should be `withError` ?
https://github.com/zio/zio-http/blob/d9d18bb5332b3452f9e395da704e7cf0b824e077/zio-http/src/main/scala/zio/http/Server.scala#L84-L84

This is also breaking an unrelated PR, so the failure is not caused by this code.

swoogles:
Went ahead and made a tiny PR to correct that method name-
https://github.com/dream11/FrameworkBenchmarks/pull/8

jdegoes:
@swoogles Is this good to merge after conflicts fixed?

swoogles:
> @swoogles Is this good to merge after conflicts fixed?

Yes I will resolve them after my capital one meetings this morning

swoogles:
Resolved conflicts locally, but now the tests are just locking up when I run them.
Trying to discern what changed since I made this PR.

jdegoes:
@swoogles I believe this was related to the benchmark. Please re-open when you are ready for this to be merged!

PR #1472, opened 2022-09-14
Follow zio2 pattern

### LINKED ISSUE #1460: Move client and server to interfaces + implementations + layers

The server and client are objects, whose methods use the environment to obtain their dependencies. This was a ZIO 1.0 pattern. In ZIO 2, it's more idiomatic for such functionality to be moved to service traits, whose classes consume dependencies through constructors, and whose construction is itself described by a layer.

There should exist `Client` and `Server` traits, which define what it means for something to be a `Client` or `Server`, and the dependencies and resources necessary to create these should be bundled up into layers.

This will allow the client and server to consume resources, have their own dependencies, and possibly even have test implementations, while at the same time following the newer ZIO 2 module pattern, which embraces constructor-based dependency injection.

None.

devsprint:


zio-http-example/src/main/scala/example/ClientServer.scala
```diff
@@ -17,6 +17,7 @@ object ClientServer extends ZIOAppDefault {
 
   val run = {
     val clientLayers = ChannelFactory.auto ++ EventLoopGroup.auto()
-    Server.start(8080, app).provideLayer(clientLayers).exitCode
+    Server.serve(app).provide(Server.default ++ clientLayers)
```
jdegoes: In all the examples, try to let ZIO do the wiring for you: `.provide(Server.default, clientLayers)`.

zio-http-example/src/main/scala/example/HelloWorld.scala
```diff
@@ -12,6 +12,10 @@ object HelloWorld extends ZIOAppDefault {
   }
 
   // Run it like any simple app
-  override val run =
-    Server.start(8090, app)
+  override val run = Server.serve(app).provide(Server.default)
+
+//  val config = ServerConfig.Config().withPort(8090)
```
jdegoes: Dead code

zio-http-example/src/main/scala/example/HelloWorldAdvanced.scala
```diff
@@ -20,24 +17,26 @@ object HelloWorldAdvanced extends ZIOAppDefault {
     case Method.GET -> !! / "utc"    => Clock.currentDateTime.map(s => Response.text(s.toString))
   }
 
-  private val server =
-    Server.port(PORT) ++              // Setup port
-      Server.paranoidLeakDetection ++ // Paranoid leak detection (affects performance)
-      Server.app(fooBar ++ app)       // Setup the Http app
+//  private val server =
+//    Server.port(PORT) ++              // Setup port
```
jdegoes: Dead code

zio-http-example/src/main/scala/example/HelloWorldAdvanced.scala
```diff
@@ -20,24 +17,26 @@ object HelloWorldAdvanced extends ZIOAppDefault {
...
   val run = ZIOAppArgs.getArgs.flatMap { args =>
     // Configure thread count using CLI
-    val nThreads: Int = args.headOption.flatMap(x => Try(x.toInt).toOption).getOrElse(0)
+    // val nThreads: Int = args.headOption.flatMap(x => Try(x.toInt).toOption).getOrElse(0)
+
+    Server.serve(fooBar ++ app).provide(Server.default)
 
     // Create a new server
-    server.make
+    /*server.make
```
jdegoes: Dead code

zio-http-example/src/main/scala/example/HtmlTemplating.scala
```diff
@@ -46,5 +46,5 @@ object HtmlTemplating extends ZIOAppDefault {
     }
   }
 
-  def run = Server.start(8090, app)
+  def run = Server.serve(app).provide(Server.default)
```
jdegoes, line 49: Hmm, we have a problem: these apps will exit immediately.

You should probably add `Server.serveForever(app)` defined as `Server.serve(app) *> ZIO.never`.

Or maybe that should go to `serve` and we could change the existing `server` into `add` or `install`. So if you just want to serve 1 app forever, you do `Server.serve`, but if you want to install a server (e.g. from another layer which requires to expose some endpoints), then you do `Server#install`.

zio-http-example/src/main/scala/example/HttpsHelloWorld.scala
```diff
@@ -25,11 +24,12 @@ object HttpsHelloWorld extends ZIOAppDefault {
     getClass().getClassLoader().getResourceAsStream("server.key"),
   )
 
-  private val server =
-    Server.port(8090) ++ Server.app(app) ++ Server.ssl(
-      ServerSSLOptions(sslctx, SSLHttpBehaviour.Accept),
-    )
+//  private val server =
```
jdegoes: Dead code

zio-http-example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -42,17 +41,18 @@ object Main extends ZIOAppDefault {
...
-      .withLeakDetection(LeakDetectionLevel.DISABLED)
-      .withConsolidateFlush(true)
-      .withFlowControl(false)
-      .withObjectAggregator(-1)
+    Server.serve(app).provide(Server.default)
+//    Server
+//      .app(app)
+//      .withPort(8080)
+//      .withError(_ => ZIO.unit)
+//      .withLeakDetection(LeakDetectionLevel.DISABLED)
```
jdegoes: Dead code

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -1,300 +1,62 @@
...
-  def withUnsafeChannelPipeline(unsafePipeline: ChannelPipeline => Unit): Server[R, E] =
-    Concat(self, UnsafeChannelPipeline(unsafePipeline))
-
-  /**
-   * Provides unsafe access to netty's ServerBootstrap. Modifying server
-   * bootstrap is generally not advised unless you know what you are doing.
-   */
-  def withUnsafeServerBootstrap(unsafeServerbootstrap: ServerBootstrap => Unit): Server[R, E] =
-    Concat(self, UnsafeServerBootstrap(unsafeServerbootstrap))
+  def port: UIO[Int]
```
jdegoes: If the port never changes for a running server, this could be `def port: Int` instead.

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -1,300 +1,62 @@
...
-    case object PARANOID extends LeakDetectionLevel
+  val test = ServerConfig.testConfig >>> live
+
+  private final case class ServerLive(
+                                       appRef: java.util.concurrent.atomic.AtomicReference[HttpApp[Any, Throwable]],
+                                       bindPort: Int
+                                     ) extends Server {
+    override def serve[R](httpApp: HttpApp[R, Throwable]): URIO[R, Unit] =
+      ZIO.environment[R].map { env =>
+        val newApp = httpApp.provideEnvironment(env)
```
jdegoes: You can add a fast-path optimization here by doing something like:

```suggestion
        val newApp = if (env == ZEnvironment.empty) httpApp.asInstanceOf[HttpApp[Any, Throwable]] else httpApp.provideEnvironment(env)
```

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -1,300 +1,62 @@
...
+        var loop = true
+        while (loop) {
+          val oldApp = appRef.get()
+          if (appRef.compareAndSet(oldApp, newApp ++ oldApp)) loop = false
+        }
+        ()
+      }
+
+    override def port: UIO[Int] = ZIO.succeed(bindPort)
```
jdegoes: Looks constant to me for the life of the server.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+import io.netty.util.ResourceLeakDetector
+import zio.http.service.{EventLoopGroup, ServerChannelFactory}
+import zio.{ULayer, ZIO, ZLayer}
+import zio.http.service.ServerSSLHandler.ServerSSLOptions
+
+import java.net.{InetAddress, InetSocketAddress}
+
+object ServerConfig {
+
+  val default: ZLayer[Any, Nothing, Config with EventLoopGroup with ServerChannelFactory] = {
```
jdegoes: I think this should simply be: `ZLayer[Any, Nothing, ServerConfig]`. 

Then we push the creation of `EventLoopGroup` and `ServerChannelFactory`, based on the `ChannelType`, into `Server` layer, not here. This will be clean and uniform and consistent with ZIO 2 idioms.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+    case object DISABLED extends LeakDetectionLevel
+
+    case object SIMPLE extends LeakDetectionLevel
+
+    case object ADVANCED extends LeakDetectionLevel
+
+    case object PARANOID extends LeakDetectionLevel
+  }
+
+  final case class Config(
```
jdegoes: Pull this out to top-level and call it `ServerConfig`.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+                           leakDetectionLevel: LeakDetectionLevel = LeakDetectionLevel.SIMPLE,
+                           error: Option[Throwable => ZIO[Any, Nothing, Unit]] = None,
+                           sslOption: ServerSSLOptions = null,
+                           address: InetSocketAddress = new InetSocketAddress(8080),
+                           acceptContinue: Boolean = false,
+                           keepAlive: Boolean = true,
+                           consolidateFlush: Boolean = false,
+                           flowControl: Boolean = true,
+                           channelInitializer: ChannelPipeline => Unit = null,
```
jdegoes: This is a function and not data, it is not possible to read this from configuration. It's also highly Netty-specific. We should think if it can be replaced by a sealed trait for common options.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+    case object SIMPLE extends LeakDetectionLevel
+
+    case object ADVANCED extends LeakDetectionLevel
+
+    case object PARANOID extends LeakDetectionLevel
+  }
+
+  final case class Config(
+                           leakDetectionLevel: LeakDetectionLevel = LeakDetectionLevel.SIMPLE,
+                           error: Option[Throwable => ZIO[Any, Nothing, Unit]] = None,
```
jdegoes: I think we should move this to `Server`, something like, `Server#onError`. Because it's a function, not data, and unlike the rest of the config, it cannot be read from external configuration.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+                           sslOption: ServerSSLOptions = null,
+                           address: InetSocketAddress = new InetSocketAddress(8080),
+                           acceptContinue: Boolean = false,
+                           keepAlive: Boolean = true,
+                           consolidateFlush: Boolean = false,
+                           flowControl: Boolean = true,
+                           channelInitializer: ChannelPipeline => Unit = null,
+                           requestDecompression: (Boolean, Boolean) = (false, false),
```
jdegoes: Holy boolean, Batman! 

Maybe we could have sealed trait / case objects for these?

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+                           objectAggregator: Int = 1024 * 100,
+                           serverBootstrapInitializer: ServerBootstrap => Unit = null,
```
jdegoes: Oops, another function. I think we need to move this out and into wire-up. If the server bootstrap needs to be customized, we can put it into a layer, and then expose a more customizable live `Server` layer which requires you to feed it a (customized) server bootstrap.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+    def withObjectAggregator(maxRequestSize: Int = 1024 * 100): Config =
+      self.copy(objectAggregator = maxRequestSize)
+
+    /**
+     * Creates a new server listening on the provided port.
+     */
+    def withPort(port: Int): Config = self.copy(address = new InetSocketAddress(port))
+
+    /**
+     * Creates a new server with netty's HttpContentDecompressor to decompress
```
jdegoes: These Scaladoc can be changed to reflect the fact that none of these methods create a server, they just modify the configuration.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+     * Http requests (@see <a href =
+     * "https://netty.io/4.1/api/io/netty/handler/codec/http/HttpContentDecompressor.html">HttpContentDecompressor</a>).
+     */
+    def withRequestDecompression(enabled: Boolean, strict: Boolean): Config =
```
jdegoes: We are using the `withXYZ` convention, which I am fine with, but we could also consider the builder convention, e.g. `config.port(23).ssl(options)....`. A bit more concise and "modern" but maybe less familiar to Java devs.

zio-http/src/test/scala/zio/http/ContentTypeSpec.scala
```diff
@@ -42,11 +42,11 @@ object ContentTypeSpec extends HttpRunnableSpec {
   )
 
   private val env =
-    EventLoopGroup.nio() ++ ChannelFactory.nio ++ ServerChannelFactory.nio ++ DynamicServer.live ++ Scope.default
+    DynamicServer.live ++ Server.test ++ ChannelFactory.nio ++ EventLoopGroup.nio(0)
```
jdegoes: Don't wire up manually, instead, use the `.provide()` macro below.

zio-http/src/test/scala/zio/http/service/StaticServerSpec.scala
```diff
@@ -68,29 +68,29 @@ object StaticServerSpec extends HttpRunnableSpec {
...
-      }
-    },
-  )
+//  def serverStartSpec = suite("ServerStartSpec")(
+//    test("desired port") {
+//      val port = 8088
+//      ZIO.scoped {
+//        (Server.port(port) ++ Server.app(Http.empty)).make.flatMap { start =>
+//          assertZIO(ZIO.attempt(start.port))(equalTo(port))
+//        }
```
jdegoes: Disabled test

zio-http/src/test/scala/zio/http/service/WebSocketSpec.scala
```diff
@@ -13,7 +13,7 @@ import zio.{Promise, Ref, UIO, ZIO, durationInt}
 object WebSocketSpec extends HttpRunnableSpec {
 
   private val env =
-    EventLoopGroup.nio() ++ ServerChannelFactory.nio ++ DynamicServer.live ++ ChannelFactory.nio
+    DynamicServer.live ++ Server.test ++ ChannelFactory.nio ++ EventLoopGroup.nio(0)
```
jdegoes: Let ZIO wire up.

jdegoes (changes requested):
Overall, looking really good! Made a few comments to clean things up further and deal with some edge cases.

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: Done

zio-http-example/src/main/scala/example/ClientServer.scala
devsprint: Done.

zio-http-example/src/main/scala/example/HtmlTemplating.scala
devsprint: No, it will not exit right away because the `serve` implementation is:
`  def serve[R](httpApp: HttpApp[R, Throwable]): URIO[R with Server, Unit] =
    ZIO.serviceWithZIO[Server](_.serve(httpApp)) *> ZIO.never`

jdegoes: Hmmm, we need two different methods because both are useful

jdegoes: In fact, let's have `install` be the only method on Server, then `serve` can be a convenience method on `Server` companion object that installs and waits forever (via never)

zio-http/src/main/scala/zio/http/Server.scala
devsprint: if the configuration is having port set to 0, then it means that netty will choose first free port on the machine. So it is dynamic, that's I added it in this way. However, it become problematic to obtain the binding port with the latest changes.

zio-http-example/src/main/scala/example/HtmlTemplating.scala
devsprint: done

zio-http/src/main/scala/zio/http/Server.scala
devsprint: Done

zio-http-example/src/main/scala/example/HelloWorldAdvanced.scala
devsprint: Done

zio-http-example/src/main/scala/example/PlainTextBenchmarkServer.scala
devsprint: Done

zio-http/src/main/scala/zio/http/Server.scala
devsprint: It is updated when the config has port set to 0. The bind port will be determined at runtime.

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: Done

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: removed for now, added a TODO in the place where it is used to find a different way to provided it. As far as I'm aware this feature is not used currently

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: Removed for now and added a TODO in the place where it is used. As far as I'm aware is not used.

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: Done

zio-http-example/src/main/scala/example/HelloWorldAdvanced.scala
devsprint: Done

zio-http/src/test/scala/zio/http/service/WebSocketSpec.scala
devsprint: Done

zio-http/src/test/scala/zio/http/service/StaticServerSpec.scala
devsprint: just temporary, to move them out in their own suite as requires different configs to run

zio-http/src/test/scala/zio/http/ContentTypeSpec.scala
devsprint: Done

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: Done

zio-http/src/test/scala/zio/http/service/StaticServerSpec.scala
devsprint: re-enabled

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -1,300 +1,86 @@
...
-
-  def bind(port: Int): UServer = Server.Address(new InetSocketAddress(port))
-
-  def bind(hostname: String, port: Int): UServer = Server.Address(new InetSocketAddress(hostname, port))
-
-  def bind(inetAddress: InetAddress, port: Int): UServer = Server.Address(new InetSocketAddress(inetAddress, port))
-
-  def bind(inetSocketAddress: InetSocketAddress): UServer = Server.Address(inetSocketAddress)
+object Server {
+  def serve[R](httpApp: HttpApp[R, Throwable]): URIO[R with Server, Unit] =
```
jdegoes: The success type should be `Nothing` to indicate this method never returns.
```suggestion
  def serve[R](httpApp: HttpApp[R, Throwable]): URIO[R with Server, Nothing] =
```

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -1,300 +1,86 @@
...
-  def withUnsafeChannelPipeline(unsafePipeline: ChannelPipeline => Unit): Server[R, E] =
-    Concat(self, UnsafeChannelPipeline(unsafePipeline))
-
-  /**
-   * Provides unsafe access to netty's ServerBootstrap. Modifying server
-   * bootstrap is generally not advised unless you know what you are doing.
-   */
-  def withUnsafeServerBootstrap(unsafeServerbootstrap: ServerBootstrap => Unit): Server[R, E] =
-    Concat(self, UnsafeServerBootstrap(unsafeServerbootstrap))
+  def port[R]: URIO[R, Int]
```
jdegoes:
```suggestion
  def port: Int
```

zio-http/src/main/scala/zio/http/ServerConfigLayer.scala
```diff
@@ -0,0 +1,15 @@
+package zio.http
+
+import zio.ZLayer
+
+object ServerConfigLayer {
```
jdegoes: I would move all these into the `ServerConfig` companion object.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -4,7 +4,7 @@ import zio.http.service.ServerSSLHandler.ServerSSLOptions
 
 import java.net.{InetAddress, InetSocketAddress}
 
-final case class ServerConfig private (
```
devsprint: fix to avoid scala 2.12.x bug: https://github.com/scala/bug/issues/7707

zio-http/src/main/scala/zio/http/Server.scala
devsprint: done

zio-http/src/main/scala/zio/http/Server.scala
devsprint: Done

zio-http/src/main/scala/zio/http/ServerConfigLayer.scala
devsprint: Done

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: @jdegoes I have pushed the change but implemented differently from your suggestion. OnError callback is more like an input parameter for the Server layer.

zio-http/src/main/scala/zio/http/ClientConfig.scala
```diff
@@ -0,0 +1,21 @@
+package zio.http
+
+import zio.http.service.ClientSSLHandler.ClientSSLOptions
+import zio.http.socket.SocketApp
+
+case class ClientConfig(
+  socketApp: Option[SocketApp[Any]] = None,
```
jdegoes, line 8: Let's save it for further PRs, but SocketApp is definitely not configuration for an HTTP client.

zio-http/src/main/scala/zio/http/LeakDetectionLevel.scala
```diff
@@ -0,0 +1,23 @@
+package zio.http
+
+import io.netty.util.ResourceLeakDetector
+
+sealed trait LeakDetectionLevel {
```
jdegoes: If "small" config data types like this appear _only_ in `ServerConfig`, it's acceptable (even desirable) to move them into the `ServerConfig` companion object, to avoid cluttering the top-level namespace with irrelevant classes.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -0,0 +1,182 @@
...
+
+    case object ADVANCED extends LeakDetectionLevel
+
+    case object PARANOID extends LeakDetectionLevel
+  }
+
+  final case class Config(
+                           leakDetectionLevel: LeakDetectionLevel = LeakDetectionLevel.SIMPLE,
+                           error: Option[Throwable => ZIO[Any, Nothing, Unit]] = None,
+                           sslOption: ServerSSLOptions = null,
```
tobia80: Could this be an Option[SslConfig] that defaults to None?

jdegoes: Good catch!

zio-http/src/main/scala/zio/http/ClientConfig.scala
devsprint: for now, I left it there.

zio-http/src/main/scala/zio/http/ServerConfig.scala
devsprint: fixed

zio-http/src/main/scala/zio/http/LeakDetectionLevel.scala
devsprint: done

jdegoes (approved):
Given the size and scope of this PR, I think we can merge this now and address remaining items as followup tickets / PRs.

jdegoes (approved):

PR #1475, opened 2022-09-14
1453 abstract away netty

scottweaver:
Proof of Concept were all the Netty bits have been abstracted away behind the concept of a implementation-agnostic 'Driver'.


project/ScalaSettings.scala
```diff
@@ -2,8 +2,8 @@ trait ScalaSettings {
   private val baseSettings = Seq(
     "-language:postfixOps",      // Added by @tusharmath
     "-deprecation",              // Emit warning and location for usages of deprecated APIs.
-    "-encoding",
-    "utf-8",                     // Specify character encoding used by source files.
+    // "-encoding",
```
thehonesttech, line 5: Should we leave this commented or can we remove them?

zio-http/src/main/scala/zio/http/Driver.scala
```diff
@@ -0,0 +1,8 @@
+package zio.http
+
+import zio._
+// import java.util.concurrent.atomic.AtomicReference
```
thehonesttech, line 4: can we remove this reference?

PR #1482, opened 2022-09-15
Improve performance of ZIO HTTP by using ZIO.refailCause in more places

### LINKED ISSUE #1478: Improve performance of ZIO HTTP by using ZIO.refailCause in more places

ZIO 2 had a massive update to the way async stack traces are generated. Now, they are virtually cost-free, except when created, which is not in the happy path. However, the creation of an async stack trace is now much more costly than before.

For this reason, it is important to pay attention to the generation of async stack traces. In particular, **any use of ZIO.fail or ZIO.failCause, directly or indirectly, will now pay the FULL COST of async stack trace generation**.

These methods work by (1) first generating an async stack trace, (2) attaching the async stack trace to the failure, and (3) calling `ZIO.refailCause` to actually fail.

Every usage of `ZIO.fail` / `ZIO.failCause` in ZIO HTTP should be **audited* to ensure it is **necessary**. When refailing with an existing cause, the method `ZIO.refailCause` should be utilized instead, because the cause already has an attached stack trace and does not need a new one. In addition, any usage of the error channel as short-circuiting or optionality should be replaced by usage of `ZIO.refailCause`. **For example, the code uses `ZIO.fail(None)` in quite a few places, which should all be changed to `ZIO.refailCause(Cause.fail(None))` (factored out into a constant to avoid re-allocating this structure endlessly).** This is because such failures do not need a stack trace, they serve only to switch over to the next route handler in the event the previous one does not match against the route.

N/A

tobia80:
Replaced ZIO.fail(None) with ZIO.refailCause (stacktrace is not needed) and replaced in some places in order to improve performance

### DIFF
--- a/zio-http/src/main/scala/zio/http/package.scala
+++ b/zio-http/src/main/scala/zio/http/package.scala
@@ -21,5 +21,5 @@ package object http extends PathSyntax with RequestSyntax with RouteDecoderModul
   object HeaderNames  extends headers.HeaderNames
   object HeaderValues extends headers.HeaderValues
 
-  val failNoStacktrace: ZIO[Any, None.type, Nothing] = ZIO.refailCause(Cause.fail(None))
+  private val failNoStacktrace: ZIO[Any, None.type, Nothing] = ZIO.refailCause(Cause.fail(None))
 }
--- a/zio-http/src/main/scala/zio/http/package.scala
+++ b/zio-http/src/main/scala/zio/http/package.scala
@@ -21,5 +21,5 @@ package object http extends PathSyntax with RequestSyntax with RouteDecoderModul
   object HeaderNames  extends headers.HeaderNames
   object HeaderValues extends headers.HeaderValues
 
-  private val failNoStacktrace: ZIO[Any, None.type, Nothing] = ZIO.refailCause(Cause.fail(None))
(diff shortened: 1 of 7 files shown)


zio-http/src/main/scala/zio/http/package.scala
```diff
@@ -20,4 +20,6 @@ package object http extends PathSyntax with RequestSyntax with RouteDecoderModul
 
   object HeaderNames  extends headers.HeaderNames
   object HeaderValues extends headers.HeaderValues
+
+  val failNoStacktrace: ZIO[Any, None.type, Nothing] = ZIO.refailCause(Cause.fail(None))
```
jdegoes: I'd make this private to the http package to avoid cluttering the user namespace with implementation details.
```suggestion
  private[http] val failNoStacktrace: ZIO[Any, None.type, Nothing] = ZIO.refailCause(Cause.fail(None))
```

adamgfraser:
```suggestion
  private val failNoStacktrace: ZIO[Any, None.type, Nothing] = ZIO.refailCause(Cause.fail(None))
```

adamgfraser (approved):

PR #1485, opened 2022-09-15
Abstract Away Netty

scottweaver:
* 'zio.http.netty' package created as the new home of all Netty-related implementations and artifacts.
* 'ChannelFuture' renamed 'NettyFutureExecutor' to better describe its purpose.
* Moved all Netty-specific implementation out of 'zio.http.Server' and into the new 'NettyDriver'
* Netty 'ChannelInitializer' and 'RequestHandler' are no longer hard coded and now can be provided using layers to the 'NettyDriver'
* Eliminated the somewhat confusing trend of using Netty names for zio http things.
* Eliminated a whole host of type aliases to make things a bit more clear.
* 'ServerChannelFactory' and 'ChannelFactory' consolidated to a single 'ChannelFactories' object.  Code has been streamlined as well.
  * It provides convenience layers for materializing both Server-specific and Client-specific into the environment by way of a 'ChannelType.Config'.
* 'EventLoopGroups' replaces 'EventLoopGroup' and provides a convenience layer similar to the ones mentioned above for ChannelFactory.
* Reduced API foot print by privatizing non-essential methods.
* Converted implicit arguments to explicit ones methods where they provided no real benefit by being implicit.  In many cases, they we being explicitly provided anyway.
* NettyRuntime replaces HttpRuntime.  'Strategy' removed in favor just implementing the abstract 'runtime' method on 'NettyRuntime'
  * 'sticky' renamed to 'usingSharedThreadPool' as it better reflects the semantics of the runtime.
  * 'default' renamed to 'usingDedicatedTrheadPool' as it better reflects the semantics of the runtime.
* 'Channel.Config' added so that there is no need for 'fromConfig' methods 'ChannelFactories' for both 'ClientConfig' and 'ServerConfig'.
* Many of the "unsafe" operations on 'ChannelHandlerContext' spread out all over the place now reside the 'ChannelHandlerContextOps' extension class.
* All client-specific Netty bits have made their way into 'zio.http.netty.client'.
* All server-specific Netty bits have made their way into 'zio.http.netty.server'.
* The 'R' type argument has been removed from all Netty internal type signatures as it served no real purpose.


zio-http/src/main/scala/zio/http/Client.scala
```diff
@@ -216,36 +218,19 @@ object Client {
...
-
-  private def channelFactory(config: ClientConfig): UIO[ChannelFactory] = {
-    config.channelType match {
-      case ChannelType.NIO    => ChannelFactory.Live.nio
-      case ChannelType.EPOLL  => ChannelFactory.Live.epoll
-      case ChannelType.KQUEUE => ChannelFactory.Live.kQueue
-      case ChannelType.URING  => ChannelFactory.Live.uring
-      case ChannelType.AUTO   => ChannelFactory.Live.auto
+  val live
+    : ZLayer[ClientConfig with Scope with ChannelFactory with EventLoopGroup with NettyRuntime, Throwable, Client] =
```
vigoo: I see that it was like this before as well, but that `Scope` as input does not seem necessary.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -75,7 +72,7 @@ final case class Response private (
   def withServerTime: Response = self.copy(attribute = self.attribute.withServerTime)
 
   private[zio] def close: Task[Unit] = self.attribute.channel match {
-    case Some(channel) => ChannelFuture.unit(channel.close())
+    case Some(channel) => NettyFutureExecutor.unit(channel.close())
```
vigoo: I think it would be nice to rename this `unit` too, it was very confusing for me first

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -34,45 +31,21 @@ object Server {
 
   val default = ServerConfig.live >>> live
 
-  val live: ZLayer[ServerConfig, Throwable, Server] = ZLayer.scoped {
+  val live = NettyDriver.default >>> base
+
+  val base: ZLayer[
+    Driver with AtomicReference[HttpApp[Any, Throwable]] with AtomicReference[Option[ErrorCallback]],
```
vigoo: This looks a bit ugly _and_ does not compile with Scala 3.2.0 (it somehow confuses the two `AtomicReference`s in `ZIO.service`. So I would wrap them in one or two case classes.

zio-http/src/main/scala/zio/http/netty/EventLoopGroups.scala
```diff
@@ -0,0 +1,67 @@
...
+    make(ZIO.succeed(new EpollEventLoopGroup(nThreads, executor)))
+
+  def uring(nThread: Int): ZIO[Scope, Nothing, EventLoopGroup] =
+    make(ZIO.succeed(new IOUringEventLoopGroup(nThread)))
+
+  def uring(nThread: Int, executor: Executor): ZIO[Scope, Nothing, EventLoopGroup] =
+    make(ZIO.succeed(new IOUringEventLoopGroup(nThread, executor)))
+
+  def auto(nThreads: Int): ZIO[Scope, Nothing, EventLoopGroup] =
+    if (Epoll.isAvailable)
```
vigoo: I have seen this same logic in another place as well. I think it would be better to have this decision logic in the configuration layer (if it is set to `auto`) and in all these places just use a `ChannelTypeWithoutAuto` (with a better name) type. 

Maybe this is out of the scope of this PR though

zio-http/src/main/scala/zio/http/netty/NettyRuntime.scala
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    def removeListener(close: GenericFutureListener[Future[_ >: Void]]): Unit = {
+      if (close != null)
+        ctx.channel().closeFuture().removeListener(close): Unit
+    }
+
+    // Close the connection if the program fails
+    // When connection closes, interrupt the program
+    // Unsafe.unsafe { implicit unsafe =>
```
vigoo: This commented out unsafe can be removed

zio-http/src/main/scala/zio/http/netty/client/ClientInboundHandler.scala
```diff
@@ -29,7 +31,8 @@ final class ClientInboundHandler[R](
     msg.touch("handlers.ClientInboundHandler-channelRead0")
     // NOTE: The promise is made uninterruptible to be able to complete the promise in a error situation.
     // It allows to avoid loosing the message from pipeline in case the channel pipeline is closed due to an error.
-    zExec.runUninterruptible(promise.succeed(Response.unsafe.fromJResponse(ctx, msg)))(ctx, unsafeClass)
+    zExec.run(ctx)(promise.succeed(Response.unsafe.fromJResponse(ctx, msg)))(unsafeClass)
```
vigoo: You should keep this `runUninterruptible`

zio-http/src/main/scala/zio/http/netty/client/ClientInboundHandler.scala
```diff
@@ -29,7 +31,8 @@ final class ClientInboundHandler[R](
...
+    zExec.run(ctx)(ZIO.unit)
```
vigoo: what's this? :)

zio-http/src/main/scala/zio/http/netty/package.scala
```diff
@@ -0,0 +1,160 @@
...
+
+import scala.annotation.tailrec
+
+package object netty {
+  import server._
+
+  private[zio] val AutoReleaseRequest = false
+
+  object Names {
+    private[zio] val ServerCodecHandler             = "SERVER_CODEC"
```
vigoo: Can we make the `Names` `private[zio]` and drop them from the vals?

zio-http/src/main/scala/zio/http/netty/server/NettyDriver.scala
```diff
@@ -0,0 +1,72 @@
...
+        serverConfig = sc,
+      )
+      d
+    }
+
+  type Provides = Driver
+    with AtomicReference[HttpApp[Any, Throwable]]
+    with AtomicReference[Option[Server.ErrorCallback]]
+
+  val default = ZLayer.scopedEnvironment {
```
vigoo: Would be easier to read if these layer definitions had their types written out

zio-http/src/test/scala/zio/http/service/RequestStreamingServerSpec.scala
```diff
@@ -60,6 +60,6 @@ object RequestStreamingServerSpec extends HttpRunnableSpec {
       Client.default,
       Scope.default,
     ) @@
-      timeout(10 seconds) @@ sequential
+      timeout(10 seconds) @@ sequential @@ shrinks(0)
```
vigoo: did you want to add this?

zio-http/src/main/scala/zio/http/Server.scala
vigoo: Now that I read the rest of the PR: I think it would be best to expose these atomic references through `Driver` as they are created in `NettyDriver`'s layer anyway.

vigoo (changes requested):

jdegoes:
@scottweaver Thanks so much for your work on this! I will let @vigoo take the lead shepherding this through but will do one quick review when everything is ready. 🙏

zio-http/src/main/scala/zio/http/netty/EventLoopGroups.scala
```diff
@@ -0,0 +1,67 @@
+package zio.http
+package netty
```
jdegoes: This pattern makes it VERY confusing to see where things are coming from, because it implicitly imports **ALL OF** `zio.http` into scope.

Please always use:

```scala
package zio.http.netty
```

combined with explicit imports.

🙏

zio-http/src/test/scala/zio/http/service/RequestStreamingServerSpec.scala
scottweaver: Ooops, I had it there when I was trying to debug issues.  The shinkage was just confusing the issue, I will remove it.

zio-http/src/main/scala/zio/http/Server.scala
scottweaver: I may be missing something here, but I am not sure how to get that to work since `ServerLive` needs both of those AtomicReferences as well.

zio-http/src/main/scala/zio/http/netty/EventLoopGroups.scala
scottweaver: I removed the `auto` members from `EventLoopGroups` and `ChannelFactories` and moved the logic to `fromConfig`.

However, not sure what was being asked for in the second part of the comment:
> and in all these places just use a ChannelTypeWithoutAuto (with a better name) type.

zio-http/src/main/scala/zio/http/netty/client/ClientInboundHandler.scala
scottweaver: have no idea :)

zio-http/src/main/scala/zio/http/Server.scala
scottweaver: I think I am pretty close to what you are asking for, lemme know.

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: Done,  Renamed to `executed`

zio-http/src/main/scala/zio/http/netty/EventLoopGroups.scala
vigoo: Yeah this is not exactly what I had in mind. What I was thinking to reduce that decision completely from the two places and move it to the `Config` reader part. So `Config` would always have a concrete resolved channel type (which cannot be `Auto`) and the two "users" can just take that decision and don't have to replicate that fallback logic themselves.

zio-http/src/main/scala/zio/http/Server.scala
vigoo: Maybe I'm missing something but the new version seems a bit unnecessarily complicated to me - can't we just add those atomic references (or some abstraction over them) to `Driver` itself? Why do we have to expose `Driver & Driver.Context` in one layer and then require `Driver & Driver.Context` in another? 

If there is a good reason for the separation I'm fine with it but its not obvious to me.

vigoo (approved):
Two of my original comments is not fully resolved. I'm approving it anyway because it is already in a very good shape but please consider my new comments.

@jdegoes you can do a quick review as well when you have time

zio-http/src/main/scala/zio/http/Driver.scala
```diff
@@ -0,0 +1,32 @@
+package zio.http
+
+import zio._
+
+import java.util.concurrent.atomic.AtomicReference
+
+trait Driver {
```
jdegoes, line 5: I agree with @vigoo that rather than expose these atomic references in this callback-style API, we should refactor a bit.

Fundamentally, Driver does not need to expose the atomic references, only expose the capabilities. e.g.:

```scala
def install(httpApp: HttpApp[Any, Throwable])(implicit unsafe: Unsafe): Unit
def setErrorCallback(errorCallback: Server.ErrorCallback)(implicit unsafe: Unsafe): Unit
```

where the first one will add it to the atomic reference inside the Netty implementation, and the second one will do whatever it does.

AtomicReference feels like an implementation detail and we don't need to know about it.

This will let you delete `Context` entirely and stop threading these through layers, etc.

jdegoes (changes requested):
Looks fantastic! One suggested refactoring and then good to merge (due to size and scope). We can do followup issues in subsequent PRs.

zio-http/src/main/scala/zio/http/Driver.scala
scottweaver: OK, I _think_ this more in the spirit of what y'all are asking for, https://github.com/zio/zio-http/pull/1485/commits/9951ead005496a8b8ab7ca167a725187b0bfb07d.

Thoughts/concerns?

zio-http/src/main/scala/zio/http/Driver.scala
```diff
@@ -2,31 +2,10 @@ package zio.http
...
-    def onErrorCallbackRef[A](f: AtomicReference[Option[Server.ErrorCallback]] => A): A = f(errorCallbackRef)
-
-  }
 
-  object Context {
+  def setErrorCallback(newCallback: Option[Server.ErrorCallback]): UIO[Unit]
 
-    private[zio] def empty = Context(new AtomicReference(Http.empty), new AtomicReference(Option.empty))
-  }
+  def setApp(newApp: HttpApp[Any, Throwable]): UIO[Unit]
```
jdegoes: `setApp` implies that there is a single app that should be set one time.

Instead, the feature needs to be: `addApp`, which adds the app together with any existing app. These are the semantics required by `Server#install`.

scottweaver: Done.

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -32,51 +32,35 @@ object Server {
...
     Throwable,
     Server,
   ] = ZLayer.scoped {
     for {
-      driver    <- ZIO.service[Driver]
-      driverCtx <- ZIO.service[Driver.Context]
-      port      <- driver.start()
-    } yield ServerLive(driverCtx, port)
+      driver <- ZIO.service[Driver]
+      port   <- driver.start()
```
jdegoes: `driver.start` should not accept any parentheses, since it is a pure value.

```scala
port <- driver.start
```

jdegoes (approved):

jdegoes (approved):

PR #1488, opened 2022-09-16
Refactor Client Interface

### LINKED ISSUE #1481: Redesign end-user interface of Client without changing implementation

`Client` features a non-compositional design that makes re-use entirely a user-side concern. In order to facilitate reuse within the library, without placing the burden on users, and also in order to pave the way for future optimizations, which may require some upfront work (e.g. ahead-of-time compiling headers or generating codecs), Client can be redesigned with a more functional take.

Currently the main interface of client is (details abbreviated):

```scala
trait Client {
  def request(
    url: String,
    method: Method = Method.GET,
    content: Body = Body.empty,
    ssl: ClientSSLOptions = ClientSSLOptions.DefaultSSL,
  ): ZIO[Any, Throwable, Response] =
```

These parameters can be divided into:

1. Those that will likely differ on every request (body, path suffix, method).
2. Those that will likely be somewhat or almost the same on every request (headers, ssl, path prefix).

They can also be sorted by:

1. Those that can be combined (path, headers).
2. Those that cannot be combined (method, body, ssl).

These sortings suggest a way to extract out a functional domain equipped with unary operators that can add information:

```scala
trait Client {
  def request(method: Method, pathSuffix: String, body: Body): ZIO[Any, Throwable, Response]

  def path(segment: String): Client

  def query(key: String, value: String): Client

  def header(key: String, value: String): Client 

  def header(header: Header): Client

  def ssl(ssl: ClientSSLOptions): Client
}
```

This allows incremental specification of some things which can be reused. For example:

```scala
val github = client.path("http://api.github.com").header(acceptJson).header(githubToken)

github.request(...)
```

The next step is to add specializations by method time to eliminate some boilerplate:

```scala
trait Client {
  def get(pathSuffix: String)(body: Body): ZIO[Any, Throwable, Response] = request(Method.GET, pathSuffix, body)
  def put...
}
```

Which simplifies usage considerably:

```scala
github.get("/users/")
```

Further, adding overloads can ensure users do not have to perform conversions between equivalent types themselves:

```scala
trait Client {
  def uri(uri: URI): Client
  def url(url: URL): Client
  ...
}
```

Adding polymorphism allows viewing a client as a function whose inputs and outputs can be modified:

```scala
trait Client[-In, +Out] {
  def request(method: Method, pathSuffix: String, body: In): ZIO[Any, Throwable, Out]

  def map[Out2](f: Out => Out2): Client = ???
  def mapZIO[Out2](f: Out => ZIO[Any, Throwable, Out2]): Client = ???
  def contramap[In2](f: In2 => In): Client = ???
  def contramapZIO...
}
```

Adding errors can facilitate edge-level retries for flaky services:

```scala
trait Client[-In, +Err, +Out] {
  def retry[Err1 >: Err](policy: Schedule[Err1, Any]): Client[In, Err1, Out] = ???
  def refineOrDie[Err2](pf: PartialFunction[Err, Err2]): Client[In, Err2, Out] = ???
  def dieOn[Err1 >: Err](f: Err1 => Boolean): Client[In, Err1, Out] = ???
}
```

Which could be utilized as follows:

```scala
github.retry(Schedule.recurs(100).whileInput(_.isInstanceOf[ServerUnavailableException])).dieOn{
  e => e.isInstanceOf[ServerUnavailableException)
}
```

The final step would be generalizing to support contextual effects in operations like `mapZIO` and `contramapZIO`, which would require adding an `R` type parameter:

```scala
trait Client[-Env, -In, +Err, +Out] {
  def mapZIO[Env1 <: Env, Err1 >: Env, Out2](f: Out => ZIO[Env1, Err1, Out2]): Client[Env1, In, Err1, Out2] = ???
  ...
}
```

At this point, the polymorphism probably demands we call this `ZClient`:

```scala
trait ZClient[-Env, -In, +Err, +Out] {
  ...
}
```

Which is made more usable with type aliases:

```scala
type Client = ZClient[Any, Body, Throwable, Response]
```

Then the default client which one accesses from a layer would simply have type `Client`.

None.

adamgfraser:
Resolves #1481.

Adds unary operators to allow configuring client. Introduction of additional convenience methods and more polymorphic version is still to be done.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Client.scala
+++ b/zio-http/src/main/scala/zio/http/Client.scala
@@ -19,8 +19,19 @@ import zio.http.socket.SocketApp
 
 import java.net.InetSocketAddress
 
-trait Client {
-  def request(request: Request): ZIO[Any, Throwable, Response]
+trait Client { self =>
+
+  def headers: Headers
+
+  def hostOption: Option[String]
+
+  def pathPrefix: Path
+
+  def portOption: Option[Int]
+
+  def queries: Map[String, Chunk[String]]
+
+  def sslOption: Option[ClientSSLOptions]
 
   def socket[R](
     url: String,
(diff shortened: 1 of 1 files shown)


zio-http/src/main/scala/zio/http/Client.scala
```diff
@@ -29,21 +40,152 @@ trait Client {
...
+      queries: Map[String, Chunk[String]],
+      sslOption: Option[ClientSSLOptions],
+      version: Version,
+    ): ZIO[Any, Throwable, Response] = {
 
-    override def request(request: Request): ZIO[Any, Throwable, Response] =
-      requestAsync(request, settings)
+      for {
+        host     <- ZIO.fromOption(hostOption).orElseFail(new IllegalArgumentException("Host is required"))
+        port     <- ZIO.fromOption(portOption).orElseSucceed(80)
```
vigoo, line 174: Maybe we should default to `443` if `sslOption` is defined

jdegoes: I agree, @adamgfraser can you do a followup?

jdegoes (approved):
Beautiful work, and a step closer to having the robust, powerful, compositional HTTP client that ZIO HTTP deserves!

zio-http/src/main/scala/zio/http/Client.scala
adamgfraser: @jdegoes Will do!

PR #1489, opened 2022-09-16
Move pure data classes to `zio.http.model`

### LINKED ISSUE #1447: Organize HTTP model classes into separate package

All high-level data types and service types are located in the `zio.http` package, which makes navigating code more difficult and introduces the possibility of name clashes, with common names like `Headers` and `Cookies` used up by data types.

Move all pure data classes that model aspects of the Http protocol into a new package `zio.http.model`, following the pattern of Tapir, Spray, and other Scala libraries.

**Why is it important?**

Better to do it now when the number of data types is small, than later after a stable release.

987Nabil:
fixes #1447


987Nabil:
I am not sure, if `ClientConfig` and `ServerConfig` should go to the model package too.

zio-http/src/main/scala/zio/http/model/Body.scala
```diff
@@ -1,11 +1,12 @@
-package zio.http
+package zio.http.model
 
 import io.netty.buffer.{ByteBuf, ByteBufUtil, Unpooled}
 import io.netty.channel.{Channel => JChannel, DefaultFileRegion}
 import io.netty.handler.codec.http.LastHttpContent
 import io.netty.util.AsciiString
 import zio._
 import zio.http.service.Ctx
+import zio.http.{HTTP_CHARSET, failNoStacktrace}
```
987Nabil: I think this imports look a bit suspicious. Maybe a hint that there should be some refactoring?

zio-http/src/main/scala/zio/http/model/Cookie.scala
```diff
@@ -1,8 +1,9 @@
-package zio.http
+package zio.http.model
 
-import zio.http.Cookie.{SameSite, Type}
 import zio.http.CookieDecoder.log
+import zio.http.model.Cookie.{SameSite, Type}
 import zio.http.service.Log
```
987Nabil: At least this service import seems odd.

zio-http/src/main/scala/zio/http/model/Headers.scala
```diff
@@ -1,7 +1,8 @@
-package zio.http
+package zio.http.model
 
 import io.netty.handler.codec.http.{DefaultHttpHeaders, HttpHeaders}
-import zio.http.headers.{HeaderConstructors, HeaderExtension, HeaderIterable}
+import zio.http.headers._
```
987Nabil: we can find `headers` imported in some of the model classes. And it seems to me, that the traits in headers are mainly for data manipulation. Maybe `headers` should be a suppackage of `model`?

zio-http/src/main/scala/zio/http/model/HttpError.scala
```diff
@@ -1,6 +1,6 @@
-package zio.http
+package zio.http.model
 
-import zio.http.HttpError.HTTPErrorWithCause
+import zio.http.model.HttpError.HTTPErrorWithCause
```
987Nabil, line 4: In line 25, the `abstract class` is not sealed. Is this leaky or intended? I'd expect it is leaking. If so I could fix this here.

zio-http/src/main/scala/zio/http/model/Patch.scala
```diff
@@ -1,4 +1,6 @@
-package zio.http
+package zio.http.model
+
+import zio.http.Header
```
987Nabil: Is this ok or should it be `import zio.http.model.Headers.Header`

zio-http/src/main/scala/zio/http/model/Request.scala
```diff
@@ -1,4 +1,4 @@
-package zio.http
+package zio.http.model
 
 import io.netty.channel.ChannelHandlerContext
 import io.netty.handler.codec.http.{DefaultFullHttpRequest, FullHttpRequest, HttpRequest}
```
987Nabil: I think the import of the `service` package in line 7 is odd.

zio-http/src/main/scala/zio/http/model/Response.scala
```diff
@@ -12,6 +12,7 @@ import zio.http.service.{
   ClientResponseStreamHandler,
 }
 import zio.http.socket.{SocketApp, WebSocketFrame}
```
987Nabil: I think the import of the packages `service` and `socket` are odd.

zio-http/src/main/scala/zio/http/model/Body.scala
vigoo: I think the default `HTTP_CHARSET` (and also the `HeaderNames` and `HeaderValues` there should also be moved to the model package. 

But `Body` itself encapsulates the implementation of writing it to a netty channel which definitely does not make it a pure data model. There used to be a more declarative encoding of this before it has been refactored to the current type (sealed trait with different cases which can be then handled on the channel implementation side).

zio-http/src/main/scala/zio/http/model/Cookie.scala
vigoo: Yeah I would not log an error on this level

zio-http/src/main/scala/zio/http/model/Headers.scala
vigoo: I agree

zio-http/src/main/scala/zio/http/model/HttpError.scala
vigoo: Make it sealed

zio-http/src/main/scala/zio/http/model/Patch.scala
vigoo: I think headers belong to the model package

zio-http/src/main/scala/zio/http/model/Request.scala
vigoo: - `ServerInboundHandler` is imported for encoding `Body`, as I wrote there it is not a pure data structure, here we are directly translating streaming bodies into netty context
- Note that `Ctx` is a type alias for `ChannelHandlerContext` and both are used in this file, that could be cleaned up
- Capturing and exposing the `ChannelHandlerContext` in `Request` is also a bit ugly but I think right now it is the only way to access the context from the `Http` level. I think it is only used for encoding streaming bodies (mentioned above) and to be able to get remote address but it is also exposed on the API so it is possible that there are other uses. Anyway I think changing this is out of the scope of this PR

zio-http/src/main/scala/zio/http/model/Response.scala
vigoo: Similar to `Request`.

Maybe `Request` and `Response` should not be moved to the `model` package at all.

zio-http/src/main/scala/zio/http/model/Patch.scala
tobia80: Not sure about Header if it must belong to model or not

jdegoes:
Thanks @987Nabil for taking this on! Much appreciated.

A few highlevel thoughts:

1. If it's a simple ADT (sealed traits + case classes + collections + primitives) that models some aspect of the HTTP protocol independent from its Netty engine, then it belongs in `zio.http.model`.
2. If it is a simple ADT, but it does not model some aspect of the HTTP protocol, but rather, its intended to be used even by end-users who do not care about the HTTP protocol, and it is not related to Netty, then it belongs in `zio.http`.
3. If it is Netty specific, then no matter what, it does not belong in `zio.http` or `zio.http.model`, even if it is a simple ADT.
4. If it is **evolving** to be a simple ADT (e.g. almost an ADT, having only a function inside it, which can be refactored later), and if the other pre-requisites are met (not Netty related, part of the HTTP protocol), then it can be located in `zio.http.model`, even though further work needs to be done to make it a simple ADT.

With these guidelines, `Request` and `Response` would be located in `zio.http`, but `Headers` would be located in `zio.http.model`, and so forth.

What does everyone think of these guidelines? cc @vigoo

vigoo:
> 

I like these guidelines. 

Based on this `Body` also would remain in`zio.htttp` which feels correct if `Request` and `Response` is also there

987Nabil:
> Thanks @987Nabil for taking this on! Much appreciated.
> 
> A few highlevel thoughts:
> 
> 1. If it's a simple ADT (sealed traits + case classes + collections + primitives) that models some aspect of the HTTP protocol independent from its Netty engine, then it belongs in `zio.http.model`.
> 2. If it is a simple ADT, but it does not model some aspect of the HTTP protocol, but rather, its intended to be used even by end-users who do not care about the HTTP protocol, and it is not related to Netty, then it belongs in `zio.http`.
> 3. If it is Netty specific, then no matter what, it does not belong in `zio.http` or `zio.http.model`, even if it is a simple ADT.
> 4. If it is **evolving** to be a simple ADT (e.g. almost an ADT, having only a function inside it, which can be refactored later), and if the other pre-requisites are met (not Netty related, part of the HTTP protocol), then it can be located in `zio.http.model`, even though further work needs to be done to make it a simple ADT.
> 
> With these guidelines, `Request` and `Response` would be located in `zio.http`, but `Headers` would be located in `zio.http.model`, and so forth.
> 
> What does everyone think of these guidelines? cc @vigoo

So, I guess 3. is out of scope for this pr and will maybe be done in #1485.
But I am not sure about 2. I guess if I would apply this, I could argue that `Headers` belongs in `zio.http` since for example the `Response` constructor takes a `Headers` parameter. I think if I apply the rule, should be used by the end-user, there is actually nothing left in the Model package. Because there is always a method that is end-user API, that will have a parameter of this type.
This is true for `Cookie` via `setCookie` on `Headers`. For `Headers` via the headers field of the `Response` apply. And I can go on basically for ever.
And what is with constant data that is not an ADT? Like `MimeDb`/`MediaType` for example.

vigoo:
> But I am not sure about 2. I guess if I would apply this, I could argue that `Headers` belongs in `zio.http` since for example the `Response` constructor takes a `Headers` parameter. I think if I apply the rule, should be used by the end-user, there is actually nothing left in the Model package

The headers would belong to the `model` package because they model an aspect of the HTTP protocol. The mention of end-user API in the second point is to distinguish between types that are "public" and going to `zio.http` and others that are internal for example dealing with Netty specific things that should go to another package (out of scope for this PR)

987Nabil:
Thanks for clarifying @vigoo

jdegoes:
@987Nabil Would love to get this merged in soon. Please let me know when conflicts are merged and final clarifications are incorporated.

987Nabil:
@jdegoes I am on it rn

987Nabil:
@jdegoes @vigoo I think this should fit now 🙂

987Nabil:
Can it be, that `fmt` does not format `Client.scala` right? I had to revert it manually to the state of main.

vigoo:
> Can it be, that `fmt` does not format `Client.scala` right? I had to revert it manually to the state of main.

I ran into caching issues with scalafix before, where it did not re-run on some modified files

vigoo (approved):

jdegoes:
Thank you, @987Nabil!

PR #1360, opened 2022-07-21
Add example for reconnecting websocket client

landlockedsurfer:
Hi,
Here's another example that implements a web socket client that uses some reconnection logic to recover from errors.
Hope this is useful!

PR #1491, opened 2022-09-16
Add example for reconnecting websocket client (see also pull request #1360)

landlockedsurfer:
See pull request #1360.


jdegoes (approved):

landlockedsurfer:
We are using ZIO HTTP for a web socket client in production and are using ZIO primitives to support send (via a ZIO queue) and close (via a ZIO promise).
It works very nicely, but the code is a little bit unwieldy.
I was thinking about reworking it by storing the channel on connect in a ZIO `Ref`, to be able to perform send and close on the connection more directly. If it works I'll submit an update to this example.

PR #1493, opened 2022-09-16
Client connection pools

### LINKED ISSUE #1479: Implement client connection pooling

The performance of the Http client can be improved through connection pooling.

Connection pooling is easier to implement with a keyed pool, which ZIO 2 does not yet support. 

We can build a `ZKeyedPool` here and upstream to ZIO 2 later:

```scala
trait ZKeyedPool[+Error, -Key, Item] {
  def get(key: Key)(implicit trace: Trace): ZIO[Scope, Error, Item]

  def invalidate(item: Item)(implicit trace: Trace): UIO[Unit]
}
object ZKeyedPool {
  def makeWith[K, R, E, A](get: K => ZIO[R, E, A], total: => Range, maxPerKey: => Int)(implicit trace: Trace): ZKeyedPool[E, K, A] = ???
}
```

Then we can utilize this connection pool in the implementation of `Client` to ensure that we have ready-to-use connections to common targets.

None.

vigoo:
Resolves #1479 

TODO:

- [x] Web socket support is broken possibly due to the dynamic handler registration
- [x] Try to revert the channel after closed web socket for reuse, or invalidate it in the pool if that is not possible
- [x] Check if the current handlers already automatically handle the connect/keep-alive headers or if it has to be added
- [x] Invalidate the failed channels in the pool
- [x] Configuration for the pools
- [x] Cleanup 
- [x] Tons of tests


zio-http/src/main/scala/zio/http/service/ZKeyedPool.scala
```diff
@@ -0,0 +1,173 @@
...
+  ) extends ZKeyedPool[Err, Key, Item] {
+
+    override def get(key: Key)(implicit trace: Trace): ZIO[Scope, Err, Item] =
+      createdPools
+        .contains(key)
+        .flatMap { alreadyCreated =>
+          if (alreadyCreated) {
+            activePools.get(key).flatMap {
+              case None       =>
+                ZSTM.retry
```
jdegoes: What is this about?

zio-http/src/main/scala/zio/http/service/ZKeyedPool.scala
```diff
@@ -0,0 +1,173 @@
...
+                for {
+                  pool <- createPool(key)
+                  _    <- activePools.put(key, pool).commit
+                  item <- acquireFrom(key, pool)
+                } yield item
+              }
+          }
+        }
+        .commit
+        .flatten
```
jdegoes: `commit.flatten` is "interruptibly racy". That is, the commit may happen, but no the effect produced by the commit. If this is not desired, use `ZIO.uninterruptibleMask { restore => restore(effect.commit).flatMap(identity) }`.

zio-http/src/main/scala/zio/http/service/ZKeyedPool.scala
```diff
@@ -0,0 +1,173 @@
...
+        .get(item)
+        .flatMap {
+          case None      => ZSTM.succeed(ZIO.unit)
+          case Some(key) =>
+            activePools.get(key).map {
+              case None       => ZIO.unit
+              case Some(pool) => pool.invalidate(item)
+            }
+        }
+        .commit
```
jdegoes: Ditto on "interruptibly racy" with this commit.flatten

zio-http/src/main/scala/zio/http/service/ZKeyedPool.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Makes a new pool with the specified minimum and maximum sizes and time to
+   * live before a pool whose excess items are not being used will be shrunk
+   * down to the minimum size. The pool is returned in a `Scope`, which governs
+   * the lifetime of the pool. When the pool is shutdown because the `Scope` is
+   * used, the individual items allocated by the pool will be released in some
+   * unspecified order.
+   */
+  def make[Key, Env: Tag, Err, Item](get: Key => ZIO[Env, Err, Item], range: => Range, timeToLive: => Duration)(implicit
```
jdegoes: As written, it allows infinitely many pools. That's like Netty although if we are using this in the HTTP client, it does mean the number of unique domains should be fixed and small (e.g. no web crawlers).

The moment you impose a cap, however, you have "cache problems", since you now need to do eviction on "unused" or "less used" resources when new requests come in.

zio-http/src/main/scala/zio/http/service/ZKeyedPool.scala
vigoo: Creating a new pool is an effect so we can't do it as part of the transaction, so there is a race condition between checking if you need to create a new one, creating it and then storing it. 
So I chose to maintain a set beside the key->pool map because we can record the intention to create a new pool within the same transaction where we check if it already exists.

So either:
- the pool is already in the map and in the set, then we use it
- the key is in the set, but the pool is not in the map, so we retry because another fiber is already creating it
- not in the set and not in the map, then we put it in the set, commit, create the pool and put it in the map

I think this could be replaced by a key->promise[value] map instead, which may be simpler.

zio-http/src/main/scala/zio/http/service/ZKeyedPool.scala
vigoo: Yes. In the context of HTTP connection pools I was thinking that we never want to limit the number of pools - that would mean that when connecting to a new host you fall back to a "no connection pool" mode. All the other options seems to be invalid:
- you cannot block until another connection pool is closed because that normally does not happen - it keeps connections alive.
- you cannot (easily) force another one to shut down because it may have active connections 

I think when dealing with dynamic hosts, you just want to use the `ConnectionPool.disabled` and you can still provide a connection pool for the static subset of your clients.

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -562,133 +556,166 @@ object ZClient {
         ).withFinalizer(_.close.orDie)
       } yield res
 
-    private def requestAsync(request: Request, clientConfig: ClientConfig)(implicit
-      trace: Trace,
-    ): ZIO[Any, Throwable, Response] = {
+    private def requestAsync(request: Request, clientConfig: ClientConfig): ZIO[Any, Throwable, Response] =
```
devsprint: is there any reason why the trace has been removed?

adamgfraser: I think just because we have merged changes on this a bunch of times and it compiles. The bigger issue is why this compiles even though there is no trace. I think the problem is we still are defining an empty trace in the package object and that needs to get deleted.

jdegoes:
@vigoo Should we just merge this?

jdegoes (approved):

vigoo:
If it's ok that it depends on a snapshot ZIO version we can merge it, otherwise wait for 2.0.3

PR #1495, opened 2022-09-17
Abstract away netty.2

scottweaver:
### [A.A.N.2] Request and Version refactorings
* 'Request' is now a simple case class.
  * 'unsafe' removed and all Netty-specific operations moved to 'zio.http.netty.Requests'.
* 'Version'. 'unsafe' removed and all Netty-specific operations moved to 'zio.http.netty.Versions'.
* 'remoteAddress' removed from 'Http' and relocated to to 'Request'.  This is a side-effect of removing access to Netty's 'ChannelHandlerContext' 'Http'.


jdegoes:
@scottweaver Fantastic to see this! Can you fix conflicts after the `http.model` refactoring, which was merged in this morning? Thank you!

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -1,199 +1,54 @@
...
 
   /**
-   * Constructor for Request
+   * Constructor for Request that provides sane defaults.
    */
-  def apply(
+  def make(
     version: Version = Version.`HTTP/1.1`,
     method: Method = Method.GET,
     url: URL = URL.root,
```
987Nabil, line 49: Is `URL.root` really a sane default?

987Nabil: Let me be more specific. I think a URL can/should not have any default value. I think the other values are sane, because they have some technical reason(ing) behind them. And they describe the _how_. But URL is the _what_ and I'd assume, that in 99.9% of the cases URL won't be `URL.root` and will be overwritten.
Idk if this is the right PR to do it (most likely not), but since we are working on a 2.0 release with a lot of changes anyway, I'd say we should change this.
I would say, maybe one could just remove the default URL parameter, or have methods to build a request based on the http method somewhat like this
```scala
object Request {
...
  def get(url:URL): Request = make(url = url)
  def post(url: URL, body: Body) = make(url = url, body = body)
```
And then the idiomatic use would be
```scala
val myRequest = Request.get(url).addHeaders(myHeaders)
```
@jdegoes or @vigoo maybe you have an opinion? I am happy to open an issue/pr, if you see some potential for improvement 🙂

scottweaver: @jdegoes Rebased and pushed.

scottweaver:
@jdegoes Rebased and pushed.

scottweaver:
Closing this PR in favor of a local branch variant so as to be able to run benchmarks correctly.

PR #1512, opened 2022-09-19
propagate tracing information

### LINKED ISSUE #1503: Propagate Tracing Information

All ZIO HTTP operators should propagate tracing information so that execution traces show the user's code and not the "guts" of ZIO HTTP.

All ZIO HTTP operators that construct a ZIO workflow should require an implicit trace. So instead of:

```scala
def foo: ZIO[Any, Nothing, Int] =
  ???
```

We should do:

```scala
def foo(implicit trace: Trace): ZIO[Any, Nothing, Int] =
  ???
```

We can enforce this by importing `import zio.stacktracer.TracingImplicits.disableAutoTrace` in each file in the project, which will disable automatic generation of traces and force us either to propagate traces or explicitly generate new ones.

None.

devsprint:
Fixes #1503

### DIFF
--- a/project/ScalaSettings.scala
+++ b/project/ScalaSettings.scala
@@ -10,7 +10,7 @@ trait ScalaSettings {
     "-language:higherKinds",     // Allow higher-kinded types
     "-unchecked",                // Enable additional warnings where generated code depends on assumptions.
     "-Xcheckinit",               // Wrap field accessors to throw an exception on uninitialized access.
-   // "-Xfatal-warnings",          // Fail the compilation if there are any warnings.
+    "-Xfatal-warnings",          // Fail the compilation if there are any warnings.
     "-Xlint:adapted-args",       // Warn if an argument list is modified to match the receiver.
     "-Xlint:constant",           // Evaluation of a constant arithmetic expression results in an error.
     "-Xlint:delayedinit-select", // Selecting member of DelayedInit.
@@ -42,6 +42,7 @@ trait ScalaSettings {
     //  "-Ywarn-infer-any",                 // Warn when a type argument is inferred to be `Any`.
     //  "-language:experimental.macros",   // Allow macro definition (besides implementation and application). Disabled, as this will significantly change in Scala 3
     //  "-language:implicitConversions",   // Allow definition of implicit functions called views. Disabled, as it might be dropped in Scala 3. Instead use extension methods (implemented as implicit class Wrapper(val inner: Foo) extends AnyVal {}
+    "-Wconf:cat=unused:info"               // log unused as info
   )
 
   val scala3Settings: Seq[String] = Seq("-Xignore-scala2-macros", "-noindent")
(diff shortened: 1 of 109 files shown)


vigoo:
Could you also add the `import zio.stacktracer.TracingImplicits.disableAutoTrace` imports to enforce this?

devsprint:
@vigoo I had it initially in the code but I have removed as the linter is marking them as unused and breaks compilation

vigoo:
I see. Maybe this option from `zio` allows that (I'm not sure): https://github.com/zio/zio/blame/series/2.x/project/BuildHelper.scala#L118

adamgfraser:
Yes we should silence that warning so we can include this import. Otherwise it is going to be very difficult to enforce this.

project/ScalaSettings.scala
```diff
@@ -10,7 +10,7 @@ trait ScalaSettings {
     "-language:higherKinds",     // Allow higher-kinded types
     "-unchecked",                // Enable additional warnings where generated code depends on assumptions.
     "-Xcheckinit",               // Wrap field accessors to throw an exception on uninitialized access.
-    "-Xfatal-warnings",          // Fail the compilation if there are any warnings.
+   // "-Xfatal-warnings",          // Fail the compilation if there are any warnings.
```
devsprint: Looking to find a solution without commenting this flag.

zio-http/src/main/scala/zio/http/netty/ChannelFactories.scala
```diff
@@ -27,6 +27,7 @@ object ChannelFactories {
     def uring(implicit trace: Trace)  = serverChannel(new IOUringServerSocketChannel())
     def kqueue(implicit trace: Trace) = serverChannel(new KQueueServerSocketChannel())
 
+    implicit val trace: Trace = Trace.empty
```
devsprint: @adamgfraser @vigoo  what will be the recommended approach to create/pass a trace at this level ?

adamgfraser: @devsprint I think using an empty trace here is fine.

jdegoes:
@devsprint Build not passing.

project/ScalaSettings.scala
devsprint: fixed

zio-http/src/test/scala/zio/http/BodySpec.scala
```diff
@@ -29,10 +29,10 @@ object BodySpec extends ZIOSpecDefault {
           ),
           suite("fromFile")(
             test("failure") {
-              val res: URIO[Any, Either[Throwable, Chunk[Byte]]] =
+              val res =
```
devsprint, line 32: @adamgfraser  I have ignored the test for now, but I think the test is not correct. It is not failing the compilation however ...

adamgfraser (approved):

PR #1516, opened 2022-09-19
refactor: Optimize Headers#get

### LINKED ISSUE #1511: Optimize Headers#get

`Headers#get` has a quick and dirty implementation that is extremely inefficient. This method is intended to get the first header having a certain name.

Implement `Headers#get` by implementing it in the subtypes of `Headers`, which is about as fast as we can get without having an index. Take care to perform minimal allocations, always using low-level APIs rather than high-level APIs.

N/A

**Additional context**

The API now returns an `Option` but we could introduce an even-lower level `Unsafe` API that returns `null`, which is package-private to `http`, and which can be used by the code inside `zio.http.api` for improved performance.

jakubjanecek:
Closes #1511

### DIFF
--- a/zio-http/src/main/scala/zio/http/model/Headers.scala
+++ b/zio-http/src/main/scala/zio/http/model/Headers.scala
@@ -27,9 +27,13 @@ sealed trait Headers extends HeaderExtension[Headers] with HeaderIterable {
 
   private[http] def encode: HttpHeaders
 
-  // TODO: Make this fast and work on CharSequence
-  final def get(key: String): Option[String] =
-    collect { case Header(k, v) if k.toString() == key => v.toString() }.headOption
+  final def get(key: String): Option[String] = Option(getUnsafe(key))
+
+  /**
+   * @return
+   *   null if header is not found
+   */
+  private[http] def getUnsafe(key: String): String
 
   override final def headers: Headers = self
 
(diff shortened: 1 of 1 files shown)


zio-http/src/main/scala/zio/http/model/Headers.scala
```diff
@@ -124,6 +141,18 @@ object Headers extends HeaderConstructors {
     override def iterator: Iterator[Header] =
       toJHeaders.entries().asScala.map(e => Header(e.getKey, e.getValue)).iterator
 
+    private[http] override def getUnsafe(key: String): String = {
+      val iterator = toJHeaders.iteratorAsString()
+      while (iterator.hasNext) {
+        val entry = iterator.next()
+        if (entry.getKey == key) {
```
jdegoes, line 148: Probably faster to compare the two `CharSequence` directly rather than converting all the keys to strings via `iteratorAsString`, right?

jdegoes (approved):
Thank you! Going to approve and merge this, feel free to do a followup PR to improve performance in the Java headers case!

zio-http/src/main/scala/zio/http/model/Headers.scala
jakubjanecek: Right, missed that, thank you.

PR #1518, opened 2022-09-19
feat: CharSequenceExtensions

### LINKED ISSUE #1510: Create CharSequenceExtensions to help consolidate CharSequence-related code

ZIO HTTP tries to use `CharSequence` which can be backed by entities other than `String`, which helps because `String` must always be decoded into bytes before it can be written to sockets, and other entities can have this logic baked in. 

However, `CharSequence` have undefined equality and hash code, and there is no way to test for substrings (etc.).

To rectify these problems, we can create `CharSequenceExtensions` object in `zio.http.internal`, which contains a few helper functions:

```scala
def equals(left: CharSequence, right: CharSequence): Boolean = compare(left, right) == 0
def compare(left: CharSequence, right: CharSequence): Int = ???
def hashCode(c: CharSequence): Int = ??? // computed like String.hashCode
def contains(sequence: CharSequence, subsequence: CharSequence): Boolean = ???
...
```

In addition, the methods in `Headers.scala` which compare char sequences and compute their hash code can be refactored to use the new helpers in `CharSequencExtensions`.

N/A

**Additional context**

Equality should be implemented to short-circuit on reference equality: `if (left eq right) true else { /* longer more expensive computation */ }`.

jakubjanecek:
Closes #1510

@jdegoes Before I polish and test the implementation, is this what you had in mind? I've basically copied code from JDK + case sensitivity.

### DIFF
--- a/zio-http/src/main/scala/zio/http/internal/CharSequenceExtensions.scala
+++ b/zio-http/src/main/scala/zio/http/internal/CharSequenceExtensions.scala
@@ -1,0 +1,79 @@
+package zio.http.internal
+
+private[http] object CharSequenceExtensions {
+
+  def equals(left: CharSequence, right: CharSequence, caseMode: CaseMode = CaseMode.Sensitive): Boolean =
+    if (left eq right) true else compare(left, right, caseMode) == 0
+
+  def compare(left: CharSequence, right: CharSequence, caseMode: CaseMode = CaseMode.Sensitive): Int = {
+    val modify = modifier(caseMode)
+
+    if (left eq right) {
+      0
+    } else {
+      var i      = 0
+      val length = Math.min(left.length, right.length)
+      while (i < length) {
+        val a = modify(left.charAt(i))
+        val b = modify(right.charAt(i))
+        if (a != b) return a - b
+
+        i += 1
+      }
+
+      left.length - right.length
+    }
+  }
+
+  def hashCode(cs: CharSequence): Int = {
+    val length = cs.length()
+    var hash   = 0
+    var i      = 0
+    while (i < length) {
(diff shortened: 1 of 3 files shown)


jdegoes:
@jakubjanecek 

These utilities are exactly those envisioned. However, due to licensing issues, all code contributed to ZIO HTTP should be de novu, created from scratch. The algorithm used to compute the hash code for a String in Java is perfectly fine to use in a fresh implementation, but the code itself may not be reused, but must be independently created.

So sorry to have to ask this, but can you create fresh implementations for all of these utilities? I wrote one myself for CharSequence equality in `Headers.scala`.

zio-http/src/main/scala/zio/http/internal/CharSequenceExtensions.scala
```diff
@@ -0,0 +1,79 @@
...
+          return true
+        }
+      }
+
+      i += 1
+    }
+    false
+  }
+
+  private def modifier(caseMode: CaseMode): Char => Char = caseMode match {
```
jdegoes: Unfortunately every time this function is invoked, it will create a new lambda (`Function1`), which will have an `invokedynamic` associated with its invocation.

I know it is not "good practice", but we need to switch off `caseMode` in the parent function, and have separate code paths for each, so that we do not need to call a lambda for each character.

jakubjanecek:
OK, I'll rework it. I got sick so give me a bit more time, thanks.

jakubjanecek:
@jdegoes Could you have another look at it? It's not as elegant but the tests pass 😉

jdegoes (approved):

jdegoes:
Thank you!

PR #1534, opened 2022-09-19
Abstract away netty.2

scottweaver:


zio-http/src/main/scala/zio/http/Driver.scala
```diff
@@ -2,10 +2,35 @@ package zio.http
...
 
-  def setErrorCallback(newCallback: Option[Server.ErrorCallback]): UIO[Unit]
+  def setErrorCallback(newCallback: Option[Server.ErrorCallback]): Unit
+
+  def addApp(newApp: HttpApp[Any, Throwable]): Unit
+}
+
+object Driver {
+
+  private[zio] final case class Context private (
```
jdegoes: This class (`Driver.Context`) should be regarded as a private implementation detail for _some_ drivers, but not others, and moved to some place Netty-specific.

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -723,10 +722,10 @@ object Http {
     i.reduce(_.defaultWith(_))
 
   /**
-   * Provides access to the request's ChannelHandlerContext
+   * [1453] Provides access to the request's ChannelHandlerContext
    */
-  def context: Http[Any, Nothing, Request, ChannelHandlerContext] =
-    Http.fromFunctionZIO[Request](request => ZIO.succeedUnsafe { implicit u => request.unsafe.context })
+  // def context: Http[Any, Nothing, Request, ChannelHandlerContext] = ???
```
jdegoes: Dead code

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -981,15 +980,16 @@ object Http {
...
+   * `zio.http.Request`
    */
-  def remoteAddress: Http[Any, IOException, Request, InetAddress] =
-    context flatMap { ctx =>
-      ctx.channel().remoteAddress() match {
-        case m: InetSocketAddress => Http.succeed(m.getAddress)
-        case _                    => Http.fail(new IOException("Unable to get remote address"))
-      }
-    }
+  // def remoteAddress: Http[Any, IOException, Request, InetAddress] =
```
jdegoes: Dead code

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -1037,10 +1037,10 @@ object Http {
   def tooLarge: HttpApp[Any, Nothing] = Http.status(Status.RequestEntityTooLarge)
 
   /**
-   * Provides low level access to an HttpApp to perform unsafe operations using
-   * the request's ChannelHandlerContext.
+   * [1453] Provides low level access to an HttpApp to perform unsafe operations
+   * using the request's ChannelHandlerContext.
    */
-  def usingContext[R, E](f: ChannelHandlerContext => HttpApp[R, E]): HttpApp[R, E] = context.flatMap(f(_))
+  // def usingContext[R, E](f: ChannelHandlerContext => HttpApp[R, E]): HttpApp[R, E] = context.flatMap(f(_))
```
jdegoes: Dead code

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -1037,10 +1037,10 @@ object Http {
   def tooLarge: HttpApp[Any, Nothing] = Http.status(Status.RequestEntityTooLarge)
 
   /**
-   * Provides low level access to an HttpApp to perform unsafe operations using
-   * the request's ChannelHandlerContext.
+   * [1453] Provides low level access to an HttpApp to perform unsafe operations
```
jdegoes: I don't know what these `[1453]` comments are about (here and elsewhere).

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -1,199 +1,55 @@
...
-   * logging or other purposes. It contains the essential properties of HTTP
-   * request: protocol version, method, URL, headers and remote address.
-   */
-  final override def toString = s"Request($version, $method, $url, $headers)"
-
-  /**
-   * Updates the headers using the provided function
-   */
-  final override def updateHeaders(update: Headers => Headers): Request = self.copy(headers = update(self.headers))
+final case class Request(
```
jdegoes, line 9: This change will need to be benchmarked since one of the reasons `Request` was a trait was to provide the ability to virtualize some of its fields, rather than be forced to materialize the body, header, method, url, version, and remote address (including its decoding and allocation cost).

Now, I am not sure if this was actually utilized anymore, or if the `trait` was more "forward thinking". We need to benchmark to make sure no regression here.

To accomodate high-performance use-cases, if necessary, we could introduce a `trait RequestLike` that defines all the same fields, and then read-only use cases could require `RequestLike` rather than `Request`.

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -1,199 +1,55 @@
...
-      }
-    }
-  }
+  val path = url.path
 
-  /**
-   * Remove trailing slash from path.
-   */
-  final def dropTrailingSlash: Request = self.copy(url = self.url.dropTrailingSlash)
+  def withMethod(method: Method): Request = self.copy(method = method)
```
jdegoes: I think `copy` is fine and we should probably delete all the `withXYZ` methods.

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -1,199 +1,55 @@
...
 }
 
 object Request {
 
   /**
-   * Constructor for Request
+   * Constructor for Request that provides sane defaults.
    */
-  def apply(
+  def make(
```
jdegoes: It's an anti-pattern (in my view) for a straightforward case class to have a `make` constructor. Rather, the default case class constructor should always be used: `Request(...)` (ordinary Scala) instead of `Request.make(...)` (weird Scala).

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -1,199 +1,55 @@
...
-    }
-  }
-
-  private[zio] def fromHttpRequest(jReq: HttpRequest)(implicit ctx: Ctx): Request = {
-    val protocolVersion = Version.unsafe.fromJava(jReq.protocolVersion())(Unsafe.unsafe)
-
-    new Request {
-      override final def headers: Headers = Headers.make(jReq.headers())
-      override final def method: Method   = Method.fromHttpMethod(jReq.method())
-      override final def url: URL         = URL.fromString(jReq.uri()).getOrElse(URL.empty)
```
jdegoes, line 188: Looking at these, I suspect the use of `trait` for `Request` was not buying any performance gains, because every field is created every time. In addition, you succeeded to delete `val unsafe` which has negative performance implications.

zio-http/src/main/scala/zio/http/Request.scala
scottweaver: I try to avoid default arguments in the base class itself, that's what I created a `make` so as to support sane defaults.

At previous gigs, we considered default values in a case class an anti-pattern.

zio-http/src/main/scala/zio/http/Http.scala
scottweaver: Oops, they were there as a reference point.  That is the ticket number.

zio-http/src/main/scala/zio/http/Driver.scala
scottweaver: It actually shouldn't exist at all.  It was replaced by new methods in the Driver.

zio-http/src/main/scala/zio/http/Request.scala
scottweaver: Deleting unsafe from all the public APIs is one of my end-goals for this effort.

zio-http/src/main/scala/zio/http/Request.scala
jdegoes: I also agree, but let's do that via `Request.default.copy(...)`.

jdegoes: Or maybe `Request.get` would be more appropriate. Then a user could do: `Request.get.copy(...)`.

zio-http/src/main/scala/zio/http/Request.scala
scottweaver: I assume your comment further down allays some the concern here?  Still, I will be running benchmarks.

jdegoes:
@scottweaver Can you resolve conflicts on `|ServerConfig`?

scottweaver:
Rebased and pushed.

zio-http/src/main/scala/zio/http/netty/Versions.scala
```diff
@@ -0,0 +1,21 @@
...
+object Versions {
+  import Version._
+  def make(version: HttpVersion)(implicit unsafe: Unsafe, trace: Trace): Version =
+    version match {
+      case HttpVersion.HTTP_1_0 => Http_1_0
+      case HttpVersion.HTTP_1_1 => Http_1_1
+      case _                    => throw new IllegalArgumentException(s"Unsupported HTTP version: $version")
+    }
+
+  def make(version: Version): HttpVersion = version match {
```
jdegoes: This can throw exceptions.

zio-http/src/main/scala/zio/http/netty/Versions.scala
```diff
@@ -0,0 +1,21 @@
+package zio.http.netty
+
+import io.netty.handler.codec.http.HttpVersion
+import zio._
+import zio.http.model.Version
+
+object Versions {
+  import Version._
+  def make(version: HttpVersion)(implicit unsafe: Unsafe, trace: Trace): Version =
```
jdegoes: `make` is good for a constructor in the companion object of a data type, but for a utility like this, `convertVersion` is probably better.

zio-http/src/main/scala/zio/http/netty/Versions.scala
jdegoes: Or maybe not.

zio-http/src/main/scala/zio/http/netty/Versions.scala
jdegoes: Maybe `convertNettyToZIO` and `convertZIOToNetty`, respectively?

zio-http/src/main/scala/zio/http/netty/Versions.scala
scottweaver: ~`Versions#make(HttpVersion)` removed and all code now inlined into `SeverInboundHandler`.~

zio-http/src/main/scala/zio/http/netty/Versions.scala
scottweaver: `Versions#make(HttpVersion)` removed and all code now inlined into `SeverInboundHandler`.

jdegoes (approved):

PR #1536, opened 2022-09-20
Refactor client ssl config data type

ghidei:


zio-http/src/main/scala/zio/http/ClientSSLConfig.scala
```diff
@@ -0,0 +1,11 @@
+package zio.http
+
+sealed trait ClientSSLConfig
+
+object ClientSSLConfig {
+  case object Default                                                                         extends ClientSSLConfig
```
jdegoes, line 6: There's overlap with the server one--does it make sense to unify them, or not?

zio-http/src/main/scala/zio/http/ClientSSLConfig.scala
```diff
@@ -0,0 +1,11 @@
+package zio.http
+
+sealed trait ClientSSLConfig
```
jdegoes, line 3: Great name change for improved consistency. 👍

jdegoes (approved):

zio-http/src/main/scala/zio/http/ClientSSLConfig.scala
ghidei: There's overlap, but I don't think it's worth unifying them, because then we'd have to prefix the branches like: `ClientFromResource`, or have something like `case class SSLConfig(..., ..., clientOrServer: ClientOrServer)`

PR #1538, opened 2022-09-20
Client decompression

fokot:


zio-http/src/main/scala/zio/http/ClientConfig.scala
```diff
@@ -13,6 +13,7 @@ case class ClientConfig(
   channelType: ChannelType = ChannelType.AUTO,
   nThreads: Int = 0,
   useAggregator: Boolean = true,
+  requestDecompression: (Boolean, Boolean) = (false, false),
```
kitlangton: It might be nice to a have a case class for this pair o' bools, just so it's clearer which is which.

fokot: In reality it has 3 values not 4 `(false, true)` and `(false, false)` are the same. 

We have 4 options
1. `case class Decompression(enabled: Boolean, strict: Boolean)`
2. `case class Decompression(strict: Boolean)` and to use `requestDecompression: Option[Decompression]`
3. `Option[Boolean]` - `None` - no decompressions and `Some(true)` - strict and `Some(false)` non strict
4. 
```sealed trait Decompression
object Decompression {
  case object NoDecompression extends Decompression
  case object StrictDecompression extends Decompression
  case object NonStrictDecompression extends Decompression
}
```
@kitlangton Which one do you prefer? Or do you have another?

jdegoes: Use a sealed trait extended by 3 case objects instead of a case class. You can add one boolean method to that trait if desired.

fokot: I changed it also in `ServerConfig`, the Boolean tuple was from there.

zio-http/src/main/scala/zio/http/ClientConfig.scala
```diff
@@ -26,6 +27,8 @@ case class ClientConfig(
   def maxThreads(nThreads: Int): ClientConfig = self.copy(nThreads = nThreads)
 
   def useObjectAggregator(objectAggregator: Boolean): ClientConfig = self.copy(useAggregator = objectAggregator)
+
+  def requestDecompression(decompression: Decompression): ClientConfig = self.copy(requestDecompression = decompression)
```
kitlangton: I think the Decompression sealed trait is great internally, as it's clearer how to use it as an implementer. However, I think accepting a bool here is fine. As `No` is sort of the default state, and now you're opting into Decompression.

```suggestion
  def requestDecompression(isStrict: Boolean): ClientConfig = {
    val decompression = if (isStrict) Decompression.Strict else Decompression.NonStrict
    self.copy(requestDecompression = decompression)
  }
    
```

kitlangton: But other than that, looks great!

fokot: Changed

kitlangton (approved):
Woo!

fokot:
I just fixed the merge conflicts after review.

vigoo (approved):

PR #1541, opened 2022-09-21
Build Cause Into Http

### LINKED ISSUE #1539: Http.catchAllCause
`Http.catchAllDefect` gives me ability to handle `Throwable`s. But they do not have nice stack trace information as `Cause`. It would be nice to have method to handle `Cause` as well. Now when I want to log the error in an application I will log every endpoint separately as there I can do `ZIO.foldCauseZIO` and get tracing information.

adamgfraser:
Resolves #1539.

To propagate execution tracing we are going to need to maintain the `Cause` of failure in the event of errors instead of just an error or a defect.

### DIFF
--- a/zio-http/src/main/scala/zio/http/HExit.scala
+++ b/zio-http/src/main/scala/zio/http/HExit.scala
@@ -1,7 +1,7 @@
 package zio.http
 
 import zio.http.HExit.Effect
-import zio.{Trace, ZIO}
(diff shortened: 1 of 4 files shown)


kitlangton (approved):

jdegoes (approved):

jdegoes:
❤️

PR #1543, opened 2022-09-21
fix client request in case of streaming.

devsprint:
Client should flush all pending messages in order to complete a request sent to server.


vigoo:
Why is it necessary to flush before start writing the body?

jdegoes (approved):

Issue #1524, opened 2022-09-19
Improve headers codecs in zio.http.api to operate on structured data


jdegoes:
**Is your feature request related to a problem? Please describe.**

The header inputs, which one can construct using `zio.http.api.HttpCodec`, and which are defined in `/src/main/scala/zio/http/api/HeaderCodecs.scala`, currently produce (/accept) a `String` value, which is unstructured.

Unstructured data is more difficult to work with and may lead to protocol errors if the user does not happen to generate or parse exactly the right header value structure.

**Describe the solution you'd like**

In order to provide a safer and more useful way to interact with header values, we can change the definitions to accept/produce structured values. For example, instead of `HeaderCodec.accept` returning type `HeaderCodec[String]`, it could return type `HeaderCodec[MimeType]`, for some sealed trait `MimeType`. This can be done by utilizing the `HeaderCodec#transform` method, and by implementing a pair of highly-optimized conversion functions: one to convert the string to a specialized, type-safe data type, and the other, to convert the type-safe data type to a string.

The `HttpCodec` data type is located in `src/main/scala/zio/http/api/HeaderCodec.scala`. The new type-safe data types for header values should be located in `zio.http.model` (maybe in the headers subpackage?).

Given the scope of this work, each person who wants to work on this should pick one of the following headers to work on, claim that in the comments below, and submit a pull request for that specific header. When completed, the box may be checked.

 - [x] accept
 - [x] acceptCharset
 - [x] acceptEncoding
 - [x] acceptLanguage
 - [x] acceptRanges
 - [x] acceptPatch
 - [x] accessControlAllowCredentials
 - [x] accessControlAllowHeaders
 - [x] accessControlAllowMethods
 - [x] accessControlAllowOrigin
 - [x] accessControlExposeHeaders
 - [x] accessControlMaxAge
 - [x] accessControlRequestHeaders
 - [x] accessControlRequestMethod
 - [x] age
 - [x] allow
 - [x] authorization
 - [x] cacheControl
 - [x] connection
 - [x] contentBase
 - [x] contentEncoding
 - [x] contentLanguage
 - [x] contentLength
 - [x] contentLocation
 - [x] contentTransferEncoding
 - [x] contentDisposition
 - [x] contentMd5
 - [x] contentRange
 - [x] contentSecurityPolicy
 - [x] contentType
 - [x] cookie
 - [x] date
 - [x] dnt
 - [x] etag
 - [x] expect
 - [x] expires
 - [x] from
[...]

PR #1548, opened 2022-09-22
added structured data for acceptEncoding header

devsprint:
#1524 - implementation for `acceptEncoding`


zio-http/src/main/scala/zio/http/model/headers/Structured.scala
```diff
@@ -0,0 +1,75 @@
+package zio.http.model.headers
+
+import zio.Chunk
+
+import scala.util.Try
+
+object Structured {
```
jdegoes: Maybe this should be called `HeaderValues` and located in `zio.http.model`, instead? I think I'd like to delete the headers package eventually.

zio-http/src/main/scala/zio/http/api/TextCodec.scala
```diff
@@ -74,4 +80,20 @@ object TextCodec {
 
     def encode(value: UUID): String = value.toString
   }
+
+  case object EncodingCodec extends TextCodec[Encoding] {
+    override def decode(value: String): Option[Encoding] = {
+      value.split(",").flatMap(Encoding.identifyEncodingFull) match {
```
jdegoes: We should avoid using `flatMap`, `split`, and other allocation-intensive methods in performance critical code. Although we can merge this in and fix later, it's best to just get into the habit of focusing on performance in the first pass for the structured header values.

zio-http/src/main/scala/zio/http/api/TextCodec.scala
```diff
@@ -74,4 +80,20 @@ object TextCodec {
 
     def encode(value: UUID): String = value.toString
   }
+
+  case object EncodingCodec extends TextCodec[Encoding] {
```
jdegoes: We do **not** want to add a new `TextCodec` for every structured type! There would be hundreds of `TextCodec` which would make maintenance impossible.

Rather, we want to use the *existing* text codecs, to make an `In`, and then `transform` that `In[String]` to `In[Encoding]` using the `In#transform` method. Does that make sense?

devsprint: it makes sense. I will update accordingly.

devsprint: updated

zio-http/src/main/scala/zio/http/model/headers/Structured.scala
devsprint: renamed to `HeaderValue` as `HeaderValues` is already used ...

zio-http/src/main/scala/zio/http/model/HeaderValue.scala
```diff
@@ -0,0 +1,90 @@
+package zio.http.model
+
+import zio.Chunk
+
+import scala.util.Try
+
+object HeaderValue {
```
tobia80: I guess HeaderValue will contain the big list of all the headers. Correct? In that case the file could become huge. Is not better create a package headervalues or something similar with all the headers there and it would cause conflicts using the same object

devsprint: you are right. I have migrated to a package.

tobia80 (approved):

zio-http/src/main/scala/zio/http/api/TextCodec.scala
devsprint: re-implemented without `split` and `flatMap`

tobia80 (approved):

jdegoes (approved):

jdegoes:
@devsprint Let me know when conflicts are fixed and we're good to merge!

devsprint:
@jdegoes conflict is resolved, ready to merge.

jdegoes:
@devsprint Sorry for the delay. Can you fix conflicts and re-open this pull request? Thank you!

devsprint:
@jdegoes - ready for merge

PR #1551, opened 2022-09-22
1524: added structured data for acceptCharset

tobia80:
Removed AcceptCharset


devsprint:
@thehonesttech  Mozilla is recommending to not support this header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset

zio-http/src/main/scala/zio/http/model/headers/values/AcceptCharset.scala
```diff
@@ -0,0 +1,72 @@
+package zio.http.model.headers.values
+
+sealed trait AcceptCharset {
```
jdegoes: Let's move this to a new object `object HeaderValues` inside `zio.http.model`.

zio-http/src/main/scala/zio/http/model/headers/values/AcceptCharset.scala
```diff
@@ -0,0 +1,72 @@
...
+   * Unicode character set
+   */
+  object `US-ASCII` extends AcceptCharset {
+    override val raw: String = "US-ASCII"
+  }
+
+  /**
+   * ISO Latin Alphabet No. 1, a.k.a. ISO-LATIN-1
+   */
+  object `ISO-8859-1` extends AcceptCharset {
```
jdegoes: I'm not sure about this convention. I see it in some Scala projects. But others would do `ISO_8859_1`, which I think I have a slight preference for.

adamgfraser: We use `_` in ZIO Core.

tobia80:
> @thehonesttech Mozilla is recommending to not support this header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset

Should I remove every accept-charset reference everywhere then?

devsprint:
> > @thehonesttech Mozilla is recommending to not support this header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Charset
> 
> Should I remove every accept-charset reference everywhere then?

Either we removed it totally or we might define a single value: UTF_8 and to ignore the raw value received.

zio-http/src/main/scala/zio/http/model/headers/values/AcceptCharset.scala
devsprint: @jdegoes the `HeaderValues` object will become very large. Keeping the `headers.values` package will allow easier parallelisation of development of the header values, i.e. less conflicts during merge ...

jdegoes (approved):

zio-http/src/main/scala/zio/http/model/headers/values/AcceptCharset.scala
jdegoes: How about this:

```scala
sealed trait HeaderValue
object HeaderValue {
  final case class Age(...) extends HeaderValue
  ...
}
```

This is a nice and logical way to organize them even if they share no similarities for now. 

It also let's us say at some point, `final case class Multiple(headerValues: Chunk[HeaderValue])` which we can't say right now.

zio-http/src/main/scala/zio/http/model/headers/values/AcceptCharset.scala
jdegoes: All right, let's replicate that here.

jdegoes (approved):

PR #1553, opened 2022-09-23
implemented Age header value

devsprint:
#1524 - implemented `Age` header value


zio-http/src/main/scala/zio/http/api/HeaderInputs.scala
```diff
@@ -1,7 +1,7 @@
 package zio.http.api
 
 import zio.http.model.HeaderNames
-
+import zio.http.model.headers.values.Age
```
jdegoes, line 4: I want to delete `zio.http.model.headers`, so let's not put more stuff in there. I think a good place is `object HeaderValue` inside` zio.http.model`, so we can put a sealed trait around all of them, eventually (called `HeaderValue`).

jdegoes (approved):

PR #1554, opened 2022-09-23
Implemented ContentLength header value

devsprint:
#1524 - `ContentLength` header value


zio-http/src/main/scala/zio/http/model/headers/values/ContentLength.scala
```diff
@@ -0,0 +1,35 @@
...
+ */
+sealed trait ContentLength
+
+object ContentLength {
+
+  /**
+   * The Content-Length header indicates the size of the message body, in bytes,
+   * sent to the recipient.
+   */
+  final case class ContentLengthValue(seconds: Int) extends ContentLength
```
fsvehla: seconds?

zio-http/src/main/scala/zio/http/model/headers/values/ContentLength.scala
```diff
@@ -0,0 +1,35 @@
...
+  case object InvalidContentLengthValue extends ContentLength
+
+  def fromContentLength(contentLength: ContentLength): String =
+    contentLength match {
+      case ContentLengthValue(length) => length.toString
+      case InvalidContentLengthValue  => ""
+    }
+
+  def toContentLength(value: String): ContentLength =
+    Try(value.trim.toInt).fold(
```
fsvehla: `.toIntOption`?

zio-http/src/main/scala/zio/http/model/headers/values/ContentLength.scala
devsprint: good catch. Fixed.

zio-http/src/main/scala/zio/http/model/headers/values/ContentLength.scala
devsprint: it is not available in scala 2.12.x

zio-http/src/main/scala/zio/http/model/headers/values/ContentLength.scala
```diff
@@ -0,0 +1,35 @@
...
+  case object InvalidContentLengthValue extends ContentLength
+
+  def fromContentLength(contentLength: ContentLength): String =
+    contentLength match {
+      case ContentLengthValue(length) => length.toString
+      case InvalidContentLengthValue  => ""
+    }
+
+  def toContentLength(value: String): ContentLength =
+    Try(value.trim.toLong).fold(
```
jdegoes, line 30: Unfortunately this will involve a tremendous amount of overhead. 

I am fine just accepting this as-is, but we will need to separately optimize all of these decodings.

jdegoes (approved):

jdegoes (approved):

PR #1556, opened 2022-09-23
Tracing Cleanup

adamgfraser:


jdegoes (approved):
It's likely adding stack tracing to `Http` will slow things down. Might be useful to do a benchmark so we know what to attribute to these changes!

jdegoes:
(Good to merge when passing.)

adamgfraser:
Yes. I tend to think that execution tracing is just an expected feature for an idiomatic ZIO library that we need to support even at some cost to performance but that is up for debate and we should quantity how large the trade off is. Let's get everything in here and then we can look at the benchmarks and make a decision.

PR #1561, opened 2022-09-23
Add api client/server spec

### LINKED ISSUE #1529: Comprehensive test suite for ZIO HTTP API endpoints

`zio.http.api` lacks tests for client and server, for happy and unhappy paths.

Develop one test suite for the client, one for the server, which test both happy and unhappy paths.

None.

**Additional context**

Goal should be relatively high level of code coverage of `zio.http.api`.

kitlangton:
Adds a simple spec that makes sure the Client and Server derived from the same API work as expected.

Edit - Closes #1529

### DIFF
--- a/zio-http/src/test/scala/zio/http/api/ServerClientIntegrationSpec.scala
+++ b/zio-http/src/test/scala/zio/http/api/ServerClientIntegrationSpec.scala
@@ -1,9 +1,9 @@
 package zio.http.api
 
-import zio.{Scope, ZIO, ZLayer}
 import zio.http.{Client, ClientConfig, Server, ServerConfig, URL}
 import zio.schema.{DeriveSchema, Schema}
 import zio.test.{ZIOSpecDefault, assertTrue}
+import zio.{Scope, ZIO, ZLayer}
 
 object ServerClientIntegrationSpec extends ZIOSpecDefault {
 
--- a/zio-http/src/test/scala/zio/http/api/ServerClientIntegrationSpec.scala
+++ b/zio-http/src/test/scala/zio/http/api/ServerClientIntegrationSpec.scala
@@ -1,0 +1,57 @@
+package zio.http.api
+
+import zio.{Scope, ZIO, ZLayer}
+import zio.http.{Client, ClientConfig, Server, ServerConfig, URL}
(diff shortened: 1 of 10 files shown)


zio-http/src/test/scala/zio/http/api/package.scala
```diff
@@ -0,0 +1,7 @@
+package zio.http
+
+import scala.language.implicitConversions
+
+package object api {
+  implicit def stringToIn(s: String): In[Unit] = In.literal(s)
```
jdegoes, line 6: Let's put this into the companion object of `In`, where it will be searched in the event a `String` is encountered but an `In` is expected.

zio-http/src/test/scala/zio/http/api/ServerClientIntegrationSpec.scala
```diff
@@ -0,0 +1,57 @@
...
+  val executorLayer = ZLayer.fromFunction(makeExecutor _)
+
+  def spec =
+    suite("ServerClientIntegrationSpec")(
+      test("server and client integration") {
+        for {
+          _        <- Server.install(usersPostHandler.toHttpApp)
+          _        <- ZIO.debug("Installed server")
+          executor <- ZIO.service[APIExecutor[usersPostAPI.Id]]
+          // QUESTION: Do we want to encode `E` in an API?
```
jdegoes, line 42: I think rather, we want to make sure the handlers can only fail with `HttpError`, and then we can wrap up any fatal error in an `InternalServerError`.

Another possibility would be to have a table in `API` that maps some `E` to `HttpError`. This would have the nice side-effect of allowing us to put error codes in docs if used properly. But probably it wouldn't be used properly since if you're failing with `HttpError` then you would just map the sum type to itself, which would not permit any kind of reasonable documentation.

jdegoes:
@kitlangton Can you take a look at the failures? 🙏

jdegoes (approved):

jdegoes:
@kitlangton Thanks for your work on this! Please do a followup PR to move the implicit string to In conversion. 🙏

kitlangton:
@jdegoes Will do! Sorry for the delay on that.

PR #1563, opened 2022-09-25
Attempt to fix performance regression

vigoo:


jdegoes:
@vigoo Will close as this did not affect performance (as I recall).

PR #1564, opened 2022-09-25
>>>>> FOR RESEARCH PURPOSES ONLY! DO NOT COMMIT! <<<<<

scottweaver:
This pr is comparing the point where performance took a noticeable hit.


scottweaver:
https://github.com/zio/zio-http/pull/1565 makes more sense,

PR #1565, opened 2022-09-25
>>>>>[~300K r/s] FOR RESEARCH PURPOSES ONLY! DO NOT COMMIT! <<<<< 

scottweaver:
This pr is comparing the point where performance took a noticeable hit.


scottweaver:
.

jdegoes:
@scottweaver There is not really anything obvious here. I suspect the best way to figure out what (probably silly) changed caused the regression is to:

1. Profile the `perf-good` branch using JProfiler, YourKit, etc.
2. Profile the current one
3. See what's different in terms of allocations or hot spot methods

Will close this for now.

PR #1581, opened 2022-09-27
feat: Implement structured Accept header

jakubjanecek:
Part of #1524.

### DIFF
--- a/zio-http/src/main/scala/zio/http/model/headers/values/Accept.scala
+++ b/zio-http/src/main/scala/zio/http/model/headers/values/Accept.scala
@@ -1,0 +1,61 @@
+package zio.http.model.headers.values
+
+import zio.http.model.MediaType
+
+import scala.util.Try
+
+/** Accept header value. */
+sealed trait Accept
+
+object Accept {
+
+  /**
+   * The Accept header value one or more MIME types optionally weighed with
+   * quality factor.
+   */
+  final case class AcceptValue(mimeTypes: List[MediaTypeWithQFactor]) extends Accept
+
+  final case class MediaTypeWithQFactor(mediaType: MediaType, qFactor: Option[Double])
+
+  /** The Accept header value is invalid. */
+  case object InvalidAcceptValue extends Accept
+
+  def fromAccept(header: Accept): String = header match {
+    case AcceptValue(mimeTypes) =>
+      mimeTypes.map { case MediaTypeWithQFactor(mime, maybeQFactor) =>
+        s"${mime.fullType}${maybeQFactor.map(qFactor => s";q=$qFactor").getOrElse("")}"
+      }.mkString(", ")
+    case InvalidAcceptValue     => ""
+  }
+
+  def toAccept(value: String): Accept = {
+    val acceptHeaderValues: Array[MediaTypeWithQFactor] = value
+      .split(',')
+      .map(_.trim)
+      .map { subvalue =>
+        val parts = subvalue.split(';')
+        if (parts.length >= 1) {
+          val qFactor = if (parts.length == 2) {
+            val qFactorParts = parts.tail.head.split('=')
+            if (qFactorParts.length == 2) {
+              Try(qFactorParts.tail.head.toDouble).toOption
+            } else None
+          } else None
+
+          MediaType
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/model/headers/values/Accept.scala
```diff
@@ -0,0 +1,57 @@
+package zio.http.model.headers.values
+
+import zio.http.model.MediaType
+
+/** Accept header value. */
+sealed trait Accept
+
+object Accept {
```
Wosin, line 10: Shouldn't we also propagate that to `HeaderInputs` ?

jakubjanecek: Right 🤦‍♂️

jakubjanecek:
I'm getting a `StackOverflowError` in the tests (https://github.com/zio/zio-http/actions/runs/3141493577/jobs/5103992655#step:8:1365) even though it works on my machine.

I also noticed that test `parsing and encoding is symmetrical` fails on `StackOverflowError` irregularly on my machine. The solution was to replace `allMediaTypes` with `(text.all ++ application.all)` (basically making the test data set smaller). It seems like some bug in `zio-test`:

```
java.lang.StackOverflowError
	at zio.test.TestArrow$.$anonfun$run$1(TestArrow.scala:165)
	at zio.test.TestArrow$.attempt(TestArrow.scala:148)
	at zio.test.TestArrow$.run(TestArrow.scala:164)
	at zio.test.TestArrow$.$anonfun$run$1(TestArrow.scala:177)
	at zio.test.TestArrow$.attempt(TestArrow.scala:148)
	at zio.test.TestArrow$.run(TestArrow.scala:164)
	at zio.test.TestArrow$.$anonfun$run$1(TestArrow.scala:177)
	at zio.test.TestArrow$.attempt(TestArrow.scala:148)
	at zio.test.TestArrow$.run(TestArrow.scala:164)
```

Does anyone know what is going on?

zio-http/src/test/scala/zio/http/model/headers/values/AcceptSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+              text.`html`             -> None,
+              application.`xhtml+xml` -> None,
+              application.`xml`       -> Some(0.9),
+              MediaType("*", "*")     -> Some(0.8),
+            ),
+          ),
+      )
+    },
+    test("parsing and encoding is symmetrical") {
+      (text.all ++ application.all).map { mediaType =>
```
Wosin: I think the issue here is that we are doing `reduce` which is handled by a recursive function, so the size of the dataset probably causes that to overflow. I think we can fix that on the `zio-test` side. In the meanwhile, can we map the whole list and then compare results? Something like :
```
val results = allMediaTypes.map(mediaType => Accept.fromAccept(Accept.toAccept(mediaType.fullType)))
assertTrue(allMediaTypes.map(_.fullType) == results) 
```

zio-http/src/main/scala/zio/http/model/headers/values/Accept.scala
```diff
@@ -0,0 +1,59 @@
...
+            .map(_ -> qFactor)
+            .getOrElse {
+              val mediaTypeParts = parts.head.split('/')
+              if (mediaTypeParts.length == 2) {
+                MediaType(mediaTypeParts.head, mediaTypeParts.tail.head) -> qFactor
+              } else null
+            }
+        } else null
+      }
+      .filter(_ != null)
```
Wosin: Shouldn't we just fail if any part of the header is incorrect, instead of filtering out failed parts ?

jakubjanecek: It felt too strict but alright, I've changed it.

Wosin (approved):
Looks good. One thing we might want to replace Tuple in `AcceptValue` with some case class to make it more meaningful.

jakubjanecek:
OK, I've added a `case class`.

jakubjanecek:
This is ready to be merged now.

jdegoes (approved):

Issue #1525, opened 2022-09-19
Create ZIO HTTP TestKit module to help with testing


jdegoes:
**Is your feature request related to a problem? Please describe.**

The current version of ZIO HTTP does not provide a testkit, which means that users are on their own for testing.

**Describe the solution you'd like**

Providing `TestServer` and `TestClient` implementations of the `zio.http.Server` and `zio.http.ZClient` interfaces. These test versions would operate similar to the test implementations of `Console`, `System`, etc., in ZIO core: they would record interactions, and make those interactions accessible; and they would allow feeding responses, errors, or other issues to test the code in question.

**Describe alternatives you've considered**

None.

jdegoes:
@swoogles Thank you! 🙏

tusharmath:
I think it makes sense to have a test client however I am not sure if there is any value in adding a test server.
The server is typically implemented using `Http` which is a ZIO function. One can simply call the function and run the http app directly in the test.

vigoo:
Seems to be done now

PR #1582, opened 2022-09-27
Create TestServer

swoogles:
This addresses part of #1525 
It does not include `TestClient`


adamgfraser (approved):
Would be good to add some documentation but otherwise looks great!

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -0,0 +1,91 @@
...
+      }
+
+    } *> setErrorCallback(errorCallback)
+
+  private def setErrorCallback(errorCallback: Option[ErrorCallback]): UIO[Unit] = {
+    ZIO
+      .environment[Any]
+      .flatMap(_ => driver.setErrorCallback(errorCallback))
+      .unless(errorCallback.isEmpty)
+      .map(_.getOrElse(()))
```
paulpdaniels, line 134: Why is the `ZIO.environment[Any]` accessor necessary? Couldn't this just be:

```scala
driver.setErrorCallback(errorCallback).unless(errorCallback.isEmpty).unit
```

swoogles: Good catch! That was a relic from some code in `Server` that I just copied here.

Made a separate tiny PR fixing it there as well-
https://github.com/zio/zio-http/pull/1605

swoogles:
@adamgfraser  Agreed, added docs!

zio-http/src/main/scala/zio/http/Network.scala
```diff
@@ -0,0 +1,31 @@
+package zio.http
+
+import zio.{Task, ULayer, ZIO, ZLayer}
+
+private[http] trait Network {
+  def findOpenPort: Task[Int]
```
jdegoes, line 6: There is inherently a race condition here, as what is open at one moment in time may not be open the next moment in time. This is why some APIs are allow you to bind to a specific port, or to bind to any port that is available; and then for the latter, you can find out what the port was after the fact.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -0,0 +1,155 @@
+package zio.http
```
jdegoes, line 1: Rather than putting this into `zio-http/src/test`, I think we need a new top-level project, e.g.: `zio-http-testkit`, so a user can rely on that (published) project in their own tests.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -0,0 +1,155 @@
...
+      .setErrorCallback(errorCallback)
+      .unless(errorCallback.isEmpty)
+      .map(_.getOrElse(()))
+
+  override def port: Int = bindPort
+}
+
+object TestServer {
+
+  val make: ZIO[Driver with Scope, Throwable, TestServer[Unit]] =
```
jdegoes, line 141: It would be more common for this to be called `layer` rather than `make`, I think.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -0,0 +1,155 @@
...
+ * @param behavior
+ *   Describes how the Server should behave during your test
+ * @param driver
+ *   The web driver that accepts our Server behavior
+ * @param bindPort
+ *   Port for HTTP interactions
+ * @tparam State
+ *   The type of state that will be mutated
+ */
+final case class TestServer[State](
```
jdegoes, line 21: I am not sure if I would bake `State` in, as opposed to allowing the handlers to use ZIO effects.

If the handlers can return ZIO effects (containing responses), then they can already be stateful, without complicating the design of `TestServer`.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @example
+   *   {{{ ZIO.serviceWithZIO[TestServer[Unit]]( _.addRequestResponse(
+   *   Request.get(url = URL.root.setPort(port = ???)), Response(Status.Ok)
+   *   ).install ) }}}
+   *
+   * @return
+   *   The TestSever with new behavior.
+   */
+  def addRequestResponse(
```
jdegoes, line 44: It might be useful to allow the user to perform `assertTrue` on the request, and to generate the response from the request.

Also, I can see these functions return new `TestServer`. This is not consistent with `TestClock`, `TestSystem`, `TestRandom`, etc. 

See, for example, `TestConsole.feedLines` and friends, which modify internal state rather than return a new test construct. I think all of these methods should **not** return a new `TestServer`, but rather, should modify internal state, to be consistent with other test control structures.

swoogles: Rodger. An earlier version of this used this approach, but then we ended up changing to this style at the last minute. I will rework it to be more in line with those other services.

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Corrected in my current branch

zio-http/src/main/scala/zio/http/Network.scala
swoogles: Good catch, I actually wondered about that at some point during development. Resolved on my branch.

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Removed. Unsurprisingly - things are much simpler now!

PR #1593, opened 2022-09-29
Add Allow header

ccerbusca:


zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala
```diff
@@ -0,0 +1,35 @@
...
+import zio.Chunk
+import zio.http.model.Method
+
+import scala.annotation.tailrec
+
+/**
+ * The Allow header must be sent if the server responds with a 405 Method Not Allowed status code to indicate
+ * which request methods can be used.
+ */
+final case class Allow(methods: Chunk[Method])
```
devsprint: @ccerbusca you have to define the values as ADT. `Method.fromString` is not failing but the return value in case of invalid method is not usable here, as it is not part oh Http standard.

ccerbusca: In case the list of request methods contains an invalid method, should the element be ignored or the whole list invalidated?

devsprint: no, I should have a `InvalidAllowMethod` value.

```
sealed trait Allow
  object Allow {
     case object Get extends Allow
     case object InvalidAllowMethod extends Allow
     .... 
  }
 ```

Comment (ccerbusca) [reply zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala]: Oh, i see. Makes sense

Comment (Wosin) [reply zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala]: I'm not really sure why we can't use `Method`. The RFC does not define the list of accepted methods AFAIK and we have `CustomMethod` defined. I think that pretty much You can have anything that matches `[a-zA-Z0-9]` and it should produce correct `Allow` header.

Comment (Wosin) [reply zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala]: E.g. Http4s will allow `DOG, CAT, TEST, 434,` to be correct `Allow` header.

File: zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala [hunk 2]
```diff
@@ -0,0 +1,36 @@
...
+
+object Allow {
+
+  def toAllow(value: String): Allow = {
+    @tailrec def loop(index: Int, value: String, acc: Allow): Allow = {
```
Wosin: Not really sure if we actually want to use recursion here, sounds like an overkill to parse headers to be honest.

zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala
devsprint: well, if http4s is doing that, that doesn't mean zio-http has to follow. It doesn't make any sense to allow any other values than Http method names.

Wosin: That's correct we don't need to follow http4s, that was just an example. Even if we don't want to allow anything apart from standard set of HTTP methods, I'm not sure why we want to model them as separate ADT instead of doing `Method.fromString` and failing if we get `CustomMethod` as a result.

zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala
ccerbusca: Replaced it with built-in functions

zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala
ccerbusca: I refactored the implementation to use an ADT. We could discuss further if you think going back to using Method.fromString is the better solution

devsprint: Looks good.

zio-http/src/main/scala/zio/http/model/headers/values/Allow.scala
devsprint: recursion was ok, split and fold are allocating memory

ccerbusca: reverted the implementation back to the recursive version

jdegoes (approved):

PR #1602, opened 2022-10-03
Restrict values for `OpenAPI.Link` parameters/body & add json rendering

987Nabil:
@jdegoes I checked on the specs again, and I think the right type is `Literal | Expression`.
Expressions have their own syntax that could be checked (maybe with a regex), but it seems quite complicated and I am not sure if checking it is necessary.


jdegoes (approved):

jdegoes:
@987Nabil This should pave the way for type-safe rendering!

PR #1606, opened 2022-10-04
CommonMark rendering for Doc

987Nabil:
@jdegoes I am not sure about the doc structure in general. We have no options for ordered lists and unordered list elements are docs and docs can't be text but only paragraphs. I think this needs an overhaul in a new ticket.


jdegoes:
@987Nabil Agreed. If you have the time and interest, please create a new ticket with these suggested changes! 🙏

jdegoes (approved):

PR #1610, opened 2022-10-04
Status.fromInt method added, EHttpApp type alias

yisraelU:
map integer to Http Status
type alias similar to RHttpApp however R is fixed to Any


jdegoes (approved):

jdegoes:
🙏

PR #1612, opened 2022-10-05
Rework test server

swoogles:
@jdegoes 
This resolves everything mentioned after https://github.com/zio/zio-http/pull/1582 was merged, aside from defining a `testkit` project, which I'd like to tackle separately.


zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -34,80 +29,74 @@ final case class TestServer[State](
...
+   *      }
+   *    }
    *   }}}
    */
-  def addHandler(pf: PartialFunction[Request, Response]): TestServer[State] = {
-    val handler: PartialFunction[(State, Request), (State, Response)] = {
-      case (state, request) if pf.isDefinedAt(request) => (state, pf(request))
-    }
-    addHandlerState(handler)
+  def addHandler(
```
jdegoes: You can make this more polymorphic:

```suggestion
  def addHandler[R, E](
      pf: PartialFunction[Request, ZIO[R, E, Response]],
  ): ZIO[R, E, Unit] = {
```

jdegoes: Hmm, well, you can do that for `R`, but not for `E`.

The most polymorphic you can do is:

```suggestion
def addHandler[R](
      pf: PartialFunction[Request, ZIO[R, Throwable, Response]],
  ): ZIO[R, Nothing, Unit] = {
```

assuming we can fail with `Throwable` in the handler.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -7,22 +7,17 @@ import zio.http.Server.ErrorCallback
...
- *   The type of state that will be mutated
  */
-final case class TestServer[State](
-  state: Ref[State],
-  behavior: PartialFunction[(State, Request), (State, Response)],
+final case class TestServer(
+  behavior: Ref[PartialFunction[Request, ZIO[Any, Nothing, Response]]],
   driver: Driver,
-  bindPort: Int,
+  bindPort: RuntimeFlags,
```
adamgfraser: Shouldn't be `RuntimeFlags`.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -34,80 +29,76 @@ final case class TestServer[State](
...
    *   The TestSever with new behavior.
    */
   def addRequestResponse(
     expectedRequest: Request,
     response: Response,
-  ): TestServer[State] = {
-    val handler: PartialFunction[(State, Request), (State, Response)] = {
-      case (state, realRequest) if {
+  ): ZIO[Any, Nothing, Unit] = {
+    val handler: PartialFunction[Request, ZIO[Any, Nothing, Response]] = {
```
adamgfraser, line 43: Can we generalize the error type to `Throwable` here?

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -34,80 +29,76 @@ final case class TestServer[State](
...
    */
-  def addHandler(pf: PartialFunction[Request, Response]): TestServer[State] = {
-    val handler: PartialFunction[(State, Request), (State, Response)] = {
-      case (state, request) if pf.isDefinedAt(request) => (state, pf(request))
-    }
-    addHandlerState(handler)
+def addHandler[R](
+      pf: PartialFunction[Request, ZIO[R, Throwable, Response]],
+  ): ZIO[R, Nothing, Unit] = {
+    pf: PartialFunction[Request, ZIO[Any, Nothing, Response]],
```
adamgfraser: I think duplicate lines here.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -7,22 +7,17 @@ import zio.http.Server.ErrorCallback
...
  * @param bindPort
  *   Port for HTTP interactions
- * @tparam State
- *   The type of state that will be mutated
  */
-final case class TestServer[State](
-  state: Ref[State],
-  behavior: PartialFunction[(State, Request), (State, Response)],
+final case class TestServer(
+  behavior: Ref[PartialFunction[Request, ZIO[Any, Nothing, Response]]],
```
adamgfraser: `Throwable`

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -7,22 +7,17 @@ import zio.http.Server.ErrorCallback
...
  * @param behavior
  *   Describes how the Server should behave during your test
  * @param driver
  *   The web driver that accepts our Server behavior
  * @param bindPort
  *   Port for HTTP interactions
- * @tparam State
- *   The type of state that will be mutated
  */
-final case class TestServer[State](
```
adamgfraser, line 21: Need to move this to new package so it is accessible to users.

zio-http/src/test/scala/zio/http/TestServer.scala
```diff
@@ -137,19 +125,26 @@ final case class TestServer[State](
...
-  val make: ZIO[Driver with Scope, Throwable, TestServer[Unit]] =
-    make(())
+  def addRequestResponse(
+    request: Request,
+    response: Response,
+  ): ZIO[TestServer, Nothing, Unit] =
+    ZIO.serviceWithZIO[TestServer](_.addRequestResponse(request, response))
 
-  def make[State](initial: State): ZIO[Driver with Scope, Throwable, TestServer[State]] =
+  val layer: ZIO[Driver with Scope, Throwable, TestServer] =
```
jdegoes: Is there a way we can return a layer here, rather than a ZIO effect?

adamgfraser: Yes this should just be `ZLayer.scoped`.

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Bah, I knew that Intellij would incorrectly suggest `RuntimeFlags` in place of `Int` when generating type annotations, but this is a new manifestation of that issue-

If you use the "Change Signature" refactoring and modify a parameter, it will automatically "fix" any existing `Int` parameter by changing it to `RuntimeFlags`.

https://user-images.githubusercontent.com/2054940/194116555-188129d9-4078-4f06-a078-a92a04971ca5.mov

Looks like I triggered that when I was removing `State` from `TestServer`

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Yep, accepting John's suggestion via the Github UI went awry here.
Fixed.

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Fixed.

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Fixed

swoogles:
Removed `TestServer.behavior`, since we don't actually need to track behavior changes inside `TestServer`. 

That is handled strictly by deploying new apps to the `Driver` each time we call `TestServer.add*`

zio-http/src/test/scala/zio/http/TestServer.scala
swoogles: Will be done in the next project/packaging bit of work.

adamgfraser (approved):

PR #1614, opened 2022-10-06
Commit for Improve headers inputs: Connection

wpoosanguansit:
This follows the prior commits for other header inputs in the repo.  The value for connection header other than close and keep-alive is treated as Invalid which is an empty string "".


Wosin (approved):
Looks good.

zio-http/src/test/scala/zio/http/internal/HttpGen.scala
```diff
@@ -267,4 +267,13 @@ object HttpGen {
 
   def allowHeader: Gen[Any, Allow] =
     Gen.chunkOfBounded(1, 9)(allowHeaderSingleValue).map(Allow.AllowMethods.apply)
+
+  def connectionHeader: Gen[Any, Connection] =
+    Gen.fromIterable(
```
adamgfraser: This will create a deterministic generator of these three values. I think we want a random generator instead so we should use `Gen.elements` here.

wpoosanguansit: Got it.  It is changed to elements.

adamgfraser: Doesn't look like it is changed.

adamgfraser (approved):
Thanks for your work on this!

jdegoes:
@wpoosanguansit Great work! Can you fix the conflicts and re-open the pull request? 

Thank you!

wpoosanguansit:
Thank you @jdegoes.  Sorry for being a newbie.  Where can I check to see what needs to be fixed to re-open a new pull request?

Issue #1559, opened 2022-09-23
Multipart form data server support


fokot:
**Is your feature request related to a problem? Please describe.**
Server can accept and parse Multipart form data requests.

**Describe the solution you'd like**
Something nice

**Describe alternatives you've considered**
We can either use https://netty.io/4.0/api/io/netty/handler/codec/http/multipart/InterfaceHttpPostRequestDecoder.html or parse it ourselves in ZIO streams.

fokot:
It looks it needs to be added here https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala#L124

fokot:
I will work on it

tusharmath:
There are two ways we can implement it —

1. Using Netty's multipart-decoder
2. Using ZIO Parser

We can also decode a FullHttpRequest or an HttpRequest where the body is a stream.

In my opinion, we should move fast on this issue by implementing a non-streaming solution first and decoding it with netty's multipart decoder. That will perform the decoding in memory at once. So it won't work for very large files, but at least build some multipart support in the library.

Tasselmi:
hi, my friends. any update?

PR #1617, opened 2022-10-09
Multipart #1559

fokot:
I created this PR, just to have some input how to solve it. It is still WIP.

Is this a good approach to use netty's `HttpPostMultipartRequestDecoder`? If so I need to implement it also in for `addAsyncBodyHandler`. Should I rather use my own sealed trait instead of netty's `InterfaceHttpData` in the body? Thiss way when we change netty to something else the api will stay the same.

If this approach is not best I can do manual parsing on the stream as they did it in [http4s](https://github.com/http4s/http4s/tree/main/core/shared/src/main/scala/org/http4s/multipart) but it is much more work. Plugging this in will be easier though as I could use just `Body.asStream`.

What do you think?

And I will delete the bigfile.txt.

### DIFF
--- a/zio-http-example/src/main/scala/example/MultipartServer.scala
+++ b/zio-http-example/src/main/scala/example/MultipartServer.scala
@@ -1,0 +1,30 @@
+package example
+
+import zio.http._
+import zio.http.model.Multipart.FileUpload
+import zio.stream.ZPipeline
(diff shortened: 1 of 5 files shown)


jdegoes:
@fokot I'm going to close as this has bit-rotted, but please re-open if you get a chance to work on this!

PR #1629, opened 2022-10-14
Add Host header value.

Wosin:


jdegoes:
Try running `fmt` from `root` to get past `scalafix`

jdegoes:
Scalafix still failing:

<https://github.com/zio/zio-http/actions/runs/3250814821/jobs/5335254150#step:6:447>

Wosin:
I think that should be fine now, the build fails locally with `RouteCodec` issues, but fmt seems to be fine.

jdegoes (approved):

PR #1642, opened 2022-10-16
Continue refactoring Middleware

afsalthaj:
* Store MiddlewareSpec in Middleware
* Add `Concat` node in Middleware (similar to MiddlewareSpec), making it consistent to Service having `Concat` node (similar to the one in API).
* Change name from `handle` to `implement` for both `API` and `MiddlewareSpec`
* Be more specific about RequestType and ResponseType in CodecTypes
* Added a draft `Http => Http` implementation given a `Middleware` (the new one). It works with composed `MiddlewareSpec`. Check BasicAuthAPIExample. Currently it depends on existing `Middelware` than explicitly using the function, to avoid having to reimplement `interceptZIO`, however pretty easy to make it `HttpApp => HttpApp`

cc @jdegoes


zio-http/src/main/scala/zio/http/api/API.scala
```diff
@@ -115,6 +116,23 @@ final case class API[Input, Output](
...
 
+  /**
+   * Add a middleware spec into the API. Note that once you add the middleware,
+   * it returns a ServiceSpec (instead of API) disallowing it compose with any
+   * other APIs.
+   *
+   * If you have more than 1 API endpoints, compose them all together using
+   * `++`, and add `MiddlewareSpec` once and for all.
+   */
+  def middleware[MI, MO](middlewareSpec: MiddlewareSpec[MI, MO]): ServiceSpec[MI, MO, Id] =
```
jdegoes: I don't think this is common enough to warrant its existence, since you can just call `.serviceSpec.middleware(...)` to get the same thing (a single API with middleware).

afsalthaj: Sure.

afsalthaj: Makes sense

zio-http/src/main/scala/zio/http/api/API.scala
```diff
@@ -115,6 +116,23 @@ final case class API[Input, Output](
...
+    ServiceSpec(self).middleware(middlewareSpec)
+
+  /**
+   * Convert API to a ServiceSpec.
+   */
+  def serviceSpec: ServiceSpec[Unit, Unit, Id] =
```
jdegoes: Let's call this one `toServiceSpec` to match other ZIO naming conventions.

afsalthaj: Sure.

zio-http/src/main/scala/zio/http/api/MiddlewareSpec.scala
```diff
@@ -16,6 +17,9 @@ final case class MiddlewareSpec[MiddlewareIn, MiddlewareOut](
   ): MiddlewareSpec[inCombiner.Out, outCombiner.Out] =
     MiddlewareSpec(self.middlewareIn ++ that.middlewareIn, self.middlewareOut ++ that.middlewareOut)
 
+  def handle[R, E](f: MiddlewareIn => ZIO[R, E, MiddlewareOut]): Middleware[R, E, MiddlewareIn, MiddlewareOut] =
```
jdegoes: I like this for symmetry. Maybe they should both be called `implement`. 

e.g.

```scala
myApi.implement(...)

myMiddlewareSpec.implement(...)
```

The weird thing is that `Middleware` doesn't store a `MiddlewareSpec`, while `Service` stores an `API`. Maybe that should change?

jdegoes: OTOH `handle` is shorter even if its use for middleware is a bit awkward.

zio-http/src/main/scala/zio/http/api/MiddlewareSpec.scala
```diff
@@ -57,11 +61,14 @@ object MiddlewareSpec {
   def addHeader(key: String, value: String): MiddlewareSpec[Unit, Unit] =
     MiddlewareSpec(HttpCodec.empty, HeaderCodec.header(key, TextCodec.constant(value)))
 
-  val auth: MiddlewareSpec[Auth.Credentials, Unit] =
+  def basicAuth(u: String, p: String): MiddlewareSpec[Auth.Credentials, Unit] =
```
jdegoes: I don't suggest we change this. In general, authentication is not done by passing hard-coded username and passpord, but by looking up information in a database (or at least config). I think we should revert this.

Also, this middleware spec is doing more than just being a `spec`: it is actually providing an implementation. That's an anti-pattern. The only point of a middleware spec is to define the (1) requirements on the request, and (2) the addendums to the response. No middleware logic should appear inside a `spec`. It must only ever appear inside the implementation for the middleware.

afsalthaj:
I think Middleware could store MiddlewareSpec and probably that may help us
implement toHttp a bit more easily. May be I am wrong.

afsalthaj:
Does this mean we shouldn’t really allow user to specific a logic through
transformOrFail in MiddlewareSpec?

zio-http/src/main/scala/zio/http/api/MiddlewareSpec.scala
afsalthaj: Probably will use `handle` for Service and `implement` for middleware.?

zio-http/src/main/scala/zio/http/api/MiddlewareSpec.scala
afsalthaj: >In general, authentication is not done by passing hard-coded username and passpord, but by looking up information in a database (or at least config). I think we should revert this.


I did this just trying to reimplement an existing `basicAuth` in the library.

Have a look at `Auth.scala` 

```scala
  /**
   * Creates a middleware for basic authentication that checks if the
   * credentials are same as the ones given
  final def basicAuth(u: String, p: String)(implicit trace: Trace): HttpMiddleware[Any, Nothing] =
    basicAuth { case credentials => (credentials.uname == u) && (credentials.upassword == p) }
```

But this can question our overly powerful `transformOrFail` in `MiddlewareSpec`.
The way it can get abused many `middlwareIn` may get _implemented_ as part of transformOrFail. In other words, boundary between transformOrFail in `MiddlewareSpec` and handlers in new `Middleware` is slightly blurred (for a user)

But may be its ok and leave it to the user to decide.

afsalthaj:
@jdegoes  I think Middleware need to store MiddlewareSpec , otherwise it is difficult to lookup the handler for a spec while looping for `toHttpApp` in MiddlewareSpec

zio-http/src/main/scala/zio/http/api/MiddlewareSpec.scala
jdegoes: All right, I suppose that's intended for a "quick & dirty" REST API. We can have that in the middleware, but let's leave the existing one there, and implement the new one in terms of the existing one. In any case, the logic should not be contained in the middleware.

I think you are right that `TransformOrFail` is maybe too powerful because it will encourage users to perform checks inside the specification, rather than inside the implementation. I think we are only using `transform` now for all the headers, so maybe we can get away with simplifying it.

jdegoes:
Ok lets have `Middleware` store `MiddlewareSpec`.

zio-http/src/main/scala/zio/http/api/ServiceSpec.scala
```diff
@@ -61,8 +61,61 @@ object ServiceSpec                        {
...
+    // FIXME: this is only WIP
+    self match {
+      case Empty                                   => MiddlewareSpec.none
+      case Single(_)                               => MiddlewareSpec.none
+      case Concat(left, right)                     => middlewareSpecOf(left).++(middlewareSpecOf(right))
+      case AddMiddleware(_, middlewareSpec0, _, _) => middlewareSpec0
+    }
+  }
+
+  def toHttpMiddleware[R, E, I, O](
```
afsalthaj, line 74: This is still WIP

PR #1644, opened 2022-10-16
Commit code for #1524 - Improve headers inputs for Expires.

wpoosanguansit:
A simple code commit for Expires headers with ValidExpires and InvalidExpires.

### DIFF
--- a/zio-http/src/main/scala/zio/http/model/headers/values/Expires.scala
+++ b/zio-http/src/main/scala/zio/http/model/headers/values/Expires.scala
@@ -1,10 +1,10 @@
 package zio.http.model.headers.values
 
-import java.time.ZonedDateTime
+import java.time._
 import java.time.format.DateTimeFormatter
 
 sealed trait Expires {
-  def value: String
+  def value: ZonedDateTime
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/model/headers/values/Expires.scala
```diff
@@ -0,0 +1,44 @@
+package zio.http.model.headers.values
+
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+
+sealed trait Expires {
+  def value: String
```
Wosin: I think we should represent that as `LocalDateTime` or any other date type to make it easier to work with the header

wpoosanguansit: I did have LocalDateTime at first but it would lose the timezone info when we encode and decode.  I am assuming the clients will be using this from toExpires and fromExpires, so I changed the value to String.  Should I change it to ZonedDateTime?

jdegoes (approved):

PR #1645, opened 2022-10-16
Commit code for #1524 - Adding Header AccessControlMaxAge.

wpoosanguansit:
Simple commit for AccessControlMaxAge that defaults to 5 seconds.

### DIFF
--- a/zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
+++ b/zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
@@ -1,8 +1,7 @@
 package zio.http.model.headers.values
 
-import scala.util.Try
-
 import java.time.Duration
+import scala.util.Try
 
 /**
  * The Access-Control-Max-Age response header indicates how long the results of
--- a/zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
+++ b/zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
@@ -15,7 +15,7 @@ import scala.util.Try
  * caps at 2 hours (7200 seconds). The default value is 5 seconds.
  */
 sealed trait AccessControlMaxAge {
-  val seconds: String
+  val seconds: Duration
 }
 
 object AccessControlMaxAge {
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
```diff
@@ -0,0 +1,42 @@
...
+ * Access-Control-Allow-Methods and Access-Control-Allow-Headers headers) can be
+ * cached.
+ *
+ * Maximum number of seconds the results can be cached, as an unsigned
+ * non-negative integer. Firefox caps this at 24 hours (86400 seconds). Chromium
+ * (prior to v76) caps at 10 minutes (600 seconds). Chromium (starting in v76)
+ * caps at 2 hours (7200 seconds). The default value is 5 seconds.
+ */
+sealed trait AccessControlMaxAge {
+  val seconds: String
```
Wosin: Wouldn't it make more sense to have this as `Long` or `Duration` ?

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  def fromAccessControlMaxAge(accessControlMaxAge: AccessControlMaxAge): String = {
+    accessControlMaxAge.seconds
+  }
+
+  def toAccessControlMaxAge(seconds: String): AccessControlMaxAge = {
+    Try(seconds.toLong).fold(
+      _ => ValidAccessControlMaxAge(),
+      long => if (long > 0) ValidAccessControlMaxAge(Duration(long, SECONDS)) else ValidAccessControlMaxAge(),
```
Wosin: Should we have a type to represent incorrect header value (like negative numbers) ??

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
wpoosanguansit: I will change it to Duration.

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
wpoosanguansit: I did have that but I will have that in again as I think it align closely with the spec, which says it is default to 5 seconds.

jdegoes (approved):

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
```diff
@@ -0,0 +1,46 @@
...
+  val seconds: Duration
+}
+
+object AccessControlMaxAge {
+
+  /**
+   * Valid AccessControlMaxAge with an unsigned non-negative negative for
+   * seconds
+   */
+  final case class ValidAccessControlMaxAge(private val duration: FiniteDuration = Duration(5, SECONDS))
```
jdegoes: This should not used Scala duration, only `java.time.Duration`.

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
```diff
@@ -0,0 +1,46 @@
...
+      extends AccessControlMaxAge {
+    override val seconds: Duration = duration
+  }
+
+  case object InValidAccessControlMaxAge extends AccessControlMaxAge {
```
jdegoes:
```suggestion
  case object InvalidAccessControlMaxAge extends AccessControlMaxAge {
```

`Invalid` is one word.

jdegoes (changes requested):
Just the one change then good to merge.

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
wpoosanguansit: Updated.

zio-http/src/main/scala/zio/http/model/headers/values/AccessControlMaxAge.scala
wpoosanguansit: Updated.

jdegoes:
Fails due to [compilation error](https://github.com/zio/zio-http/actions/runs/3265979566/jobs/5369042470#step:8:316)

wpoosanguansit:
Update the call toSeconds to getSeconds() as toSeconds is not found in java8.

jdegoes (approved):

PR #1662, opened 2022-10-19
Remove the complex compile time errors and type signatures, keeping typesafety and type inference

afsalthaj:
This is mainly trying to address @kitlangton 's comment which is as follows

image

The commit tries to fix this step by step, and the first step forward is to remove the type parameter Id, and yet trying to keep type-safety (and type inference) both at server side and client side.

The core change is

image

We will still have compile time errors if you don't pass all the endpoint implementations corresponding to the `EndpointSpec` (naming is yet to be improved, I think)

Similarly we will still have compile time errors if you pass a different `type` of Input to the executor in the client side.

I couldn't find any type inference issues, since all examples worked as is!

## The new compile time error

We still have a verbose compile time error  but it doesn't point to `Id` and type memberish jargons. In the below given example, it says you have implemented only `zio.http.api.EndpointSpec[(Int, Int, String),Unit]` and we need `zio.http.api.EndpointSpec[(Int, Int, String),Unit],R,E]` as well. Verbose, but better in my opinion, and we could further improve.

```scala
[error] /Users/afsalthaj/redesign/zio-http/zio-http-example/src/main/scala/example/APIExamples.scala:40:25:
 inferred type arguments[zio.http.api.APIExamples.getUsers.type,Any,Nothing]
  do not conform to method toHttpApp's type parameter bounds 
  [EndPointSpecs1 <: zio.http.api.APIExamples.getUsers.type with zio.http.api.APIExamples.getUser1.type,R,E]

````

### DIFF
--- a/zio-http/src/main/scala/zio/http/api/APIExecutor.scala
+++ b/zio-http/src/main/scala/zio/http/api/APIExecutor.scala
@@ -11,17 +11,17 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
  * failing with some kind of RPC error.
  */
 trait APIExecutor[+MI, +MO, +Ids] { self =>
-  def apply[Id, A, B](
-    invocation: Invocation[Id, A, B],
-  )(implicit ev: Ids <:< Id, trace: Trace): ZIO[Any, Throwable, B]
+  def apply[A, B](
+    invocation: Invocation[A, B],
+  )(implicit trace: Trace): ZIO[Any, Throwable, B]
 
   def middlewareInput(implicit trace: Trace): Task[MI]
 
   def mapMiddlewareInput[MI2](f: MI => MI2): APIExecutor[MI2, MO, Ids] =
     new APIExecutor[MI2, MO, Ids] {
-      def apply[Id, A, B](
-        invocation: Invocation[Id, A, B],
-      )(implicit ev: Ids <:< Id, trace: Trace): ZIO[Any, Throwable, B] =
+      def apply[A, B](
+        invocation: Invocation[A, B],
+      )(implicit trace: Trace): ZIO[Any, Throwable, B] =
         self.apply(invocation)
 
       def middlewareInput(implicit trace: Trace): Task[MI2] = self.middlewareInput.map(f)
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by jdegoes** [comment]
This is not equivalent: there can be many endpoints with exactly the same type signature.

**Comment by afsalthaj** [comment]
@jdegoes 
> This is not equivalent: there can be many endpoints with exactly the same type signature.

Just added 1 more commit addressing the issue using the singleton type you were mentioning.
https://github.com/zio/zio-http/pull/1662/commits/2bcefb74d0df51125e4c2d6ca56bd2f743f5c3c8

File: zio-http/src/main/scala/zio/http/api/APIExecutor.scala [hunk 1]
```diff
@@ -11,17 +11,17 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
...
   def middlewareInput(implicit trace: Trace): Task[MI]
 
   def mapMiddlewareInput[MI2](f: MI => MI2): APIExecutor[MI2, MO, Ids] =
     new APIExecutor[MI2, MO, Ids] {
-      def apply[Id, A, B](
-        invocation: Invocation[Id, A, B],
-      )(implicit ev: Ids <:< Id, trace: Trace): ZIO[Any, Throwable, B] =
+      def apply[A, B](
+        invocation: Invocation[A, B],
+      )(implicit trace: Trace): ZIO[Any, Throwable, B] =
```
jdegoes: This is now completely type unsafe

zio-http/src/main/scala/zio/http/api/APIExecutor.scala
```diff
@@ -11,17 +11,17 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
  * failing with some kind of RPC error.
  */
 trait APIExecutor[+MI, +MO, +Ids] { self =>
-  def apply[Id, A, B](
-    invocation: Invocation[Id, A, B],
-  )(implicit ev: Ids <:< Id, trace: Trace): ZIO[Any, Throwable, B]
+  def apply[A, B](
+    invocation: Invocation[A, B],
+  )(implicit trace: Trace): ZIO[Any, Throwable, B]
```
jdegoes: Type unsafe, allowing all invocations to be executed

zio-http/src/main/scala/zio/http/api/APIExecutor.scala
afsalthaj: Will revert this change.

afsalthaj:
Closing this in favour of https://github.com/zio/zio-http/pull/1664

PR #1663, opened 2022-10-19
fix(hexit): add lazy evaluation on defaultwith and orelse

ex0ns:
Following a discussion on Discord, the original case was a bit more convoluted:

```scala
val app1 =
    Http.collectZIO[Request] {
      case Method.GET -> !! / "hello" / name  => 
        ZIO.succeed(Response.text(s"Hi ${name}!"))
    }

class ThisCase:
  def unapply(req:Request): Option[Int] =
    println("unapply!")
    Some(2)
 
val thisCase1 = ThisCase()
val app2 =  Http.collect[Request] { case thisCase1(n) => Response.ok }
val app = app1 ++ app2
```

This was in fact caused by the `execute` method of the `Http` class and more precisely the `combine` between one effectful app (`collectZIO` and any other app).

### DIFF
--- a/zio-http/src/test/scala/zio/http/HExitSpec.scala
+++ b/zio-http/src/test/scala/zio/http/HExitSpec.scala
@@ -3,7 +3,7 @@ package zio.http
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
-import zio.{ZIO, durationInt}
+import zio.{Ref, ZIO, durationInt}
 
 object HExitSpec extends ZIOSpecDefault with HExitAssertion {
   def spec: Spec[Environment, Any] = {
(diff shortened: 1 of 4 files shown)


jdegoes:
I really like these changes. Can you check benchmarks before / after to see what (if any) the impact is?

ex0ns:
Yes I will do it, but it seems that benchmark-zio-http.sh is currently broken (on Linux, because of the `-i` flag in `sed`) but also because it looks for a `build.sbt.base` in the Framework repo that does not seem to be here.
I'll post the result (and fixes to the script) as soon as I have managed to get them running (and maybe make it compatible with a local version of zio-http because as of now it will look for a remote commit)

zio-http/src/test/scala/zio/http/HExitSpec.scala
```diff
@@ -3,7 +3,7 @@ package zio.http
 import zio.test.Assertion._
 import zio.test.TestAspect._
 import zio.test._
-import zio.{ZIO, durationInt}
+import zio.{Ref, ZIO, durationInt}
```
ex0ns: Ref is unused, need to remove

ex0ns:
I've run the benchmarks using the script available at the root of this project, but I'm unsure about the results:

```
This PR:
::set-output name=request_result::Requests/sec: 627528.94
::set-output name=concurrency_result::Concurrency: 256
::set-output name=request_result::Requests/sec: 631153.29
::set-output name=concurrency_result::Concurrency: 1024
::set-output name=request_result::Requests/sec: 623639.56
::set-output name=concurrency_result::Concurrency: 4096
::set-output name=request_result::Requests/sec: 510371.08
::set-output name=concurrency_result::Concurrency: 16384

Commit ecad015252daa5623a7dd72e85d6960fc718e2dc (head/main)
::set-output name=request_result::Requests/sec: 631441.28
::set-output name=concurrency_result::Concurrency: 256
::set-output name=request_result::Requests/sec: 635530.30
::set-output name=concurrency_result::Concurrency: 1024
::set-output name=request_result::Requests/sec: 636401.91
::set-output name=concurrency_result::Concurrency: 4096
::set-output name=request_result::Requests/sec: 544178.02
::set-output name=concurrency_result::Concurrency: 16384
```

I'm not sure that those benchmark are even relevant for the changes made in this PR (as they don't use any kind of HExit.effect as far as I could tell). Are there other benchmarks that I could have miss ?

ex0ns:
Why was this close ? Is it taken care of in another PR ?

PR #1664, opened 2022-10-19
Use singleton type over type member

afsalthaj:


zio-http/src/main/scala/zio/http/api/ServiceSpec.scala
```diff
@@ -48,8 +48,8 @@ object ServiceSpec                        {
     mo: Combiner.WithOut[MO1, MO2, MO3],
   ) extends ServiceSpec[MI3, MO3, AllIds]
 
-  def apply[A <: EndpointSpec[_, _]](api: A): ServiceSpec[Unit, Unit, api.Id] =
-    Single(api.asInstanceOf[EndpointSpec.WithId[api.Id, Any, Any]])
+  def apply[Id](api: EndpointSpec[_, _]): ServiceSpec[Unit, Unit, Id] =
```
jdegoes: This is unconstrained phantom type and therefore unsafe: can you do api.type in return value instead??

jdegoes:
```suggestion
  def apply(api: EndpointSpec[_, _]): ServiceSpec[Unit, Unit, api.type] =
```

afsalthaj: 👍

afsalthaj: @jdegoes  Done

afsalthaj: The only remaining thing here is to reduce the lengthy compile time error (its better than before IMO) using macros.

afsalthaj: Hopefully that could be a separate PR from Kit

jdegoes:
@afsalthaj Can you fix conflicts?

afsalthaj:
Sure

afsalthaj:
@jdegoes  Done!

PR #1665, opened 2022-10-19
Add ETag Header.

Wosin:


zio-http/src/test/scala/zio/http/model/headers/ETagSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import zio.Scope
+import zio.http.model.headers.values.ETag
+import zio.http.model.headers.values.ETag.{InvalidETagValue, StrongETagValue, WeakETagValue}
+import zio.test._
+
+object ETagSpec extends ZIOSpecDefault {
+  override def spec: Spec[TestEnvironment with Scope, Any] = suite("ETag header suite")(
+    test("parse ETag header") {
+      assertTrue(ETag.toETag("""W/"testEtag"""") == WeakETagValue("testEtag"))
```
fsvehla, line 11: assertions need to be combined with `&&` or only the last one will be executed.

jdegoes (approved):

PR #1674, opened 2022-10-21
Support for server header

geoffjohn11:


zio-http/src/main/scala/zio/http/model/headers/values/Server.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * No server name
+     */
+    object EmptyServerValue extends Server
+
+    def fromServer(server: Server): String =
+      server match {
+        case ServerValue(name) => name
+        case EmptyServerValue => ""
```
jdegoes: Is this empty, or is it invalid?

zio-http/src/main/scala/zio/http/model/headers/values/Server.scala
```diff
@@ -0,0 +1,32 @@
...
+ * Server header value.
+ */
+sealed trait Server
+
+  object Server {
+    
+    /**
+     * A server value with a name
+     */
+    final case class ServerValue(name: String) extends Server
```
jdegoes: Maybe call this `Name`?

zio-http/src/main/scala/zio/http/model/headers/values/Server.scala
geoffjohn11: In the MDN [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server) the value for server header is a name value with type String.  Some response look like "Apache/2.4.1 (Unix)" but twitter.com had a server header response value that was simply "tsa_b".  Therefore in the implementation any non-empty string value is accepted, and anything else is considered empty.  I'm not sure what a non-empty invalid server value would be, so an Invalid state wasn't implemented.  Any thoughts?

zio-http/src/main/scala/zio/http/model/headers/values/Server.scala
geoffjohn11: I changed it to ServerName, is your preference for "Name"?

PR #1677, opened 2022-10-23
Fix bug in outgoing in api.Middleware.apply, fix HttpCodec.Empty, and add  Optional Codec for Query and Header

afsalthaj:
* Fix bug in Middleware.apply outgoing function 
* Make sure HttpCodec.Empty is not an Atom
* Add Optional HttpCodec and implement EncoderDecoder for optional Header and Query
* Move all middlewares related to Csrf to MiddlewareSpec companion object
* Fix the bug in headers get
* Make sure cooke is optional
* Remove existing Csrf trait


afsalthaj:
@jdegoes  Could you take a look now? I have resolved conflicts. Benchmarks job failures are more about permission as far as I see.

jdegoes (approved):

PR #1681, opened 2022-10-24
Create TestClient

swoogles:
The WebSocket behavior is not implemented yet. 
It will require some changes to `Channel` that will be more invasive than this PR.

### DIFF
--- a/zio-http-testkit/src/main/scala/zio/http/TestClient.scala
+++ b/zio-http-testkit/src/main/scala/zio/http/TestClient.scala
@@ -151,7 +151,7 @@ object TestClient {
   ): ZIO[R with TestClient, Nothing, Unit] =
     ZIO.serviceWithZIO[TestClient](_.addHandler(handler))
 
-  val layer: ZLayer[Any, Throwable, TestClient] =
+  val layer: ZLayer[Any, Nothing, TestClient] =
     ZLayer.scoped {
       for {
         behavior <- Ref.make[HttpApp[Any, Throwable]](Http.empty)
--- a/zio-http-testkit/src/main/scala/zio/http/TestClient.scala
+++ b/zio-http-testkit/src/main/scala/zio/http/TestClient.scala
@@ -7,32 +7,37 @@ import zio.http.socket.SocketApp
 /**
  * Enables tests that use a client without needing a live Server
  *
- * @param behavior Contains the user-specified behavior that takes the place of the usual Server
+ * @param behavior
+ *   Contains the user-specified behavior that takes the place of the usual
+ *   Server
  */
 case class TestClient(behavior: Ref[HttpApp[Any, Throwable]]) extends Client {
+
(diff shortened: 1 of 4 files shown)


zio-http-testkit/src/main/scala/zio/http/TestClient.scala
```diff
@@ -0,0 +1,160 @@
...
+  ): ZIO[R, Nothing, Unit] =
+    for {
+      r                <- ZIO.environment[R]
+      previousBehavior <- behavior.get
+      newBehavior                  = handler.andThen(_.provideEnvironment(r))
+      app: HttpApp[Any, Throwable] = Http.collectZIO(newBehavior)
+      _ <- behavior.set(previousBehavior ++ app)
+    } yield ()
+
+  val headers: Headers                   = Headers.empty
```
jdegoes, line 69: What are these for?

zio-http-testkit/src/main/scala/zio/http/TestClient.scala
```diff
@@ -0,0 +1,160 @@
...
+import zio.http.socket.SocketApp
+
+/**
+ * Enables tests that use a client without needing a live Server
+ *
+ * @param behavior
+ *   Contains the user-specified behavior that takes the place of the usual
+ *   Server
+ */
+case class TestClient(behavior: Ref[HttpApp[Any, Throwable]]) extends Client {
```
jdegoes:
```suggestion
final case class TestClient(behavior: Ref[HttpApp[Any, Throwable]]) extends Client {
```

zio-http-testkit/src/main/scala/zio/http/TestClient.scala
```diff
@@ -0,0 +1,160 @@
...
+   *   {{{
+   *    TestClient.addHandler{case request  if request.method == Method.GET => ZIO.succeed(Response.ok)}
+   *   }}}
+   */
+  def addHandler[R](
+    handler: PartialFunction[Request, ZIO[R, Throwable, Response]],
+  ): ZIO[R with TestClient, Nothing, Unit] =
+    ZIO.serviceWithZIO[TestClient](_.addHandler(handler))
+
+  val layer: ZLayer[Any, Throwable, TestClient] =
```
jdegoes: The error is `Throwable` but there is no way for this layer to fail?

zio-http-testkit/src/main/scala/zio/http/TestClient.scala
jdegoes: Ah they come from `Client`

zio-http-testkit/src/main/scala/zio/http/TestClient.scala
swoogles: Thanks, removed!

jdegoes (approved):

PR #1683, opened 2022-10-25
Rework `Doc` and rendering

987Nabil:
Since we mainly want to render the `Doc`s to CommonMark, this rework focuses on making the data structure reflect the capabilities of CommonMark and having an intuitive API.

Also it fixes some rendering problems, since for example there can not be a line break between a `*` or `**` to format text, and the text that should be formatted. So I changed the rendering logic to ensure this.

rfc @jdegoes


jdegoes:
It would be nice to preserve plaintext support, for a future ZIO CLI integration, and HTML is pretty important too, so we can add a route that generates HTML documentation for the API. 

Any reason why to delete all this code as opposed to just improving the CommonMark rendering and the data types to better match CommonMark?

987Nabil:
I did not see the usecase and it was easier to do the changes with only one rendering method to take care of. But no problem, I'll bring it back/rewrite it. 🙂

987Nabil:
@jdegoes I added back the plain text rendering and reworked the html rendering.
The html has no style with intention. I think it is better to keep it plain here and if we want to provide some common httpApp or something like that, to offer html rendering of docs, we can define at this place templates that take a doc as an input.
So you could have something like predefined themes or the users can style it if they want to.

jdegoes (approved):

jdegoes:
Thank you!

PR #1686, opened 2022-10-25
feat: Implement structured Authorization header

jakubjanecek:
Part of #1524.

### DIFF
--- a/zio-http/src/main/scala/zio/http/model/headers/values/Authorization.scala
+++ b/zio-http/src/main/scala/zio/http/model/headers/values/Authorization.scala
@@ -1,0 +1,138 @@
+package zio.http.model.headers.values
+
+import zio.http.model.headers.values.Authorization.AuthScheme.{Basic, Bearer, Digest, Unparsed}
+
+import java.net.URI
+import java.nio.charset.StandardCharsets
+import java.util.Base64
+import scala.annotation.tailrec
+import scala.util.Try
+
+/** Authorization header value. */
+sealed trait Authorization
+
+object Authorization {
+
+  sealed trait AuthScheme
+
+  object AuthScheme {
+    final case class Basic(username: String, password: String)            extends AuthScheme
+    final case class Digest(
+      response: String,
+      username: String,
+      realm: String,
+      uri: URI,
+      opaque: String,
+      algorithm: String,
+      qop: String,
+      cnonce: String,
+      nonce: String,
+      nc: Int,
+      userhash: Boolean,
+    ) extends AuthScheme
+    final case class Bearer(token: String)                                extends AuthScheme
+    final case class Unparsed(authScheme: String, authParameters: String) extends AuthScheme
+  }
+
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zio/http/model/headers/values/Authorization.scala
```diff
@@ -0,0 +1,147 @@
...
+      username: String,
+      realm: String,
+      uri: URI,
+      opaque: String,
+      algorithm: String,
+      qop: String,
+      cnonce: String,
+      nc: Int,
+      userhash: Boolean,
+      rest: Map[String, String],
```
Wosin: IMO, we should rename that to something else than `rest` as it may misleading (especially can be similar to REST).

zio-http/src/main/scala/zio/http/model/headers/values/Authorization.scala
```diff
@@ -0,0 +1,147 @@
...
+    } else InvalidAuthorizationValue
+  }
+
+  // https://datatracker.ietf.org/doc/html/rfc7616
+  private def parseDigest(value: String) = try {
+    var state: State         = State.Key
+    var index                = 0
+    var params               = HashMap[String, String]()
+    var rest                 = HashMap[String, String]()
+    var currentKey: String   = null
```
Wosin: Is there a way to get rid of those `var` and `nulls`?

zio-http/src/main/scala/zio/http/model/headers/values/Authorization.scala
jakubjanecek: I removed it altogether. I thought I would need it but it's not necessary.

zio-http/src/main/scala/zio/http/model/headers/values/Authorization.scala
jakubjanecek: Sure there is. However I was going for low-level implementation to avoid unnecessary allocations. If you think it's not necessary here I can reimplement it differently.

jakubjanecek: So how about now?

jakubjanecek:
@Wosin Will you have another look at this please?

devsprint:
@jakubjanecek would you have time to complete it during this week?

jakubjanecek:
@devsprint It was ready to be merged but merge conflicts appeared.

I think I've fixed everything and rebased onto `master` but some unrelated test is failing. I'd like this to be merged so that I don't have to fix merge conflicts again.

devsprint:
@jakubjanecek I will take a look into it as the master or the other PRs that open are not failing during tests.

devsprint:
That's a flaky tests. I will fix in a separate PR.

devsprint:
@jdegoes can you have a look?

devsprint:
Closed in favor of #1834

PR #1694, opened 2022-10-26
Textcodec fixes

### LINKED ISSUE #1685: Improve performance of TextCodec#isDefinedAt
`TextCodec#isDefinedAt` needs to be optimized in a performant fashion for all subtypes. Currently, the method uses a slow implementation for the subtypes `UUIDCodec` and `BooleanCodec`. 

These implementations should be rewritten so that they do not allocate any memory (directly or indirectly), but rather validate that a string is a UUID or boolean by step-by-step iteration through characters in the string.

Once `isDefinedAt` is optimized for these subtypes, then the `decode` method of `TextCodec` can be rewritten to use a combination of `isDefinedAt` and `apply`.

Wosin:
Closes #1685.

Open to propositions on how to make UUID verification prettier.

### DIFF
--- a/zio-http/src/main/scala/zio/http/api/internal/TextCodec.scala
+++ b/zio-http/src/main/scala/zio/http/api/internal/TextCodec.scala
@@ -128,10 +128,7 @@ private[api] object TextCodec {
           count += 1
         else if (char == 45) {
           if (
-            group > 4 ||
-            (group == 0 && count != 8) ||
-            ((group == 1 || group == 2 || group == 3) && count != 4) ||
-            (group == 4 && count != 12)
+            group > 4 || (group == 0 && count != 8) || ((group == 1 || group == 2 || group == 3) && count != 4) || (group == 4 && count != 12)
           ) {
             defined = false
             i = value.length
--- a/zio-http/src/main/scala/zio/http/api/internal/TextCodec.scala
+++ b/zio-http/src/main/scala/zio/http/api/internal/TextCodec.scala
@@ -21,8 +21,7 @@ private[api] sealed trait TextCodec[A] extends PartialFunction[String, A] { self
   // TODO: Implement this using `isDefinedAt` and `apply` but only after all
   // subtypes properly & performantly implement `isDefinedAt`.
   final def decode(value: String): Option[A] =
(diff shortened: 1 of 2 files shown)


zio-http/src/test/scala/zio/http/api/internal/UUIDCodecSpec.scala
```diff
@@ -0,0 +1,35 @@
...
+      test("Should fail to validate String with leading hyphen") {
+        assertTrue(!UUIDCodec.isDefinedAt("-" + correctUUID))
+      },
+      test("Should work if UUID has only 0s") {
+        assertTrue(UUIDCodec.isDefinedAt("00000000-0000-0000-0000-000000000000"))
+      },
+      test("Should fail if UUID has incorrect character") {
+        assertTrue(!UUIDCodec.isDefinedAt("0000000Z-0000-0000-0000-000000000000"))
+      },
+      test("Should fail if UUID has two hyphens") {
```
bburdiliak, line 31: I would replace the illegal character Z with a legal one to really test the double hyphen case.

bburdiliak (changes requested):

zio-http/src/test/scala/zio/http/api/internal/UUIDCodecSpec.scala
Wosin: Oh, great catch. Will do!

zio-http/src/main/scala/zio/http/api/internal/TextCodec.scala
```diff
@@ -104,7 +104,7 @@ private[api] object TextCodec {
     def encode(value: Boolean): String = value.toString
 
     // TODO: Make faster by hand-writing validation:
-    def isDefinedAt(value: String): Boolean = decode(value).isDefined
+    def isDefinedAt(value: String): Boolean = value == "1" || value == "0" || value == "true" || value == "false"
```
jdegoes, line 107: This does not line up with `apply` since `apply` also accepts yes/on, no/off.

jdegoes:
@Wosin Would love to get this in. Please just re-open when you have a chance to make the requested changes!

PR #1698, opened 2022-10-27
Fix parameter ordering

swoogles:
Also add type annotations to be more explicit and keep us accurate in the future


Smudgey:
Thanks for clearing that up & fixing quickly!

jdegoes (approved):

PR #1706, opened 2022-10-27
Rewrite withAccept* Middleware

### LINKED ISSUE #1702: Middleware Rewrites - withAccept*
Part of #1689 

- [x] `withAccept` - parameters: `(value: CharSequence)`
- [x] `withAcceptEncoding` - parameters: `(value: CharSequence)`
- [x] `withAcceptLanguage` - parameters: `(value: CharSequence)`
- [x] `withAcceptPatch` - parameters: `(value: CharSequence)`
- [x] `withAcceptRanges` - parameters: `(value: CharSequence)`

Smudgey:
Resolves #1702

### DIFF
--- a/zio-http/src/main/scala/zio/http/api/Middleware.scala
+++ b/zio-http/src/main/scala/zio/http/api/Middleware.scala
@@ -3,6 +3,7 @@ package zio.http.api
 import zio._
 import zio.http._
 import zio.http.api.MiddlewareSpec.CsrfValidate
+import zio.http.model.headers.values.{Accept, AcceptEncoding}
 import zio.http.model.{Cookie, Status}
 import zio.schema.codec.JsonCodec
 
@@ -104,11 +105,11 @@ object Middleware {
   def addCookieZIO[R](cookie: ZIO[R, Nothing, Cookie[Response]]): Middleware[R, Unit, Cookie[Response]] =
     fromFunctionZIO(MiddlewareSpec.addCookie)(_ => cookie)
 
-  def withAccept(value: CharSequence): Middleware[Any, Unit, Unit] =
-    fromFunction(MiddlewareSpec.withAccept(value))(_ => ())
+  def withAccept(value: CharSequence): Middleware[Any, Unit, Accept] =
+    fromFunction(MiddlewareSpec.withAccept)(_ => Accept.toAccept(value.toString))
 
-  def withAcceptEncoding(value: CharSequence): Middleware[Any, Unit, Unit] =
-    fromFunction(MiddlewareSpec.withAcceptEncoding(value))(_ => ())
+  def withAcceptEncoding(value: CharSequence): Middleware[Any, Unit, AcceptEncoding] =
+    fromFunction(MiddlewareSpec.withAcceptEncoding)(_ => AcceptEncoding.toAcceptEncoding(value.toString))
 
   def withAcceptLanguage(value: CharSequence): Middleware[Any, Unit, Unit] =
     fromFunction(MiddlewareSpec.withAcceptLanguage(value))(_ => ())
--- a/zio-http/src/main/scala/zio/http/api/Middleware.scala
+++ b/zio-http/src/main/scala/zio/http/api/Middleware.scala
@@ -3,7 +3,7 @@ package zio.http.api
 import zio._
 import zio.http._
 import zio.http.api.MiddlewareSpec.CsrfValidate
-import zio.http.model.headers.values.{Accept, AcceptEncoding, AcceptLanguage, AcceptPatch}
+import zio.http.model.headers.values.{Accept, AcceptEncoding, AcceptLanguage, AcceptPatch, AcceptRanges}
 import zio.http.model.{Cookie, Status}
 import zio.schema.codec.JsonCodec
 
@@ -117,8 +117,8 @@ object Middleware {
   def withAcceptPatch(value: CharSequence): Middleware[Any, Unit, AcceptPatch] =
     fromFunction(MiddlewareSpec.withAcceptPatch)(_ => AcceptPatch.toAcceptPatch(value.toString))
 
-//  def withAcceptRanges(value: CharSequence): Middleware[Any, Unit, String] =
-//    fromFunction(MiddlewareSpec.withAcceptRanges)(_ => ???)
+  def withAcceptRanges(value: CharSequence): Middleware[Any, Unit, AcceptRanges] =
+    fromFunction(MiddlewareSpec.withAcceptRanges)(_ => AcceptRanges.to(value.toString))
 
   /**
    * Generates a new CSRF token that can be validated using the csrfValidate
--- a/zio-http/src/main/scala/zio/http/api/Middleware.scala
+++ b/zio-http/src/main/scala/zio/http/api/Middleware.scala
@@ -104,6 +104,21 @@ object Middleware {
   def addCookieZIO[R](cookie: ZIO[R, Nothing, Cookie[Response]]): Middleware[R, Unit, Cookie[Response]] =
     fromFunctionZIO(MiddlewareSpec.addCookie)(_ => cookie)
 
+  def withAccept(value: CharSequence): Middleware[Any, Unit, Unit] =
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/api/MiddlewareSpec.scala
```diff
@@ -158,6 +158,33 @@ object MiddlewareSpec {
   def requireHeader(name: String): MiddlewareSpec[String, Unit] =
     MiddlewareSpec(HeaderCodec.header(name, TextCodec.string), HttpCodec.empty)
 
+  def withAccept(value: CharSequence): MiddlewareSpec[Unit, Unit] =
+    MiddlewareSpec(HttpCodec.empty, HeaderCodec.header(HeaderNames.accept.toString, TextCodec.constant(value.toString)))
```
987Nabil: I think this should actually look more like this.
```suggestion
  def withAccept: MiddlewareSpec[Unit, Accept] =
    MiddlewareSpec(HttpCodec.empty, HeaderCodec.accept)
```
This is typed and lets the user know what header will be added

Smudgey: It does make more sense to me for this to be typed, seems wrong to have `MiddlewareSpec[Unit, Unit]`

However, using `HeaderCodec.accept` throws away the `value: CharSequence` required in this implementation (listed in the issue)

Smudgey: The Implementation I've changed to is
```scala
  def withAccept(value: CharSequence): Middleware[Any, Unit, Accept] =
    fromFunction(MiddlewareSpec.withAccept)(_ => Accept.toAccept(value.toString))
```
Spec is the one you've suggested

Tests are failing because the value isn't being passed through, still debugging to confirm its an issue with my solution and/or tests

Smudgey: I understand better now that the argument isn't needed in the `MiddlewareSpec` as its only defining a structure. The implementations in `Middleware` handle values passed in. However, I'm still not sure why the implementation I suggested above isn't working, it correctly adds the Accept header, but the value is empty each time.

Looking through `Accept.toAccept` I was thinking it may be an issue with the value i'm passing in (which is `"accept"` in the test) but even using a valid field such as `"json"` makes no difference

987Nabil: I just saw, that in your test you do something with a request. But this adds the accept header to a response. That might be the problem

Smudgey: Pushed what I have for now, I've been trying several different variations based on how similar files are tested. I've tried adding the middleware with `@@` similar to `WebSpec` but the types don't line up, I believe thats using the old Middleware.

I've tried HeaderValues as well but doesn't make much difference compared to just getting the headers as a list. I updated `withAccept` and `withAcceptEncoding` to correctly return appropriate types in the Spec, with the implementation taking in a given value and returning it.

With regards to the test I understand I was re-using a request. My understanding here though is that I am creating an 'empty' `Request` which I pass to my app, using the for comp to retrieve a `Response` and grab the headers to assert on. I've updated the first two tests so that no modification is happening to a `request` val and an empty `Request` is passed in directly.

I found `identifyEncoding` in `zio.http.model.headers.values.AcceptEncoding` and noticed that there are specific encodings required, such as `"compress"`. Passing this in actually works, which makes sense because there are a limited number of acceptable encodings & I was testing by passing in a random String. The other outcome of this means that when it was failing, because I've set the error type to `Unit` there was no actual error & what was being created with `Accept.toAcceptEncoding` was just returning an empty value. This makes me believe an error type should be specified, but i'm not entirely sure thats the right thing here.

After this discovery, i've noticed the same with for `withAccept`. Essentially it came down to me using these headers incorrectly while testing, because passing in a valid value such as `application/json` works fine.

So made some progress, which is nice, I just need to figure out what would be expected for cases where invalid values are passed for accept based headers. Then I need to go through and update the other Accept headers i'm implementing.

Smudgey: One suggestion for handling the error case is to match for the invalid case returned and abort the request, continue if it's valid (having seen in Kit's presentation yesterday this is possible)

Smudgey: `zio.http.model.headers.values` contains definitions and functions that I've been using when converting to an accept header. They also provide a 'from' variant and essentially do all the validation that I've discovered that requires the input to be valid, otherwise returning an Invalid case. There doesn't seem to be one for `AcceptRanges` so I'm going to attempt to implement that as well.

jdegoes: So a lot of these middleware add response headers. In fact, they add constant response headers (in many cases). 

Anything which adds a constant response header will probably have type `Unit` for the middleware output. Why? Because the middleware implementation itself is required to **produce** a middleware output value. The output value is translated into a patch, which is applied to the response.

So if a middleware is always adding the same constant header to the response, then there is no need for the middleware implementation to produce any specific value. Because there is no specific input needed to add the constant header to the response. 

I hope that makes it clear why, for middleware that add a constant header to response, it is enough to have the middleware output type be `Unit`.

Now, what if a middleware produces a header dynamically? One which is NOT fixed? Well in that case, the middleware output will not be `Unit` anymore, it will be some header value (like `Accept`), which the middleware implementation will be required to produce, and which will be translated dynamically to the response header.

Does this make sense?

Smudgey: Yes, this makes sense and I believe makes it clear why the response header is typed here rather than `Unit`

Smudgey: @987Nabil @jdegoes I've applied all planned changes

zio-http/src/main/scala/zio/http/api/Middleware.scala
```diff
@@ -104,6 +105,21 @@ object Middleware {
   def addCookieZIO[R](cookie: ZIO[R, Nothing, Cookie[Response]]): Middleware[R, Unit, Cookie[Response]] =
     fromFunctionZIO(MiddlewareSpec.addCookie)(_ => cookie)
 
+  def withAccept(value: CharSequence): Middleware[Any, Unit, Accept] =
```
jdegoes, line 112: The thing I don't like about these middlewares is that they produce constant (static) headers but they are typed with non-Unit values, which suggests (to me) that they are more dynamic.

If indeed these will only ever produce the same static header, you can do `mapOut(_.unit(/* static value*/))` in order to turn the type parameter from `Accept` into `Unit`. You will have to provide a static `Accept` value, but it looks like you have that so won't be a problem.

jdegoes:
@Smudgey Can you please resolve conflicts?

Smudgey:
@jdegoes Conflicts resolved - seems since I started on this work someone else implemented the AcceptRanges Header codec, so I've taken that and discarded my own

zio-http/src/main/scala/zio/http/api/Middleware.scala
Smudgey: @jdegoes I've thought about this several times since Friday and to be honest, I can see it either way. Ultimately, i'm happy to defer to your opinion here. Are you asking me to change the return type to `Unit`? or are you okay with this PR as-is?

devsprint:
Closed in favor of #1845

Smudgey:
Hey @devsprint can I ask why this was closed in favour of a different PR? I've been awaiting feedback for this since the end of the hackathon event, it's not been abandoned. If there are changes you believe need to be made I'm willing to work on it 👍

devsprint:
@Smudgey  We have moved all your work in a different PR and updated it to resolve the conflicts, and to keep the contribution as yours. I did this as we want to finish all the `Headers` and `Middleware` changes during this week.  Thank you for your contribution.

Smudgey:
@devsprint Understood, thank you!

PR #1708, opened 2022-10-27
Rename root for better rendering in tooling

987Nabil:
IntelliJ for example renders the name of the main module on the top of the window. With this change, it is easy to see that the window belongs to zio-http.

I think this has no implications for the code.


adamgfraser (commented):
Maybe just `zio-http`?

jdegoes (approved):

987Nabil:
@adamgfraser zio-http already exists

Issue #1689, opened 2022-10-26
Implement an existing ZIO HTTP middleware in terms of new declarative middleware API


jdegoes:
**Background**

ZIO HTTP has long supported _middleware_, which are functions that transform one `Http` into another `Http`. Middleware allow an orthogonal way of decomposing complex server functionality into modular aspects--such as logging, metrics, authentication, and so forth.

The existing middleware machinery lives in `zio.http.Middleware` and the associated companion object, with many concrete middleware implementations being defined in a subpackage that is brought into the `Middleware` companion object. 

In an effort to better support the needs of `zio.http.api`, a new declarative encoding of middleware has been introduced, which lives in the package `zio.http.api`. Unlike the previous definition of `Middleware`, the new definition of middleware divides the middleware into two separate pieces:

1. A specification. Currently, this is called `MiddlewareSpec`, and it describes two things:
  1. What the middleware requires from an incoming HTTP `Request`. For example, an authentication middleware may require an `Authentication` header. This is described by a `zio.http.api.HttpCodec`, which is limited to look at the headers and query string parameters of the request (and soon, the full path of the incoming request).
  2. What  modifications the middleware makes to the HTTP `Response` generated by the handler. For example, one middleware might add a `X-Trace-Id` header to the response, for purposes of distributed tracing. This is described by a `zio.http.api.HttpCodec`, which is limited to modify the status and the headers of the response.
2. An implementation. An implementation, which in the happy path, takes in the middleware input, and produces the middleware output. An implementation could also decide to abort calling the route handler for an endpoint, rather than producing the normal modification to the HTTP response. This abort behavior is useful for error conditions, for example, an authentication middleware does NOT want the route handler to be invoked if authentication fails. Most middleware will not use abort.

The specifications for middleware are intended to live in the `zio.http.api.MiddlewareSpec` companion object. The implementations for middleware are defined to live in the `zio.http.api.Middleware` companion object.

Here is the specification for a middleware that adds a fixed `Cookie` header to responses:

```scala
def addCookie: MiddlewareSpec[Unit, Cookie[Response]] =
    MiddlewareSpec(
      HttpCodec.empty,
      HeaderCodec.cookie.transformOrFail(
        str => Cookie.decode[Response](str).left.map(_.getMessage),
        _.encode(false).left.map(_.getMessage),
      ),
    )
```

[...]

PR #1709, opened 2022-10-27
Move requestLogging Middleware to API

BalduinLandolt:


BalduinLandolt:
@irinaschubert here is the PR draft

jdegoes:
@BalduinLandolt It's very difficult to do this one in the new middleware. I might suggest taking an easier ticket that doesn't require altering any of the declarative middleware machinery!

jdegoes:
I will close this for now because it is not super clear on how to do this one. Please re-open if you have ideas about implementing it (but it is not a simple ticket).

BalduinLandolt:
No problem! I came to the same conclusion (see my comment in #1689). I suppose it would require an input HttpCodec that gives access to the entire Request object.... but that's beyond my skill and time availability - for now :)

PR #1710, opened 2022-10-27
AddHeader* part done

### LINKED ISSUE #1701: Middleware Rewrites - add/set/remove Headers
Part of #1689 

- [x] `addHeader` - parameters: `(header: Header)`
- [ ] `addHeader` - parameters: `(name: CharSequence, value: CharSequence)`
- [x] `addHeaders` - parameters: `(headers: Headers)`
- [ ] `removeHeader` - parameters: `(name: String)`
- [ ] `removeHeaders` - parameters: `(headers: List[String])`
- [ ] `setHeaders` - parameters: `(headers: Headers)`

vder:
heres the partial solution for first 3 middlewares but im not sure how to remove and set headers
resolves #1701

### DIFF
--- a/zio-http/src/main/scala/zio/http/api/Middleware.scala
+++ b/zio-http/src/main/scala/zio/http/api/Middleware.scala
@@ -265,8 +265,7 @@ object Middleware {
     fromFunction(MiddlewareSpec.addHeader(header))(_ => ())
 
   def addHeaders(headers: Headers): Middleware[Any, Unit, Unit] =
-    fromFunction(MiddlewareSpec.addHeaders(headers))(_ => ()) 
-  
+    fromFunction(MiddlewareSpec.addHeaders(headers))(_ => ())
 
   /**
    * Generates a new CSRF token that can be validated using the csrfValidate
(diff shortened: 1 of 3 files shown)


jdegoes:
Currently there is no way to remove a header in the new middleware machinery. I'd skip the removal methods for now.

Please run `fmt` to fix errors like [this one](https://github.com/zio/zio-http/actions/runs/3343401391/jobs/5541916280#step:6:55). Thank you!

vder:
I've one more doubt regarding this code. I've added test where I add  header to the request with already defined headers and the original headers where replaced. is that expected behavior?

jdegoes:
`addHeaders` will only add headers to the `Response`.  If the response already contains headers, then possibly they will be replaced. I think we should probably fix that by modifying `Response.Patch` appropriately.

jdegoes:
@vder Build failed, can you look into it?

jdegoes:
There are a couple conflicts in the imports section. Can you please resolve? Thank you!

devsprint:
Closed in favor of #1847

PR #1712, opened 2022-10-28
Add length to `Body`

987Nabil:
Does this make sense?

### DIFF
--- a/zio-http/src/main/scala/zio/http/Body.scala
+++ b/zio-http/src/main/scala/zio/http/Body.scala
@@ -44,6 +44,8 @@ trait Body { self =>
   def isComplete: Boolean
 
   def write(ctx: Ctx)(implicit trace: Trace): Task[Boolean]
+
+  def length: Task[Long] = asChunk.map(_.length.toLong)
 }
 
 object Body {
@@ -157,6 +159,8 @@ object Body {
 
       true
     }
+
+    override def length: Task[Long] = ZIO.attempt(file.length())
   }
 
   /**


zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -44,6 +44,8 @@ trait Body { self =>
   def isComplete: Boolean
 
   def write(ctx: Ctx)(implicit trace: Trace): Task[Boolean]
+
+  def length: Task[Long] = asChunk.map(_.length.toLong)
```
tusharmath, line 48: Typically body length will be available in the headers. This way we could have an issue when the content is being read as a stream. More precisely, when the server is in Streaming mode, it won't be able to read the content more than once. So if the user has consumed the body once by calling the `length` method, it won't be able to actually read any bytes anymore.


What is the use-case for it?

987Nabil: Is the content length added by netty when creating responses? I was thinking if it makes sense to have this, so you could write a middleware that reads the length of a response body and adds it to the response.
Now I thought maybe netty takes care of this anyway. However, body also has asChunk, which has has the same problem or? If you call asChunk and then write, it will blow for streaming.

tusharmath: Yes, netty will take care of it if the size is known. Otherwise, if the size is unknown, the response will be chunked and the connection will close automatically once the data is completely sent.

PR #1725, opened 2022-10-28
Add Warning Header Codec

sf17490:
For ticket #1524 .

This is my first attempt at contributing so I'm expecting some revisions will probably need to be made - in particular I'm wondering if this is the best way to validate the warnDate. 

Keen for feedback!


sf17490:
I will close this PR and open another one so I can correct the git authorship and sign the Contributor licence agreement

PR #1727, opened 2022-10-28
Test client socket implementation

swoogles:
Definitely room for future improvement here, but we've managed to get `SocketApp`s running on both the `LiveClient` and `TestClient` implementations via our Contract test.

Major thanks to @dkarlinsky for all the assistance during the Hackathon!

### DIFF
--- a/zio-http/src/main/scala/zio/http/Channel.scala
+++ b/zio-http/src/main/scala/zio/http/Channel.scala
@@ -5,6 +5,7 @@ import zio.{Task, Trace, UIO}
  * An immutable and type-safe representation of one or more netty channels. `A`
  * represents the type of messages that can be written on the channel.
  */
+// TODO Remove all non-necessary methods here
 trait Channel[-A] {
 
   /**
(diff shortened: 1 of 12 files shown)


zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -350,7 +350,7 @@ trait ZClient[-Env, -In, +Err, +Out] { self =>
     url: String,
     app: SocketApp[Env1],
     headers: Headers = Headers.empty,
-  )(implicit trace: Trace, unsafe: Unsafe): ZIO[Env1 with Scope, Err, Out] =
```
swoogles, line 353: It seemed that this was an improper use of `Unsafe`, but please let us know if we're mistaken.

zio-http/src/main/scala/zio/http/Channel.scala
```diff
@@ -5,6 +5,7 @@ import zio.{Task, Trace, UIO}
  * An immutable and type-safe representation of one or more netty channels. `A`
  * represents the type of messages that can be written on the channel.
  */
+// TODO Remove all netty-specific methods here and reduce footprint
```
swoogles, line 8: I think we can either restrict this trait or use a new, smaller one to accomplish the same goals here.

jdegoes (approved):

Issue #1693, opened 2022-10-26
Implement zio.http.api.internal.RichTextCodec#describe


jdegoes:
**Background**

`RichTextCodec` defines a codec for some structured type `A`, which may be decoded from a string, or encoded to a string. Codecs can be thought of as combinations of two things: a parser, which can produce an `A` from a string input (decoding); and a printer, which can render an `A` to a string (encoding). However, they are more powerful than this, because they are pure data, it is possible to use them to generate documentation (the purpose of this ticket).

**Task**

This task is to implement the `describe` method of `RichTextCodec`, which is intended to generate human-readable documentation for any `RichTextCodec`, ultimately for purpose of embedding into OpenAPI documentation.

This can be done by recursively pattern matching on `RichTextCodec` (which is a sealed trait), finding out which subtype it is, and implementing the subtype-specific way of describing the codec in a human-readable way.

Together with this, you should implement a test suite that proves the implementation can handle all the `RichTextCodec` constructors and operators, so that we have confidence in the correctness of documentation generation.

gregor-rayman:
I'd like to work on this

PR #1732, opened 2022-10-29
RichTextCodec.describe

gregor-rayman:
Still work in progress. For #1693


jdegoes:
@gregor-rayman I love where you are going with this.

In order to generate BNF, we would need two things:

1. Labels associated with terms (particularly recursive ones).
2. Handling of cyclic graphs through an identity map.

An example RFC in some variant of BNF:

```
Accept = #( media-range [ accept-params ] )

     media-range    = ( "*/*"
                      / ( type "/" "*" )
                      / ( type "/" subtype )
                      ) *( OWS ";" OWS parameter )
     accept-params  = weight *( accept-ext )
     accept-ext = OWS ";" OWS token [ "=" ( token / quoted-string ) ]
```

This happens to not be recursive but relies heavily on parts being named.

I made a very light implementation that does not handle recursion and does not use an intermediate ADT. Do you have interest in pushing your version forward?

gregor-rayman:
I'd love to pushing this forward. 

Currently the labels are simple numbers, with few exceptions: the main term is called main, some character sets have predefined labels (digits, letters, whitespace, ...).  If the codecs had names, that would for sure help. 

Now I generate the labels only if: 
  * the codecs are used recursively (I had to change `repeat` from `def` to `lazy val` to avoid infinite recursion)
  * the codec is not trivial (e.g. CharIn or literal) and is used multiple times.

To further proceed in this direction, I'd add optional name to a codec. The name would always be used as a label, even for non recursive single usage. This could also help to identify recursively used codes generated dynamically (def vs lazy val - if those are needed anyway), because the `describe` method could treat them as the same codec. (Naming two different codecs the same should be probably considered an error and lead to an invalid description)

jdegoes:
@gregor-rayman That's fantastic news! We could have something extremely impressive, capable of generating fantastic documentation for all HTTP headers, query parameters, etc.

I agree with your step of adding labels. I suggest adding a new subtype of `RichTextCodec` called `Labelled` that stores the label and the codec.

Then we can have:

```scala
sealed trait RichTextCodec[A] {
  ...
  def labeled(name: String): RichTextCodec[A] = ...
  ...
}
```

Maybe with a symbolic operator `??` (often used to attach docs to ADTs in ZIO libraries).

Then you can use a `List[TextCodec[_]]` (???) to keep track of which term you are in as you recurse. 

I think you are on the right track to build an intermediate form that looks like BNF. With the labels, you will then have productions in the grammar that you can use to render in the classic `BINARY_OP = '+' | '-' | ...` form.

If something doesn't have a name, you can use the reference identity of the text codec to identify it, which assumes that people use `lazy val` for recursive codecs (I think that assumption is fine).

Let's try to produce a single `code` block that contains the whole grammar. 

Let me know if you need help!

gregor-rayman:
I am almost finished :)

The cycles as still a bit tricky. Another issue is quoting literals, because now "a|z" can be interpreted as literal "a" or literal "z"; or as literal "a|z". But that will be the second step.

gregor-rayman:
I have created draft PR #1741.

PR #1734, opened 2022-10-29
Add withContentLength Middleware

### LINKED ISSUE #1703: Middleware Rewrites - withContent*
Part of #1689 

- [x] `withContentBase` - parameters: `(value: CharSequence)`
- [x] `withContentDisposition` - parameters: `(value: CharSequence)`
- [x] `withContentEncoding` - parameters: `(value: CharSequence)`
- [x] `withContentLanguage` - parameters: `(value: CharSequence)`
- [x] `withContentLength` - parameters: `(value: Long)`
- [x] `withContentLocation` - parameters: `(value: CharSequence)`
- [x] `withContentMd5` - parameters: `(value: CharSequence)`
- [x] `withContentRange` - parameters: `(value: CharSequence)`
- [x] `withContentSecurityPolicy` - parameters: `(value: CharSequence)`
- [x] `withContentTransferEncoding` - parameters: `(value: CharSequence)`
- [x] `withContentType` - parameters: `(value: CharSequence)`

987Nabil:
fixes #1703 also implements some header codecs for #1524 

I guess some more tests would be nice, but I did not have the time. I might add them to this PR later on or in another one.


zio-http/src/main/scala/zio/http/model/headers/values/ContentType.scala
```diff
@@ -0,0 +1,169 @@
...
+
+sealed trait ContentType extends Product with Serializable { self =>
+  def toStringValue: String = ContentType.fromContentType(self)
+}
+
+object ContentType {
+  // based on Mozilla's list of common media types
+  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+
+  case object `audio/aac`                                                                 extends ContentType
```
ccerbusca: can't you use MediaType instead of defining all of these?

zio-http/src/main/scala/zio/http/api/Middleware.scala
```diff
@@ -260,6 +261,150 @@ object Middleware {
     }
   }
 
+  /**
+   * Adds the content type header to the response based on a ZIO effect
+   */
+  def withContentBase(contentBase: ContentBase): Middleware[Any, Unit, ContentBase] =
```
jdegoes: I think you should call `.unit(<constant header>)` on all of these and provide the constant header. This indicates (in my mind) that there is no dynamic nature to the header, which is correct.

zio-http/src/main/scala/zio/http/model/headers/values/ContentType.scala
987Nabil: I kinda like the idea of having some constants for common type. Should they be defined in the header vale? Maybe not. But, as far as I can see, they currently don't exist.

zio-http/src/main/scala/zio/http/api/Middleware.scala
987Nabil: So you mean, that the header should not be returned if it is not computed, but is given as a value when calling the middleware constructor?
Even though I kinda understand why, is there a harm in returning it? Could I use/access it at a later point to do something useful? 🤔

zio-http/src/main/scala/zio/http/model/headers/values/ContentType.scala
ccerbusca: They do exist, defined in the trait `MimeDB`, which is used in the MediaType implementation. You could try and make the constants there available instead

ccerbusca: But even then, not using MediaType directly means you don't have support for optional parameters, which both examples in [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#syntax) have

987Nabil: Thanks. I'll take a look and make sure it will behave as it should.

987Nabil: @ccerbusca please check again :)

zio-http/src/main/scala/zio/http/api/Middleware.scala
987Nabil: @jdegoes done

ccerbusca (approved):

devsprint:
@987Nabil can you please resolve the conflict ? Thank you

987Nabil:
Will do sometime today

987Nabil:
Done @devsprint

devsprint:
@987Nabil can you please fix the conflict again ? 🙏

987Nabil:
I'll try to do it today. But later I am outside.

devsprint:
Closed in favor of #1836

PR #1738, opened 2022-10-30
Fixes #1737: Constructing RichTextCodec.repeat causes stack overflow

### LINKED ISSUE #1737: Constructing RichTextCodec.repeat causes stack overflow
**Describe the bug**
Calling the method `repeat` on `RichTextCodec` leads to stack overflow 

**To Reproduce**
Just use the lazy variable `RichTextCodec.whitespaces`, which  internally uses `repeat`

**Expected behaviour**
No stack overflow. Probably some part is too eager.

gregor-rayman:
Wrapped `repeat` with `defer`:
```
  final def repeat: RichTextCodec[Chunk[A]] =
    (self ~ RichTextCodec.defer(repeat)).transformOrFailRight(
```
(creating double laziness)

Perhaps changing `that` in `~` to by name instead of by value were a better solution. (Similar to `|`)

```
  final def ~[B](that: => RichTextCodec[B])(implicit combiner: Combiner[A, B]): RichTextCodec[combiner.Out] =
    RichTextCodec.Zip(self, RichTextCodec.defer(that), combiner)

```

### DIFF
--- a/zio-http/src/main/scala/zio/http/api/internal/RichTextCodec.scala
+++ b/zio-http/src/main/scala/zio/http/api/internal/RichTextCodec.scala
@@ -35,7 +35,7 @@ sealed trait RichTextCodec[A] { self =>
    * Returns a new codec that is the sequential composition of this codec and
    * the specified codec, producing the values of both as a tuple.
    */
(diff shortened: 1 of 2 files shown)


jdegoes:
Thanks for reporting! I fixed this in my latest PR, so will close this for now.

PR #1745, opened 2022-11-02
De nettify response

scottweaver:
DeNettify Response, Body + Performance Tweaks
============================================

Response
--------
 `Response` is now a trait that now has specific implementations for specific uses cases.
 - The `Attribute` class is completely gone and the values it contained have been divvied up between the different implementations of Response .  The same has been done to `Response`'s previous constructor args e.g. `HttpError` .
- `BasicResponse`:  Name says it all.  If the response isn't one of successive special cases, it is this.
- `ErrorResponse`:  Used when you want create a response directly from an `HttpError`.
- `SocketAppResponse`: Used when want to create a Response from a `SocketApp`.
- `NettyResponse`: Used when you are doing Netty things, specifically when you need to propagate a `ChannelHandlerContext` around.  I have corralled all of the Netty bits into here so, eventually this will be abstracted away.
`Response#unsafe#fromJResponse` and `Response#unsafe#fromStreamingJResponse` have been moved to `NettyResponse`'s companion and exist as smart constructors renamed to `make`.
- All of the Netty-specific encoding logic and encoded response caching aka freezing have been push up into new and existing Netty classes: `NettyBodyWriter` , `NettyResponseEncoder` and `ServerInboundHandler`. 

Body
----
- `Body` has also seen a decent amount of work as well as it also now follows a similar pattern as that of Response i.e. having concrete implementations for specific use cases.
- `EmptyBody`: Does what it says on the tin.  Empty like my soul.
- `AsciiStringBody`: For building from an `AsciiString`.
- `StringBody`: A less bourgy version of `AsciiStringBody` that builds a Body from a bog-standard `String` .
- `ByteBuffBody`: For building a Body from a Netty `ByteBuf`.
- `ChunkBody` : For building from...wait for it...a `Chunk`.
- `FileBody`: If you guessed creating a `Body` from a File your mother would be proud of you, as am I.
- `StreamBody`: ... I am running out of snarky, passive-aggressive comments at this point.  Builds a `Body` from a ZIO `Stream`.
- `UnsafeBytes`  trait added that indicates that a specific `Body` implementation is capable of producing an `Array[Byte]` in a possibly unsafe way that does not require the ZIO runtime.
- `Body.asArray` is now implemented per implementation so as provide optimal implementation relative to the `Body` impl.

Body is a now sealed, so that the job of encoding is now offloaded to an implementation-specific encode-terpeter™. In the case of Netty, it is the NettyBodyWriter .

Code Clean Up
-------------
- Moved all nested methods to be non-nested, private.
- Convert single-use methods' logic to being inlined at call sites.

Performance Tweaks
------------------
- Memoize `Http` and `ZEnvironment`
- Move 'ensuring' logic off of effects in `ServerInboundHandler` and onto the `Fiber.addObserver` in `NettyRuntime`.


zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -38,10 +41,7 @@ final case class Response private (
    * detect the changes and encode the response again, however it will turn out
    * to be counter productive.
    */
-  def freeze(implicit trace: Trace): Task[Response] =
-    for {
-      encoded <- encode()
-    } yield self.copy(attribute = self.attribute.withEncodedResponse(encoded, self))
+  def freeze: Response = self.copy(frozen = true)
```
tusharmath: `freeze` is not a pure function any more.

scottweaver: The fact that operation only sets a boolean flag would have me disagree with that statement.

As for its semantic meaning, that of caching a static value for performance purposes is executed at the "end of the world".   In this case during the execution of HTTP and ZIO DSLs when a request is handled by Netty.

Would you mind expanding on your statement, maybe I am misunderstanding your concerns?

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -170,19 +209,27 @@ object Body {
   /**
    * Helper to create Body from Stream of bytes
    */
-  def fromStream(stream: ZStream[Any, Throwable, Byte]): Body = {
-    new Body {
-      override def isComplete: Boolean = false
+  def fromStream(stream: ZStream[Any, Throwable, Byte]): Body = new StreamBody(stream)
 
-      override def asChunk(implicit trace: Trace): Task[Chunk[Byte]] = stream.runCollect
+  private[zio] final case class StreamBody(stream: ZStream[Any, Throwable, Byte]) extends AnyVal with Body {
```
jdegoes: Extending AnyVal can easily slow things down due to wrapping / unwrapping that occurs in any generic context.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -170,19 +209,27 @@ object Body {
...
-          _ <- stream.runForeachChunk(c => ZIO.succeed(ctx.writeAndFlush(Unpooled.wrappedBuffer(c.toArray))))
-          _ <- ZIO.attempt(ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT))
-        } yield true
+    override def isComplete: Boolean = false
+
+    override def asChunk(implicit trace: Trace): Task[Chunk[Byte]] = stream.runCollect
+
+    override def asStream(implicit trace: Trace): ZStream[Any, Throwable, Byte] = stream
+
+    override private[zio] def asChunkUnsafe(implicit unsafe: Unsafe): Chunk[Byte] = {
```
jdegoes: This would be called `unsafeAsChunk`

zio-http/src/main/scala/zio/http/Body.scala
scottweaver: Will remove it.

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: I have removed the `AtomicReference` that was most likely the concern.

The new PoC uses a `ConcurrentMap` to hold frozen responses. 

There may or may not be a need to look at a more complete caching approach though I don't see `frozen`  Responses as a very common uses-cases outside of some simple, static content.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -170,19 +209,27 @@ object Body {
...
+      implicit val trace = Trace.empty
+      val rtm            = Runtime.default
+      val exit           = rtm.unsafe.run(asChunk)
```
jdegoes: This unsafe run is going to be quite slow. I feel like rather than doing an unsafe run here, we should do it closer to Netty, where there exists the possibility we can collapse many unsafe run into a single one.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -170,19 +209,27 @@ object Body {
...
+
+      exit match {
+        case Exit.Success(chunk) => chunk
+        case Exit.Failure(cause) => throw cause.squash
```
jdegoes: In general, squashing a cause is not a good idea. It only makes sense when there are incorrectly recoverable errors in the defect channel.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -207,14 +264,17 @@ object Body {
...
-      ZIO
-        .attempt(unsafeAsync { (ctx, msg, isLast) =>
-          ctx.writeAndFlush(msg)
-          if (isLast) ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT): Unit
-          else ctx.read(): Unit
-        })
-        .as(true)
+    override private[zio] def asChunkUnsafe(implicit unsafe: Unsafe): Chunk[Byte] = {
+      implicit val trace = Trace.empty
+      val rtm            = Runtime.default
```
jdegoes: Ditto here. Should just let Netty driver do this since the code is identical in both places.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -14,13 +14,16 @@ import zio.{Cause, Promise, Task, Trace, Unsafe, ZIO}
...
 
-final case class Response private (
+sealed abstract case class Response private (
   status: Status,
   headers: Headers,
   body: Body,
   private[zio] val attribute: Response.Attribute,
   private[zio] val httpError: Option[HttpError],
+  private[zio] val encodedResponse: AtomicReference[Option[Response.EncodedResponse]],
+  frozen: Boolean = false,
```
jdegoes: I think you should delete all the private variables, and make it a `trait` that emulates a `case class`, including `copy`, `hashCode`, `==`, `apply`, and `unapply`. 

Then add a `def cached: Response` that returns a subtype like `Response.Cached`. The cached response can store `Map[Any, Any]` and drivers can, if a response is cached, choose to accelerate rendering by precomputing the response.

Alternately, there is no need for `Map[Any, Any]` if the Netty driver uses its own map, which can be a map of `Response` to `Response.EncodedResponse`.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -77,44 +77,40 @@ final case class Response private (
...
-        }
-      }
-  } yield res
+  def copy(
+    status: Status = self.status,
+    headers: Headers = self.headers,
+    body: Body = self.body,
+    frozen: Boolean = self.frozen,
+    attribute: Response.Attribute = self.attribute,
+    encodedResponse: AtomicReference[Option[Response.EncodedResponse]] = self.encodedResponse,
```
jdegoes: Because the mutable data is copied along, it is easy to violate referential transparency. Another reason to delete these from `Response`.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -125,8 +121,9 @@ object Response {
     headers: Headers = Headers.empty,
     body: Body = Body.empty,
     httpError: Option[HttpError] = None,
+    encodedResponse: Option[Response.EncodedResponse] = None,
   ): Response =
-    Response(status, headers, body, Attribute.empty, httpError)
+    new Response(status, headers, body, Attribute.empty, httpError, new AtomicReference(None)) {}
```
jdegoes: This allocates three 64 bit pointers as well as the atomic reference for each `Response`, when it is really only necessary for a cached response.

zio-http/src/main/scala/zio/http/netty/NettyRuntime.scala
```diff
@@ -14,7 +14,7 @@ private[zio] trait NettyRuntime { self =>
 
   def runtime(ctx: ChannelHandlerContext): Runtime[Any]
 
-  def run(ctx: ChannelHandlerContext, interruptOnClose: Boolean = true)(
+  def run(ctx: ChannelHandlerContext, ensured: () => Unit = () => (), interruptOnClose: Boolean = true)(
```
jdegoes: This default parameter will allocate a new lambda at the call site whenever it is not specified. I would instead move the `ensured` to the call site if possible.

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: See newest PoC approach using `ConcurrentMap`.

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: `AtomicRef` removed, `Response` is back to being fairly pure now.

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: I already implemented the second alternative with a `ConcurrentMap`.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -14,13 +14,15 @@ import zio.{Cause, Promise, Task, Trace, Unsafe, ZIO}
...
+import java.util.concurrent.atomic.AtomicReference
 
-final case class Response private (
+sealed abstract case class Response private (
   status: Status,
   headers: Headers,
   body: Body,
   private[zio] val attribute: Response.Attribute,
   private[zio] val httpError: Option[HttpError],
+  frozen: Boolean = false,
```
jdegoes: We still have `attribute` and `httpError`, can't we eliminate them too?

jdegoes: Note that we almost never use this `Boolean` field so it may make sense to have `def frozen = false` in the trait, and override for more specific classes.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -77,44 +76,38 @@ final case class Response private (
...
-        }
-      }
-  } yield res
+  def copy(
+    status: Status = self.status,
+    headers: Headers = self.headers,
+    body: Body = self.body,
+    frozen: Boolean = self.frozen,
+    attribute: Response.Attribute = self.attribute,
+    // encodedResponse: AtomicReference[Option[Response.EncodedResponse]] = self.encodedResponse,
```
jdegoes: Dead code

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -77,44 +76,38 @@ final case class Response private (
...
+  ): Response = {
+    // Reset cached encoded response if status, headers or body are being changed.
+    // if (status != self.status || headers != self.headers || body != self.body) {
```
jdegoes: Dead code + comment

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: I have a subsequent branch started as a follow up to this one  in which I intend to address these.  I am afraid this one is starting to grow a bit bigger than I care for.

scottweaver: Makes sense.  I can put that on the "TODO" list for the next PR.

zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -28,216 +35,249 @@ private[zio] final case class ServerInboundHandler(
...
-              _         <- ZIO.attempt(ctx.writeAndFlush(jResponse))
-              flushed <- if (!jResponse.isInstanceOf[FullHttpResponse]) response.body.write(ctx) else ZIO.succeed(true)
-              _       <- ZIO.attempt(ctx.flush()).when(!flushed)
-            } yield ()
-
-        _ <- ZIO.attemptUnsafe(implicit u => setContentReadAttr(false))
-      } yield log.debug("Full write performed")
-    }
+      case jReq: HttpRequest =>
+        log.debug(s"HttpRequest: [${jReq.method()} ${jReq.uri()}]")
```
jdegoes, line 63: Debug statement, intended?

zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -28,216 +35,249 @@ private[zio] final case class ServerInboundHandler(
...
-      val nettyHttpVersion = nettyReq.protocolVersion()
-      val protocolVersion  = nettyHttpVersion match {
-        case HttpVersion.HTTP_1_0 => Version.Http_1_0
-        case HttpVersion.HTTP_1_1 => Version.Http_1_1
-        case _ => throw new IllegalArgumentException(s"Unsupported HTTP version: ${nettyHttpVersion}")
-      }
+        }
 
-      // TODO: We need to bring this back, probably not part of Request.
-      // val remoteAddress = ctx.channel().remoteAddress() match {
```
jdegoes, line 108: `Http` can gain a new subtype that allows accessing channel-related information that is backend independent, such as remote address.

zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
scottweaver: That's been there pre-ziverge.  I can remove it if you feel it is unneccesary.

zio-http/src/main/scala/zio/http/Response.scala
scottweaver: @jdegoes I went with removing the `frozen` field all together and replaced it with a simple, empty marker trait, `Frozen`.

scottweaver: I am rethinking this approach now that I started refactoring `Response` a bit more.

scottweaver: `Response` has been refactored to address this.

scottweaver: 'Response` now completely refactored and should address these issues.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -43,58 +45,83 @@ trait Body { self =>
...
   /**
    * Helper to create empty Body
    */
-  val empty: Body = new Body {
-    override def asChunk(implicit trace: Trace): Task[Chunk[Byte]]              = ZIO.succeed(Chunk.empty[Byte])
+  val empty: Body = EmptyBody
+
+  private[zio] object EmptyBody extends Body with UnsafeWriteable with UnsafeBytes {
+
+    override def asArray(implicit trace: Trace): Task[Array[Byte]] = ZIO.succeed(Array.empty)
```
jdegoes: `= ZIO.succeedNow(Array.empty)`

Moreover, since empty arrays are effectively immutable, you can store the whole thing into a private `val` in the companion object (`val zioEmptyArray = ZIO.succeedNow(Array.empty[Byte])`).

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -43,58 +45,83 @@ trait Body { self =>
...
+
+    override def asChunk(implicit trace: Trace): Task[Chunk[Byte]] = ZIO.succeed(Chunk.empty[Byte])
```
jdegoes: Ditto here.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -43,58 +45,83 @@ trait Body { self =>
...
+
+    override private[zio] def unsafeAsArray(implicit unsafe: Unsafe): Array[Byte] = Array.empty[Byte]
   }
 
   /**
    * Helper to create Body from AsciiString
    */
-  def fromAsciiString(asciiString: AsciiString): Body = new Body {
-    override def isComplete: Boolean = true
+  def fromAsciiString(asciiString: AsciiString): Body = AsciiStringBody(asciiString)
```
jdegoes, line 80: `AsciiString` is a netty thing, that exists for smaller / faster encoding / decoding. It would be a shame for such a Netty detail to leak into `zio.http.Body`. 

I think the best we should probably do is `CharSequenceBody`. Then on the Netty backend we could type case to see if it's an `AsciiString`. Then make some effort to actually use `AsciiString` as the `CharSequence` where useful to improving performance.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -43,58 +45,83 @@ trait Body { self =>
...
+
+  private[zio] final case class AsciiStringBody(asciiString: AsciiString)
+      extends Body
+      with UnsafeWriteable
+      with UnsafeBytes {
+
+    override def asArray(implicit trace: Trace): Task[Array[Byte]] = ZIO.succeed(asciiString.array())
```
jdegoes: I would make sure you are always using `ZIO.succeedNow`, both here and everywhere else.

jdegoes: Well, everywhere where laziness is neither necessary nor desired.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -43,58 +45,83 @@ trait Body { self =>
...
+
   }
 
   /**
    * Helper to create Body from ByteBuf
    */
-  def fromByteBuf(byteBuf: => ByteBuf): Body = new Body {
+  def fromByteBuf(byteBuf: => ByteBuf): Body = new ByteBufBody(byteBuf)
+
+  private[zio] final class ByteBufBody(byteBuf: => ByteBuf) extends Body with UnsafeWriteable with UnsafeBytes {
```
jdegoes: Is the laziness desired here?

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -43,58 +45,83 @@ trait Body { self =>
...
+
+    def byteBufData = byteBuf
```
jdegoes: If you don't make this a `lazy val`, then it will be recomputed every time. Which is fine if it's only computed once, but if it's computed more than once, then it could be useful to make it a `lazy val`.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -106,28 +133,45 @@ object Body {
...
+
   }
 
   /**
    * Helper to create Body from contents of a file
    */
-  def fromFile(file: => java.io.File, chunkSize: Int = 1024 * 4): Body = new Body {
+  def fromFile(file: => java.io.File, chunkSize: Int = 1024 * 4): Body = new FileBody(file, chunkSize)
+
+  private[zio] final class FileBody(file: => java.io.File, chunkSize: Int = 1024 * 4)
```
jdegoes: Same comments on laziness.

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -555,7 +555,10 @@ object ZClient {
               headers = headers,
             ),
           clientConfig = settings.copy(socketApp = Some(app.provideEnvironment(env))),
-        ).withFinalizer(_.close.orDie)
+        ).withFinalizer {
+          case resp: Response.CloseableResponse => resp.close.orDie
+          case _                                => ZIO.unit
```
jdegoes, line 560: 👍

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -0,0 +1,72 @@
...
+import io.netty.handler.codec.http.LastHttpContent
+import zio._
+import zio.http.Body
+import zio.http.Body._
+
+object NettyBodyWriter {
+
+  def write(body: Body, ctx: ChannelHandlerContext): Task[Boolean] = body match {
+    case body: ByteBufBody            =>
+      ZIO.attempt(ctx.write(body.byteBufData): Unit).as(false)
```
jdegoes: Much faster to do:
```suggestion
      ZIO.attempt { ctx.write(body.byteBufData); false })
```

Fewer ZIO ops = faster.

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -0,0 +1,72 @@
...
+    case body: FileBody               =>
+      ZIO.attempt {
+        val file = body.fileData
+        // Write the content.
+        ctx.write(new DefaultFileRegion(file, 0, file.length()))
+
+        // Write the end marker.
+        ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT): Unit
```
jdegoes: All these `: Unit` may invoke Scala's conversion of any value into unit, which will involve a function call in the best case, and an additional allocation in the worst case. I would instead just do `val _ = ...` if you just want to discard the value without a warning from the Scala compiler.

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -0,0 +1,72 @@
...
+        true
+      }
+    case AsyncBody(async)             =>
+      ZIO
+        .attempt(async { (ctx, msg, isLast) =>
+          ctx.writeAndFlush(msg)
+          if (isLast) ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT): Unit
+          else ctx.read(): Unit
+        })
+        .as(true)
```
jdegoes: Try to delete those `.as(true)`

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -0,0 +1,72 @@
...
+    case AsciiStringBody(asciiString) =>
+      ZIO.attempt(ctx.write(Unpooled.wrappedBuffer(asciiString.array())): Unit).as(false)
+    case StreamBody(stream)           =>
+      for {
+        _ <- stream.runForeachChunk(c => ZIO.succeed(ctx.writeAndFlush(Unpooled.wrappedBuffer(c.toArray))))
+        _ <- ZIO.attempt(ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT))
+      } yield true
```
jdegoes: `for` comprehensions yield a final `map` that can be eliminated. In performance-sensitive code, it's best to avoid them. Here you can fuse the `true` into the previous `ZIO.attempt`, and also eliminate the `ZIO.succeed` in favor of `ZIO.succeedNow`.

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -0,0 +1,72 @@
...
+    case AsyncBody(async)             =>
+      ZIO
+        .attempt(async { (ctx, msg, isLast) =>
+          ctx.writeAndFlush(msg)
+          if (isLast) ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT): Unit
+          else ctx.read(): Unit
+        })
+        .as(true)
+    case AsciiStringBody(asciiString) =>
+      ZIO.attempt(ctx.write(Unpooled.wrappedBuffer(asciiString.array())): Unit).as(false)
```
jdegoes: Watch out for these `: Unit` and `.as(...)`

zio-http/src/main/scala/zio/http/netty/NettyResponseEncoder.scala
```diff
@@ -0,0 +1,56 @@
...
+import io.netty.buffer.Unpooled
+import io.netty.handler.codec.http._
+import zio._
+import zio.http._
+
+import java.util.concurrent.ConcurrentHashMap
+
+private[zio] object NettyResponseEncoder {
+
+  private val frozenCache = new ConcurrentHashMap[Response, HttpResponse]()
```
jdegoes: Consider:
```suggestion
  private val frozenCache = new ConcurrentHashMap[Response, UIO[HttpResponse]]()
```

This way you can put the `HttpResponse` into `ZIO.succeedNow` and return it directly from `encode`.

zio-http/src/main/scala/zio/http/netty/NettyResponseEncoder.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  def encode(response: Response): ZIO[Any, Throwable, HttpResponse] = {
+    val body = response.body
+    if (body.isComplete)
```
jdegoes: Even in the cached case, you are executing `body.asArray.flatMap(...)`, which is a lot of unnecessary computation.

Instead, just check to see if the `frozenCache` has the response in it, and if it does, return that (pre-computed) effect directly.

Note that `Response#hashCode` and `Response#equals` will need to be very fast for such frozen path optimization to not interfere with the normal and common dynamic response case.

zio-http/src/main/scala/zio/http/netty/NettyResponseEncoder.scala
```diff
@@ -0,0 +1,56 @@
...
+      body.asArray.flatMap(bytes => ZIO.attemptUnsafe(implicit unsafe => fastEncode(response, bytes)))
+    else {
+      val jHeaders         = response.headers.encode
+      val hasContentLength = jHeaders.contains(HttpHeaderNames.CONTENT_LENGTH)
+      if (!hasContentLength) jHeaders.set(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED)
+      ZIO.succeed(new DefaultHttpResponse(HttpVersion.HTTP_1_1, response.status.asJava, jHeaders))
+    }
+  }
+
+  def fastEncode(response: Response, bytes: Array[Byte])(implicit unsafe: Unsafe): HttpResponse =
```
jdegoes, line 30: Note that here `Array[Byte]` is the basis, but if the response comes from `Chunk`, then calling `Chunk#toArray` (which is effectively what `Body#asArray` will do) will have a performance impact. 

It's possible the best way to handle this is to make a `ByteBuf` that is directly backed by a `Chunk[Byte]`. Rather than always trying to convert everything even to an `Array`.

Note that `zio.http.api` will go through _ZIO Schema_ codecs, which always use `Chunk[Byte]` (or streams).

zio-http/src/main/scala/zio/http/netty/NettyRuntime.scala
```diff
@@ -56,19 +56,23 @@ private[zio] trait NettyRuntime { self =>
       case Exit.Success(_)     =>
         log.debug(s"Completed Fiber: [${fiber.id}]")
         removeListener(close)
+        ensured()
```
jdegoes, line 59:
```suggestion
       if (ensured ne noopEnsuring) ensured()
```

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -12,12 +12,14 @@ import zio.stream.ZStream
...
 import java.nio.charset.Charset
 import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
+import java.nio.file._
 
 /**
  * Holds Body that needs to be written on the HttpChannel
  */
-trait Body { self =>
-  final def asArray(implicit trace: Trace): Task[Array[Byte]] = asChunk.map(_.toArray)
+sealed trait Body { self =>
```
jdegoes, line 20: Great to see this become `sealed`

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -1,120 +1,259 @@
...
-  def freeze(implicit trace: Trace): Task[Response] =
-    for {
-      encoded <- encode()
-    } yield self.copy(attribute = self.attribute.withEncodedResponse(encoded, self))
+  def body: Body
+
+  def copy(
+    status: Status = self.status,
+    headers: Headers = self.headers,
+    body: Body = self.body,
```
jdegoes, line 30: Should you add `frozen: Boolean`?

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -1,120 +1,259 @@
...
+      with InternalState {
+
+      override val parent: Response = self
+
+      override def serverTime: Boolean = true
+    }
+  }
+
+  // TODO: This needs to eventually be implementation agnostic.
+  private[zio] class NettyResponse(
```
jdegoes, line 226: 😱

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -1,120 +1,259 @@
...
+      override def frozen: Boolean  = true
+    }
+
+    override final def updateHeaders(update: Headers => Headers): Response = copy(headers = update(headers))
+
+    override final def withServerTime: Response = new NettyResponse(body, channelContext, headers, status)
+      with InternalState {
+      override val parent: Response = self
+
+      override def serverTime: Boolean = true
```
jdegoes, line 251: Some effort here and in request / body is made to preserve the structure of the original class. But it could be easier for maintenance and not really slower in general (due to happy path) just defaulting to a generic case class that has "all possible fields". Then `copy` on `Response` would just always return that, and it could be overriden selectively if useful in some happy paths. Would keep code much smaller.

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
scottweaver: None of the operations on `ChannelHandlerContext` will ever throw an exception, instead it will notify the underlying `ChannelFuture`.  So, in the interest of performance I will look at converting all calls to the context operations wrapped with either `attempt` or `succeed` to be `succeedNow`.

zio-http/src/main/scala/zio/http/netty/NettyResponseEncoder.scala
scottweaver: There are places where `fastEncode` is used where it completely bypasses needing ZIO runtime.  Wrapping the cached response in `UIO` would prevent us from using these optimized code pathes.

Alternatively, we could store a tuple `(HttpResponse, UIO[HttpResponse])` that way we get the best of both approach with an additional allocation of a tuple.

jdegoes (approved):

PR #1747, opened 2022-11-02
Sketching WS API redesign (WIP)

dkarlinsky:
Very rough draft - WIP, doesn't yet compile :)
 * Switch to a new SocketAppEvent ADT for socket app to handle
 * Socket app function returns SocketAppAction instead of manipulating Channel
 * SocketApp can use SocketAppChannel to send messages directly (e.g. from another fiber)

### DIFF
--- a/zio-http-example/src/main/scala/example/WebSocketEcho.scala
+++ b/zio-http-example/src/main/scala/example/WebSocketEcho.scala
@@ -4,7 +4,7 @@ import zio._
 import zio.http.ChannelEvent.ChannelRead
 import zio.http._
 import zio.http.model.Method
-import zio.http.socket.{WebSocketChannelEvent, WebSocketFrame}
+import zio.http.socket.{SocketApp, SocketAppAction, SocketAppEvent, WebSocketChannelEvent, WebSocketFrame}
 
 object WebSocketEcho extends ZIOAppDefault {
   private val socket: Http[Any, Throwable, WebSocketChannelEvent, Unit] =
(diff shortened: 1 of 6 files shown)


jdegoes:
Promising! I think I can see where this can go, in a way that makes it unopinionated and also fast. Will write more tomorrow.

zio-http-example/src/main/scala/example/WebSocketEcho.scala
```diff
@@ -19,10 +19,19 @@ object WebSocketEcho extends ZIOAppDefault {
         ch.write(WebSocketFrame.text(text)).repeatN(10) *> ch.flush
     }
 
+  private val socketApp = SocketApp {
+    case SocketAppEvent.FrameReceived(WebSocketFrame.Text("FOO")) =>
```
strelec, line 23: Why not flatten this into TextFrameReceived?

swoogles:
> Promising! I think I can see where this can go, in a way that makes it unopinionated and also fast. Will write more tomorrow.

Any further thoughts on this @jdegoes ?
I can't do much more with the Socket test capabilities until we know how this will be hammered out.

zio-http/src/main/scala/zio/http/socket/SocketApp.scala
```diff
@@ -1,14 +1,15 @@
...
 import zio.http._
 import zio.http.model.Headers
+import zio.http.socket.SocketApp.{SocketAppHandler, SocketHandler}
 import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
 
 final case class SocketApp[-R](
   decoder: SocketDecoder = SocketDecoder.default,
   protocol: SocketProtocol = SocketProtocol.default,
-  message: Option[ChannelEvent[WebSocketFrame, WebSocketFrame] => ZIO[R, Throwable, Any]] = None,
+  message: SocketAppEvent => Option[ZIO[R, Throwable, SocketAppAction]] = SocketAppHandler.noOp.lift
```
tusharmath, line 12: What if I want to perform a side-effect before deciding what socket action to take?

dkarlinsky: @tusharmath this is basically a lifted `PartialFunction[SocketAppEvent, ZIO[R, Throwable, SocketAppAction]]` - I wanted the user API to accept a `PartialFunction` but found the lifted version easier to work with internally.
If there is no pure match that can decide on whether to handle an event, your effect can always evaluate to `SocketAppAction.NoOp`
To tell the truth, I'm not at all sure about this encoding - was just a rough sketch that I wanted to share with @swoogles

jdegoes:
@dkarlinsky Wrote a strawman proposal here:

https://github.com/zio/zio-http/issues/2100

Please take a look, feedback welcome!

PR #1753, opened 2022-11-05
Test client socket improvements and tests

swoogles:
Added a few more contract tests


zio-http/src/main/scala/zio/http/package.scala
```diff
@@ -1,13 +1,16 @@
...
-  type UHttpApp                              = HttpApp[Any, Nothing]
-  type RHttpApp[-R]                          = HttpApp[R, Throwable]
-  type EHttpApp                              = HttpApp[Any, Throwable]
-  type UHttp[-A, +B]                         = Http[Any, Nothing, A, B]
+  type HttpApp[-R, +E] = Http[R, E, Request, Response]
+  type UHttpApp        = HttpApp[Any, Nothing]
+  type RHttpApp[-R]    = HttpApp[R, Throwable]
+  type EHttpApp        = HttpApp[Any, Throwable]
+  type UHttp[-A, +B]   = Http[Any, Nothing, A, B]
+  type HttpSocket      = Http[Any, Throwable, WebSocketChannelEvent, Unit]
```
swoogles: This addition probably warrants the most scrutiny. I found it useful, but I know expanding the API here should not be done lightly.

jdegoes: I would probably not introduce this just yet given the looming changes for web socket. As for the other ones, I'm not personally a fan of the benefits (of less typing) given the costs (of indirection) but don't have a huge preference either way.

swoogles: Removed

jdegoes (approved):

PR #1754, opened 2022-11-06
Change withContentType to add the header instead of replacing it

NavidJalali:
This seems like a bug to me, bug calling `withContentType` basically replaces all the headers with the content type header 💀


NavidJalali:
Okay I just realized this whole thing is being rewritten

PR #1872, opened 2022-12-08
Add HTML output encoding to protect against Cross-site scripting (XSS)

frevib:


jdegoes (approved):

jdegoes:
Can you please run `scalafmt` and `scalafix` from the root project? (Or `fmt`)

devsprint:
the intention is to provide a text alternative to a status code. There should be a simpler fix, without requiring to add a new dependency to the project.

frevib:
@devsprint If you want to echo back user input in the HTML, you have to encode your output. Else you could use a static text.

frevib:
Probably you know this, but this is a serious security issue. Consumers of ZIO HTTP are vulnerable to their sessions being stolen or other things that can be done with XSS. Expect this exploit is already being used in the wild. I can help where possible to fix this asap.

devsprint:
@frevib I'm aware of the XSS security issues. The intention was to provide an alternate text message to a http status code. What I'm saying is that particular issue could be fixed in a different way without adding a new dependency to the project. HTML response is in the hand of the users of the library, so they can take their own actions on how to protect their app from XSS security issues

frevib:
@devsprint I understand, better not having another dependency. When users misconfigure their 404 page, they will be shown ZIO HTTP's vulnerable 404 page.

jdegoes:
@frevib We definitely need to fix this and I like your approach. Do you know if maybe there is a small Scala implementation available under a permissive license that would be suitable? An advantage is that it could work on Scala Native or other platforms.

frevib:
@jdegoes I've searched but I couldn't find a lightweight Scala implementation. The creators of this dependency, [OWASP](https://github.com/OWASP), are an authority in security. So you can trust them to do it right.

As for this problem, we only need to encode a couple characters that are used in HTML entities. `Encode.forHtml(path)` only does very simple character replacement under the hood. These are the characters that need to be encoded: https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#output-encoding-rules-summary, and the OWASP encoder does exactly that: https://github.com/OWASP/owasp-java-encoder/blob/main/core/src/main/java/org/owasp/encoder/XMLEncoder.java#L106

If you want to avoid the dependency on OWASP I suggest writing a simple `encode` function that simply replaces the "dangerous" characters for it's HTML-encoded parts. I you're OK with this, I'll update this PR with this solution. Let me know.

jdegoes:
@frevib That sounds like a great plan. Maybe we can make it package private but accessible to the whole code base, so as other use cases arise, we can take advantage of your work? Thank you! 🙏

frevib:
This should be it. It's not much code, but be careful when to use it. As far as I can see, ZIO HTTP only echos back user input inside HTML context. That means the output is inserted somewhere between HTML entities, like `<div>output</div>` or `<p>output</p>`. If user input would echo'ed inside JavaScript context, or CSS or something else, this newly written `encodeHtml` will not protect you against XSS. XSS is a nasty security issue for which there is no silver bullet. Only way to deal with it, is to understand it and be aware when coding.

zio-http/src/main/scala/zio/http/security/OutputEncoder.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  private def encodeHtmlChar(char: Char): String = char match {
+    case '&'     => "&amp"
+    case '<'     => "&lt;"
+    case '>'     => "&gt;"
+    case '"'     => "&quot;"
+    case '\''    => "&#x27;"
+    case '/'     => "&#x2F;"
+    case _ @char => char.toString
+  }
```
tusharmath, line 50: We can optimize the performance a bit here —

1. Initialize with `StringBuilder` with the expected length of the output string.
2. Save the output string allocations as a Charsequence inside of `OutputEncoder` once.

frevib: Thx for the review, changes have been made. Not sure if your 2. is met, could you have a look?

tusharmath: If you save the encoded versions of each char in a val you won't be allocating it for every call.

frevib: Got it, pushed the changes.

tusharmath (approved):

strelec:
I commented on https://github.com/zio/zio-http/issues/1873.

PR #1876, opened 2022-12-08
Refactoring introducing rich text codec

devsprint:


zio-http/src/main/scala/zio/http/api/internal/HeaderValueCodecs.scala
```diff
@@ -0,0 +1,134 @@
...
+      _.foldLeft(Chunk.empty[(String, Option[Double])]) {
+        case (acc, Left(_))      => acc
+        case (acc, Right(value)) =>
+          acc :+ value
+      },
+      _.map(Right(_)),
+    )
+
+  val acceptLanguageCodec: RichTextCodec[AcceptLanguage] = rawAcceptLanguageCodec.transform(
+    raw => AcceptedLanguages(raw.map { case (name, value) => AcceptedLanguage(name, value) }),
```
devsprint: @jdegoes codec can return directly the Header values defined for each header, so we can avoid to call `transform`  and to implement the 2 methods `from...` and `to..` . Basically the all header values will be just simple ADT.

zio-http/src/main/scala/zio/http/api/internal/HeaderValueCodecs.scala
```diff
@@ -0,0 +1,134 @@
...
+    },
+    {
+      case (name, Some(value)) => Left((Chunk.fromArray(name.toArray), (";q=", value)))
+      case (name, None)        => Right(Chunk.fromArray(name.toArray))
+    },
+  )
+
+  private val rawAcceptLanguageCodec: RichTextCodec[Chunk[(String, Option[Double])]] =
+    ((langComplete ~ RichTextCodec.comma) | langComplete).repeat.transform(
+      _.foldLeft(Chunk.empty[(String, Option[Double])]) {
```
devsprint: this codec is returning only first match, it is not returning the second match that is separated by comma from the first

zio-http/src/main/scala/zio/http/api/internal/RichTextCodec.scala
```diff
@@ -186,6 +187,33 @@ object RichTextCodec {
...
+
+  val slash: RichTextCodec[Char] = char('/')
+
+  val star: RichTextCodec[Char] = char('*')
+
+  val colon: RichTextCodec[Char] = char(':')
+
+  val semicolon: RichTextCodec[Char] = char(';')
+
+  def commaSeparatedMultiValues[A](codec: RichTextCodec[A]): RichTextCodec[Chunk[A]] =
```
jdegoes: This is called `rep1sep` in some libraries and would be a method on `RichTextCodec`, with it accepting the separator character (as a separate `RichTextCodec[Unit]`).

Probably worth doing it that way so the code will be more familiar.

jdegoes: Also, generally a good idea to save companion object for constructors rather than unary operators like this one (which properly belong on the data type, for easy dot-syntax).

zio-http/src/main/scala/zio/http/api/HttpCodec.scala
```diff
@@ -210,8 +208,11 @@ object HttpCodec extends HeaderCodecs with QueryCodecs with RouteCodecs {
     def erase: Method[Any] = self.asInstanceOf[Method[Any]]
   }
 
-  private[api] final case class Header[A](name: String, textCodec: TextCodec[A], optional: Boolean)
-      extends Atom[CodecType.Header, A] {
+  private[api] final case class Header[A](
+    name: String,
+    textCodec: Either[TextCodec[A], RichTextCodec[A]],
```
jdegoes: This is good for now, eventually we want to delete the `Either` and just use `RichTextCodec` because the other one is too weak for documentation and client purposes.

jdegoes:
@devsprint Please re-open when you're ready to get this merged in!

PR #1882, opened 2022-12-14
Partial function model

### LINKED ISSUE #1627: Authentication breaks, if you combine `Http`s with different ways to authenticate
**Describe the bug**
Combining two apps with different auth middlewares leads to unexpected 401 returns.

**To Reproduce**
This happens in version `2.0.0-RC10`
Let's assume we have have 3 Http instances, one without auth, one with basic auth and one with oauth. We use the zio-http defined auth middleware.
Let's start a server like this
```scala
Server.start(8080, noAuthApp ++ basicAuthApp ++ oauthApp)
```
The endpoints of the first apps two behave correctly, but the last one always returns 401, independent of the auth headers content. If you change the order, everything after the first authenticated `Http` will return 401.

You can find a reproducer [here](https://github.com/987Nabil/zio-http-middleware-add-up)

When we build the server like this
```scala
Server.make(Server.port(8080), Server.app(noAuthApp) ++ Server.app(basicAuthApp) ++ Server.app(oauthApp))
```
it is even worse. Now all endpoints return always 401.

**Expected behaviour**
The routes of a `Http` are authenticated based on the middleware added to them, even when combined with others

### LINKED ISSUE #1860: Middlewares seems to  influence other http applications.
**Describe the bug**
I don't know if it is a bug or a feature but middlewares seems to influence other http applications.

It is an issue in my case since I have a fallback function to send a custom 404 mesage.

**To Reproduce**
If we take the following code:
```

val app1 = ???
val app2 = ???
val app3 = ???

val app = app1 ++ app2 @@ bearerAuth(token => true) ++ app3
```

The app3 will also have the bearer middleware but not the app1.

**Expected behaviour**
app3 doesn't need token.

### LINKED ISSUE #1871: Refactor model of Http
The current model of `Http[R, E, A, B]` is:

```scala
A => ZIO[R, Option[E], A]
```

Unfortunately, this means that a middleware **cannot know if a handler wants to handle a route until it's too late**.

Middleware incoming actions must be run **if and only if** a handler wants to handle a route. But there is no ability to do that, because in order to see if a handler wants to handle a route, it must be invoked with the input. At which point, it's too late.

The current middleware always runs, **every single time**, regardless of whether a handler wants to handle a route, **which is incorrect and leads to numerous issues**.

In order to solve this problem, a handler must be able to indicate **whether or not** it wishes to handle a route, independently of actually handling a route.

Therefore, the model must change to:

```scala
PartialFunction[A, ZIO[R, E, A]]
```

This proposal is **strictly less powerful** than the old model. If this proves to be troublesome, we can split it into a `isDefinedAt: A => ZIO[R, E, Boolean]` and `apply: A => ZIO[R, E, A]` pair. But hopefully that won't be necessary.

vigoo:
# Switching to partial function model

- Resolves https://github.com/zio/zio-http/issues/1871
- Resolves https://github.com/zio/zio-http/issues/1860
- Resolves https://github.com/zio/zio-http/issues/1627

The primary change of this _pull request_ is switching `Http`'s execution from being `A => HExit[R, E, B]` to `PartialFunction[A, HExit[R, E, B]`. This allows us to always be able to decide if a given `Http` will handle a given input or not.

This change cannot be done without some trade-offs. The following sections summarize these.

## HExit
The `HExit` type represents the result of a `Http` distinguishing values that need to be calculated by a _ZIO effect_ and pure success/failures. 

Previously `HExit.empty` represented an _"unhandled result"_. This is no longer possible - a `Http` cannot dynamically decide to not handle an input, so `HExit.empty` and the `HExit#defaultWith` combinator has been removed.

## Http

The core change in `Http` is that it's 

```scala
def execute(a: A): HExit[R, E, B]
```

method became

```scala
def execute: PartialFunction[A, HExit[R, E, B]]
```

Some new helper functions are added to evaluate this as a regular function into a `HExit`, or to a `ZIO` effect returning the `HExit`.

It is not possible to express all combinators that we had as a _partial function_ so some restrictions had to be made, and several new subclasses of `Http` were created to support as much of the combinators as possible.

### Total vs partial Http

A subset of the `Http` subclasses are subclasses of `Http.Total`. These are guaranteed to handle _all_ their inputs, so their `execute` function is a total function.

Some operators that were available for combining arbitrary `Http` values are now requiring that one or both their parameters to be a `Http.Total`. 

Here is the complete list of those that require `Http.Total` on the right hand side:

- `flatMap/>>=`
- `andThen/>>>`
- `zipRight/*>`
- `catchAll`
- `catchAllCause`
- `catchAllDefect`
- `catchNonFatalOrDie`
- `catchSome`
- `catchSomeDefect
- `flatten`
- `foldCauseHttp`
- `foldHttp`

Only available on `Http.Total` (on the left hand side):

- `codecMiddleware/\/`
- `compose/<<<`
- `contraFlatMap`
- `contramapZIO`

## Removed operators
Removed:

- `tap`
- `tapAll`
- `collect`
- `collectZIO`
- `Http.fromOptionFunction`

Note that although I removed the `tap` variants that were _"tapping"_ into another `Http`, I kept the effectful `tapZIO` etc which make more sense in my opinion (as the only way a `Http` given in the RHS of `tap` can do anything is to return an effect that calculates its output which is then discarded).

The `collect` and `collectZIO` is only removed as an operator on a `Http` value (they are still available as constructors). The reason is that in case the LHS returns with an effectful `HExit` we cannot decide the RHS's applicability.

The `fromOptionFunction` constructor was removed because it required the capability to _"dynamically"_ decide to not handle an input - it's `None` failure value corresponds to `HExit.empty`.

## Other differences
The `foldHttp` and `foldCauseHttp` operators no longer allows you to observe _"empty"_ (unhandled) cases. Not handling an input has been lifted to the partial function application level, so within a `Http` it is not possible to handle it.

## Http.Total in result types
Because of the above limitations it is important to mark everything as `Http.Total` that is - to limit the effect of these restrictions for the user. 

The following changes are supporting this:

### Total in result types
Many `Http` constructors now have `Http.Total` as the result type, so when starting to build a `Http` value from these we have `Http.Total`, without the restrictions of being used as RHS in some operators.

### Duplicated operators
Many operators on `Http` preserve whether it is total or not - for example `map` on `Http` gives back `Http` but `map` on a `Http.Total` returns `Http.Total`. This helps composing `Http` values together without loosing tha ability to use them in the restricted operators.

To support this, the `Http.Total` trait itself overrides many of the operators, providing a more specific return type.

Also some of the `Http` subclasses had to be duplicated, having a partial and a total variant:

- `Race` and `RaceTotal`
- `Chain` and `ChainTotal`
- `FoldHttp` and `FoldHttpTotal`
- `Combine` and `CombineTotal`
- `Aspect` and `AspectTotal`
- `OrElse` and `OrElseTotal`
- `ApplyMiddleware` and `ApplyMiddlewareTotal`

The primary reason for these duplicates is to provide the more usable variant of the operators for the user (keep things `Total` when we can) - but they also allow more performant execution implementation in some cases.

Note that `Aspect/AspectTotal` is a new construct, supporting the `provide*` and `tap*` methods on both partial and total `Http`s.

### New operators

I added a new operator that makes a partial `Http` to a `Http.Total`:

- `withFallback`

It's the same as `++` (`defaultWith`) but it requires the RHS to be total, and by doing that it can always return a `Http.Total`, making it a good choice to use in places where we have to make a `Http` total to pass it as a parameter.

## Middleware

The whole change originates from some unexpected behavior of how `Middleware`s are applied to inner parts of a complex `Http` value.

To fix this, the semantics of `Middleware` had been changed, and the new `Http.Total` concept supports this on the type level as well.

Previously `Middleware` was simply a wrapped function `Http => Http`. This is too much power for the new partial function based model - if we apply a middleware to a `Http` we still need to be able to calculate `isDefinedAt` on the resulting `Http`.

So the new `Middleware` semantics are:

- `Middleware` is a `Http.Total => Http.Total` function
- An applied `Middleware` only transforms an `Http` when it is defined on it's input.

The second rule makes middlewares applicable to _any_ `Http` (not just `Http.Total`) even though the definition is a transformation from total to total:

- For `Http.Total`, applying the `Middleware` is equivalent to performing the transformation in definition-time
- For partial `Http`s though, the execution is will be a _partial function_ that is defined for the same inputs as the original partial function, and **when it is defined**, it's value is transformed by the middleware.

So **middlewares no longer can change the set of accepted inputs** - and this is reflected in its type (`Http.Total => Http.Total`). 

It also means that if a middleware runs any effect, for example an effectful authorization function, it is guaranteed that this effect will only run if the targeted `Http` handles the input. This is exactly the fix needed for the linked issues.

#### Update
The above described `Middleware` is still not strict enough as it allows it to manipulate the _input_ of a `Http`, so checking the unapplied `Http`'s `isDefinedAt` is no longer possible. 

For that reason I introduced type-level tracking of whether a `Middleware` is manipulating the input channel or not, and if it is a pure function or not.

This is currently tracked in the last type parameter of `Middleware` and it can be one of the following cases:
- `IT.Id[A]` means that it is not changing the input
- `IT.Contramap[AIn, AOut]` means it is changing the input with a pure function and that can be used to implement `isDefinedAt` in the result `Http`
- `IT.Impossible[A]` represents the cases when we cannot implement `isDefinedAt`, for example because the middleware used `contramapZIO`. 

Middlewares with `IT.Impossible` input transformer can only applied to `Http.Total`.

The names may not be the best and can be changed. 

I'm also thinking about whether we need one more case for _extending_ the set of accepted values in a middleware. An example for that is `Cors` which adds handling of some extra requests (`OPTION` requests). 

### Removed middlewares and operators
Making `Middleware` less powerful means I had to delete some of the existing middlewares and middleware operators:

- `allow`
- `allowZIO`
- `collect`
- `collectZIO`
- `empty`

These were all transforming `Http` into another accepting less inputs than the original.

## API
No user visible changes were required on the higher level API (`zio.http.api`).


zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -552,87 +461,49 @@ sealed trait Http[-R, +E, -A, +B] { self =>
...
-      case Combine(self, other) => {
-        self.execute(a) match {
-          case HExit.Empty            => other.execute(a)
-          case exit: HExit.Success[_] => exit.asInstanceOf[HExit[R, E, B]]
-          case exit: HExit.Failure[_] => exit.asInstanceOf[HExit[R, E, B]]
-          case exit @ HExit.Effect(_) => exit.defaultWith(other.execute(a)).asInstanceOf[HExit[R, E, B]]
-        }
-      }
-    }
+  private[zio] def execute: PartialFunction[A, HExit[R, E, B]]
```
jdegoes: This will necessitate a `PartialFunction` allocation every time `execute` is invoked. The overhead can be eliminated by `extend`ing `PartialFunction`.

vigoo: Made this change; I had to rename the existing `apply` that was returning a `ZIO` to `toZIO`.

vigoo:
I realized there is a problem with middleware application for partial `Http`s, will fix tomorrow. Do not merge

vigoo:
Let's close this as "attempt 1", I'll open a new PR with new description when I'm ready with my next attempt.

PR #1893, opened 2022-12-22
propagate a client request interrupt - fixes #1863

### LINKED ISSUE #1863: Client Request Interruption Not Working as Expected
**Describe the bug**
I have a ZIO HTTP client that connects to a ZIO HTTP server where the server handler sleeps for 10 seconds.  If the client has a timeout of a couple seconds or after a couple seconds interrupts the request, I'd expect the server handler's `onExit` to get a `Failure` as this is what happens when a request is made from curl and cancelled before it is completed.  However, the server handler instead succeeds normally.

**To Reproduce**
See: https://github.com/jamesward/zio-http-client-interrupt
Run the server & client and observe that the timeout request and the interrupted client request do not cause the server handler to produce failures.

**Expected behaviour**
Interrupting a client request should cause a failure on the server handler (I think but maybe I'm doing it wrong).

jamesward:
This might be completely wrong but did work in my use case and didn't break any zio-http tests.  Note that there should probably be a test to validate this but that might be beyond my abilities to create.

### DIFF
--- a/zio-http/src/main/scala/zio/http/ZClient.scala
+++ b/zio-http/src/main/scala/zio/http/ZClient.scala
@@ -1,6 +1,6 @@
 package zio.http
 
-import io.netty.channel.{ChannelHandler, Channel => JChannel}
+import io.netty.channel.{ChannelFuture, ChannelHandler, Channel => JChannel}
 import io.netty.handler.codec.http._
 import io.netty.handler.codec.http.websocketx.WebSocketClientProtocolHandler
 import io.netty.handler.flow.FlowControlHandler
@@ -587,7 +587,10 @@ object ZClient {
         jReq       <- encode(request)
         _          <- internalRequest(request, jReq, onResponse, clientConfig)(Unsafe.unsafe, trace)
           .catchAll(cause => onResponse.fail(cause))
-        res        <- onResponse.await
+        res        <- onResponse.await.onInterrupt {
+          // propagate the interrupt to the Promise
+          onResponse.interrupt
+        }
       } yield res
 
     /**
@@ -606,7 +609,7 @@ object ZClient {
             channelScope <- Scope.make
             _            <- ZIO.uninterruptibleMask { restore =>
               for {
-                channel      <-
+                channel <-
                   restore {
                     connectionPool
                       .get(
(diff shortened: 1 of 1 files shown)


jamesward:
I've done a bit more testing and discovered this fix isn't quite right.  I'll be updating it soon.

jamesward:
Ok, I think this is better now.  It now disconnects the pipeline on interrupt.

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -633,25 +637,39 @@ object ZClient {
...
-                        connectionPool
-                          .invalidate(channel)
-                          .when(channelState == ChannelState.Invalid)
-                      }
-                      .flatMap { _ =>
-                        channelScope.close(exit)
-                      }
-                      .uninterruptible
+                    if (exit.isInterrupted) {
+                      // do not try to reset the channel because it is disconnected
```
vigoo, line 651: Should not we invalidate the channel in the connection pool in case of interrupt? 
If it is not reset it is not reusable for the next connection - this should work the same as the case when reset fails below, I think.

jamesward: I'm not sure. I was hoping there'd be some tests that validate the connection pool.  But maybe not. If the pipeline is disconnected, can the connection be reused? I don't fully understand how all this works.

vigoo: Unfortunately the existing connection pool tests are not covering interruption. It would be nice to have tests for it, for the full matrix of possibilities (streaming/non-streaming, http 1.1/1.0)

When the request finishes normally, there are some cleanup steps to be done on the netty pipeline - some handlers that are dynamically added during handling the request need to be removed, while the ones that are "static" for the connection has to remain there. I think when interrupting we cannot guarantee that this state is restored. Or maybe we can but then we need to explicitly do it.

vigoo: The reason I cannot give 100% straight answer is that even though I wrote that connection pool code I don't fully remember how it works anymore, need to refresh my understanding of Netty :)

jamesward:
Closing in favor of: #1966

PR #1916, opened 2023-01-06
Redesigned model

### LINKED ISSUE #1627: Authentication breaks, if you combine `Http`s with different ways to authenticate

### LINKED ISSUE #1860: Middlewares seems to  influence other http applications.

### LINKED ISSUE #1871: Refactor model of Http

### LINKED ISSUE #1930: Remove HExit
If we merge https://github.com/zio/zio-http/pull/1916 and https://github.com/zio/zio/pull/7693 gets released `HExit` can be replaced with `Exit` from ZIO.

vigoo:
# Redesigned core model

- Resolves https://github.com/zio/zio-http/issues/1871
- Resolves https://github.com/zio/zio-http/issues/1860
- Resolves https://github.com/zio/zio-http/issues/1627
- Resolves https://github.com/zio/zio-http/issues/1930

This is my second attempt to fix the core problem described in https://github.com/zio/zio-http/issues/1871 (the previous was https://github.com/zio/zio-http/pull/1882). 

## The primary changes
- splitting the `Http[R, E, A, B]` type into `Handler` and `Route`
- `HExit` can no longer represent "unhandled input"
- the final runnable http server has to have `Response` as its error channel
- middlewares are reimplemented on top of the new core types with more straightforward semantics and less generic combinators

#### Handler
The old `Http[R, E, A, B]` type represented something like `A => ZIO[R, Option[E], B]` in which it combined transformation from input to output with possibly not handling the given input ("routing") and defined powerful combinators to express complex routes and transformations with it.

In the new model we define:

```scala
trait Handler[-R, +Err, -In, +Out]

type RequestHandler[-R, +Err] = Handler[R, Err, Request, Response]
type AppHandler[-R]           = RequestHandler[R, Response]
```

as a possibly effectful transformation from `In` to `Out` with no possibility to not handle an input (other than failing). So `Handler` is basically a `In => ZIO[R, Err, Out]` function.

(Note that in the real implementation I'm still using `HExit` on both the `Handler` and `Router` level to make it possible to avoid the ZIO runtime in some cases, but that is just an optimization.)

All the constructors and operators from `Http` that are not are depending on partial results are implemented on `Handler` (so no `collect`, `orElse`, `++`, etc).

#### Route
The remaining features of the old `Http` type are moved into the type `Route`:

```scala
trait Route[-R, +Err, -In, +Out]

type HttpRoute[-R, +Err] = Route[R, Err, Request, Response]
type App[-R]             = HttpRoute[R, Response]
```

`Route` corresponds to something like `In => ZIO[R, Err, Option[Handler[R, Err, In, Out]]]`: it can either evaluate to a `Handler`, or  state that it is not handling the request, or fail, and it can do this in an effectful way if needed. Route has the "usual" zio-http constructors like `collect` and they can be combined with `++`. 

A `Handler` can always be converted to a `Route` (`.toRoute`). A `Route` can be converted to a `Handler` by providing a default route (`.toHandler(default)`). 

The zio-http server itself requires `App[R]` which is `Route[R, Response, Request, Response]`. A `HttpRoute[R, Err]` can be converted to this using the default error reporting with `route.withDefaultErrorResponse`.

#### Implementation notes
- Both `Handler` and `Route` is now implemented with direct executable encoding unlike the previous `Http` type. It was easier to experiment this way
- `HExit` can no longer represent `Empty` (just like in my previous PR). I kept this limitation from the previous attempt to make sure `Handler` can never represent unhandled state. As `Route` still needs to represent it and I did not want to wrap everything in option, I'm representing it with `HExit.succeed(null)` and all methods directly returning such `HExit`s  are protected by `zio.Unsafe` .

### Middlewares
The old "middleware" concept which was a `Http => Http` function is slightly modified in this model.

First of all we define `HandlerAspect` and `RouteAspect` as transformers of the two types:

```scala
trait HandlerAspect[-R, +Err, +AIn, -AOut, -BIn, +BOut]

type RequestHandlerMiddleware[-R, +Err] = HandlerAspect[R, Err, Request, Response, Request, Response]
type AppHandlerMiddleware[-R]           = RequestHandlerMiddleware[R, Response]

trait RouteAspect[-R, +Err, +AIn, -AOut, -BIn, +BOut]

type HttpRouteMiddleware[-R, +Err] = RouteAspect[R, Err, Request, Response, Request, Response]
type AppMiddleware[-R]             = HttpRouteMiddleware[R, Response]
```

These are equivalent to `ZIOAspect` and they have the same expression power:
- they can be `apply`d (`@@`) to a handler or route
- they can be combined with `>>>` (`andThen`)

No other generic combinators are defined on this level to keep it simple.

On top of these we reimplement all the "non-generic" middlewares as either handler or router aspects. Only those need to be router which need access to the routing itself. For example:
- `metrics`, to be able to count unhandled cases
- `CORS` because it adds new routes
- `dropTrailingSlash` because it "doubles" the allowed routes
- `allow` and `allowZIO` because they dynamically change the routing

Everything else (of the predefined middlewares in zio-http) turns out to be expressible as handler aspects. 

To make it more approachable, most are defined using the `HttpRouteMiddleware` and `RequestHandlerMiddleware` type aliases, and most of them are implemented directly as functions on handlers, instead of composing very generic middlewares.

All these "middlewares", both handler and route aspect based, are collected to the `Middleware` object as before to make it more backward compatible.

#### Applying middlewares
The only thing remaining is to define how these middlewares are applied in a complex application, and this is where the new model must solve the reported linked issues.

I think by having this well defined, type level separation between *routes* and *handlers* the middleware application rule can be simple enough to not be surprising for users (but this is the weakest point of the whole change, of course):

With the actually used type aliases:
- Applying a `RequestHandlerMiddleware[R, E] to a `RequestHandler[R, E]` gives back a `RequestHandler[R, E]`
- Applying a `HttpRouteMiddleware[R, E] to a `HttpRoute[R, E]` gives back a new `HttpRoute[R, E]`
- Applying a `RequestHandlerMiddleware[R, E] to a `HttpRoute[R, E]` gives back a new `HttpRoute[R, E]` with **no change in the routing logic**, and in case it routes to a handler, the given handler gets modified with the middleware.

More generally:
- Applying a _handler aspect_ to a `Route` requires that the `In` type is not changed by the aspect
- The handler aspect is always applied after the routing so for example if it performs an effect, or contramap's the input, it can never affect the routing

### Results
This model solves the linked issues (added unit tests for them), and it does not really reduce the power of zio-http - including effectful routing etc. that were mentioned in the linked issue.

Freedom of combining middlewares via various combinators has been limited but I did it intentionally.

We can also rename either `Handler` or `Route` back to `Http` but I wanted to have fresh names to make it easier to discuss the new model. 

Some boilerplate (`.toRoute` conversions etc) could be reduced by adding more operators or implicit conversions but I'm not sure about it - and for now I wanted to keep everything as clear as possible for discussion.

All existing tests are passing.

## Remaining:
- [x] restore fallback feature of the file/resource handlers
- [x] mark unsafe operators of Route 
- [x] Restore the scaladoc
- [x] Write explanation into the PR
- [ ] Update docs
- [x] Review implicit traces


zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,486 @@
...
+  def collectHandler[In]: CollectHandler[In] = new CollectHandler[In](())
+
+  def collectHExit[In]: CollectHExit[In] = new CollectHExit[In](())
+
+  def collectRoute[In]: CollectRoute[In] = new CollectRoute[In](())
+
+  def collectZIO[In]: CollectZIO[In] = new CollectZIO[In](())
+
+  def empty: Route[Any, Nothing, Any, Nothing] =
+    (_: Any) => HExit.succeed(null)
```
jdegoes: `HExit.succeed(null)`?

Is this truly an empty with respect to `++`?

vigoo: Yes as I mentioned somewhere in the long PR description I removed `HExit.empty` (so users can't use it to define `Handler`s but did not introduce a `HExitOrEmpty` or used `Option` but in `Route` `HExit.succeed(null)` represents the unhandled case.

Maybe it's an unnecessary premature optimization, I would be happy to not do that way.

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,486 @@
...
+
+trait Route[-R, +Err, -In, +Out] { self =>
+  protected def run(in: In): HExit[R, Err, Handler[R, Err, In, Out]]
+
+  final def >>>[R1 <: R, Err1 >: Err, In1 >: Out, Out1](
+    handler: Handler[R1, Err1, In1, Out1],
+  ): Route[R1, Err1, In, Out1] =
+    Route.fromHandlerHExit[In] { in =>
+      self.run(in).map { firstHandler =>
+        if (firstHandler eq null) null.asInstanceOf[Handler[R1, Err1, In, Out1]]
```
jdegoes: I see that `null` handlers are used as empty. 

Another possibility is to make `Handler` a `sealed trait`, so it can be extended by `Empty` and `Handled`. The API won't need to change, I don't think, but then you can allow the user code to participate in writing combinators like this one, which should properly pattern match on handler and treat the `Empty` case separately.

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,486 @@
...
+        else HExit.succeed(aspect(handler))
+      }
+    }
+
+  final def ++[R1 <: R, Err1 >: Err, In1 <: In, Out1 >: Out](
+    that: Route[R1, Err1, In1, Out1],
+  ): Route[R1, Err1, In1, Out1] =
+    self.defaultWith(that)
+
+  final def defaultWith[R1 <: R, Err1 >: Err, In1 <: In, Out1 >: Out](
```
jdegoes: This might be called `orElse` as it represents a kind of fallback.

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,486 @@
...
+import zio._
+import zio.http.model.{Headers, MediaType, Status}
+import zio.http.socket.{SocketApp, WebSocketChannelEvent}
+import zio.stream.ZStream
+
+import java.io.{File, FileNotFoundException}
+import java.nio.file.Paths
+import java.util.zip.ZipFile
+
+trait Route[-R, +Err, -In, +Out] { self =>
```
jdegoes: As strange as it might be, executable encodings like this one can sometimes be slower than the declarative encoding due to dynamic dispatch. Possibly worth making this `sealed` in case we want to change our minds later.

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,486 @@
...
+    trace: Trace,
+  ): SocketApp[R] =
+    SocketApp(event =>
+      self.toZIO(event).catchAll {
+        case Some(value) => ZIO.fail(value)
+        case None        => ZIO.unit
+      },
+    )
+
+  final def toZIO(in: In)(implicit trace: Trace): ZIO[R, Option[Err], Out] =
```
jdegoes: Maybe `execute` for consistency with the old `Http`?

zio-http/src/main/scala/zio/http/Route.scala
vigoo: That's a good idea, I will give it a try; Maybe `Handler` could be kept as the name of the "handled handler" and just add a base trait called `OptionalHandler`? (I just feel `Handler` looks better in the types than `Handled`)

zio-http/src/main/scala/zio/http/Route.scala
vigoo: I agree, I was following the old API here

zio-http/src/main/scala/zio/http/Route.scala
vigoo: `execute` was 

```scala
  final private[zio] def execute(a: A)(implicit trace: Trace): HExit[R, E, B]
```

so this is not exactly the same. And I wanted to prefer consistency in providing the different `toXyz` methods here.

zio-http/src/main/scala/zio/http/Route.scala
vigoo: I was thinking about it, but then decided its faster to build up the new model with executable encoding (as I was not sure how exactly it will look like). 

So let's make it `sealed` now, and experiment with declarative encoding in a follow-up PR if we decide to merge this?

zio-http/src/main/scala/zio/http/package.scala
```diff
@@ -1,17 +1,20 @@
...
-  type RHttpApp[-R]    = HttpApp[R, Throwable]
-  type EHttpApp        = HttpApp[Any, Throwable]
-  type UHttp[-A, +B]   = Http[Any, Nothing, A, B]
 
-  type ResponseZIO[-R, +E]                   = ZIO[R, E, Response]
-  type UMiddleware[+AIn, -BIn, -AOut, +BOut] = Middleware[Any, Nothing, AIn, BIn, AOut, BOut]
+  type RequestHandler[-R, +Err] = Handler[R, Err, Request, Response]
+  type AppHandler[-R]           = RequestHandler[R, Response]
+
+  type RequestHandlerMiddleware[-R, +Err] = HandlerAspect[R, Err, Request, Response, Request, Response]
```
jdegoes: I suggest, maybe, that we define a new trait called `Middleware[R, Err]`, and have it extend `HandlerAspect[R, Err, Request, Response, Request, Response]`. Then we'll have a reasonable place to put all the `Middleware` (into the companion object of `Middleware`).

vigoo: I'm not sure I understand this suggestion. 

- if I understand correctly `MiddlewareAspect` is what above defined as `RequestHandlerMiddleware` - but in other cases we used `XYAspect` as something that transforms `XY` so that name does not feel right to me
- What is the reason for making it a trait?
- Currently there are a couple of traits defining all kind of middlewares (ones working on handlers, also some working on routes): `RequestHandlerMiddlewares`, `HttpRouteMiddlewares`, `Cors`, `Metrics`, etc. And there is a single`Middleware` object merging all these to make them easier to find. What subset would go into `Middleware` by this suggestion, and where would the remaining ones go?

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: Yes, makes sense.

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: Yes!

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: Did it have a name before? Usually `toXYZ` methods do not require anything as input, e.g. `Queue#toStream`.

zio-http/src/main/scala/zio/http/package.scala
jdegoes: I am not 100% sure, but it would be nice to retain the terminology of middleware because it's well-established, whereas aspect terminology, however applicable and consistent, is less familiar and less domain-specific.

In addition, having a `trait` `Middleware` improves organization: because we are always teaching Scala developers to look into the companion object of `XYZ` to find constructors for `XYZ`.

Using `Middleware` as a "package", which is what we are doing in this pull request, helps to organize the middlewares, but not in any way that is discoverable; and in a way that is quite different than the typical type/companion object relationship.

Now, you mention we have some handler middleware, and route middleware. 

Well, that's unfortunate, because the `Middleware` companion object to a `Middleware` type (which extends `HandlerAspect`) should only contain handler middleware. Not route middleware.

But if there are only a couple route middleware, then maybe we can just implement them as methods on route. When I looked, I saw only really 1 route aspect ("route middleware") which is a sign the majority of what a user would think of as "middleware" moved to handler middleware. If true, then the middleware name could be reused just for route middleware, and that would also result in less code breakage.

vigoo: There is not only one route middleware, but much less than handler ones. Basically those middlewares that somehow extend or limit the set of handled requests, or needs to observe the event of not handling one:


- `metrics`, to be able to count unhandled cases
- `CORS` because it adds new routes (handles `OPTIONS`)
- `dropTrailingSlash` because it "doubles" the allowed routes
- `allow` and `allowZIO` because they dynamically change the routing

vigoo: I was thinking of also exposing each "kind" of middleware in a companion object of the corresponding type alias for example `RequestHandlerMiddleware` (by pulling in the same underlying trait that I mixed in `Middleware` now).

Not sure if that would help or make it even worse :)

zio-http/src/main/scala/zio/http/Route.scala
vigoo: It was the `Http#apply`

vigoo: I felt confusing that `apply` is the ZIO version and there are other similar functions with `in` parameter but different result (`HExit`, etc) with other names and wanted to make more consistent. But maybe it was not a good idea.

adamgfraser: Maybe these should be `runXYZ`? You are running the route by giving it an input.

zio-http/src/main/scala/zio/http/package.scala
adamgfraser: We could also just wrap them like we did for streams, sinks, and pipelines with channels. Maybe route middleware is `RouteMiddleware` and handler middleware is just `Middleware` if we think that is the much more common case?

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,503 @@
...
+import zio._
+import zio.http.model.{Headers, MediaType, Status}
+import zio.http.socket.{SocketApp, WebSocketChannelEvent}
+import zio.stream.ZStream
+
+import java.io.{File, FileNotFoundException}
+import java.nio.file.Paths
+import java.util.zip.ZipFile
+
+sealed trait Route[-R, +Err, -In, +Out] { self =>
```
jdegoes: This is a "polishing" question:

The name `Route` is not exactly accurate, it's more like `RoutingLogic`, which is too weird and long to be the right name here.

Given the name is an approximation anyway, could it make sense to instead call this `Http`? 

The rational would be:

1. Fewer changes required in the many libraries that already use ZIO HTTP.
2. It's a little nice for the "entry point" data type in a library to have a name that is similar to the library's name (e.g. `ZIO` in ZIO Core).

Thoughts?

vigoo: I was thinking the same but wanted to keep a separate new name while the new model is being discussed. (Although in general I don’t think that’s a very good name for what this trait does either :)

vigoo: The question is what do we call the related types then. For example the Http that is “really” http (request to response), and the http vs handler middleware’s

adamgfraser: Makes sense to me. An `Http` is a program, that, given an input, either handles it with a handler, or chooses not to handle it.

zio-http/src/main/scala/zio/http/Handler.scala
```diff
@@ -0,0 +1,632 @@
...
+import zio.stream.ZStream
+
+import java.io.{File, FileNotFoundException}
+import java.nio.charset.Charset
+import java.nio.file.Paths
+import java.util.zip.ZipFile
+import scala.reflect.ClassTag
+import scala.util.control.NonFatal
+
+sealed trait OptionalHandler[-R, +Err, -In, +Out]
```
jdegoes: Looking at this it made me wonder, why have the `sealed trait` here instead of `Route`?

If `Route` is a sealed trait, then every route is either a pattern and a handler, or the empty route, which has neither pattern nor handler.

(I know this was my suggestion but it seems cleaner to push the sum type to the route.)

Thoughts?

vigoo: The route itself cannot be the empty case because the route evaluates the input possibly with an effect to figure out the handler/unhandled. So in that case route would be a pattern to not handler but route. I’m not sure that’s better - for the current set of operators it’s not needed at least

adamgfraser: Is the empty case a route that doesn't want to handle anything?

The other possibility this opens that I don't think the current model does is you could have a route which as a result of its input returns another route. Basically `Http` becomes a "tree" that eventually either chooses not to handle something or has leaves with handlers. Don't think that is necessary for anything currently but it is kind of interesting and opens up another mode of composition.

vigoo: Actually I think this change makes my "middleware semantics" invalid, or at lest harder to understand - that you can apply a "handler middleware" to a route and it is just get applied "immediately on the next level" without affecting the routing. I was trying to keep the composition on the route level to minimum (to just have orElse, basically) to keep it simple and to solve the original issue of weird behavior of middlewares.

So even though it may make it possible to express more interesting composed routes let's think it through how it relates to applying middlewares and not running into surprises.

adamgfraser: 💯

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: How about this proposal:

 - `Http` &mdash; a program that, when given an input, either handles it with a handler, or chooses not to handle it. Mental model: `Set[(RoutePattern, RouteHandler)]` (my personal favorite), or `Route => Option[RouteHandler]` (classically).
 - `Handler` &mdash; Handles an input to produce an output.
 - `HttpAspect` &mdash; Transforms an `Http`.
 - `HandlerAspect` &mdash; Transforms a `Handler`.

This doesn't allow reuse of `Middleware` terminology at all, which is unfortunate. But it would be really weird to use `Middleware` for `Handler` and not `Http`, yet the vast majority of aspects are `Handler` transformers, not `Http` transformers.

By the way, another possibility occurs to me: `Route` (`Http`) already allows application of `HandlerAspect`, because there is a way to lift any `HandlerAspect` into an `HttpAspect`: by choosing some subset of handlers on which to apply the aspect (currently: "forall").

I think that means `HandlerAspect` could extend `HttpAspect`.

That means that `myApp @@ <xyz>` can work for any `<xyz>`, be they handler aspects or http aspects. That, in turn, implies we could use the `Middleware` name for `HttpAspect`, and mix into the companion object not merely all of the `HttpAspect` (which extend or contract the table size) but all of the `HandlerAspect`. 

In this alternative:

 - `Http` &mdash; a program that, when given an input, either handles it with a handler, or chooses not to handle it. Mental model: `Set[(RoutePattern, RouteHandler)]` (my personal favorite), or `Route => Option[RouteHandler]` (classically).
 - `Handler` &mdash; Handles an input to produce an output.
 - `Middleware` &mdash; Transforms an `Http`. Companion object includes constructor for ALL `Middleware`, including those in companion object of `HandlerAspect`.
 - `HandlerAspect` &mdash; Transforms a `Handler`. Extends `Middleware` using "for all" semantic.

Thoughts?

zio-http/src/main/scala/zio/http/Handler.scala
jdegoes: I think that `Http` (`Route`) is basically a routing table, something like:

| Http Method | Path                 | Handler               |
| ------------ | -------------------- | --------------------- |
| GET         | /users                | getAllUsers            |
| GET         | /users/{id}           | getUserById            |
| POST        | /users                | createNewUser           |
| PUT         | /users/{id}           | updateUserById          |
| DELETE      | /users/{id}           | deleteUserById         |
| GET         | /orders                | getAllOrders            |
| GET         | /orders/{id}           | getOrderById            |
| POST        | /orders                | createNewOrder           |
| PUT         | /orders/{id}           | updateOrderById          |
| DELETE      | /orders/{id}           | deleteOrderById         |

Why can't a table be empty? 

Right now we fake emptiness by saying it's a route that returns a handler that doesn't want to handle its input (null or Unhandled). But that seems equivalent to me (but worse than) just supporting an empty route.

A "tree" almost suggests to me a way to unify this and the endpoint stuff. But doing that while preserving the current API is not possible. 

It's worth discussing this a bit more as I think `OptionalHandler` is really the only unsightly bit in this new model.

adamgfraser: 👍

zio-http/src/main/scala/zio/http/Route.scala
adamgfraser: I think I like the second. The idea of applying the aspect to all handlers indicates how potentially we could support the world where the `Http` is not a "flat" tree, basically the table has nesting in your table analogy. That seems like it could be useful. Like you want to apply some logic to all the `users` handlers which could be done very cleanly if all the users routes can themselves be their own table which is then embedded into a larger table.

zio-http-example/src/main/scala/example/AuthenticationServer.scala
```diff
@@ -39,19 +40,19 @@ object AuthenticationServer extends ZIOAppDefault {
   }
 
   // Http app that is accessible only via a jwt token
-  def user: UHttpApp = Http.collect[Request] { case Method.GET -> !! / "user" / name / "greet" =>
+  def user: HttpRoute[Any, Nothing] = Http.collect[Request] { case Method.GET -> !! / "user" / name / "greet" =>
```
jdegoes: Maybe we can keep the old type aliases to avoid the updating code??

zio-http-example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -30,11 +30,11 @@ object PlainTextBenchmarkServer extends ZIOAppDefault {
     .withServer(STATIC_SERVER_NAME)
     .freeze
 
-  private def plainTextApp(response: Response) =
-    Http.fromHExit(HExit.succeed(response)).whenPathEq(plaintextPath)
+  private def plainTextApp(response: Response): HttpRoute[Any, Nothing] =
+    Handler.response(response).toRoute.whenPathEq(plaintextPath)
```
jdegoes: `toRoute`, some old temrinology still exists??

vigoo: Fixed

zio-http-example/src/main/scala/example/AuthenticationServer.scala
vigoo: I think these old aliases are conflicting with the new ones:

```scala
type HttpRoute[-R, +Err] = Http[R, Err, Request, Response]
type App[-R]             = HttpRoute[R, Response]
```

Probably the `HttpRoute` name is not very good either anymore (I defined it when it was `Route[R, Err, Req, Resp]`).

So we should review all the names together. I don't like much these old aliases but maybe it is more important to keep them for people who already use them.

zio-http-example/src/main/scala/example/PlainTextBenchmarkServer.scala
vigoo: although I think `.toRoute` was a nicer name than `.toHttp` (and `Route` vs `Handler` is a better distinction than `Http` vs `Handler` which does not make much sense to me)

zio-http/src/main/scala/zio/http/HExit.scala
```diff
@@ -1,7 +1,7 @@
 package zio.http
 
 import zio.http.HExit.Effect
-import zio.{Cause, Trace, ZIO}
+import zio.{Cause, Tag, Trace, ZEnvironment, ZIO, ZLayer}
```
jdegoes: Are we going to be able to replace this in favor of `zio.Exit` with appropriate optimizations applied?

vigoo: I opened a ticket about it few days ago. Should we do it in this PR now that 2.0.6 is out?

vigoo: (I think we would be able to simply replace it with `Exit` now)

vigoo: Pushed it here, with some workaround to override `flatMap/map/mapBoth` (also opened https://github.com/zio/zio/pull/7714 but it creates stack safety issues in streams)

zio-http/src/main/scala/zio/http/package.scala
```diff
@@ -1,17 +1,20 @@
...
 
-  type ResponseZIO[-R, +E]                   = ZIO[R, E, Response]
-  type UMiddleware[+AIn, -BIn, -AOut, +BOut] = Middleware[Any, Nothing, AIn, BIn, AOut, BOut]
+  type RequestHandler[-R, +Err] = Handler[R, Err, Request, Response]
+  type AppHandler[-R]           = RequestHandler[R, Response]
+
+  type RequestHandlerMiddleware[-R, +Err] = HandlerMiddleware[R, Err, Request, Response, Request, Response]
+  type AppHandlerMiddleware[-R]           = RequestHandlerMiddleware[R, Response]
+
+  type HttpRoute[-R, +Err] = Http[R, Err, Request, Response]
```
jdegoes: @vigoo Let's not use any `Route` terminology, and let's minimize name changes (so keep `HttpApp` instead of `App`, etc.).

vigoo: Old names for reference:

```scala
  type HttpApp[-R, +E] = Http[R, E, Request, Response]
  type UHttpApp        = HttpApp[Any, Nothing]
  type RHttpApp[-R]    = HttpApp[R, Throwable]
  type EHttpApp        = HttpApp[Any, Throwable]
  type UHttp[-A, +B]   = Http[Any, Nothing, A, B]

  type ResponseZIO[-R, +E]                   = ZIO[R, E, Response]
  type UMiddleware[+AIn, -BIn, -AOut, +BOut] = Middleware[Any, Nothing, AIn, BIn, AOut, BOut]
```

vigoo: so the old `HttpApp` is what currently is `HttpRoute`. If that's `HttpApp` we still need to call somehow the "final" thing, which fixes the `Err` to `Response` - that's what I called `App[R]`. Do you have a better idea?

vigoo: Maybe we just don't need type aliases for that and use `HttpApp[R, Response]` in the code.

jdegoes (approved):

jdegoes (approved):

jamesward:
One impact of this change is that previously interrupting a ZIO handing a request would close the connection, not returning a response.  I was using this behavior in https://github.com/jamesward/easyracer
Now the interrupted ZIO causes an *Internal Server Error* response (or the default error handler) to be returned instead.  Maybe that is the desired behavior, but I wanted to check.

vigoo:
> Maybe that is the desired behavior, but I wanted to check.

I did not intentionally change this, but not sure if it was by design. @jdegoes what do you think?

jdegoes:
@vigoo I think interruption maps most directly to immediate close. Can we do that in a followup?

jamesward:
I've filed a bug to track: https://github.com/zio/zio-http/issues/2022

vigoo:
@jamesward thanks!

PR #1939, opened 2023-01-14
Name route codecs

987Nabil:
For documentation (generators) just a type of a route segment is not
enough. We need a name to generate meaningful docs.


987Nabil:
@jdegoes follow up work from our last session.

jdegoes (approved):

PR #1942, opened 2023-01-17
Fixing specified dependency and CI badge in README.md

fmarl:
Hey,

just a small fix of the CI badge as well as the dependency in the README.md and index.md after I learned that this is auto-generated. The sbt dependency seems to be fixed in the README.md but not in the index.md.


khajavi:
@fxttr
Thank you again for your contribution.
But it seems it does not make sense anymore as the issue is resolved. Anyway, if you don't think so, please resolve the conflict.

PR #1944, opened 2023-01-18
Remove Scope From Input Type Of Layers

adamgfraser:
Since layers are resourceful they should not depend on a `Scope`.


vigoo:
I tried to do this once and somehow having multiple `EventLoopGroups` or `ChannelFactories` (don't remember exactly) at the same time caused some collision in Netty in some global state. But seems like now all tests are passing so it is not an issue anymore, good news!

vigoo (approved):

PR #1948, opened 2023-01-20
Helper methods on Client without requiring Body

### LINKED ISSUE #1929: Simpler Client.get without body
Doing simple GET requests require to pass an empty body explicitly:

```scala
client.get("...")(Body.empty)
```

However having bodies for GET requests is maybe a 1% use case, in the vast majority of the time you don't send a body on a GET request.

I'd like to be able to just call `client.get("...")` with no body by default.

vigoo:
Resolves https://github.com/zio/zio-http/issues/1929

Also adds some other methods to the API than `GET` and `PUT`

### DIFF
--- a/zio-http/src/main/scala/zio/http/ZClient.scala
+++ b/zio-http/src/main/scala/zio/http/ZClient.scala
@@ -110,6 +110,9 @@ trait ZClient[-Env, -In, +Err, +Out] { self =>
         )
     }
 
+  final def delete(pathSuffix: String)(body: In)(implicit trace: Trace): ZIO[Env, Err, Out] =
+    request(Method.DELETE, pathSuffix, body)
+
   final def dieOn(
     f: Err => Boolean,
   )(implicit ev1: Err IsSubtypeOfError Throwable, ev2: CanFail[Err], trace: Trace): ZClient[Env, In, Err, Out] =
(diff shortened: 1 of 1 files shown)


zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -744,6 +756,18 @@ object ZClient {
     }
   }
 
+  final implicit class ClientSyntax(val client: Client) extends AnyVal {
```
jdegoes: You could generalize this a bit and just require the In type to be `Body` rather than fixing all the type parameters.

However, it strikes me this will probably work better with tooling if added as a method on `ZClient` that requires implicit evidence of type equality (like `List#flatten`).

vigoo: I wanted to do it like that (methods with implicit evidence) but it did not work. Because of the combination of multiple parameter lists and implicits, I think. It could not choose the proper method when having the two defined on the same interface:

```scala
def get(pathSuffix: String)(body: In)(implicit trace: Trace)
def get(pathSuffix: String)(implicit trace: Trace, ev: ...)
```

jdegoes: All right 👍

jdegoes: Oh wait, another thing we could do: why are the original methods which require a body curried? Maybe we could just express them as multiple parameter?

It is not like most people will use curly braces `{ }` to define the body.

vigoo: Not sure why they are curried. I'll change it.

vigoo: Probably the idea was to define functions for different endpoints like `val store = client.post("/user") _` but I think one would either write a wrapper class anyway, or generate it from OpenAPI etc.

adamgfraser (approved):

jdegoes (approved):

PR #1949, opened 2023-01-20
Better handling of default ports in URLs

### LINKED ISSUE #1897: Host header value always includes port
**Describe the bug**
The `host` header is populated with this code

```scala
val headers = req.url.host match {
  case Some(host) => encodedReqHeaders.set(HttpHeaderNames.HOST, req.url.port.fold(host)(port => s"$host:$port"))
  case _          => encodedReqHeaders
}
```

But if `url.host` is defined then `url.port` is also defined because they both come from `Location.Absolute` where they are non-optional.

This makes the client impossible to use with S3 pre-signed URLs, which include the `host` header in the signed portion and excludes the port from the value when it is default for the scheme. Excluding the port when it matches the scheme default is also how browsers behave. So an S3 presigned URL which works in the browser will not work via zio-http client.

**Expected behaviour**
The port is only included in the `host` header value when it is different from the default port for the scheme, or when forced to via configuration

vigoo:
Resolves https://github.com/zio/zio-http/issues/1897

### DIFF
--- a/zio-http/src/main/scala/zio/http/URL.scala
+++ b/zio-http/src/main/scala/zio/http/URL.scala
@@ -2,7 +2,7 @@ package zio.http
 
 import io.netty.handler.codec.http.QueryStringEncoder
 import zio.Chunk
-import zio.http.URL.{Fragment, Location}
+import zio.http.URL.{Fragment, Location, portFromScheme}
 import zio.http.model.Scheme
 
 import java.net.{MalformedURLException, URI}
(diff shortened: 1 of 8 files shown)


zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -25,10 +25,19 @@ final case class URL(
 
   def encode: String = URL.encode(self)
 
-  def host: Option[String] = kind match {
+  def host: Option[String]                 = kind match {
     case URL.Location.Relative      => None
     case abs: URL.Location.Absolute => Option(abs.host)
   }
+  def hostWithOptionalPort: Option[String] =
```
jdegoes, line 32: Maybe this is telling us that `host` should return a `Host` that has `hostname: String` and `port: Option[Int]`?

vigoo: We already have `URL.Location.Absolute(scheme, host, port)` which is similar.. Not sure. Can we open a separate ticket for possibly redesigning the `URL` model?

jdegoes: Sounds good.

jdegoes (approved):

PR #1950, opened 2023-01-20
Fix client connection and memory leaks

### LINKED ISSUE #1025: Http Client Resource Leak
**Describe the bug**
Http client does not clean up resources after the call has succeeded. This means that after a specific number of requests depending on the configurations of the machine, no more requests can be made. 

**To Reproduce**
Create an infinite stream of Http requests using the client.

I tested it with the following versions:

ZIO Version = "1.0.13"
ZIO Http Version = "1.0.0.0-RC24"

```scala
import zio._
import zio.console.putStrLn
import zhttp.service.{ChannelFactory,EventLoopGroup}
import zhttp.service.Client
import zhttp.service.Client.ClientRequest
import zhttp.http.URL
import zhttp.http.Method
import zio.stream.ZStream
import zio.stream.ZTransducer

object Main extends App {

  val sleep = "http://localhost:8000/healthcheck"

  def get(url: URL) = 
  Client
    .request(ClientRequest(url, Method.GET))
    .flatMap(result => result.bodyAsString.map((_, result.status.asJava.code)))

  def stream(url: URL, batchSize: Int) = 
    ZStream
      .repeat(())
      .transduce(ZTransducer.fromEffect(
        ZIO.collectAllPar(List.fill(batchSize)(get(url)))
      ))

  val app = for {
    url <- ZIO.fromEither(URL.fromString(sleep))
    _ <- stream(url, 100).zipWithIndex.foreach(s => putStrLn(s.toString))
  } yield ()

    app
    .catchAllCause(c => putStrLn(c.toString))
    .provideCustomLayer(env)
    .exitCode
}
```

https://user-images.githubusercontent.com/5600005/153767685-cd50175e-f2fb-4bbe-a613-27a94706b789.mp4

As well as

ZIO Version = "2.0.0-RC2"
ZIO Http Version = "2.0.0-RC3"

```scala
import zio.*
import zio.Console.printLine
import zhttp.service.{ChannelFactory,EventLoopGroup}
import zhttp.service.Client
import zhttp.service.Client.ClientRequest
import zhttp.http.URL
import zhttp.http.Method
import zio.stream.ZStream

object Main extends ZIOAppDefault:

  val sleep = "http://localhost:8000/healthcheck"

  def get(url: URL) = 
    Client
      .request(ClientRequest(url, Method.GET))
      .flatMap(result => result.bodyAsString.map((_, result.status.asJava.code)))

  def stream(url: URL, batchSize: Int) = 
    ZStream
      .repeat(get(url))
      .grouped(batchSize)
      .flatMap(calls => ZStream.fromZIO(ZIO.collectAllPar(calls)))

  val app = for {
    url <- ZIO.fromEither(URL.fromString(sleep))
    _ <- stream(url, 100).zipWithIndex.runForeach(printLine(_))
  } yield ()

  override def run: URIO[zio.ZEnv, ExitCode] =
    app
    .catchAllCause(printLine(_))
    .provideCustomLayer(env)
    .exitCode
```

**Expected behaviour**
The stream goes on forever.

**Actual behaviour**
The stream halts after it cannot acquire the necessary resources to continue. (I could not reproduce this, but @girdharshubham managed to get an actual netty exception, or sometimes a 0 exit).

https://user-images.githubusercontent.com/5600005/153767798-f64fb66b-d3dd-4830-b5e7-a5db09c0e71a.mp4

image

image

**Additional context**
ATM I am not sure if its hitting the file descriptor limit or maximum running process limit.

### LINKED ISSUE #1588: Http client connection leak
**Describe the bug**
When using the http client a lot, resources are eventually exhausted.
A new connection is established on every client request and they are not closed. It doesn't matter if client is reused or not.
This is tested on 2.0.0-RC11

**To Reproduce**
Run this code and check connections with ss as explained in the output.

```
package test

import zhttp.service.server.ServerChannelFactory
import zhttp.service.{ChannelFactory, Client, EventLoopGroup, Server}
import zio._
import zio.test.TestAspect.withLiveClock
import zio.test._

object HttpTest extends ZIOSpecDefault {

  def spec = test("Http client test") {
    ZIO.scoped {
      for {
        _ <- Server.port(12345).make.orDie
        client <- Client.make[Any]
        _ <- client.request("http://localhost:12345")
        _ <- client.request("http://localhost:12345")
        _ <- client.request("http://localhost:12345")
        _ <- client.request("http://localhost:12345")
        _ <- client.request("http://localhost:12345")
        _ <- client.request("http://localhost:12345")
        _ <- ZIO.debug("When test is now hanging, run the following code (on Linux) and see that there are 6 connections open")
        _ <- ZIO.debug("ss -o state established '( sport = :12345 )'")
        _ <- ZIO.never
      } yield assertTrue(1 == 1)
    }
  }.provideShared(EventLoopGroup.auto(), ChannelFactory.auto, ServerChannelFactory.auto) @@ withLiveClock
}
```

### LINKED ISSUE #1867: HTTP Client OOM
**Describe the bug**
I'm trying to consume 10M events from ZIO-Kafka and send HTTP request for each using ZIO HTTP client. unfortunately, I got the below exception when It reaches the consumption of the 2M+ events

```
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "ZScheduler-Supervisor"
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "ZScheduler-2"
```

note that consuming the 10M and committing them using offset-batches without sending the HTTP requests works fine without using any additional memory and the GC performs great.

**Screenshots**

***sending the HTTP request for each event (fails at the consumption of 2M events)***
jconsole
Screen Shot 2022-12-07 at 11 51 05 AM

jmap 
Screen Shot 2022-12-07 at 11 51 58 AM

***without sending the HTTP request for each event***
jconsole
Screen Shot 2022-12-07 at 11 56 22 AM

jmap
Screen Shot 2022-12-07 at 11 56 40 AM

**Desktop:**
 - OS: macOS Monterey  
 - Version ZIO-HTTP version 0.0.3 (I've tried to use 2.0.0R but due to the  limitation of client-config I switched to 0.0.3 as it contains more configs for the client ex: connection-pool)

vigoo:
Resolves https://github.com/zio/zio-http/issues/1867
Resolves https://github.com/zio/zio-http/issues/1588
Resolves https://github.com/zio/zio-http/issues/1025

- Fixes the memory leak when using connection pools
- Fixes the connection leak when not using connection pools

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
+++ b/zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
@@ -1,22 +1,17 @@
 package zio.http.netty.client
 
 import io.netty.bootstrap.Bootstrap
-import io.netty.channel.{
-  Channel => JChannel,
-  ChannelFactory => JChannelFactory,
-  ChannelInitializer,
-  EventLoopGroup => JEventLoopGroup,
-}
+import io.netty.channel.{Channel => JChannel, ChannelFactory => JChannelFactory, ChannelInitializer, EventLoopGroup => JEventLoopGroup}
 import io.netty.handler.codec.http.{HttpClientCodec, HttpContentDecompressor}
 import io.netty.handler.logging.LoggingHandler
 import io.netty.handler.proxy.HttpProxyHandler
+import zio._
 import zio.http.URL.Location
 import zio.http._
 import zio.http.logging.LogLevel
 import zio.http.netty.NettyFutureExecutor
 import zio.http.service._
 import zio.http.service.logging.LogLevelTransform.LogLevelWrapper
-import zio.{Duration, Scope, ZIO, ZKeyedPool, ZLayer}
 
 import java.net.InetSocketAddress
 
@@ -31,6 +26,8 @@ trait ConnectionPool {
   ): ZIO[Scope, Throwable, JChannel]
 
   def invalidate(channel: JChannel): ZIO[Any, Nothing, Unit]
+
+  def enableKeepAlive: Boolean
 }
 
 object ConnectionPool {
@@ -97,6 +94,7 @@ object ConnectionPool {
     }
 
     ZIO.attempt {
+      println("Creating new channel")
       val bootstrap = new Bootstrap()
         .channelFactory(channelFactory)
         .group(eventLoopGroup)
@@ -136,6 +134,9 @@ object ConnectionPool {
 
     override def invalidate(channel: JChannel): ZIO[Any, Nothing, Unit] =
       ZIO.unit
+
+    override def enableKeepAlive: Boolean =
+      false
   }
 
   case class PoolKey(
@@ -148,6 +149,7 @@ object ConnectionPool {
 
   private class ZioConnectionPool(
     pool: ZKeyedPool[Throwable, PoolKey, JChannel],
+    activeChannels: Ref[Set[JChannel]],
   ) extends ConnectionPool {
     override def get(
       location: Location.Absolute,
@@ -162,16 +164,27 @@ object ConnectionPool {
           pool
             .get(PoolKey(location, proxy, sslOptions, maxHeaderSize, decompression)),
         ).tap { channel =>
-          restore(
-            NettyFutureExecutor.executed(channel.closeFuture()),
-          ).zipRight(
-            pool.invalidate(channel),
-          ).forkDaemon
+          for {
+            channels        <- activeChannels.get
+            updatedChannels <-
+              if (channels.contains(channel)) ZIO.succeed(channels)
+              else
+                restore(
+                  NettyFutureExecutor.executed(channel.closeFuture()),
+                ).zipRight(
+                  pool.invalidate(channel),
+                ).forkDaemon
+                  .as(channels + channel)
+            _               <- activeChannels.set(updatedChannels)
+          } yield ()
         }
       }
 
     override def invalidate(channel: JChannel): ZIO[Any, Nothing, Unit] =
-      pool.invalidate(channel)
+      pool.invalidate(channel) *>
+        activeChannels.update(_ - channel)
+
+    override def enableKeepAlive: Boolean = true
   }
 
   val disabled: ZLayer[EventLoopGroup with ChannelFactory, Nothing, ConnectionPool] =
@@ -265,7 +278,8 @@ object ConnectionPool {
           ),
         (key: PoolKey) => size(key.location),
       )
-    } yield new ZioConnectionPool(keyedPool)
+      activeChannels <- Ref.make(Set.empty[JChannel])
+    } yield new ZioConnectionPool(keyedPool, activeChannels)
 
   private def createDynamic(
     min: Int,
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
```diff
@@ -97,6 +99,7 @@ object ConnectionPool {
     }
 
     ZIO.attempt {
+      println("Creating new channel")
```
adamgfraser: Stray debug statement?

vigoo: yes :)

zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
```diff
@@ -148,6 +153,7 @@ object ConnectionPool {
 
   private class ZioConnectionPool(
     pool: ZKeyedPool[Throwable, PoolKey, JChannel],
+    activeChannels: Ref[Set[JChannel]],
```
adamgfraser: Why isn't this being handled by the pool? Is the pool not being eager enough in releasing resources?

vigoo: There is an external event (the channel's "closed" future) that we want to listen for, and invalidate the channel in the pool when it happens - for example the channel was closed by a network error, etc.

The memory leak was that a new listener was added to this event on every `get` from the pool - but we only want to do it once.

Maybe it would be better to add this into `createChannel` but then we need to pass a promise of the pool to it. I could not decide which one is better.

jdegoes: I would say the `Promise` option (actually you only need to pass `Promise#await`, not the whole `Promise`). Because this one has significant overhead below and is not interruption safe with respect to consistency of the pool and `activeChannels` (which could be fixed, but another approach may be simpler).

vigoo: Done

zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
```diff
@@ -282,20 +293,30 @@ object ConnectionPool {
...
+                NettyFutureExecutor.executed(channel.closeFuture()),
+              ).zipRight(
+                poolPromise.await.flatMap(_.invalidate(channel)),
+              ).forkDaemon
+            }
+          },
         (key: PoolKey) => min(key.location) to max(key.location),
         (key: PoolKey) => ttl(key.location),
       )
+      activeChannels <- Ref.make(Set.empty[JChannel])
```
adamgfraser: Can't this be deleted now?

zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
adamgfraser: Yes this makes more sense to me too. We need to invalidate the item from the pool when the channel closed future completes. The promise is basically just a way to give us a self reference to the pool in the pool constructor.

zio-http/src/main/scala/zio/http/netty/client/ConnectionPool.scala
vigoo: deleted it

adamgfraser (approved):

PR #1953, opened 2023-01-21
Friendlier Forms in ZIO HTTP

### LINKED ISSUE #1559: Multipart form data server support

### LINKED ISSUE #1560: Multipart form data client support
Client can send multipart POST requests

scottweaver:
Friendly Forms MK II
===============

This is pretty much complete rewrite of the previous Friendly Forms API that only initially supported 'application/x-www-form-urlencoded'.

User API: Form
---------------

'zio.http.forms.Form' provides the user the tools they need to:

- Encode 'application/x-www-form-urlencoded' to a 'zio.http.Body'.
- Decode a 'zio.http.Body' that is encoded as 'application/x-www-form-urlencoded' to a 'zio.http.forms.Form'.
- Encode 'multipart/form-data  ' to a 'zio.http.Body'.
- Decode a 'zio.http.Body' that is encoded as 'multipart/form-data' to a 'zio.http.forms.Form'.

### DIFF
--- a/zio-http/src/main/scala/zio/http/forms/package.scala
+++ b/zio-http/src/main/scala/zio/http/forms/package.scala
@@ -1,9 +1,9 @@
 package zio.http
 
-import zio.http.model.Headers
-
 import java.net._
 import java.nio.charset.Charset
+
+import zio.http.model.Headers
 
 package object forms {
   private[forms] val CRLF = "\r\n"
--- a/zio-http/src/main/scala/zio/http/forms/package.scala
+++ b/zio-http/src/main/scala/zio/http/forms/package.scala
@@ -6,8 +6,6 @@ import java.net._
 import java.nio.charset.Charset
 
 package object forms {
-  private[zio] val `UTF-8` = Charset.forName("UTF-8")
-
   private[forms] val CRLF = "\r\n"
 
 }
--- a/zio-http/src/main/scala/zio/http/forms/package.scala
+++ b/zio-http/src/main/scala/zio/http/forms/package.scala
@@ -1,0 +1,13 @@
+package zio.http
+
+import zio.http.model.Headers
+
+import java.net._
+import java.nio.charset.Charset
+
+package object forms {
+  private[zio] val `UTF-8` = Charset.forName("UTF-8")
+
+  private[forms] val CRLF = "\r\n"
+
+}
--- a/zio-http/src/main/scala/zio/http/Body.scala
+++ b/zio-http/src/main/scala/zio/http/Body.scala
@@ -2,18 +2,14 @@ package zio.http
 
 import java.io.FileInputStream
 import java.nio.charset._
-import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
 import java.nio.file._
-import zio.http.model.Headers
-
-import zio.http.forms._
-import zio.http.model.MediaType
 
 import zio._
 
 import zio.stream.ZStream
 
-import zio.http.model.HTTP_CHARSET
+import zio.http.forms._
+import zio.http.model.{HTTP_CHARSET, Headers, MediaType}
 
 import io.netty.buffer.{ByteBuf, ByteBufUtil}
 import io.netty.channel.{Channel => JChannel}
@@ -95,8 +91,10 @@ object Body {
    */
   def fromAsciiString(asciiString: AsciiString): Body = AsciiStringBody(asciiString, Headers.empty)
 
-  private[zio] final case class AsciiStringBody(asciiString: AsciiString, override val requestHeaders: Headers)
-      extends Body
+  private[zio] final case class AsciiStringBody(
+    asciiString: AsciiString,
+    override val requestHeaders: Headers = Headers.empty,
+  ) extends Body
       with UnsafeWriteable
       with UnsafeBytes {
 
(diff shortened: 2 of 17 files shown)


zio-http/src/main/scala/zio/http/Form.scala
```diff
@@ -0,0 +1,71 @@
...
+   *
+   * @param value
+   */
+  private[zio] final case class URLEncodedForm(private[zio] encoding: Charset, fields: Map[String, String])
+      extends Form {
+    def toEncodedString: String = encodeFields(encoding, fields).map { case (k, v) => s"$k=$v" }.mkString("&")
+  }
+
+  object URLEncodedForm {
+    def decode(encoded: String, encoding: Charset): Task[Form] = {
```
jdegoes: This should be pure as it does not perform side-effects. Also, it should be called `From.fromDecoded`, and we can delete the `object URLEncodedForm`.

zio-http/src/main/scala/zio/http/Form.scala
```diff
@@ -0,0 +1,71 @@
...
+   * Represents a `application/x-www-form-urlencoded` form.
+   *
+   * `Request.form` will automatically detect and set the `Content-Type` header
+   * to `application/x-www-form-urlencoded` for this request.
+   *
+   * @param value
+   */
+  private[zio] final case class URLEncodedForm(private[zio] encoding: Charset, fields: Map[String, String])
+      extends Form {
+    def toEncodedString: String = encodeFields(encoding, fields).map { case (k, v) => s"$k=$v" }.mkString("&")
```
jdegoes: This could be pushed to `trait Form` since every form can be rendered to an encoded string.

zio-http/src/main/scala/zio/http/Form.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import java.net.URLEncoder
+import java.nio.charset.Charset
+import java.net.URLDecoder
+import zio._
+
+/**
+ * Represents a form that can be either multipart or url encoded.
+ */
+sealed trait Form { self =>
```
jdegoes: Perhaps `FormData` instead of `Form`?

zio-http/src/main/scala/zio/http/Form.scala
```diff
@@ -0,0 +1,71 @@
...
+            new IllegalArgumentException(s"Invalid form field.  Expected: '{name}={value}'', Got '$pair' instead."),
+          )
+
+      }
+
+      fields.map(fields => URLEncodedForm(encoding, fields.toMap))
+    }
+  }
+
+  def makeUrlEncoded(encoding: Charset, fields: Seq[(String, String)]): Form = {
```
jdegoes: The order of these two parameters should be flipped, because the `fields` parameter is more important and always required, whereas the other one is optional.

Also these are quite lengthy names. Probably just `make` would suffice.

zio-http/src/main/scala/zio/http/Form.scala
scottweaver: That was my first preference, but there is already a `FormData` elsewhere in the codebase :(

zio-http/src/main/scala/zio/http/Form.scala
scottweaver: There will be a variant for multipart forms as well so we need a way to differentiate the two.  I am open to suggestions :)

scottweaver:
While this initial implementation conveys what I feel the form API should look like from a usability perspective it ends up being too naive to support the more complex cases one might encounter when dealing `multipart/formdata`.

I will continue experimenting to come up with an implementation powerful enough for the multipart case yet remains simple and performant in the urlencoded case.

zio-http/src/main/scala/zio/http/Form.scala
jdegoes: Maybe we can consolidate?

zio-http/src/main/scala/zio/http/Form.scala
jdegoes: I think `.apply` should get the most "general" constructor, maybe with a little overloading.

Esoteric constructors should use the naming convention `.fromXYZ`.

Sound ok?

jdegoes:
@scottweaver Please re-open when you're ready for another look!

scottweaver:
Pretty much a rewrite from the ground up that supports both `application/x-www-form-urlencoded` and `multipart/form-data`.

The only functionality remaining is support of 'multipart/mixed' for multipart forms.

zio-http/src/main/scala/zio/http/Form.scala
scottweaver: Refactor includes the naming suggestions.

zio-http/src/main/scala/zio/http/Form.scala
scottweaver: I am using now using `FormData`.  However, I don't see a reasonable way to combine it as the other variant is only related to `ContentDisposition`.

zio-http/src/main/scala/zio/http/forms/package.scala
```diff
@@ -0,0 +1,13 @@
+package zio.http
+
+import zio.http.model.Headers
+
+import java.net._
+import java.nio.charset.Charset
+
+package object forms {
+  private[zio] val `UTF-8` = Charset.forName("UTF-8")
```
vigoo: Maybe we should use `StandardCharsets.UTF8` instead of creating new aliases for it

vigoo (approved):

PR #1964, opened 2023-01-26
Return remote address to Request at server

narma:


zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -175,11 +176,10 @@ private[zio] final case class ServerInboundHandler(
       case _                    => throw new IllegalArgumentException(s"Unsupported HTTP version: ${nettyHttpVersion}")
     }
 
-    // TODO: We need to bring this back, probably not part of Request.
-    // val remoteAddress = ctx.channel().remoteAddress() match {
-    //   case m: InetSocketAddress => Some(m.getAddress)
-    //   case _                    => None
-    // }
+    val remoteAddress = ctx.channel().remoteAddress() match {
```
vigoo, line 179: @scottweaver do you remember was this commented out?

narma: I've found notice and related changes in commit 274db5b32 - 
`* Perf iteration 3.  Stop eagerly fetching the remote address.`

Don't know how to run benchmarks and make comparison, but at least can we enable this feature under configuration flag? We still use this feature at our app with version 2.0.0-RC10.

vigoo: Ah I see. The performance reports in the checks below looks good, so I think this is safe to add back

vigoo (approved):

PR #1965, opened 2023-01-27
Add in[I: Schema] to EndpointSpec

vigoo:


zio-http/src/main/scala/zio/http/api/EndpointSpec.scala
```diff
@@ -92,6 +92,9 @@ final case class EndpointSpec[Input, Output](
   ): EndpointSpec[combiner.Out, Output] =
     copy(input = self.input ++ codec)
 
+  def in[Input2: Schema]: EndpointSpec[Input2, Output] =
+    copy(input = HttpCodec.Body(implicitly[Schema[Input2]]))
```
vigoo, line 96: @jdegoes All these `HttpCodec` constructors are private and have no public way to create them, but we have public combinators that are accepting them. Maybe this needs a review

jdegoes (approved):

PR #1966, opened 2023-01-27
Fixing interruption in client

### LINKED ISSUE #1863: Client Request Interruption Not Working as Expected

vigoo:
Resolves #1863 

Based on @jamesward 's PR (https://github.com/zio/zio-http/pull/1893) but I rewrote it as several things changed since then.


jamesward:
Glad to see progress on this, with tests!  For reference, here are related issues/PRs:
https://github.com/zio/zio-http/pull/1893
https://github.com/zio/zio-http/issues/1863

jdegoes (approved):

vigoo:
@jdegoes It's failing on CI but I could not reproduce it locally yet. Can be a real issue though so I'll continue investigating. 
We can ignore the test and merge it as it's definitely already better than before, or just keep it open until the remaining issue is fixed.

jdegoes:
Unfortunately if I merge, the CI will be broken. Best to just disable the test if we think it's not that serious and can be fixed later.

vigoo:
Yes that’s what I meant but cannot commit it until next weekOn 2023. Feb 1., at 10:06, John A. De Goes ***@***.***> wrote:﻿
Unfortunately if I merge, the CI will be broken. Best to just disable the test if we think it's not that serious and can be fixed later.

adamgfraser (approved):

jamesward:
I've tested this change in my [Easy Racer ZIO app](https://github.com/jamesward/easyracer/tree/main/scala-zio) and it is working great!

vigoo:
> I've tested this change in my [Easy Racer ZIO app](https://github.com/jamesward/easyracer/tree/main/scala-zio) and it is working great!

That's good to hear! I'm just trying to fix some edge case that only happens on GitHub CI. Hopefully we can merge it soon.

vigoo:
Continued in https://github.com/zio/zio-http/pull/1994

PR #1979, opened 2023-02-07
Call error callback for ZIO die failures

### LINKED ISSUE #1973: Server doesn't call errorCallback
**Describe the bug**

When you do  `Server.serve(app, Some(errorCallback))` the `errorCallback` is never called on fatal error.

**How to reproduce**

`case Method.GET -> !!   =>   ZIO.die(new RuntimeException("I'm dead"))`

**Additional context**

I've tracked down the issue till `ServerInboundHandler`.  Def  `exceptionCaught` is not called on fatal error produced on endpoint call. 

Method `writeResponse` catches all the errors and doesn't call `errCallback` on the error. 

**Version**

Specification-Version: 0.0.4
Implementation-Version: 0.0.4

vigoo:
Resolves #1973

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -264,13 +264,18 @@ private[zio] final case class ServerInboundHandler(
     runtime.run(ctx, ensured) {
       val pgm = for {
         response <- exit.sandbox.catchAll { error =>
-          ZIO.succeedNow {
-            error.failureOrCause
-              .fold[Response](
-                identity,
-                cause => HttpError.InternalServerError(cause = Some(FiberFailure(cause))).toResponse,
-              )
-          }
+          error.failureOrCause
+            .fold[UIO[Response]](
+              response => ZIO.succeedNow(response),
+              cause =>
+                (cause.dieOption match {
+                  case Some(failure) =>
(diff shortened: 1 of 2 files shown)


zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -264,13 +264,18 @@ private[zio] final case class ServerInboundHandler(
...
-              .fold[Response](
-                identity,
-                cause => HttpError.InternalServerError(cause = Some(FiberFailure(cause))).toResponse,
-              )
-          }
+          error.failureOrCause
+            .fold[UIO[Response]](
+              response => ZIO.succeedNow(response),
+              cause =>
+                (cause.dieOption match {
```
adamgfraser: Maybe the error callback should be a function `Cause[Nothing] => ZIO[Any, Nothing, Any]`?

This is also pushing a little on whether the exit should really be `ZIO[Any, Nothing, Response]`.

vigoo: Maybe. I think probably it's original goal was only to report netty exceptions, but I think the linked user request to also call it with ZIO die failures from handlers is reasonable. But I did not think of changing the callback type itself.

vigoo: Changed to `Cause[Nothing]`

jdegoes (approved):

PR #1981, opened 2023-02-07
Fixing streaming response when compression is enabled

### LINKED ISSUE #1977: stream responses reported as compressed, even though they are not
**Describe the bug**
With the configuration `responseCompression`, the server transparently compresses the responses in the client sends the appropriate `Accept-Encoding` header. This however does not work for streamed  / chunked responses. The server however still sends the header `content-encoding:gzip` or `deflate`. 

**To Reproduce**

Here a sample application (using zio-http version 0.0.4, Scala 3):

```
import zio.*
import zio.Console.printLine
import zio.http.*
import zio.http.model.{HTTP_CHARSET, Headers, Method, Status}
import zio.stream.ZStream

import scala.collection.Seq


object Main extends ZIOAppDefault:

  private val text: HttpApp[Any, Nothing] = Http.collect[Request] { case Method.GET -> !! / "text" =>
    Response.text("Hello World!\n")
  }

  private val stream = Http.collect[Request] { case Method.GET -> !! / "stream" =>
    Response(
      Status.Ok,
      Headers(
        Seq(
          Headers.Header("Content-Type", "text/plain"),
        ),
      ),
      Body.fromStream(
        ZStream
          .unfold[Long, String](0L) { s =>
            if s < 100 then Some((s"$s\n", s + 1)) else None
          }
          .grouped(10).map(_.mkString)
      ),
    )
  }

  override val run: ZIO[Any, Throwable, Unit] =
    for {
      _ <- Server
        .serve(
          text ++ stream
        )
        .provide(
          Server.live,
          ZLayer.succeed(
            ServerConfig.default
              .port(8088)
              .maxThreads(0)
              .responseCompression(),
          ),
        )
    } yield ()
```

Then call the server with curl:

`curl  localhost:8088/stream -v -H 'Accept-Encoding: gzip,deflate' --output infinity.gz` and check that the file `infinity.gz` is not compressed. Also opening the URL `localhost:8088/stream` in a browser will fail. 

`curl  localhost:8088/text -v -H 'Accept-Encoding: gzip,deflate' --output text.gz produces correctly compressed file, which can be checked with `gunzip text`. Opening ` localhost:8088/text` in a browser will work.

**Expected behaviour**
Ideally the server would properly compresses also streamed/chunked responses. If this is not feasible, the `content-encoding` header should be omitted, when the compression has not been applied to the response body.

 - OS:any, tested on Linux
 - Browser: chrome, safari
 - Version [0.0.4, but also reproduced on 0.0.3]

vigoo:
Resolves #1977


jdegoes:
We should add a test here to prevent regression (when it's ready for review).

jdegoes (approved):

PR #1993, opened 2023-02-12
Simpler port

Wosin:
An idea to modify the current Config and Server layers to make it easier to start server on custom port.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Server.scala
+++ b/zio-http/src/main/scala/zio/http/Server.scala
@@ -30,10 +30,10 @@ object Server {
     ZIO.serviceWithZIO[Server](_.install(httpApp, errorCallback)) *> ZIO.service[Server].map(_.port)
   }
 
-  def defaultWithPort(port: Int): ZLayer[Any, Throwable, Server] =
+  def defaultWithPort(port: Int)(implicit trace: Trace): ZLayer[Any, Throwable, Server] =
     defaultWith(_.port(port))
 
-  def defaultWith(f: ServerConfig => ServerConfig): ZLayer[Any, Throwable, Server] =
+  def defaultWith(f: ServerConfig => ServerConfig)(implicit trace: Trace): ZLayer[Any, Throwable, Server] =
     live(f(ServerConfig.default))
 
   val default: ZLayer[Any, Throwable, Server] = {
@@ -41,8 +41,7 @@ object Server {
     ServerConfig.live >>> live
   }
 
-  def live(conf: ServerConfig): ZLayer[Any, Throwable, Server] = {
-    implicit val trace = Trace.empty
+  def live(conf: ServerConfig)(implicit trace: Trace): ZLayer[Any, Throwable, Server] = {
     ServerConfig.live(conf) >>> Server.live
   }
 
--- a/zio-http/src/main/scala/zio/http/Server.scala
+++ b/zio-http/src/main/scala/zio/http/Server.scala
@@ -30,6 +30,9 @@ object Server {
     ZIO.serviceWithZIO[Server](_.install(httpApp, errorCallback)) *> ZIO.service[Server].map(_.port)
   }
 
+  def defaultWithPort(port: Int): ZLayer[Any, Throwable, Server] =
+    defaultWith(_.port(port))
+
   def defaultWith(f: ServerConfig => ServerConfig): ZLayer[Any, Throwable, Server] =
     live(f(ServerConfig.default))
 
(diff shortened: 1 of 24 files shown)


adamgfraser (changes requested):
The `()` after every usage of `Server.default` is very unidiomatic since this would normally indicate that this is a side effecting method. It also means that the `default` server is not a value that will automatically be memoized and shared during layer construction.

This can already be done as:

```scala
ServerConfig.live(ServerConfig.default.port(???)) >>> Server.live
```

This can potentially be shortened by introducing an overloaded `live` constructor on `Server` that takes a `ServerConfig` so the user does not need to manually compose the layers:

```scala
def live(serverConfig: ServerConfig): ZLayer[Any, Throwable, Server] =
  ServerConfig.live(serverConfig) >>> Server.live

Server.live(ServerConfig.default.port(???))
```

Or going a step further:

```scala
def defaultWith(f: ServerConfig => ServerConfig): ZLayer[Any, Throwable, Server] =
  live(f(ServerConfig.default))

Server.defaultWith(_.port(???))
```

Or even:

```scala
def defautlWithPort(port: Int): ZLayer[Any, Throwable, Server] =
  defaultWith(_.port(port))

Server.defaultWithPort(???)
```

These would not impact the ergonomics of the existing usage of the server and would support other modifications to the server configuration in a compositional way in addition to just the port.

Wosin:
Okay, that definitely looks better. Let me fix that.

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -6,7 +6,7 @@ import zio.http.netty.server._
 import zio.stacktracer.TracingImplicits.disableAutoTrace // scalafix:ok;
 
 trait Server {
-  def install[R](httpApp: App[R], errorCallback: Option[ErrorCallback] = None)(implicit
+  def   install[R](httpApp: App[R], errorCallback: Option[ErrorCallback] = None)(implicit
```
adamgfraser: Stray formatting change?

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -30,11 +30,21 @@ object Server {
...
+
+  def defaultWith(f: ServerConfig => ServerConfig): ZLayer[Any, Throwable, Server] =
+    live(f(ServerConfig.default))
+
   val default: ZLayer[Any, Throwable, Server] = {
     implicit val trace = Trace.empty
     ServerConfig.live >>> live
   }
 
+  def live(conf: ServerConfig) = {
```
adamgfraser: Very minor but no need for curly braces here.

zio-http/src/test/scala/zio/http/api/ServerClientIntegrationSpec.scala
```diff
@@ -55,7 +55,7 @@ object ServerClientIntegrationSpec extends ZIOSpecDefault {
       },
     ).provide(
       Server.live,
-      ServerConfig.live,
+      ServerConfig.live(),
```
adamgfraser: Don't think this change should be necessary.

zio-http/src/main/scala/zio/http/ServerConfig.scala
```diff
@@ -122,9 +122,9 @@ final case class ServerConfig(
 }
 
 object ServerConfig {
-  val default: ServerConfig = ServerConfig()
+  def default: ServerConfig = ServerConfig()
```
adamgfraser: Don't think these changes should be necessary.

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -30,11 +30,22 @@ object Server {
...
+
+  def defaultWith(f: ServerConfig => ServerConfig): ZLayer[Any, Throwable, Server] =
+    live(f(ServerConfig.default))
+
   val default: ZLayer[Any, Throwable, Server] = {
     implicit val trace = Trace.empty
     ServerConfig.live >>> live
   }
 
+  def live(conf: ServerConfig): ZLayer[Any, Throwable, Server] = {
```
adamgfraser: If this is already going to be a `def` we can just accept a `Trace` as a parameter.

adamgfraser (approved):
Looks good!

zio-http/src/main/scala/zio/http/Server.scala
```diff
@@ -30,11 +30,21 @@ object Server {
...
+
+  def defaultWith(f: ServerConfig => ServerConfig)(implicit trace: Trace): ZLayer[Any, Throwable, Server] =
+    live(f(ServerConfig.default))
+
   val default: ZLayer[Any, Throwable, Server] = {
     implicit val trace = Trace.empty
     ServerConfig.live >>> live
   }
 
+  def live(conf: ServerConfig)(implicit trace: Trace): ZLayer[Any, Throwable, Server] = {
```
adamgfraser, line 44: Don't need the curly braces here.

jdegoes (approved):

PR #2005, opened 2023-02-21
fixes #2000, #2001

### LINKED ISSUE #2000: Declarative Endpoint returns 404 when specifying body
When using the declarative endpoint feature and specifying a body type using the `in` operator, requests to that endpoint will always return a `404 Not Found` error.

**Small reproducer:**
```scala
import zio.*
import zio.http.codec.HttpCodec.*
import zio.http.endpoint.*
import zio.http.{ Middleware, Server }
import zio.schema.{ DeriveSchema, Schema }

  case class Test(value: String)

  given Schema[Test] = DeriveSchema.gen[Test]

  val endpoint =
    Endpoint
      .post("test")
      .in[Test]
      .out[Test]
      .implement(ZIO.succeed)

  override def run =
    Server
      .serve(endpoint.toApp @@ Middleware.debug)
      .provide(
        Server.defaultWithPort(8080)
      )

}
```

with the following sample request:
image

**Expected behaviour**
Sending a correct post request to the `/test` endpoint should return a 200 OK response 

**Additional context**
Happens on the latest snapshot of `zio-http` with `Scala 3.2.2`

### LINKED ISSUE #2001: Apps created from endpoints do not correctly fall back on other apps
When creating apps from endpoints and then combining the apps together, requests to the first endpoint in the chain will succeed, while requests to subsequent endpoints will always fail. In the reproducer below, get requests to `/test` will succeed while requests to `/test2` will fail. 

Small reproducer:
```scala
import zio.*
import zio.http.codec.HttpCodec.*
import zio.http.endpoint.*
import zio.http.{ Middleware, Server }

  val endpoint =
    Endpoint
      .get("test")
      .out[String]
      .implement(_ => ZIO.succeed("Success1"))

  val endpoint2 =
    Endpoint
      .get("test2")
      .out[String]
      .implement(_ => ZIO.succeed("Success2"))

  override def run =
    Server
      .serve((endpoint.toApp ++ endpoint2.toApp) @@ Middleware.debug)
      .provide(
        Server.defaultWithPort(8080)
      )

}
```

**Expected behaviour**
GET requests to the `/test2` endpoint should succeed

**Additional context**
Happens on the latest snapshot version of `zio-http` with `Scala 3.2.2`

jdegoes:


zio-http/src/main/scala/zio/http/endpoint/Routes.scala
```diff
@@ -33,22 +33,19 @@ sealed trait Routes[-R, +E, M <: EndpointMiddleware] { self =>
...
+    val routingTree     = RoutingTree.fromRoutes(self)
     val requestHandlers = Memoized[Routes.Single[R, _ <: E, _, _, M], EndpointServer[R, _ <: E, _, _, M]] {
       handledApi =>
         EndpointServer(handledApi)
     }
 
     Http
-      .collectZIO[Request] { case (request: Request) =>
-        val handler = handlerTree.lookup(request)
+      .collectZIO[Request] {
```
vigoo: You could use `fromOptionalHandlerZIO` and then no need to evaluate the lookup function twice

vigoo (approved):

zio-http/src/main/scala/zio/http/endpoint/Routes.scala
jdegoes: Thanks, exactly what I was looking for!

vigoo (approved):

PR #2012, opened 2023-02-21
fix: Scala 2.12 visibility

### LINKED ISSUE #2006: Unable to use `NettyDriver` in Scala 2.12
**Describe the bug**
It seems like there's currently an issue with using the `NettyDriver` in Scala 2.12. This has forced Tapir to drop Scala 2.12 support for zio-http. I'm not that well-versed in 2.12, but it seems like the root cause of the issue being that the `NettyDriver` trait is `private[zio]` but the companion object defining the ZLayers not having any access modifiers, resulting in the following error:

```
[error] /Users/adamw/projects/tapir/server/zio-http-server/src/test/scala/sttp/tapir/server/ziohttp/ZioHttpTestServerInterpreter.scala:65:7: object NettyDriver in package server cannot be accessed in package zio.http.netty.server
```
I think there are good reasons to keep `NettyDriver` private, but perhaps we can move the constructors to the `zio.http.netty.server` package object instead, or at least expose them from it?

frekw:
Fixes #2006.

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/server/package.scala
+++ b/zio-http/src/main/scala/zio/http/netty/server/package.scala
@@ -1,11 +1,30 @@
 package zio.http.netty
 
-import zio.http._
+import io.netty.channel.{Channel => NettyChannel, _}
 import zio._
-import java.util.concurrent.atomic.AtomicReference
+import zio.http._
+import zio.http.netty._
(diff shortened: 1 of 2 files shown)


vigoo:
Closing it as #2009 is merged already

PR #2021, opened 2023-02-26
Ability to provide context from middlewares, alternative approach

vigoo:


vigoo:
Closing in favor of #2011

PR #2026, opened 2023-03-01
Removing Netty from the public API

### LINKED ISSUE #1453: Do not expose Netty data types in public APIs

It is not easy to replace Netty by a different library, or to contemplate a Scala Native version of ZIO HTTP.

Ensuring that no Netty data types are exposed in the public API. Any Netty data types used internally can be made private or package private, while any Netty data types that a user must interact with can be wrapped in data types that leak minimal details about the implementation of said data type. 

**Why is it important?**

Avoiding leaking Netty helps improve the long-term stability of the API, and open up the door to running on Scala Native.

vigoo:
The remaining steps for https://github.com/zio/zio-http/issues/1453


zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -238,80 +230,41 @@ object Response {
...
   }
 
-  // TODO: This needs to eventually be implementation agnostic.
-  private[zio] class NettyResponse(
+  private[zio] class NativeResponse(
     val body: Body,
-    val channelContext: ChannelHandlerContext,
     val headers: Headers,
     val status: Status,
+    onClose: () => Task[Unit],
```
jdegoes, line 237: Clever!

jdegoes (approved):

jdegoes:
@vigoo There are might be a few tweaks we can do later, overall, this looks amazing and we should just try to get it in as soon as possible given the scope. 🙏

PR #2035, opened 2023-03-11
Update link to RunSettings

denisgarci:
Update link to point to the correct line number.


vigoo:
Thanks, but the README is generated from this file: https://github.com/zio/zio-http/blob/main/docs/index.md
could you make the change there too, and then run `sbt generateReadme`?

denisgarci:
@vigoo Thanks for the guidance, done.

vigoo (approved):

PR #2043, opened 2023-03-15
Make request logging middleware universial

guersam:
The current `Middleware.requestLogging` adds `Throwable` error type to the given `Http` so that it cannot be applied in the outermost place.

This PR removes the error constraint by simply ignoring the errors from body extraction.

### DIFF
--- a/zio-http/src/main/scala/zio/http/middleware/RequestLogging.scala
+++ b/zio-http/src/main/scala/zio/http/middleware/RequestLogging.scala
@@ -35,7 +35,7 @@ private[zio] trait RequestLogging { self: RequestHandlerMiddlewares =>
     logResponseBody: Boolean = false,
     requestCharset: Charset = StandardCharsets.UTF_8,
     responseCharset: Charset = StandardCharsets.UTF_8,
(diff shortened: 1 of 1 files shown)


zio-http/src/main/scala/zio/http/middleware/RequestLogging.scala
```diff
@@ -55,8 +55,8 @@ private[zio] trait RequestLogging { self: RequestHandlerMiddlewares =>
                   LogAnnotation(name.toString, value.toString)
               }.toSet
 
-            val requestBody  = if (request.body.isComplete) request.body.asChunk.map(Some(_)) else ZIO.none
-            val responseBody = if (response.body.isComplete) response.body.asChunk.map(Some(_)) else ZIO.none
+            val requestBody  = if (request.body.isComplete) request.body.asChunk.option else ZIO.none
+            val responseBody = if (response.body.isComplete) response.body.asChunk.option else ZIO.none
```
vigoo, line 59: Maybe we could log any errors happening here instead of just ignoring them

guersam: I thought the error will occur in the userland again and should be handled there. What do you think?

vigoo (approved):

PR #2051, opened 2023-03-17
Header model cleanup

### LINKED ISSUE #2048: Improve zio.http.model.headers.values data types
There are a few aspects of the `zio.http.mode.header.value` data types which are not idiomatic and which should be improved before release of 1.0:

 - **Eliminating the failure value.** Currently, each header value has a failure value, which is used in case the header value cannot be parsed according to the RFC spec for that header value. However, `HttpCodec` already has a separte mechanism to report an error during parsing, which is the `transformOrFail` family of functions on `HttpCodec`. By using this mechanism, to fail directly in `HttpCodec` rather than fail in "value space", we can simplify the ADT, and prevent users from having to manually handle what is really a low-level failure that no endpoint implementation should have to deal with directly.
 - **Reducing verbosity.** A common convention in the current set of header values is to replicate the name of the `sealed trait` for the header value inside each `case class` that extends it in the companion object. However, as these possibilities for the header value are already namespaced in the companion object, there is no need to replicate that information twice, and doing so requires more typing and interferes with readability (it's also not very idiomatic in the ZIO ecosystem). Instead, we should use short names, so for example, instead of `AccessControlMaxAge.ValidAccessControlMaxAge`, we could simply have `AccessControlMaxAge.Valid` (or actually, in this case, once we eliminate the `Invalid` possibility, we will only have one possibility, so we can change `AccessControlMaxAge` into a `case class`. As another example, instead of `Upgrade.UpgradeProtocols`, we could have `Upgrade.Protocols`; or instead of `Upgrade.UpgradeValue`, we could have `Upgrade.Value`. 
 -  **Turning sealed traits into case classes**. Once the "invalid" term is removed from the sealed trait of many of the header values, then there is only a single "value" value. In such cases, we should turn the sealed trait into a `case class`.

vigoo:
Resolves https://github.com/zio/zio-http/issues/2048


jdegoes (approved):
💥

PR #2055, opened 2023-03-20
Unifying headers

### LINKED ISSUE #1411: CORS Middleware: Return only requested allowed headers, rather than all configured allowed headers
### Issue
This is not a bug, but a suggestion for improving the CORS implementation. Other CORS implementations, namely Play, which I am currently using in production, only return the _Requested_ headers in the Access-Control-Allow-Headers preflight response header, rather than returning the internal configured whitelist.

### Example

Configured Allowed Headers:
`X-Custom-Head-1`, `X-Custom-Head-2`, `X-Custom-Head-3`
Request header:
`Access-Control-Requested-Headers: X-Custom-Head-1`
Current Response:
`access-control-allow-headers: X-Custom-Head-1,X-Custom-Head-2,X-Custom-Head-3`
Desired Response:
`access-control-allow-headers: X-Custom-Head-1`

This is made a bit worse by the fact that there is currently no way to allow all request headers; so all of my responses include a long long list of whitelisted headers.

### Option to allow all request headers?

Currently this is not possible when Credentials are enabled. While the `*` header is included in the default config, this header name is interpreted literally when credentials are enabled. Current defaults:
```scala
Set(CONTENT_TYPE, AUTHORIZATION, "*")
```
> In requests with credentials, it is treated as the literal header name "*" without special semantics.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#directives

Perhaps a field could be added to `CorsConfig` named `anyHeader: Boolean`, to go along with the existing `anyOrigin` and `anyMethod`. If set, this option would simply take any requested headers from the request's `Access-Control-Requested-Headers` and include them in the response's access-control-allow-headers. [This is what Play does](https://github.com/playframework/playframework/blob/78bcdb4e1d1fe0a350ab2a56a38649571415383f/web/play-filters-helpers/src/main/scala/play/filters/cors/AbstractCORSPolicy.scala#L268) and seems to be compliant with the spec.

I'd like to add that if the maintainers decide that either of these suggestions are worth pursuing, I'd be happy to submit a PR.

### LINKED ISSUE #2050: Unify headers
Currently, ZIO HTTP has duplication and inconsistency in how it exposes headers.

1. We have typed structures like `Cookie[T]`, and random parts of headers such as Credentials.
2. We have some APIs that deal with header names and values as `String`.
3. We have some APIs that deal with header names and values as `CharSequence`.
4. We have typed representations of most header values in `zio.http.model.headers.values`.

We need to unify these and provide a standardized way of consuming, producing, and transforming headers, one which allows optimization, via pre-encoding structured information / strings into bytes.

vigoo:
Resolves #2050 
Resolves #1411


zio-http/src/main/scala/zio/http/model/Header.scala
```diff
@@ -0,0 +1,4238 @@
...
+
+    def name: CharSequence
+
+    def parse(value: String): Either[String, HeaderValue]
+
+    def render(value: HeaderValue): String
+  }
+
+  object HeaderType {
+    type Aux[HV] = HeaderType { type HeaderValue = HV }
```
jdegoes:
```suggestion
    type Typed[HV] = HeaderType { type HeaderValue = HV }
```

zio-http/src/main/scala/zio/http/model/Header.scala
```diff
@@ -0,0 +1,4238 @@
...
+
+import zio._
+
+import zio.http._
+import zio.http.model.Header.Authorization.AuthScheme
+import zio.http.model.Header.HeaderType.Aux
+
+sealed trait Header {
+  def headerName: CharSequence
+  def renderedValue: CharSequence
```
jdegoes: Add `def headerType: HeaderType` here? Due to covariance, this could also be overridden to be the specific type of header in the subtype.

zio-http/src/main/scala/zio/http/model/Header.scala
```diff
@@ -0,0 +1,4238 @@
...
+    def parse(value: String): Either[String, HeaderValue]
+
+    def render(value: HeaderValue): String
+  }
+
+  object HeaderType {
+    type Aux[HV] = HeaderType { type HeaderValue = HV }
+  }
+
+  sealed trait HeaderValue[H <: Header] { self: H =>
```
jdegoes: Do we need a distinction between `Header` and `HeaderValue`?

Conceptually, I view a header as a combination of two things: the type (e.g. `Content-Type`), and the value (e.g. `application/json`). 

If we can move from `Header` + `HeaderValue` to just `Header`, it will be a simpler API.

vigoo: I wanted two things with it:

- Have a simple untyped `Header` interface with no type parameters, to be used to get the name and value as strings when for example converting to netty data types, etc. 
- The other to have the actual type as a type parameter, so type inference works properly (without it, ```header.headerType.render(header)``` was not type checking for example)

vigoo: It could be just `Header[H]` and used as `Header[_]` in places where we don't care. Is that better?

zio-http/src/main/scala/zio/http/netty/model/Conversions.scala
```diff
@@ -82,17 +82,25 @@ private[netty] object Conversions {
...
-  private def encodeHeaderListToNetty(headersList: List[Product2[CharSequence, CharSequence]]): HttpHeaders = {
+  private def encodeHeaderListToNetty(headers: Iterable[Header]): HttpHeaders = {
     val (exceptions, regularHeaders) =
-      headersList.partition(h => h._1.toString.contains(HeaderNames.setCookie.toString))
-    val combinedHeaders              = regularHeaders
-      .groupBy(_._1)
-      .map { case (key, tuples) =>
-        key -> tuples.map(_._2).mkString(",")
-      }
+      headers.partition(h => h.headerName.toString.contains(HeaderNames.setCookie.toString))
```
jdegoes: We end up going `toString` on the header name so many times, but obviously if it's not a string, that has a cost.

I wonder if our headers should store both `String` and (potentially optimized) `CharSequence` representation?

vigoo: This whole function completely destroys any performance gain these pooled char sequences could give, especially the original version. I tried to improve it a bit already, but it should be further optimized. (For example we can get rid of the above `toString`s, or move this concatenation logic to the header type and do it when a `Header` gets added to `Headers` instead of here.

jdegoes: I have an idea that when you construct `Header.ContentType("application/json")` it's immediately encoded into `Array[Byte]`. So if you use `x match { Header.ContentType(s) => ... }` it's decoded from the bytes. i.e. eager encoding, lazy decoding. 

Such things could make `CharSequence` obsolete. For now all the `CharSequence` ceremony seems like we pay a lot for something that is so brittle, it ends up being destroyed. Maybe we should switch to strings for now and rip out a big source of complexity?

vigoo: Right now we use Netty's built-in cached HTTP header names which are `AsciiString`s. By the way, their `toString` is cached so there is no point building our own cached one on top of it. Netty's request decoder also creates `AsciiString` header names and values, and for some it uses these cached names (seems to me that only for a few default ones like `Host`).

The header parser also seems to copy the source array when creating these `AsciiStrings` so it's not just pointing to the native buffer or anything like that. 

That said I don't really see much advantage of using `CharSequence`s everywhere other than making the API more painful. Using `String`s everywhere would simplify many things.

vigoo: The `Header.Custom`, which holds the untyped headers coming from Netty, could still refer to the `CharSequence`s internally, to avoid converting them to strings if the headers are not used for anything.

zio-http/src/main/scala/zio/http/model/Header.scala
```diff
@@ -0,0 +1,4453 @@
...
+                }
+                i = i + 1
+              }
+              equal
+            }
+          }
+
+          eqs(self.customName, k) && eqs(self.value, v)
+
+        case _ => false
```
guersam, line 105: Maybe unrealistic, but what if the user extract a header value as `Custom`, then compare it with a modeled one with the same name?

vigoo: Those are different case classes so equality won't be true. But on the `Headers`/`HeaderGetter`/`HeaderModifier` level it is handled so even though all headers from a Netty request or response are `Header.Custom` (not parsed), you can get a typed header value and then it gets parsed, also when modifying headers it does not matter whether it is in typed or unencoded form in the end the last one per key gets encoded.

jdegoes (approved):

jdegoes:
Massive cleanup! ❤️

PR #2057, opened 2023-03-22
Set scheme when setting url/uri to ZClient

zeal18:
the issue comes when setting a uri to the client like this:
```scala
val configuredClient = client.uri(URI("https://example.com"))
```
it sets the port to 443 which is correct, but doesn't set the scheme which results to defaulting it to the `http` value in the `ClientLive.requestInternal` implementation.

I think the client should be completely redesigned and rewritten but it requires much more time, so the PR just fixes the issue

### DIFF
--- a/zio-http/src/main/scala/zio/http/ZClient.scala
+++ b/zio-http/src/main/scala/zio/http/ZClient.scala
@@ -416,21 +416,27 @@ trait ZClient[-Env, -In, +Err, +Out] { self =>
   final def ssl(ssl: ClientSSLConfig): ZClient[Env, In, Err, Out] =
     copy(sslConfig = Some(ssl))
 
-  final def uri(uri: URI): ZClient[Env, In, Err, Out] =
+  final def uri(uri: URI): ZClient[Env, In, Err, Out] = {
+    val scheme = Scheme.decode(uri.getScheme)
+
     copy(
       hostOption = Option(uri.getHost),
       pathPrefix = pathPrefix ++ Path.decode(uri.getRawPath),
-      portOption = Option(uri.getPort).filter(_ != -1).orElse(Scheme.decode(uri.getScheme).map(_.port)),
+      portOption = Option(uri.getPort).filter(_ != -1).orElse(scheme.map(_.port)),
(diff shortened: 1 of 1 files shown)


vigoo (approved):

jdegoes:
@zeal18 Can you please describe your other problems with the client, as we are in the process of finalizing the API for 1.0?

jdegoes (approved):

zeal18:
> @zeal18 Can you please describe your other problems with the client, as we are in the process of finalizing the API for 1.0?

@jdegoes, sure, my other problems with the client could be:
* the interface should define just a single `request` method to prevent inconsistencies, for example, current interface defines `def portOption: Option[Int]` method, but it's not used in the implementation of the main `request` method: `port <- ZIO.fromOption(portOption).orElseSucceed(sslConfig.fold(80)(_ => 443))` it gets it's portOption from arguments and falls back to the sslConfig. The same applies for the `hostOption`, it could be "configured" but the main implementation ignores it. This wasn't a case when the main `request` method was protected, but now it not any more
* the main `request` method requires host and port as optional arguments and then fails when some of them are `None`s, I think it's better to require them on the interface than failing
* `SocketApp` is required by `ClientDriver.requestOnChannel` and `ClientDriver` is required by `ZClient.live`, but `ClientDriver.shared` requires the `ClientConfig` which also requires the `SocketApp`, so there are two ways of providing the `SocketApp` and one of them could be eliminated removing the `ZClient.socket` method from the interface. The same applies for the `ClientSSLConfig`
* `ZClientAspect` is not allowed to modify environment, error type, input and output types of the client, it would be nice to be able to do that
* probably such methods like `refineOrDie`, `retry` and `dieOn` could be replaced by aspects, when it's needed to be applied to a client globally
* `ClientDriver.requestOnChannel` method requires both `location: URL.Location.Absolute` and `req: Request` arguments, but `Request` contains a `URL` which could be different

the list is very rough and should be reestimated, and of cause I can help with some improvements if you want to improve something of this)

vigoo:
Thanks @zeal18 very useful list of issues!

PR #2063, opened 2023-03-26
Fail on decoding errors

### LINKED ISSUE #2056: `decoderResult` is not honored before executing the HttpApp
**Describe the bug**
When the client sends a request with the header size larger than the configured `maxHeaderSize`, netty will get a failure DecoderResult like `failure(io.netty.handler.codec.http.TooLongHttpHeaderException: HTTP header is larger than 1 bytes.)`, 
but the ZIO Http's HttpApp will still be executed and gets an empty request body.

**To Reproduce**

```scala

import zio.Scope
import zio.http._
import zio.http.model._
import zio.test._

object NettyMaxHeaderLengthTest extends ZIOSpecDefault {
  override def spec: Spec[TestEnvironment with Scope, Any] =
    test("should get a failure instead of an empty body") {
      for {
        port <- Server.install(
                  Http
                    .collectZIO[Request] { request =>
                      request.body.asString.map { body =>
                        val responseBody = if (body.isEmpty) "<empty>" else body
                        Response.text(responseBody)
                      } // this should not be run, as the request is invalid
                    }
                    .withDefaultErrorResponse // it should execute the logic within this line
                )
        url     = s"http://localhost:$port"
        headers = Headers.userAgent("a looooooooooooooooooooooooooooog header")

        res  <- Client.request(url, headers = headers, content = Body.fromString("some-body"))
        data <- res.body.asString
      } yield {
        assertTrue(res.status == Status.InternalServerError && data == "<empty>")
      }
    }.provide(
      Client.default,
      Server.live,
      ServerConfig.live(ServerConfig(maxHeaderSize = 1))
    )
}

```

**Expected behaviour**
As the request is invalid, the HttpApp should not be executed. And the error handling logic should be executed (the `withDeafultErrorResponse` in the above snippet)

**Screenshots**
image

**Additional context**
This above snippet was run with zio-http `0.0.4` and `0.0.5`. But it can also be reproduced in the earlier versions(e.g., `2.0.0-RC10`).

vigoo:
Resolves #2056 

Note that it is not exactly what the bug report asks for: the decoding error is converted to a `Die` failure which means it gets an internal server error response, and can be observed with the server's error callback, while the bug description asks for running to the `withDefaultErrorResponse` handler in this case. To do so, we would need to model decoding errors in our `Request` class and do the check within the `Http` app which does not sound like a good idea to me.

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -80,7 +80,7 @@ private[zio] final case class ServerInboundHandler(
         }
 
         ensureHasApp()
-        val exit = app.runZIOOrNull(req)
+        val exit = failOnDecodingError(jReq) *> app.runZIOOrNull(req)
         if (!attemptImmediateWrite(ctx, exit, time))
           writeResponse(ctx, env, exit, jReq)(releaseRequest)
         else
@@ -90,7 +90,7 @@ private[zio] final case class ServerInboundHandler(
         val req = makeZioRequest(ctx, jReq)
 
         ensureHasApp()
-        val exit = app.runZIOOrNull(req)
+        val exit = failOnDecodingError(jReq) *> app.runZIOOrNull(req)
         if (!attemptImmediateWrite(ctx, exit, time)) {
 
           if (
@@ -336,6 +336,9 @@ private[zio] final case class ServerInboundHandler(
     ZIO.attempt {
       ctx.channel().close()
     }.unit.orDie
+
+  private def failOnDecodingError(request: HttpRequest): ZIO[Any, Nothing, Unit] =
+    ZIO.die(request.decoderResult().cause()).when(request.decoderResult().isFailure).unit
 }
 
 object ServerInboundHandler {
--- a/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -337,8 +337,8 @@ private[zio] final case class ServerInboundHandler(
       ctx.channel().close()
     }.unit.orDie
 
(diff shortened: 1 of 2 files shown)


alphaho:
Thanks @vigoo for the swift fix on the issue.

On leveraging the `withDefaultErrorResponse` to handle the decode failure, I'm thinking different users may have different needs on how they handle it. For example, some may like to return 500, while others may want to return 400 instead.

Instead of model the decoding errors in the `Request` class (which I totally agree it's a bad idea), would it be better to model these kinds of Http-intrinsic errors in a `Cause` type (maybe with a different name to avoid confusion) similar to what we have in ZIO? So that we can provide both `mapError` and `mapCause` to handle different errors from different levels. (As I've found that there's only `mapError` available in the `Http` class)

vigoo:
@alphaho The problem is that a `Http` passed to the server is like a function `Request => ZIO[R, Response, Response]`. The error we are talking about here happens when decoding Request` so no matter what kind of combinator you would use to capture this problem within `Http`, currently there is no way to get to that point because we don't have a `Request` to "call" the `Http` with.

(Imagine a ZIO equivalent like

```scala
def handleRequest(request: Request): ZIO[Any, Response, Response] =>
   ???.mapCause(cause => if (isDecodingError(cause)) { ... })
```
)

That's what I meant that it can only be handled on the `Http` level if the decoding error is represented in the zio-http `Request` model somehow - that would make `Http` modelled by `Either[DecodingError, Response] => ZIO[R, Response, Response]` which in my opinion makes everything more complex and does not give much in exchange.

A less elegant solution for customizing this behavior could be to be able to produce a `Response` in the exception callback of the server. What do you think?

jdegoes (approved):

jdegoes:
I think a decoding error is precisely a non-recoverable error and should be placed on the `Die` channel. Of course, we need good combinators to allow users to surface some of those errors for purpose of generating a useful `Response` (which could be new tickets).

alphaho:
@vigoo thanks for the explanation, I now understand the problem. And I agree that neither option is good enough, and both can make it less pleasant.
I've opened a new issue(#2065) to track down the problem

PR #2064, opened 2023-03-27
Async DNS resolver

### LINKED ISSUE #1730: Feature: Async DNS Resolver for zio-http client support
twitter [finagle](https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/InetResolver.scala) and apache http client supports DSNResolver to resolve the host address into list of ip address asynchronously and concurrently  which would be used by the underlying TCP connection. This reduces dns resolution during the http call.

The zio http-client should support DNS resolver like finagle . See the example application using finagle [here](https://github.com/rajcspsg/FinagleClientDemo/tree/master/scala2_12). (It is in java but not too many lines of code)
If you check the twitter code for DNS resolver InetResolver.scala there are different schemes to resolve the host to Ip addresses. It would be nice if we could replicate the similar way so that finagle users could migrate smoothly to zio http-client.

FInagle and apache http client supports DNS resolver but those are not pure functional approach.

**Additional context**
Please let me know if you have any questions and I'm happy to discuss on this. I discussed this with John at zio hackathon.

vigoo:
Resolves #1730

### DIFF
--- a/zio-http/src/main/scala/zio/http/DnsResolver.scala
+++ b/zio-http/src/main/scala/zio/http/DnsResolver.scala
@@ -1,15 +1,18 @@
-package zio.http.internal
-
-import zio._
+package zio.http
 
 import java.net.{InetAddress, UnknownHostException}
 import java.time.Instant
+
+import zio._
 
 trait DnsResolver {
   def resolve(host: String): ZIO[Any, UnknownHostException, Chunk[InetAddress]]
 }
 
 object DnsResolver {
+  def resolve(host: String): ZIO[DnsResolver, UnknownHostException, Chunk[InetAddress]] =
+    ZIO.serviceWithZIO(_.resolve(host))
+
   private final case class SystemResolver() extends DnsResolver {
     override def resolve(host: String): ZIO[Any, UnknownHostException, Chunk[InetAddress]] =
       ZIO
@@ -18,7 +21,7 @@ object DnsResolver {
         .map(Chunk.fromArray)
   }
 
-  private final case class CacheEntry(
+  private[http] final case class CacheEntry(
     resolvedAddresses: Promise[UnknownHostException, Chunk[InetAddress]],
     previousAddresses: Option[Chunk[InetAddress]],
     lastUpdatedAt: Instant,
@@ -28,6 +31,13 @@ object DnsResolver {
   object ExpireAction {
     case object Refresh extends ExpireAction
     case object Drop    extends ExpireAction
+
+    val config: Config[ExpireAction] =
+      Config.string.mapOrFail {
+        case "refresh" => Right(Refresh)
+        case "drop"    => Right(Drop)
+        case other     => Left(Config.Error.InvalidData(message = s"Invalid expire action: $other"))
+      }
   }
 
   private sealed trait CachePatch
@@ -75,6 +85,10 @@ object DnsResolver {
         result    <- getResult
       } yield result
 
+    /** Gets a snapshot of the cache state, for testing only */
+    private[http] def snapshot(): ZIO[DnsResolver, Nothing, Map[String, CacheEntry]] =
+      entries.get
+
     private def startResolvingHost(
       host: String,
       targetPromise: Promise[UnknownHostException, Chunk[InetAddress]],
@@ -111,7 +125,8 @@ object DnsResolver {
                           None,
                           now,
                         )
-                        startResolvingHost(host, newEntry.resolvedAddresses).as(Some(CachePatch.Update(host, newEntry)))
+                        startResolvingHost(host, newEntry.resolvedAddresses)
+                          .as(Some(CachePatch.Update(host, newEntry)))
                       } else {
                         ZIO.some(CachePatch.Remove(host))
                       }
@@ -168,7 +183,7 @@ object DnsResolver {
       }
   }
 
-  object CachingResolver {
+  private object CachingResolver {
     def make(
       resolver: DnsResolver,
       ttl: Duration,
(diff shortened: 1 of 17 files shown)


zio-http/src/main/scala/zio/http/internal/DnsResolver.scala
```diff
@@ -0,0 +1,188 @@
...
+        patch match {
+          case CachePatch.Remove(host)               =>
+            entries - host
+          case CachePatch.Update(host, updatedEntry) =>
+            entries.updated(host, updatedEntry)
+        }
+      }
+  }
+
+  object CachingResolver {
```
jdegoes: Better to hide this and instead only `DnsResolver#cached`.

zio-http/src/main/scala/zio/http/internal/DnsResolver.scala
```diff
@@ -0,0 +1,188 @@
+package zio.http.internal
+
+import zio._
+
+import java.net.{InetAddress, UnknownHostException}
+import java.time.Instant
+
+trait DnsResolver {
```
jdegoes: This is a pure interface, could be put into `zio.http` instead of internal.

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -0,0 +1,285 @@
...
+  final case class DnsResolverConfig(
+    ttl: Duration,
+    unknownHostTtl: Duration,
+    maxCount: Int,
+    maxConcurrentResolutions: Int,
+    expireAction: ExpireAction,
+    refreshRate: Duration,
+  )
+
+  object DnsResolverConfig {
```
jdegoes, line 219: I would probably just call this `Config` since it's already located inside `DnsResolver`, although I realize that makes it necessary to qualify `zio.Config`.

Compare:

```scala
DnsResolver.DnsResolverConfig
```

vs

```scala
DnsResolver.Config
```

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  def cached(
+    ttl: Duration = 10.minutes,
+    unknownHostTtl: Duration = 1.minute,
+    maxCount: Int = 4096,
+    maxConcurrentResolutions: Int = 16,
+    expireAction: ExpireAction = ExpireAction.Refresh,
+    refreshRate: Duration = 2.seconds,
+    implementation: DnsResolver = SystemResolver(),
+  ): ZLayer[Any, Nothing, DnsResolver] =
```
jdegoes, line 249: This is "cached system resolver".

Could be useful to add `cached` as a method on `DnsResolver`.

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -0,0 +1,285 @@
...
+        Config.int("max-concurrent-resolutions").withDefault(default.maxConcurrentResolutions) ++
+        ExpireAction.config.nested("expire-action").withDefault(default.expireAction) ++
+        Config.duration("refresh-rate").withDefault(default.refreshRate)).map {
+        case (ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate) =>
+          DnsResolverConfig(ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate)
+      }
+  }
+
+  def cached(
+    ttl: Duration = 10.minutes,
```
jdegoes, line 242: It is odd to have defaults here as well as on `Config`. 

I wonder if you could just do two variants of this: one which takes a DNS resolver config, and one which loads the settings _from_ config using `ZIO.config`.

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -0,0 +1,285 @@
...
+          maxConcurrentResolutions,
+          expireAction,
+          refreshRate,
+        )
+    }
+
+  val configured: ZLayer[Any, Config.Error, DnsResolver] =
+    ZLayer(ZIO.config(DnsResolverConfig.config.nested("zio.http.dns"))) >>> fromConfig
+
+  val default: ZLayer[Any, Nothing, DnsResolver] = ZLayer.succeed(DnsResolverConfig.default) >>> fromConfig
```
jdegoes, line 266: Is there a good rerason to have both `configured` and `default`? Wouldn't a user always want `configured` (so we could rename that to `default`)?

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  lazy val fromConfig: ZLayer[DnsResolverConfig, Nothing, DnsResolver] =
```
jdegoes, line 268: Maybe this coudl be called `configured`? `fromXYZ` methods usually take an `XYZ` as a parameter. I am not sure about this one.

vigoo: I was following the existing pattern in other layers like `Client` and `Server`. We should find a common one and make sure everything has the same constructors and layers.

zio-http/src/main/scala/zio/http/DnsResolver.scala
vigoo: The other layers in the library are currently having similar config case classes and default layers to work with - but here I also added a ZIO `Config` based mode to initialize because I think we should have that; but I agree that it felt that now we have too many ways to initialize this thing. 

We should find a common pattern for having configurable services in the library and refactor all of them to the common pattern.

zio-http/src/main/scala/zio/http/DnsResolver.scala
vigoo: Yeah probably this one is not needed

zio-http/src/main/scala/zio/http/DnsResolver.scala
vigoo: Yes, but we have https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/ClientConfig.scala etc (although not in the companion object of `Client` - one more thing to unify)

jdegoes (approved):

PR #2066, opened 2023-03-29
WIP

jdegoes:


zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -114,16 +110,11 @@ private[zio] final case class ServerInboundHandler(
...
 
-  override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
-    if (errCallback != null) {
-      runtime.run(ctx, NettyRuntime.noopEnsuring)(errCallback(Cause.die(cause)))
-    } else {
-      cause match {
-        case ioe: IOException if ioe.getMessage.contentEquals("Connection reset by peer") =>
-        case t => super.exceptionCaught(ctx, t)
-      }
+  override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit =
```
vigoo, line 113: By removing the error callback, we no longer have any way to:

- log the error
- measure the error (with a counter for example)
- customize the response (that was not possible before either)

I think if we don't want to provide any user-defined callback for this, we should at least have a default, ZIO Core log and metric here. 

Alternatively, we could modify `Http` to be something like `Exit[Throwable, Requiest] => ZIO[R, Response, Response]` and pass errors to the `App`.

vigoo (commented):
I have concerns about removing the error callback, the rest of it looks nice

zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
jdegoes: I like the idea of modifying the `Http` to be able to handle server failures.

vigoo (approved):

PR #2069, opened 2023-03-29
Implement ClientSSLConfig.config

adamgfraser:


zio-http/src/main/scala/zio/http/ClientSSLConfig.scala
```diff
@@ -21,25 +21,32 @@ import zio.Config
...
+
+    tpe
+      .switch(
+        "Default"                -> default,
+        "FromCertFile"           -> fromCertFile,
+        "FromCertResource"       -> fromCertResource,
+        "FromTrustStoreFile"     -> fromTrustStoreFile,
+        "FromTrustStoreResource" -> fromTrustStoreResource,
+      )
+      .nested("zio.http.ClientSSLConfig")
```
vigoo: I think this nesting is not good the way this is used now in `ZClient.Config` (where it gets explicitly `nested` too)

jdegoes: 👍

vigoo (approved):

PR #2072, opened 2023-03-30
Ability to catch server errors and defects in Http

### LINKED ISSUE #2056: `decoderResult` is not honored before executing the HttpApp

### LINKED ISSUE #2065: allow users to handle the decoding errors
As discussed in #2063, we would need a better way to allow users to handle the decoding errors, so that they can return different response based on their needs, which is not possible at the moment.

Having new combinators to handle Http-intrinsic errors, e.g., a decoding error due to the header being too large.

(Thanks @vigoo for summarizing the alternatives)
1. Update `Http` to be modelled by `Either[DecodingError, Response] => ZIO[R, Response, Response]`
2.  model decoding errors in the `Request` class and do the check within the Http app
3. be able to produce a `Response` in the exception callback of the server.

### LINKED ISSUE #2071: Add principled, modular means of handling server errors
Conceptually, an `Http` will be modified as:

```scala
Either[Cause[Nothing], In] => ZIO[Env, Err, Out]
```

Practically, however, we will do it by a pair of functions, to avoid boxing:

```scala
serverErrorHandler: Cause[Nothing]=> ZIO[Env, Err, Out]
inputHandler: In => ZIO[Env, Err, Out]
```

When trying to decide how to implement the primitive operators, we look to the conceptual model (the first one, above).

This will allow us to, inside `Http` (rather than `Server`), expose server errors, decoding errors, HTTP composition errors (e.g. NPE in wiring up two Http), and so forth, each `Http` handling such unexpected / fatal / server errors in its own modular way.

vigoo:
Resolves #2071
Resolves #2056
Resolves #2065

### DIFF
--- a/zio-http/src/main/scala/zio/http/Http.scala
+++ b/zio-http/src/main/scala/zio/http/Http.scala
@@ -96,10 +96,29 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
   )(implicit trace: Trace): Http[R1, Err1, In1, Out1] =
     self.defaultWith(that)
 
-  final def catchAllDefect[R1 <: R, Out1 >: Out](f: Cause[Nothing] => ZIO[R1, Nothing, Out1])(implicit
+  final def catchAllZIO[R1 <: R, Err1 >: Err, Out1 >: Out](
+    f: Err => ZIO[R1, Err1, Out1],
+  )(implicit trace: Trace): Http[R1, Err1, In, Out1] =
+    self match {
+      case Http.Empty(errorHandler)           => Http.Empty(errorHandler)
+      case Http.Static(handler, errorHandler) =>
+        Http.Static(handler.catchAll(err => Handler.fromZIO(f(err))), errorHandler)
+      case route: Route[R, Err, In, Out]      =>
+        new Route[R1, Err1, In, Out1] {
+          override def run(in: In): ZIO[R1, Err1, Http[R1, Err1, In, Out1]] =
+            route.run(in).map(_.catchAllZIO(f))
+
+          override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
+            route.errorHandler
+        }
+    }
+
+  final def catchAllCauseZIO[R1 <: R, Out1 >: Out](f: Cause[Err] => ZIO[R1, Nothing, Out1])(implicit
     trace: Trace,
   ): Http[R1, Err, In, Out1] =
-    self.withErrorHandler(Some(f))
+    self
+      .catchAllZIO(err => f(Cause.fail(err)))
+      .withErrorHandler(Some(f))
 
   /**
    * Named alias for `++`
--- a/zio-http/src/main/scala/zio/http/Http.scala
+++ b/zio-http/src/main/scala/zio/http/Http.scala
@@ -24,7 +24,7 @@ import zio._
 
 import zio.stream.ZStream
 
-import zio.http.Http.{Empty, Route}
+import zio.http.Http.{Empty, FailedErrorHandler, Route}
 import zio.http.model.Header.HeaderType
 import zio.http.model.{Header, Headers, MediaType, Status}
 import zio.http.socket.{SocketApp, WebSocketChannelEvent}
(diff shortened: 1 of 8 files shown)


zio-http/src/main/scala/zio/http/middleware/Auth.scala
```diff
@@ -150,19 +150,26 @@ private[zio] trait Auth {
...
-                  Http.fromHttpZIO[Request] { in =>
-                    route
-                      .asInstanceOf[Http.Route[R1, Err1, Request, Response]]
-                      .run(in)
-                      .provideSomeEnvironment[R0](_.union[Context](ZEnvironment(context)))
-                      .map { (http: Http[R1, Err1, Request, Response]) =>
-                        self.apply(http)
-                      }
-                  }
+                case empty @ Http.Empty(_)              => empty.asInstanceOf[Http.Empty[R0, Response]]
```
vigoo: These casts in this file are probably invalid and we need to provide environment for the error handler too - to be checked

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -36,14 +36,20 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
    */
   final def >>>[R1 <: R, Err1 >: Err, In1 >: Out, Out1](
     handler: Handler[R1, Err1, In1, Out1],
+    errorOutputMapper: Out => Out1,
```
vigoo, line 39: This operator became weird by this parameter, and in fact - at least in zio-http's codebase - every time it is used, `Out1` is `Unit`. So maybe we should replace it with a stricter, single parameter version.

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -70,33 +76,46 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
...
         new Route[R1, Err1, In1, Out1] {
           override def run(in: In1): ZIO[R1, Err1, Http[R1, Err1, In1, Out1]] =
             route.run(in).map(_.defaultWith(that))
+
+          override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
+            route.errorHandler
         }
     }
 
+  private[http] def errorHandler: Option[Cause[Nothing] => ZIO[R, Nothing, Out]]
```
jdegoes, line 142: Why put it into an Option?

vigoo: We cannot have a default error handler producing arbitrary Out. So having one is optional and on the edge where it is always known to be Response we can choose a default handler if needed

vigoo: And also this way we don’t have to catchAll ZIOs on each level just where we specify the handler

zio-http/src/main/scala/zio/http/Http.scala
jdegoes: Yes, I agree. This is a bit too weird!

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -70,33 +96,46 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
   )(implicit trace: Trace): Http[R1, Err1, In1, Out1] =
     self.defaultWith(that)
 
+  final def catchAllDefect[R1 <: R, Out1 >: Out](f: Cause[Nothing] => ZIO[R1, Nothing, Out1])(implicit
```
jdegoes: I think `catchAllCause` should let you get the defects. No need for a `catchAllDefect` method in particular.

vigoo: Fixed

jdegoes (approved):

PR #2074, opened 2023-03-30
WIP

jdegoes:
- [X] Polish ZClient
 - [x] Polish URL 
 - [ ] Polish Http
 - [ ] Polish Handler

### DIFF
--- a/zio-http/src/main/scala/zio/http/QueryParams.scala
+++ b/zio-http/src/main/scala/zio/http/QueryParams.scala
@@ -27,16 +27,12 @@ import zio.http.model.Charsets
 /**
  * A collection of query parameters.
  */
-final case class QueryParams private[http] (map: Map[String, Chunk[String]])
-    extends scala.collection.Map[String, Chunk[String]] {
+final case class QueryParams(map: Map[String, Chunk[String]]) {
   self =>
 
-  override def -(key: String): QueryParams = QueryParams(map - key)
+  def -(key: String): QueryParams = QueryParams(map - key)
 
-  override def -(key1: String, key2: String, keys: String*): QueryParams =
-    QueryParams(map.--(Chunk(key1, key2) ++ keys))
-
-  override def +[V1 >: Chunk[String]](kv: (String, V1)): Map[String, V1] = map.+(kv)
+  def -(key1: String, key2: String, keys: String*): QueryParams = QueryParams(map - (key1, key2, keys: _*))
 
   def ++(that: QueryParams): QueryParams =
     QueryParams(that.map.foldLeft(map) { case (map, (k, v)) =>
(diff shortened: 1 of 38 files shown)


zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -64,19 +64,29 @@ case class QueryParams private[http] (map: Map[String, Chunk[String]])
...
 
   override def iterator: Iterator[(String, Chunk[String])] = map.iterator
 
+  def normalize: QueryParams =
+    if (isEmpty) self
+    else QueryParams(map.filter(i => i._1.nonEmpty && i._2.nonEmpty))
+
+  def toForm: Form = Form.fromQueryParams(self)
+
+  def toMap: Map[String, Chunk[String]] = map
```
vigoo: Looks a bit confusing that

- we are extending `Map[String, Chunk[String]]`
- also have a public field `map: Map[String, Chunk[String]]`
- and a `toMap: Map[String, Chunk[String]]` 

and `self` is not the same map as the "inner map"; one could ask what is the difference (apart from having extra methods on `QueryParams`), are all map operations equivalent to regardless when you call `toMap`, etc.

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -28,60 +28,43 @@ import zio.http.socket.SocketApp
 import java.net.{InetSocketAddress, URI} // scalafix:ok;
 
 trait ZClient[-Env, -In, +Err, +Out] extends HeaderOps[ZClient[Env, In, Err, Out]] { self =>
```
vigoo, line 30: Now that I moved config to `ZClient.Config`, I was wondering it would look better to have `Client` as the "main" companion object, hand have the layers and config class in that instead of `ZClient`. Would be more symmetric to `Server`.

zio-http/src/main/scala/zio/http/QueryParams.scala
jdegoes: Agreed, will clean up 👍

zio-http/src/main/scala/zio/http/ZClient.scala
jdegoes: How about `val Client = ZClient` or is that too ZIO 1.0-ish?

adamgfraser: I think that's the best we came up with when using this more polymorphic variant pattern.

vigoo: We already have that. I proposed to switch them and make `Client` the "real" `object`, because right now in type signatures you have to use `ZClient.Config`

adamgfraser: Don't we just need a type alias? `type Client = ZClient.type`

jdegoes: We could do `trait Client[...] extends Client.Poly[...]`

jdegoes:
@vigoo Appreciate a review if you get a chance, would like to split this up since it's already so big.

vigoo (approved):

PR #2075, opened 2023-03-31
octet-stream as default multipart form contentType

liewhite:
In many clients, content-type in `multipart/form-data` is not specified, for best compatibility, we should use `application/octet-stream` as content-type when not specified in body.


liewhite:
Wrong email, recreate it

PR #2086, opened 2023-04-08
Add mapErrorZIO to Handler and Http

ex0ns:
Feel free to discard this PR but I was confused when migrating to 0.0.5 as those methods were missing.

I have included some tests that I used to understand the changes between 2.X and 0.0.5. 

I am a bit confused about `errorHandler` and how to deal with it in the `mapErrorZIO` case.


vigoo:
Thanks! Please run `sbt fmt`

vigoo (approved):

PR #2097, opened 2023-04-14
Endpoint#outErrorWith

### LINKED ISSUE #2091: Allow specifying different status codes for an error ADT in Endpoint
Currently `.outError` on Endpoint is not very flexible when working with an ADT modeling different kinds of errors that could map to different status codes. It would be nice to have a method  `.outErrorWith` that takes a function that maps the ADT to a status code.

Something like

```scala
Endpoint
  .get("example")
  .out[String]
  .outErrorWith[Error] {
     case Error.Example1 => Status.BadRequest
     case Error.Example2 => Status.NotFound
     case _              => Status.InternalServerError
  }
```

vigoo:
Resolves #2091

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
@@ -17,6 +17,7 @@
 package zio.http.codec
 
 import scala.language.implicitConversions
+import scala.reflect.ClassTag
 
 import zio._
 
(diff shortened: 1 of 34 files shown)


zio-http/src/main/scala/zio/http/codec/SimpleCodec.scala
```diff
@@ -49,4 +49,16 @@ object SimpleCodec {
 
     def isDefinedAt(input: A): Boolean = true
   }
+
+  final case class Custom[Input, Output](f: PartialFunction[Input, Output], g: Output => Input)
```
jdegoes: I am not sure this is the right place to solve the problem, or, indeed, that it is necessary to solve it by adding something new (as opposed to re-using alternation).

Basically, can we write the following function:

```scala
def outErrors[Err2](map: Map[Status, Err2])(implicit ...)
```

I think we can write it using only `|` and `transformOrFail`.

(Why `Map` and why not a pair of functions? Because the status code contains no details so every status code can be mapped to something like a case object of a sealed trait; there's no extra content to include in the error.)

I also have my suspicious that what is wanted here is not so much "multiple status codes", but multiple error types.

In which case one could imagine something like:

```scala
def outErrors[Sum, Term1 <: Sum: Tag, Term2 <: Sum: Tag](
  term1: HttpCodec[CodecType.Status & CodecType.Content, Term1],
  term2: ...
)
```

So you do something like:

```scala
endpoint.outErrors(error1, error2, error3)
```

where:

```scala
val error1 = StatusCodec.status[MyError1](Status.NotFound)
```

where `status` takes an implicit `Schema[MyError1]` and uses that to define the body.

vigoo: I first wanted to do something similar to this but I was thinking of somehow automatically combining multiple `outError`s having the same supertype into the supertype instead of an `Either`. Then chose to try to solve it differently. 

Your proposed version definitely looks much cleaner. Although the `SimpleCodec` with being a partial function already, seemed like something that was inteded to support this.

jdegoes: Maybe we could do with: `HttpCodec.enum(..., ..., ...)` just as a way to simplify the subtyping stuff.

Although used for errors, it could be used for other things too.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -81,6 +81,10 @@ private[codec] object EncoderDecoder                   {
           case error: HttpCodecError =>
             // TODO: Aggregate all errors in disjunction:
             lastError = error
+
+          case other: Throwable =>
+            other.printStackTrace()
```
jdegoes: Debug code?

vigoo: yes, removing

zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -187,6 +190,200 @@ final case class Endpoint[Input, Err, Output, Middleware <: EndpointMiddleware](
...
+  ](
+    codec1: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub1],
+    codec2: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub2],
+    codec3: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub3],
+    codec4: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub4],
+    codec5: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub5],
+    codec6: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub6],
+    codec7: HttpCodec[HttpCodecType.Status & HttpCodecType.Content, Sub7],
+  )(implicit alt: Alternator[Err, Err2]): Endpoint[Input, alt.Out, Output, Middleware] = {
+    val codec = (codec1 | codec2 | codec3 | codec4 | codec5 | codec6 | codec7).transformOrFail(
```
jdegoes: These look fantastic! However, can you pull this functionality out into `HttpCodec.enum(.., ..)`?

Then these implementations will be trivially delegating to `HttpCodec.enum`, and other code can use the feature for various sum types.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -273,6 +274,197 @@ object HttpCodec
     with StatusCodecs {
   implicit def stringToLiteral(string: String): PathCodec[Unit] = literal(string)
 
+  def `enum`[AtomTypes, Value, Sub1 <: Value: ClassTag, Sub2 <: Value: ClassTag](
```
jdegoes: Maybe best to call this `enumeration` so we don't have to escape for Scala 3.

jdegoes (changes requested):
Looks great! Just one comment on the name `enum` (which I should have foreseen) and then good to merge.

zio-http/src/test/scala/zio/http/endpoint/EndpointSpec.scala
```diff
@@ -302,6 +302,59 @@ object EndpointSpec extends ZIOSpecDefault {
...
+            response <- routes.toApp.runZIO(request).mapError(_.get)
+            body     <- response.body.asString.orDie
+          } yield assertTrue(response.status.code == 999, body == "\"path(users, 123)\"")
+        },
+        test("status depending on the error subtype") {
+          val routes =
+            Endpoint
+              .get(literal("users") / int("userId"))
+              .out[String]
+              .outErrors[TestError, TestError.UnexpectedError, TestError.InvalidUser](
```
jdegoes: I wonder if we can improve this to:

```suggestion
              .outErrors[TestError](
```

by using partial application trick (intermediate class with an `apply`)?

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -273,6 +274,202 @@ object HttpCodec
...
+          case sub4: Sub4 => Right(Left(Left(Left(Left(Right(sub4))))))
+          case sub5: Sub5 => Right(Left(Left(Left(Right(sub5)))))
+          case sub6: Sub6 => Right(Left(Left(Right(sub6))))
+          case sub7: Sub7 => Right(Left(Right(sub7)))
+          case sub8: Sub8 => Right(Right(sub8))
+          case _          => Left(s"Unexpected error type")
+        },
+    )
+
+  def statusAndContent[Body](status: zio.http.Status)(implicit
```
jdegoes: Not sure about this one, but perhaps:

```suggestion
  def error[Body](status: zio.http.Status)(implicit
```

Also we could think about putting it inside `StatusCodecs`, although I'm not 100% sure about that one either.

vigoo: It is not only a status codec, so I would leave it there

zio-http/src/test/scala/zio/http/endpoint/EndpointSpec.scala
vigoo: Done (both here and in the `HttpCodec.enumeration`)

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -273,6 +274,14 @@ object HttpCodec
     with StatusCodecs {
   implicit def stringToLiteral(string: String): PathCodec[Unit] = literal(string)
 
+  def enumeration[AtomType, Value]: Enumeration[AtomType, Value] =
```
jdegoes: Due to contravariance of `AtomType`, can we just let it widen automatically?

jdegoes: That would reduce it down to a single type parameter, unless there's some problem with that approach.

zio-http/src/test/scala/zio/http/endpoint/EndpointSpec.scala
```diff
@@ -302,6 +302,59 @@ object EndpointSpec extends ZIOSpecDefault {
...
+            response <- routes.toApp.runZIO(request).mapError(_.get)
+            body     <- response.body.asString.orDie
+          } yield assertTrue(response.status.code == 999, body == "\"path(users, 123)\"")
+        },
+        test("status depending on the error subtype") {
+          val routes =
+            Endpoint
+              .get(literal("users") / int("userId"))
+              .out[String]
+              .outErrors[TestError](
```
jdegoes, line 332: Looks superb! About as easy as you could hope for given the constraints we're working with.

jdegoes (approved):

PR #2099, opened 2023-04-14
Graceful shutdown

### LINKED ISSUE #2053: Allow for in-flight requests to finalize before shutting down the server

The current behaviour of `zio-http` is to interrupt in-flight requests when the server is shutting down. E.g., if we send a SIGINT signal to the app below while the request is being processed, we get the following in curl:

App:
```scala
import zio._
import zio.http._
import zio.http.model.Method

  private val app = Http.collectZIO[Request] { case Method.GET -> !! / "hello" =>
    ZIO.sleep(10.seconds).as(Response.text("hello!"))
  }

  override val run =
    Server.serve(app).provideLayer(Server.default)
}
```

Terminal:
```shell
curl -vvv http://localhost:8080/hello
*   Trying 127.0.0.1:8080...
* Connected to localhost (127.0.0.1) port 8080 (#0)
> GET /hello HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.86.0
> Accept: */*
>
* Empty reply from server
* Closing connection 0
curl: (52) Empty reply from server
```

It would be good if zio-http allowed configuring the shutdown timeout. Once a termination signal is received, the following sequence begins:

1. Stops accepting new requests
2. Waits for in-flight requests to finish processing (up to the `shutdownTimeout`)
3. Closes open connections

One way this could be done is by forcing a shutdown timeout using `Server.serve(app).onInterrupt(ZIO.sleep(30.seconds))`. However, this approach has the following limitations:

1. The server continues to accept requests during the shutdown period
2. The server remains online for the full duration of the shutdown period instead of the time it takes to complete in-flight requests

**Additional context**

This feature is supported fully by [Akka-http](https://doc.akka.io/docs/akka-http/current/server-side/graceful-termination.html#akka-coordinated-shutdown) and partially by http4s (http4s continues to accept requests after the termination signal is received. It would be great if zio-http also provided out-of-the-box 

I had a look through the issues and saw that there is an [open feature request for graceful termination](https://github.com/zio/zio-http/issues/1505), however I'm not sure whether that addresses the full scope of this feature request.

vigoo:
Resolves #2053

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -18,6 +18,7 @@ package zio.http.netty.server
 
 import java.io.IOException
 import java.net.InetSocketAddress
+import java.util.concurrent.atomic.AtomicInteger
 
 import scala.annotation.tailrec
 
@@ -47,6 +48,8 @@ private[zio] final case class ServerInboundHandler(
   private var app: App[Any]          = _
   private var env: ZEnvironment[Any] = _
 
+  val inFlightRequests: AtomicInteger = new AtomicInteger(0)
+
   def refreshApp(): Unit = {
     val pair = appRef.get()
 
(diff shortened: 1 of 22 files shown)


zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -47,6 +48,8 @@ private[zio] final case class ServerInboundHandler(
   private var app: App[Any]          = _
   private var env: ZEnvironment[Any] = _
 
+  val inFlightRequests: AtomicInteger = new AtomicInteger(0)
```
jdegoes: Should probably be `LongAdder`.

jdegoes (approved):

PR #2101, opened 2023-04-14
Create initial version of HttpCliApp

jorge-vasquez-2301:
This PR is to create the initial version of `HttpCliApp`, which will allow to generate a zio-cli `CliApp` from zio-http `Endpoints`, by calling `HttpCliApp.fromEndpoints`


jdegoes:
Good to merge when the build passes! We can do followups separately, although I will comment in this PR.

zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpCliApp.scala
```diff
@@ -0,0 +1,771 @@
...
+import zio.cli._
+import zio.json.ast._
+
+import zio.schema._
+
+import zio.http._
+import zio.http.codec._
+import zio.http.endpoint._
+
+class description(val text: String) extends StaticAnnotation // TODO move this to zio-schema
```
jdegoes, line 17: TODO: Move to ZIO Schema.

zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpCliApp.scala
```diff
@@ -0,0 +1,771 @@
...
+
+object HttpCliApp {
+  private[cli] final case class CliRequest(
```
jdegoes, line 20: I would pull all these nested classes outside of HttpCliApp and make them into separate files, for better organization.

jorge-vasquez-2301:
@jdegoes build has finally passed 🙂

zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpCliApp.scala
```diff
@@ -0,0 +1,771 @@
...
+   *   The server host where the generated CLI will send requests to
+   * @param port
+   *   The server port where the generated CLI will send requests to
+   * @return
+   *   a [[zio.cli.CliApp]]
+   */
+  def fromEndpoints[M <: EndpointMiddleware](
+    name: String,
+    version: String,
+    summary: String,
```
jdegoes, line 710: If in `zio-cli`, any of these take more rich docs, then we could accept the richer types here, rather than just `String`, so you could do formatting if you wanted to. The user can always lift a string to the richer doc.

jdegoes (approved):

jdegoes:
The other change I would make is wrap `CliApp` in `HttpCliApp` so we have a `case class` with proper companion object, just as a way to organize the code.

jorge-vasquez-2301:
Thanks for your comments @jdegoes! I'll address them in a follow up PR. Could you please approve this one for now so it can be merged?

PR #2102, opened 2023-04-14
feature: retrieve repeatable headers from response

ex0ns:
I'm not sure about the name of the method yet, but I would like to avoid `getHeaders` and `headers` is already taken.
Should we keep `Seq` as the return type ?

### DIFF
--- a/zio-http/src/main/scala/zio/http/internal/HeaderGetters.scala
+++ b/zio-http/src/main/scala/zio/http/internal/HeaderGetters.scala
@@ -16,6 +16,8 @@
 
 package zio.http.internal
 
+import zio.Chunk
+
 import zio.http.Header.HeaderType
 import zio.http.Headers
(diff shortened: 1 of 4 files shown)


vigoo (approved):
I'm OK with the name and `Seq`.

zio-http/src/main/scala/zio/http/internal/HeaderGetters.scala
```diff
@@ -37,6 +38,18 @@ trait HeaderGetters { self =>
       parsed.toOption
     }
 
+  final def headers(headerType: HeaderType): Seq[headerType.HeaderValue] =
```
jdegoes: The rule is:

Accept the most generic collection type, provide the most specific collection type.

So here, maybe use Chunk so the user has some guarantee about the resulting Seq.

ex0ns:
Tests are failing in 2.12 when using `Chunk.from`, moving to `Chunk.fromIterator` works

PR #2115, opened 2023-04-24
Fix usage of Config.nested

### LINKED ISSUE #2106: Default configuration nesting is broken
The `configured` layers have a default nesting set up which does not work as expected.

For example `Server.configured` expects the server configuration to be at `nested("zio.http.server")` but that does not mean it will be read from a HOCON's `zio.http.server` node, for example. It should be `nested("zio", "http", "server")`.

We have to review all uses for `nested` and fix this.

vigoo:
Resolves https://github.com/zio/zio-http/issues/2106

### DIFF
--- a/zio-http/src/main/scala/zio/http/DnsResolver.scala
+++ b/zio-http/src/main/scala/zio/http/DnsResolver.scala
@@ -254,8 +254,10 @@ object DnsResolver {
     )
   }
 
-  def configured(path: String = "zio.http.dns"): ZLayer[Any, zio.Config.Error, DnsResolver] =
-    ZLayer(ZIO.config(Config.config.nested(path))) >>> live
+  def configured(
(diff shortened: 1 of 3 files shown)


adamgfraser (approved):

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -254,8 +254,10 @@ object DnsResolver {
     )
   }
 
-  def configured(path: String = "zio.http.dns"): ZLayer[Any, zio.Config.Error, DnsResolver] =
-    ZLayer(ZIO.config(Config.config.nested(path))) >>> live
+  def configured(
```
adamgfraser, line 257: Maybe variable arguments instead of `NonEmptyChunk`?

```scala
def configured(path: String, paths: String*): ZLayer[Any, Error, DnsResolver] =
  ???
```

PR #2116, opened 2023-04-24
Significantly improve performance of streaming multipart/form-data

### LINKED ISSUE #2113: [3.0.0-RC1] Streaming multipart performance issue for any significant binary sizes
**Describe the bug**
Using `asMultipartFormStream` to process a multipart form request using streams is showing poor performance. It also does not seem to actually emit any streaming binary data for files in a multipart request until all of the request is internally processed or buffered in some way.

**To Reproduce**
Set up two endpoints, one using `req.body.asStream` and one using `req.body.asMultipartFormStream`. 
Running the streams through `ZSink.count` to simply count the bytes and respond with the count.
```scala
  private def uploadRawBodyStream(req: Request)    =
    for {
      res <- req.body.asStream
               .run(ZSink.count)
    } yield Response.text(s"bytes counted $res")

  private def uploadMultipartStream(req: Request)    =
    for {
      form   <- req.body.asMultipartFormStream
      res    <- form.fields.flatMap { case sb: FormField.StreamingBinary => sb.data }
               .run(ZSink.count)
    } yield Response.text(s"bytes counted $res")
```

Using httpie to send a file such as an image:
`http -p=mb --multipart localhost:8080/upload/multipart "file@./large.png"`

Im my case I'm sending a `4204247` byte `image/png`.

**Expected behaviour**
Performance to be closer to processing the request body stream directly, although of course a bit of overhead is understandable, but the level I'm seeing is making it not realistic for practical use. Using an actually big file, such as `100 MB`, is essentially impossible.

**Screenshots**
Here's the timings for a request towards both of these endpoints.
*Note*: this is with a `4MB` image.
Screenshot 2023-04-24 at 09 29 45

Here's CPU and Heap footprint for a single request using the `StreamingForm` processing:
Screenshot 2023-04-24 at 09 33 24
*Horizontal axis is 1 minute*

Same for directly processing the body stream:
Screenshot 2023-04-24 at 09 34 03

 - OS: macOS Ventura 13.2.1

vigoo:
Resolves https://github.com/zio/zio-http/issues/2113

### DIFF
--- a/zio-http/src/main/scala/zio/http/StreamingForm.scala
+++ b/zio-http/src/main/scala/zio/http/StreamingForm.scala
@@ -22,6 +22,7 @@ import zio.{Chunk, Queue, Trace, Unsafe, ZIO}
 
 import zio.stream.{Take, ZStream}
 
+import zio.http.StreamingForm.Buffer
 import zio.http.internal.{FormAST, FormState}
 
 final case class StreamingForm(source: ZStream[Any, Throwable, Byte], boundary: Boundary, bufferSize: Int = 8192) {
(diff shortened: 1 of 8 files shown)


adamgfraser (commented):
Why `flatMapPar` with parallelism of `1`?

zio-http/src/main/scala/zio/http/StreamingForm.scala
```diff
@@ -70,103 +74,135 @@ final case class StreamingForm(source: ZStream[Any, Throwable, Byte], boundary:
...
+        inNonStreamingPart = false,
+      )
   }
 
-  private def initialState(boundary: Boundary): State =
-    State(boundary, FormState.fromBoundary(boundary), None, immutable.Queue.empty, inNonStreamingPart = false)
+  private def initialState(boundary: Boundary, bufferSize: Int): State =
+    State(boundary, FormState.fromBoundary(boundary), None, Buffer.empty(bufferSize), inNonStreamingPart = false)
+
+  private case class Buffer(buffer: Array[Byte], length: Int) {
```
jdegoes: Could consider using `ByteBuffer` here if it makes sense.

jdegoes: I mean, instead of an array + index.

jdegoes: This isn't really immutable because of `buffer(length) = byte`, so copying along the buffer suggests the wrong thing.

vigoo: Yeah it was originally immutable, changing per byte and now it is half mutable; will make it fully mutable instead.

vigoo:
> Why `flatMapPar` with parallelism of `1`?

Good question, I just based it on some older code which used `mapZIOPar` and when I wrote that I still remembered why but not anymore. (but it was blocking forever without it). I will check what's going on there

jdegoes (approved):

jdegoes (approved):

PR #2123, opened 2023-04-27
WIP: create csrf request handler middleware

williamareynolds:
Added a csrfGenerate and a csrfValidate middleware similar to what used to be in the EndpointMiddlewares in 0.0.4. I opted to create these outside of the Endpoint API for the time being since I'm less familiar with it, just so I can get some early feedback.


williamareynolds:
I'm going to save the EndpointMiddleware version of this for a separate PR. I'll add some tests for what I have here, then mark it complete.

jdegoes:
@williamareynolds Please re-open if you get a chance to push this forward.

Issue #2122, opened 2023-04-26
g8 template


TomTriple:
I updated the g8 template for ZIO Http: [https://github.com/TomTriple/zio-http-start.g8](https://github.com/TomTriple/zio-http-start.g8) - actually I think it should be hosted at ZIO. In this case we could remove the template from Dream11 and update [the docs](https://zio.dev/zio-http/setup#using-dream11s-g8-template) to using: 

`sbt new zio/zio-http.g8`

Do we want that starter to be hostet at ZIO ? 

At the moment a resulting start page renders as: 

![Screenshot 2023-05-01 at 19-05-25 ZIO Http](https://user-images.githubusercontent.com/202410/235494368-e5162e9b-f3e9-4682-87b5-4be40c102563.png)

dejvid:
Wow that is great!

vigoo:
Would be nice to have one using the new `Endpoint` API too :)

jdegoes:
Agreed! Please let us know what you need to make this happen. 

@khajavi Please assist Tom. 🙏

TomTriple:
I updated the starter project once again. Of course, if anyone wants a preview simple use `sbt new TomTriple/zio-http-start.g8` to create a project from the current template. As of now it's a tiny tour of zio-http that hopefully also serves as a good project starting point. 

![zio-starter](https://user-images.githubusercontent.com/202410/236624172-a1ca8e00-94c9-4ff3-9223-d374b53ec3a3.png)

@khajavi I think we could now host that at ZIO organization at `zio/zio-http.g8`? When that is done we should update the docs and remove the old g8 template from Dream11.

khajavi:
@TomTriple Will check that shortly today!

khajavi:
@TomTriple 
The `BuildHelper` is deprecated and now we use [ZIO SBT](https://github.com/zio/zio-sbt) plugins. After removing that, we can transfer that to the ZIO organization.  Anyway, If there's anything I can help you with.

TomTriple:
@khajavi 
I removed the `BuildHelper` as your proposal and added `ZIO-SBT` (`zio-sbt-ecosystem`). Tests for the generated application worked with Scala 2.13, 3.2.x and the newest 3.3.0-RC5. 

Please note, that this template still can be improved in many ways! 

[...]

PR #2133, opened 2023-05-01
link to updated g8 template

TomTriple:
Before this PR makes sense we would need to host a new g8 starter at ZIO (please see https://github.com/zio/zio-http/issues/2122), which I hope would make sense.


TomTriple:
The PR now makes sense as we have the repo available

khajavi:
@TomTriple Thank you!

PR #2135, opened 2023-05-03
Fix mapErrorZIO

vigoo:


adamgfraser (approved):
Might call this `Err2` and use the convention that `Err1 >: Err` but otherwise looks great!

vigoo:
> Might call this `Err2` and use the convention that `Err1 >: Err` but otherwise looks great!

I don't think it is consistent now in other places so just merged it

adamgfraser:
👍 

We've at best been "weakly consistent" about that pattern. 😃

PR #2138, opened 2023-05-03
added basic implementation for Server sent events on the server side.

### LINKED ISSUE #2137: Server Sent Events support
Add support for Server Sent Events on the server side.

devsprint:
Fixes: #2137  - basic implementation for SSE on server side.

### DIFF
--- a/zio-http-example/src/main/scala/example/SSEServer.scala
+++ b/zio-http-example/src/main/scala/example/SSEServer.scala
@@ -1,0 +1,25 @@
+package example
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter.ISO_LOCAL_TIME
+
+import zio.{ExitCode, Schedule, URIO, ZIOAppDefault, durationInt}
+
+import zio.stream.ZStream
+
+import zio.http._
+import zio.http.sse.ServerSentEvent
+
+object SSEServer extends ZIOAppDefault {
+
+  val stream: ZStream[Any, Nothing, ServerSentEvent] =
+    ZStream.repeatWithSchedule(ServerSentEvent(ISO_LOCAL_TIME.format(LocalDateTime.now)), Schedule.spaced(1.second))
+
+  val app: Http[Any, Nothing, Request, Response] = Http.collect[Request] { case Method.GET -> !! / "sse" =>
+    Response.sse(stream)
+  }
+
+  val run: URIO[Any, ExitCode] = {
+    Server.serve(app.withDefaultErrorResponse).provide(Server.default).exitCode
+  }
+}
--- a/zio-http-example/src/main/scala/example/SSEServer.scala
+++ b/zio-http-example/src/main/scala/example/SSEServer.scala
@@ -8,7 +8,6 @@ import zio.{ExitCode, Schedule, URIO, ZIOAppDefault, durationInt}
 import zio.stream.ZStream
 
 import zio.http._
-import zio.http.sse.ServerSentEvent
 
 object SSEServer extends ZIOAppDefault {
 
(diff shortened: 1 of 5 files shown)


zio-http/src/main/scala/zio/http/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,44 @@
...
+      sb.append("retry: ").append(r).append('\n')
+    }
+    sb.append('\n').toString
+  }
+}
+
+object ServerSentEvent {
+  def apply(data: String): ServerSentEvent = new ServerSentEvent(data)
+
+  def heartbeat(): ServerSentEvent = new ServerSentEvent("")
```
jdegoes:
```suggestion
  def heartbeat: ServerSentEvent = ServerSentEvent("")
```

zio-http/src/main/scala/zio/http/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,44 @@
...
+    }
+    retry.foreach { r =>
+      sb.append("retry: ").append(r).append('\n')
+    }
+    sb.append('\n').toString
+  }
+}
+
+object ServerSentEvent {
+  def apply(data: String): ServerSentEvent = new ServerSentEvent(data)
```
jdegoes: Why not rely on `case class` apply?

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -405,7 +408,17 @@ object Response {
...
 
-  private lazy val contentTypeJson: Headers = Headers(Header.ContentType(MediaType.application.json).untyped)
-  private lazy val contentTypeHtml: Headers = Headers(Header.ContentType(MediaType.text.html).untyped)
-  private lazy val contentTypeText: Headers = Headers(Header.ContentType(MediaType.text.plain).untyped)
+  /**
+   * Creates a response with content-type set to text/event-stream
+   * @param data
+   *   \- stream of data to be sent as Server Sent Events
+   */
+  def sse(data: ZStream[Any, Nothing, ServerSentEvent]): Response =
```
jdegoes: Maybe: `fromServerSentEvents`? (Not sure, if `sse` is commonly understood by those interested in using the protocol, then it's probably fine as `sse`.)

zio-http/src/main/scala/zio/http/sse/ServerSentEvent.scala
devsprint: done

zio-http/src/main/scala/zio/http/sse/ServerSentEvent.scala
devsprint: right

zio-http/src/main/scala/zio/http/Response.scala
devsprint: even I think it is commonly used, `fromServerSentEvents` is much better.

zio-http-example/src/main/scala/example/SSEServer.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter.ISO_LOCAL_TIME
+
+import zio.{ExitCode, Schedule, URIO, ZIOAppDefault, durationInt}
+
+import zio.stream.ZStream
+
+import zio.http._
+import zio.http.sse.ServerSentEvent
```
jdegoes: Just one more quick comment here: if there's only ever going to be one thing in `sse`, then the package doesn't really pull its weight, and we should consider moving its sole data type to `zio.http`. What do you think?

devsprint: I was thinking that maybe we have to provide some encoding/decoding support for Server Sent Event, but I guess we can move it for now in `zio.http` and update it later if is need it.

jdegoes (approved):

jdegoes:
Thanks for your work on this!

PR #2139, opened 2023-05-03
Improve dx for ZClient

987Nabil:
- don't append trailing slash for empty pathSuffixes
- add no params method for get/head/delete

### DIFF
--- a/zio-http/src/main/scala/zio/http/ZClient.scala
+++ b/zio-http/src/main/scala/zio/http/ZClient.scala
@@ -18,8 +18,8 @@ package zio.http
 
 import zio._
 import zio.http.URL.Location
-import zio.http.netty.NettyConfig
 import zio.http.internal.HeaderOps
+import zio.http.netty.NettyConfig
 import zio.http.netty.client._
 import zio.http.socket.SocketApp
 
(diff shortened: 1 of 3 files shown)


jdegoes (changes requested):
Superb! Can you add a couple tests to make sure no one accidentally breaks the features you added?

987Nabil:
Sure!

987Nabil:
@jdegoes I added this now in the client impl., but I actually think it might be better to have this as the general behaviour of `URL`. Wdyt? This would be a breaking change.

jdegoes:
Yes, please push it into URL with tests. 🙏

987Nabil:
@jdegoes After taking a look at this again, I think it correct as is.
Having a call like `Client.url(...).get("")`adding a trailing slash is unexpected and with the current master state unavoidable.
But, `url / ""` adding a trailing slash seems reasonable, because that is what you wrote in the DSL. And if you want to ensure that there is no trailing slash, you have `URL#dropTrailingSlash`.
So I think this PR is good to go (at least from my side)

jdegoes (approved):

Issue #2118, opened 2023-04-25
Add examples to zio.endpoint.Endpoint


jdegoes:
An endpoint should be capable of storing several examples that can be used when generating OpenAPI documentations (or for testing, command-line clients, etc.).

It remains to be seen exactly how to do this. Most likely, the best way is to add examples directly to BodyCodec, so that the user has to specify the content type **as well as the examples** at the same time. The examples can be `List` and we can have the existing variations simply specify **no example**, while other variations take _varargs_, so the user can specify one or more examples for the body codec.

jdegoes:
A better way to do this:

Add `HttpCodec#examples` that accepts a `List[A]`, using declarative encoding, and just stores them. And then other code can process such examples.

Then we can add `examplesIn` and `examplesOut` to `Endpoint` so developers can attach examples to input and output.

PR #2144, opened 2023-05-04
Offer API to add examples to `HttpCodec` and `Endpoint` (#2118)

987Nabil:
/claim #2118

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
@@ -25,7 +25,7 @@ import zio.stream.ZStream
 
 import zio.schema.Schema
 
-import zio.http.{Status, _}
+import zio.http._
 
 /**
  * A [[zio.http.codec.HttpCodec]] represents a codec for a part of an HTTP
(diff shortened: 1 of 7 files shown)


987Nabil:
@adamgfraser the build failure looks to me, like a bug in the assertTrue macro for Scala 3. I guess that the overloaded method names are a problem. Do you know of such a bug?

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -195,6 +195,14 @@ sealed trait HttpCodec[-AtomTypes, Value] {
   ): Z =
     encoderDecoder.encodeWith(value)(f)
 
+  def examples(examples: Iterable[Value]): HttpCodec[AtomTypes, Value] =
+    HttpCodec.WithExamples(self, Chunk.fromIterable(examples))
+
+  def examples(examples: Value*): HttpCodec[AtomTypes, Value] =
```
jdegoes: To improve overloading resolution:

```suggestion
  def examples(example1: Value, examples: *): HttpCodec[AtomTypes, Value] =
```

jdegoes (approved):

PR #2145, opened 2023-05-05
Special support for byte stream on endpoint api in/out

### LINKED ISSUE #2141: Support binary element types for Endpoint API
Currently, the Endpoint API only supports JSON as input and output.

However, with multi-part encoding support on the way (#2140), it makes sense to revisit this limitation, and add support for binary streams for both request and response, whether multi-part form or not.

This would allow the user to use the Endpoint API in order to describe file uploads and file downloads.

To implement this, we must check the input and output elements. If such an element is a stream, and if the captured schema is in fact identically equal to the schema for a byte, then we know the stream is a binary stream, and we can leverage this to perform binary encoding / decoding: either for the full content, or for the part of a multi-part form request / response.

When this ticket is done, it will be possible to use Endpoint to describe multi-part (multiple named input / output elements), where some of those are binary streams, and which are encoded not using JSON (as with ordinary content), but with binary media types. Consequently, it will be possible, and there should be tests to demonstrate, the Endpoint API being used for file uploads / downloads, as well as other pieces of metadata.

vigoo:
Resolves https://github.com/zio/zio-http/issues/2141

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/ContentCodecs.scala
+++ b/zio-http/src/main/scala/zio/http/codec/ContentCodecs.scala
@@ -24,6 +24,6 @@ private[codec] trait ContentCodecs {
   def content[A](implicit schema: Schema[A]): ContentCodec[A] =
     HttpCodec.Content(schema)
 
-  def contentStream[A](implicit schema: Schema[A]): ContentCodec[ZStream[Any, Nothing, A]] =
+  def contentStream[A](implicit schema: Schema[A]): ContentCodec[ZStream[Any, Throwable, A]] =
     HttpCodec.ContentStream(schema)
 }
--- a/zio-http/src/main/scala/zio/http/codec/ContentCodecs.scala
+++ b/zio-http/src/main/scala/zio/http/codec/ContentCodecs.scala
@@ -20,10 +20,18 @@ import zio.stream.ZStream
 
 import zio.schema.Schema
 
+import zio.http.MediaType
+
 private[codec] trait ContentCodecs {
   def content[A](implicit schema: Schema[A]): ContentCodec[A] =
-    HttpCodec.Content(schema)
+    HttpCodec.Content(schema, mediaType = None)
 
-  def contentStream[A](implicit schema: Schema[A]): ContentCodec[ZStream[Any, Throwable, A]] =
-    HttpCodec.ContentStream(schema)
+  def content[A](mediaType: MediaType)(implicit schema: Schema[A]): ContentCodec[A] =
+    HttpCodec.Content(schema, mediaType = Some(mediaType))
+
+  def contentStream[A](implicit schema: Schema[A]): ContentCodec[ZStream[Any, Nothing, A]] =
+    HttpCodec.ContentStream(schema, mediaType = None)
+
+  def contentStream[A](mediaType: MediaType)(implicit schema: Schema[A]): ContentCodec[ZStream[Any, Nothing, A]] =
+    HttpCodec.ContentStream(schema, mediaType = Some(mediaType))
 }
--- a/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
@@ -494,7 +494,7 @@ object HttpCodec
   }
 
   private[http] final case class Status[A](codec: SimpleCodec[zio.http.Status, A], index: Int = 0)
-      extends Atom[HttpCodecType.Status, A]                           {
+      extends Atom[HttpCodecType.Status, A]                         {
     self =>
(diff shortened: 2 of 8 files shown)


zio-http/src/main/scala/zio/http/codec/ContentCodecs.scala
```diff
@@ -24,6 +24,6 @@ private[codec] trait ContentCodecs {
   def content[A](implicit schema: Schema[A]): ContentCodec[A] =
     HttpCodec.Content(schema)
 
-  def contentStream[A](implicit schema: Schema[A]): ContentCodec[ZStream[Any, Nothing, A]] =
+  def contentStream[A](implicit schema: Schema[A]): ContentCodec[ZStream[Any, Throwable, A]] =
```
vigoo: Maybe this should not be changed and the input stream's errors should be handled somehow before passing the stream to the implementation, but I did not see an obvious place to do that.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -326,11 +333,26 @@ private[codec] object EncoderDecoder                   {
     }
 
     private def encodeContentType(inputs: Array[Any]): Headers = {
-      val _ = inputs // TODO: Support multiple content types
-      if (jsonEncoders.length == 0) Headers.empty
-      else if (jsonEncoders.length == 1) {
-        Headers(Header.ContentType(MediaType.application.json))
-      } else throw new IllegalStateException("A request on a REST endpoint should have at most one body")
+      if (isByteStream(flattened.content)) {
+        Headers(Header.ContentType(MediaType.application.`octet-stream`)) // TODO: customizable content type
```
vigoo: We should be able to attach a custom media type to the body codec somehow, not sure if it should be part of this PR or not

zio-http/src/main/scala/zio/http/codec/ContentCodecs.scala
jdegoes: This should definitely not be changed.

The reason is that an endpoint handler cannot deal with `Throwable` errors. It only knows about typed errors, which are part of the protocol itself. `Throwable` errors indicate something fatal happened: for example, that an element of the stream was improperly encoded. No recovery is possible at the level of the endpoint handler. So it's a defect and must be handled as a fallback higher up (request handler, HTTP).

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -517,13 +517,13 @@ object HttpCodec
     def withIndex(index: Int): Content[A] = copy(index = index)
   }
   private[http] final case class ContentStream[A](schema: Schema[A], index: Int = 0)
-      extends Atom[HttpCodecType.Content, ZStream[Any, Nothing, A]] {
+      extends Atom[HttpCodecType.Content, ZStream[Any, Throwable, A]] {
```
jdegoes: Same here, this should be `Nothing` and we should `orDie` on throwables.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
jdegoes: I think it should be part of this PR. Basically, store an `Option[MediaType]` inside the codec, and add variants that allow it to be specified (overridden) when the content codec is specified.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -326,11 +333,26 @@ private[codec] object EncoderDecoder                   {
...
+      } else {
+        val _ = inputs // TODO: Support multiple content types
+        if (jsonEncoders.length == 0) Headers.empty
+        else if (jsonEncoders.length == 1) {
+          Headers(Header.ContentType(MediaType.application.json))
+        } else throw new IllegalStateException("A request on a REST endpoint should have at most one body")
```
jdegoes, line 345: This, btw, is where the multi-part stuff would go.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -326,11 +333,26 @@ private[codec] object EncoderDecoder                   {
...
+      }
     }
+
+    private def isByteStream(codecs: Chunk[BodyCodec[_]]): Boolean =
+      if (codecs.length == 1) {
+        codecs.headOption match {
```
jdegoes: `headOption` boxes. You already know there is one, so do `codecs(0)`.

jdegoes (changes requested):
Looking good! Just a few changes and we can get this merged. 💪

jdegoes (approved):

PR #2146, opened 2023-05-05
rename html to html5

TomTriple:
A top level import in a e.g. `UserRoutes.scala` like
```scala
import zio.http._
import zio.http.html._

case class UserRoutes(userService:IUserService) {
    private def withLayout(content: Dom) = 
        html(
            head(
                title("User")
            ), 
...

```
results in a: 

```scala
[error] /tmp/aaa/blupp/src/main/scala/com/example/blupp/UserRoutes.scala:13:9: reference to html is ambiguous;
[error] it is imported twice in the same scope by
[error] import zio.http.html._
[error] and import zio.http._
[error]         html(
[error]         ^
[error] one error found
```

A possible workaround would be `import zio.http.html.{html => html5, _}`. Otherwise, in order to improve DX, I think it's worthwhile to rename the element `html` to `html5`. With that change both toplevel imports shown above would work without any surprises.

### DIFF
--- a/zio-http-example/src/main/scala/example/HtmlTemplating.scala
+++ b/zio-http-example/src/main/scala/example/HtmlTemplating.scala
@@ -3,7 +3,7 @@ package example
 import zio._
 import zio.http._
 
-// Importing everything from `zio.html`
+// Importing everything from `zio.http.html`
 import zio.http.html._
 
 object HtmlTemplating extends ZIOAppDefault {
--- a/zio-http-example/src/main/scala/example/HtmlTemplating.scala
+++ b/zio-http-example/src/main/scala/example/HtmlTemplating.scala
@@ -1,19 +1,18 @@
 package example
 
 import zio._
-
 import zio.http._
 
-object HtmlTemplating extends ZIOAppDefault {
-  // Importing everything from `zio.html`
-  import zio.http.html._
+// Importing everything from `zio.html`
+import zio.http.html._
 
+object HtmlTemplating extends ZIOAppDefault {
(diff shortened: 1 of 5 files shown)


987Nabil:
how about renaming the package or changing the package structure?
I find it kinda ugly to have `html5` in the api.

TomTriple:
> I find it kinda ugly to have `html5` in the api.

That's a fast, pragmatic solution. When I think about the traditional webapps I wrote I usually have one surrounding application template that would be the *only place* that contains a `html5` element. But I see what you do not like. What better options do we have? 

- rename `html` to `htmldoc`
- rename the package to `zio.http.tags` 
- externalize to a new project `zio.tags` 
- leave it as it is

Maybe @vigoo can offer some guidance?

vigoo:
I think it would be better to leave the `html` function as is, and rename the package. That way we maintain the direct mapping between the names of these constructors and the constructed tags. I would also keep "html" in the package name so would not choose the proposed `tags`. How about `htmltags`? @jdegoes do you have any name suggestion?

jdegoes:
`htmltags` is two words, which is uncommon for package names. How about simply `template`? Or `tags`, though not common, is probably better than `html` to avoid the clash (don't love either, however).

In any case, I'm in favor of doing this, and doing it quickly before the next RC.

jdegoes:
@TomTriple Would love to get this in, please re-open if you get a chance to work on this!

TomTriple:
Hey @jdegoes - are you still in favor of renaming the package from `zio.http.html` to `zio.http.template`?

Issue #2143, opened 2023-05-04
Endpoint input decoding error handling


rhueara-ca:
**Describe the bug**
For the Endpoint API, I could not find a way to handle input decoding errors.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a endpoint and add an implementation to it:
```
val postResource = Endpoint
  .post("resource")
  .in[ResourceObject]
  .out[Unit](Status.NoContent)
  .outErrors[CustomFailure](
    HttpCodec.error[CustomFailure.Duplicate](Status.Conflict),
    HttpCodec.error[CustomFailure.Unexpected](Status.InternalServerError),
  )
val postResourceRoute = postResource.implement { (rs: ResourceObject) =>
  CreateResource(rs).map(_ => ())
}
```
2. Start the application
3. Send a request to "/resource" with an invalid JSON to be decoded into a `ResourceObject` instance (for example "{")
4. The server does not send back any response. When catching the error at `App` level, what is catched is `Cause[Response]` with value `Empty`

**Expected behaviour**
When catching the error at `App` level, an error related to malformed body is catched (maybe the existing `HttpCodecError.MalformedBody` error). Or some other way to handle decoding errors via the `Endpoint` API.

jdegoes:
This should be catchable (after conversion to request handler) via `catchAllCause` as a defect, namely some variation of `HttpCodecError`.

/bounty $75

rhueara-ca:
I'll try to push a fix for this soon.

PR #2147, opened 2023-05-05
Throw request body decoding error

rhueara-ca:
Throw `MalformedBody` exception for any errors occurred during request body decoding.

/claim #2143


zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -236,10 +236,14 @@ private[codec] object EncoderDecoder                   {
       if (jsonDecoders.length == 0) {
         ZIO.unit
       } else if (jsonDecoders.length == 1) {
-        jsonDecoders(0)(body).map { result => inputs(0) = result }
+        jsonDecoders(0)(body).map { result => inputs(0) = result }.mapError { err =>
+          HttpCodecError.MalformedBody(err.getMessage())
```
jdegoes: I think you should modify `MalformedBody` so that it takes another `Throwable` (or `Option[Throwable]`), so we can attach the original error. Otherwise, we preserve the message, but nothing else, and there could be useful details in that `Throwable`.

rhueara-ca: Thank you for your feedback. I've applied your suggestion and also changed the test case to match the error initial Throwable as a use case example.

rhueara-ca:
I forgot to rollback something on the test case. Also fixed some `fmtCheck` errors after pulling `main` changes.

jdegoes (approved):

jdegoes:
Thank you!

Issue #2127, opened 2023-04-29
Streaming multipart hangs for binary files above a certain threshold unless rechunking the body stream


jgranstrom:
**Describe the bug**
Note: using `3.0.0-RC1+9-429c3938-SNAPSHOT` which includes https://github.com/zio/zio-http/pull/2116

Sending a multipart binary file that is above a certain size, close to the StreamingForm buffer of `8192`, I assume minus some header size, will cause the stream to hang indefinitely. 

**To Reproduce**
Test server:
```
Server.serve(
    Http.collectZIO {
      case req @ Method.POST -> !! => (for {
        form <- req.body.asMultipartFormStream
        _ <- form.collectAll
      } yield Response.text("done")
        ).catchAll(e => ZIO.succeed(Response.text(e.toString)))
    }
  )
```

Sending an image with size `8155` is fine:
`http --multipart localhost:8000 "file@./8155.png"`
Sending an image with size `8156`, or any size above that hangs indefinitely:
`http --multipart localhost:8000 "file@./8156.png"`

Creating the form manually with a custom buffer size, such as:
`form <- StreamingForm(req.body.asStream, boundary, 8193)`
Will now happily process one more byte, so `8156` will work, but anything above that will hang

Creating the form manually by first rechunking the body stream, such as:
`form <- StreamingForm(req.body.asStream.rechunk(1024), boundary)`
Seems to handle arbitrary sizes without hanging, although difficult to test exhaustively.

**Expected behaviour**
Any file size to be handled by streaming form without hanging. If rechunking the body stream is required for the form processing to work properly, it should probably be done by `asMultipartFormStream`

andrzejressel:
So I've decided to take a look at it and it seems to be going deeper and just setting rechunk does not fix entirelly

So I've run `MultipartFormDataStreaming`, I've hardcoded `3000` in `asMultipartFormStream` and I've done rechunking `StreamingForm.fields`. While it works with files slightly above 3000 (like  32K), it completelly broke when I've send 80 MB file.

By broke I mean that that after waiting for several seconds I got these logs:
```
Starting to read multipart/form stream
Finished reading multipart/form stream, received 0 bytes of data
```

So I guess my file was silently dropped somewhere in the code.

andrzejressel:
All right, so error with exec is because, exe is not detected as binary (for some reason it's set as `non-binary`).

In any case @jgranstrom I believe the issue is on your code side (although without getting deeper it can be hard co debug). As Scaladoc says
[...]

PR #2148, opened 2023-05-05
Fork offerAll call to not suspend fiber when Queue is full

ccerbusca:
After some time spent debugging, I think what causes issues is the `.offerAll` call on the bounded queue. It suspends until all elements have been offered. If large multipart bodies are streamed, it suspends the fiber and no new bytes are being read, which means the boundary may not be reached for the streaming binary file to complete. (At least that's what I think is happening). 

Forking the `.offerAll` call as a daemon fixed the issues for different buffer sizes, small or big.

Looking forward to some feedback on this one, as I'm not sure if I'm right

/claim #2127

### DIFF
--- a/zio-http/src/main/scala/zio/http/StreamingForm.scala
+++ b/zio-http/src/main/scala/zio/http/StreamingForm.scala
@@ -59,7 +59,7 @@ final case class StreamingForm(source: ZStream[Any, Throwable, Byte], boundary: 
                     case Some(queue) =>
                       val takes = buffer.addByte(crlfBoundary, byte)
                       if (takes.nonEmpty) {
-                        runtime.unsafe.run(queue.offerAll(takes)).getOrThrowFiberFailure()
+                        runtime.unsafe.run(queue.offerAll(takes).forkDaemon).getOrThrowFiberFailure()
                       }
                     case None        =>
                   }


ccerbusca:
Small note: There already is a test that verifies large multipart decoding that strangely passed before, where client like httpie would hang. Could not wrap my head around why though.

And now I see that the tests that passed on my laptop, do not pass in CI. Will investigate that

ccerbusca:
Forking that method call will not preserve order of bytes I assume. Should have seen that one coming. Issue is still with the bounded queue filling up I think. What do you think about increasing the capacity from 3 to some way higher number?

vigoo:
Hm I've set that to 3 because we only offer 1 or 2 elements at the same time (either a chunk, or in the end a chunk + end). The inner stream should be able to take one chunk from this small queue and then the main stream can proceed. 

So I think we still miss some understanding of what's going on (at least me :)) but it is nice that you've found that it blocks on that offer.

Forking the offer as you already discovered is not a good solution.

ccerbusca:
> Hm I've set that to 3 because we only offer 1 or 2 elements at the same time (either a chunk, or in the end a chunk + end). The inner stream should be able to take one chunk from this small queue and then the main stream can proceed.
> 
> So I think we still miss some understanding of what's going on (at least me :)) but it is nice that you've found that it blocks on that offer.
> 
> Forking the offer as you already discovered is not a good solution.

I'll continue investigating the issue in that case

ccerbusca:
closing this one in favor of #2150

PR #2150, opened 2023-05-05
Fix multipart streaming for files above bufferSize

### LINKED ISSUE #2127: Streaming multipart hangs for binary files above a certain threshold unless rechunking the body stream

andrzejressel:
From my understanding `newQueue <- Queue.bounded[Take[Nothing, Byte]](3)` expects 3 following elements:

- Start of boundary
- Full body
- End marker

or 

- Start of boundary
- Partial data from buffer
- ????

In `???` is where the problem is - normally it would be pushed to user and handled - but it doesn't work like that when body is in one chunk. Instead a lot of partial data is pushed and queue lock after 2.

Important thing to understand this issue is how zio mapping work with chunk.

I thought `mapAccum` and `collectZIO` works on single elements. While that's the case it's important to realise that these methods are batched. For every chunk all elements are firstly `mapAccum` and then they're all pushed to `collectZIO`. While it work in streaming mode, because netty chunk is the same as buffer it does not work in non streaming mode, because all the data is pushed at one in one big chunk.

Fixes #2127 
/claim #2127

### DIFF
--- a/zio-http/src/main/scala/zio/http/StreamingForm.scala
+++ b/zio-http/src/main/scala/zio/http/StreamingForm.scala
@@ -29,7 +29,7 @@ final case class StreamingForm(source: ZStream[Any, Throwable, Byte], boundary: 
 
   private val fullQueueErrorMessage =
     """
-      |Internal form queue is full for 1 minute. Please ensure streaming parts are handled, if so report bug to https://github.com/zio/zio-http/issues
+      |Internal form queue is full for 10 seconds. Please ensure streaming parts are handled, if so report bug to https://github.com/zio/zio-http/issues
       |""".stripMargin
 
   /**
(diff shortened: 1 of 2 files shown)


jdegoes:
@vigoo Can you take a look at this one?

andrzejressel:
😞

Issue #2100, opened 2023-04-14
Strawman for websockets


jdegoes:
```scala
trait Channel[-In, +Out] {
  def awaitShutdown: UIO[Unit]

  def receive: Task[Out]

  def send(in: In): Task[Unit]

  def shutdown: UIO[Unit]

  // map, contramap, etc.
}
```

```scala
type WebsocketChannel = Channel[ChannelEvent[WebsocketFrame], ChannelEvent[WebsocketFrame]]
```

```scala
final case class SocketApp[-Env, +Err, +Out](run: WebsocketChannel => ZIO[Env, Err, +Out])
```

NavidJalali:
I'm kinda interested

adamgfraser:
I will do this.

PR #2152, opened 2023-05-06
Refactor Websockets

adamgfraser:
/claim #2100


zio-http/src/main/scala/zio/http/socket/SocketApp.scala
```diff
@@ -20,11 +20,7 @@ import zio._
 
 import zio.http._
 
-final case class SocketApp[-R](
-  decoder: SocketDecoder = SocketDecoder.default,
-  protocol: SocketProtocol = SocketProtocol.default,
-  message: Option[ChannelEvent[WebSocketFrame, WebSocketFrame] => ZIO[R, Throwable, Any]] = None,
-) { self =>
+final case class SocketApp[-R](run: WebSocketChannel => ZIO[R, Throwable, Any]) { self =>
```
jdegoes: I think at this point, `SocketApp` is not pulling its weight, and we could just use:

```scala
type SocketApp[-R] = 
   Handler[-R, Throwable, WebSocketChannel, Any]
```

zio-http/src/main/scala/zio/http/socket/WebSocketChannel.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import zio.http.ChannelEvent.{ChannelRead, ChannelRegistered, ChannelUnregistered, ExceptionCaught, UserEventTriggered}
+import zio.http._
+import zio.http.netty.NettyChannel
+import zio.http.socket.WebSocketChannelEvent
+
+import io.netty.buffer.{ByteBufUtil, Unpooled}
+import io.netty.handler.codec.http.websocketx.{WebSocketFrame => JWebSocketFrame, _}
+
+object WebSocketChannel {
```
jdegoes: It is odd for this companion object to appear in the public API, even though it has no public methods.

zio-http/src/main/scala/zio/http/socket/package.scala
```diff
@@ -21,10 +21,10 @@ package object socket {
   /**
    * A channel that allows websocket frames to be written to it.
    */
-  type WebSocketChannel = Channel[WebSocketFrame]
+  type WebSocketChannel = Channel[WebSocketChannelEvent, WebSocketChannelEvent]
```
jdegoes: Possibly we could move these two type aliases to `zio.http` and delete the whole `.socket` package? It seems to thin to keep.

zio-http/src/test/scala/zio/http/WebSocketSpec.scala
```diff
@@ -24,35 +24,41 @@ import zio.test.{TestClock, assertCompletes, assertTrue, assertZIO, testClock}
...
-        msg <- MessageCollector.make[ChannelEvent.Event[WebSocketFrame]]
+        msg <- MessageCollector.make[WebSocketChannelEvent]
         url <- DynamicServer.wsURL
         id  <- DynamicServer.deploy {
           Handler
-            .fromFunctionZIO[WebSocketChannelEvent] {
-              case ev @ ChannelEvent(ch, ChannelRead(frame)) => ch.writeAndFlush(frame) *> msg.add(ev.event)
-              case ev @ ChannelEvent(_, ChannelUnregistered) => msg.add(ev.event, true)
-              case ev @ ChannelEvent(_, _)                   => msg.add(ev.event)
+            .fromFunctionZIO[WebSocketChannel] { case channel =>
```
jdegoes: These could use a `Handler` helper, perhaps. `Handler.webSocket { ... }`

zio-http/src/test/scala/zio/http/WebSocketSpec.scala
```diff
@@ -24,35 +24,41 @@ import zio.test.{TestClock, assertCompletes, assertTrue, assertZIO, testClock}
...
         res <- ZIO.scoped {
           Http
-            .collectZIO[WebSocketChannelEvent] {
-              case ChannelEvent(ch, UserEventTriggered(HandshakeComplete))   =>
-                ch.writeAndFlush(WebSocketFrame.text("FOO"))
-              case ChannelEvent(ch, ChannelRead(WebSocketFrame.Text("FOO"))) =>
-                ch.writeAndFlush(WebSocketFrame.text("BAR"))
-              case ChannelEvent(ch, ChannelRead(WebSocketFrame.Text("BAR"))) =>
-                ch.close()
+            .collectZIO[WebSocketChannel] { case channel =>
```
jdegoes: This too could use a domain-specific helper. `Http.webSocket`??

zio-http/src/main/scala/zio/http/ChannelEvent.scala
```diff
@@ -70,4 +49,19 @@ object ChannelEvent {
     case object HandshakeTimeout  extends UserEvent
     case object HandshakeComplete extends UserEvent
   }
+
+  def channelRead[A](msg: A): ChannelEvent[A] =
```
jdegoes: Is the `channel*` prefix adding anything to these constructors? Also it's inconsistent since it's not being used for `exceptionCaught` and `userEventTriggered`.

jdegoes:
No place to put this feedback, so will put it here: I see the socket package contains `SocketProtocol` and `WebSocketFrame` and `CloseStatus`. These are all web-socket specific. I think it could make sense to move them to `zio.http`, however, with different names and maybe different organization. CloseStatus feels like it should be inside the case object for `SocketProtocol` (since it looks like it's only used there). `SocketProtocol` feels like it should be called `WebSocketConfig`.

With these changes we would end up with only a couple data types moved to `zio.http` and the deletion of `zio.http.socket`.

jdegoes (changes requested):
Looks really nice. A huge improvement in usability and a big simplification to the API! Just a few minor suggestions to push it a little further.

adamgfraser:
@jdegoes This is ready for another review.

docs/dsl/socket/socket.md
```diff
@@ -7,12 +7,18 @@ Websocket support can be added to your Http application using the same `Http` do
 
 ```scala mdoc:silent
 import zio.http._
-import zio.http.socket._
 import zio._
 
-val socket = Http.collectZIO[WebSocketChannelEvent] {
-  case ChannelEvent(ch, ChannelEvent.ChannelRead(WebSocketFrame.Text("foo"))) =>
-    ch.writeAndFlush(WebSocketFrame.text("bar"))
+val socket = Http.webSocket { channel =>
```
jdegoes: This looks really beautiful.

docs/examples/advanced/websocket-server.md
```diff
@@ -7,60 +7,51 @@ sidebar_label: "WebSocket Server"
...
-      case ChannelEvent(_, ChannelRead(WebSocketFrame.Close(status, reason))) =>
-        Console.printLine("Closing channel with status: " + status + " and reason: " + reason)
-
-      // Print the exception if it's not a normal close
-      case ChannelEvent(_, ExceptionCaught(cause))                            =>
-        Console.printLine(s"Channel error!: ${cause.getMessage}")
+  val httpSocket: Http[Any, Throwable, WebSocketChannel, Unit] =
+    Http.webSocket { channel =>
+      channel
+        .receive
```
jdegoes: This makes me want:

```scala
channel.receiveForever(f)
```

which of course does `receive.flatMap(f).forever` for you.

jdegoes: Or `receiveAll(f)`, maybe

zio-http-testkit/src/main/scala/zio/http/TestChannel.scala
```diff
@@ -1,51 +1,32 @@
...
-  override def close(await: Boolean)(implicit trace: Trace): Task[Unit] =
-    counterpartEvents.offer(ChannelEvent.ChannelUnregistered).unit
-
-  override def contramap[A1](f: A1 => WebSocketFrame): Channel[A1] = ???
-
-  override def flush(implicit trace: Trace): Task[Unit] =
-    // There's not queuing as would happen in a real Netty server, so this will always be a NoOp
-    ZIO.unit
-
-  // TODO Is this ID meaningful in a test?
```
jdegoes, line 22: Nice to see this all gone

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -422,6 +421,7 @@ object ZClient {
     requestDecompression: Decompression,
     localAddress: Option[InetSocketAddress],
     addUserAgentHeader: Boolean,
+    socketProtocol: WebSocketConfig,
```
jdegoes: Worth renaming:

```suggestion
    webSocketConfig: WebSocketConfig,
```

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -446,6 +446,9 @@ object ZClient {
 
     def requestDecompression(isStrict: Boolean): Config =
       self.copy(requestDecompression = if (isStrict) Decompression.Strict else Decompression.NonStrict)
+
+    def withSocketProtocol(socketProtocol: WebSocketConfig): Config =
```
jdegoes:
```suggestion
    def withWebSocketConfig(socketProtocol: WebSocketConfig): Config =
```

zio-http/src/main/scala/zio/http/Handler.scala
```diff
@@ -516,14 +535,23 @@ sealed trait Handler[-R, +Err, -In, +Out] { self =>
...
+    ev2: WebSocketChannel <:< In,
     trace: Trace,
-  ): SocketApp[R] =
-    SocketApp(event => self.runZIO(event).mapError(ev2))
+  ): ZIO[R, Nothing, Response] =
+    ZIO.environment[R].flatMap { env =>
+      Response.fromSocketApp(self.asInstanceOf[SocketApp[R]].provideEnvironment(env))
+    }
+
+  def toRoute(implicit
```
jdegoes:
```suggestion
  def toHttpimplicit
```

zio-http/src/main/scala/zio/http/Handler.scala
```diff
@@ -834,6 +862,15 @@ object Handler {
   def tooLarge: Handler[Any, Nothing, Any, Response] =
     Handler.status(Status.RequestEntityTooLarge)
 
+  val unit: Handler[Any, Nothing, Any, Unit] =
+    fromExit(Exit.unit)
+
+  /**
+   * Constructs a handler from a function that uses a web socket.
+   */
+  final def webSocket[Env](f: WebSocketChannel => ZIO[Env, Throwable, Any]): SocketApp[Env] =
```
jdegoes: You can be polymorphic on the type of the ZIO effect to preserve it during the handler. In fact because we have monadic composition on handlers, you could use that to achieve a kind of composition of a slightly generalized socket app (with a success type parameter).

```suggestion
  final def webSocket[Env, A](f: WebSocketChannel => ZIO[Env, Throwable, Any]): Handler[Env, Throwable, WebChannelEvent, A] =
```

zio-http/src/main/scala/zio/http/Http.scala
```diff
@@ -622,6 +616,12 @@ object Http {
   def getResourceAsFile(path: String)(implicit trace: Trace): Http[Any, Throwable, Any, File] =
     getResource(path).map(url => new File(url.getPath))
 
+  /**
+   * Constructs an Http from a function that uses a web socket.
+   */
+  def webSocket[Env](f: WebSocketChannel => ZIO[Env, Throwable, Any]): Http[Env, Throwable, WebSocketChannel, Unit] =
```
jdegoes: You can make this polymorphic too in the success value of the web socket.

zio-http/src/main/scala/zio/http/Handler.scala
adamgfraser: We already have a `toHttp` operator with a different signature that propagates the `In` type instead of one that takes a `Handler` with an `In` type of `WebSocketChannel` and converts it into an `Http` with an `In` type of `Any` so that it can be further composed. This is where we are feeling a little bit the deletion of `SocketApp` which had these more specialized operators.

zio-http/src/main/scala/zio/http/Response.scala
```diff
@@ -331,9 +330,9 @@ object Response {
    * Creates a new response for the provided socket
    */
   def fromSocket[R](
-    http: Handler[R, Throwable, ChannelEvent[WebSocketFrame, WebSocketFrame], Unit],
+    http: Handler[R, Throwable, WebSocketChannel, Unit],
```
jdegoes: We can allow any success value here:

```suggestion
    http: Handler[R, Throwable, WebSocketChannel, Any],
```

zio-http/src/main/scala/zio/http/Handler.scala
adamgfraser: Maybe `toHttpFromSocketApp`?

zio-http/src/main/scala/zio/http/netty/socket/NettySocketProtocol.scala
```diff
@@ -27,47 +27,47 @@ import io.netty.handler.codec.http.websocketx.{
...
-      case CloseStatus.MandatoryExtension    => WebSocketCloseStatus.MANDATORY_EXTENSION
-      case CloseStatus.InternalServerError   => WebSocketCloseStatus.INTERNAL_SERVER_ERROR
-      case CloseStatus.ServiceRestart        => WebSocketCloseStatus.SERVICE_RESTART
-      case CloseStatus.TryAgainLater         => WebSocketCloseStatus.TRY_AGAIN_LATER
-      case CloseStatus.BadGateway            => WebSocketCloseStatus.BAD_GATEWAY
-      case CloseStatus.Empty                 => WebSocketCloseStatus.EMPTY
-      case CloseStatus.AbnormalClosure       => WebSocketCloseStatus.ABNORMAL_CLOSURE
-      case CloseStatus.TlsHandshakeFailed    => WebSocketCloseStatus.TLS_HANDSHAKE_FAILED
-      case CloseStatus.Custom(code, message) => new WebSocketCloseStatus(code, message)
+      case WebSocketConfig.CloseStatus.NormalClosure         => WebSocketCloseStatus.NORMAL_CLOSURE
```
jdegoes, line 55: Does it make sense for a `Channel` to have a `def close(s: CloseStatus)` instead of this being part of config?

zio-http-testkit/src/test/scala/zio/http/TestClientSpec.scala
```diff
@@ -56,43 +55,30 @@ object TestClientSpec extends ZIOSpecDefault {
...
-            messageSocketServer ++ channelSocketServer
+                case _ => ZIO.unit
+              }.forever
+            }
 
           for {
-            _        <- TestClient.installSocketApp(httpSocketServer)
-            response <- ZIO.serviceWithZIO[Client](_.socket(pathSuffix = "")(httpSocketClient.toSocketApp))
+            _        <- TestClient.installSocketApp(socketServer)
+            response <- ZIO.serviceWithZIO[Client](_.socket(pathSuffix = "")(socketClient.toSocketApp))
```
jdegoes: `toSocketApp` can be deleted now, can't it?

zio-http/src/test/scala/zio/http/WebSocketSpec.scala
```diff
@@ -22,47 +22,47 @@ import zio.test.TestAspect.{diagnose, nonFlaky, timeout, withLiveClock}
...
-            }
-            .toSocketApp
-            .toRoute
+          Handler.webSocket { channel =>
+            channel.receive.flatMap {
+              case event @ Read(frame)  => channel.send(Read(frame)) *> msg.add(event)
+              case event @ Unregistered => msg.add(event, true)
+              case event                => msg.add(event)
+            }.forever
+          }.toRoute
```
jdegoes: `.toHttp`?

zio-http/src/test/scala/zio/http/WebSocketSpec.scala
```diff
@@ -79,21 +79,25 @@ object WebSocketSpec extends HttpRunnableSpec {
...
-          .toRoute
-          .deployWS
+        serverHttp   = Http.webSocket { channel =>
+          channel.receive.flatMap {
+            case Unregistered =>
+              isStarted.succeed(()) <&> isSet.succeed(()).delay(5 seconds).withClock(clock)
+            case _            =>
+              ZIO.unit
+          }.forever
+        }.toSocketApp.toRoute.deployWS
```
jdegoes: `toSocketApp` can be deleted here and elsewhere, I think.

jdegoes (changes requested):
Nearly there, just a few minor tweaks!

zio-http/src/main/scala/zio/http/netty/socket/NettySocketProtocol.scala
adamgfraser: It is part of the configuration in the underlying Netty implementation so I think this is going to be hard to do.

zio-http-testkit/src/test/scala/zio/http/TestClientSpec.scala
adamgfraser: `socketClient` here is defined as an `Http` so we need to go from an `Http` to a `Handler`, which normally requires specifying a default handler.

zio-http/src/test/scala/zio/http/WebSocketSpec.scala
adamgfraser: This is expecting an `HttpApp` that doesn't have `WebSocketChannel` as its input type anymore.

zio-http/src/test/scala/zio/http/WebSocketSpec.scala
adamgfraser: This is the same issue as above. We need to go from an `Http` with input type of `WebSocketChannel` to an `Http` that no longer has this requirement so it can be further composed.

adamgfraser:
Picked up all your comments except a couple of them about deleting some of the converter methods.

Agree it would be nice to clean this up further. I think we need one method that goes from a `Handler` that needs a `WebSocketChannel` to one that doesn't by returning a `Response` that is a `SocketAppResponse` containing the handler.

And then ideally we have a method to turn a web socket application defined as an `Http` into a `Handler` without having to provide a default value. Or possibly they should always be defined as handlers?

zio-http/src/main/scala/zio/http/Handler.scala
adamgfraser: Or `ToHttpApp`.

adamgfraser:
Okay, I think this is cleaned up now. I said that you should always construct your web socket logic as a `Handler` versus an `Http`. I think this makes sense to separate the handling logic from routing. Then there are just a couple of methods on `Handler`, `toResponse` and `toHttpApp` that are special for web sockets and embed the web socket in the response. All usages of `toSocketApp` are deleted.

jdegoes (approved):

jdegoes:
Beautiful work!

zio-http/src/main/scala/zio/http/ChannelEvent.scala
```diff
@@ -18,49 +18,28 @@ package zio.http
...
-  /**
-   * Immutable and type-safe representation of events that are triggered on a
-   * channel.
-   */
-  sealed trait Event[+A] { self =>
-    def map[B](f: A => B): Event[B] = self match {
-      case ChannelRead(msg)              => ChannelRead(f(msg))
-      case ChannelRegistered             => ChannelRegistered
-      case ChannelUnregistered           => ChannelUnregistered
+  def map[B](f: A => B): ChannelEvent[B] =
```
eshu, line 26: In the `match` statement below there is only `Read` shows a different different from identity. Maybe it would be to make the default implementation as identity function and override it for `Read` only?

Or make implementation like
```
    def map[B](f: A => B): Event[B] = self match {
      case ChannelRead(msg)              => ChannelRead(f(msg))
      case other                                     => other
    }
```
?

Or I am completely wrong and Scala compiler optimises this statement and it won't check all 5 conditions in the worst case?

PR #2153, opened 2023-05-06
Enable Client connection pooling by default

### LINKED ISSUE #2149: Enable Client connection pooling by default
Currently the default client in `zio-http` does not use connection pooling by default. We should consider enabling that by default. Most users would benefit from the increased performance, while not having to set up the connection pool by themselves.

ccerbusca:
closes #2149

/claim #2149


vigoo (approved):

jdegoes (approved):

jdegoes:
This may be hanging CI, but I don't understand why.

vigoo:
No, some tests are really hanging - reproducible

ccerbusca:
@jdegoes i disabled the client connection pool for that particular test suite. It seems the tests hang at some point when trying to send a new request. Might be an issue with the connection pool? Would merit it's own ticket for further investigation

ccerbusca:
Weird, just in case I ran the tests locally and they did not hang, but still seem to hang on the pipeline

PR #2154, opened 2023-05-06
Allow appending query parameters in path, and fix Combiner to avoid nesting of types

afsalthaj:
Note: None of the suggestions are breaking compatibility.

### DIFF
--- a/zio-http-example/src/main/scala/example/CliExamples.scala
+++ b/zio-http-example/src/main/scala/example/CliExamples.scala
@@ -46,7 +46,9 @@ trait TestCliEndpoints {
     Endpoint
       .get(
         "users" / int("userId") ?? Doc.p("The unique identifier of the user") /
-          "posts" / int("postId") ?? Doc.p("The unique identifier of the post") :? paramStr("The user's name"),
+          "posts" / int("postId") ?? Doc.p("The unique identifier of the post") ^? paramStr("user-name") ?? Doc.p(
+            "The user's name",
+          ),
       )
       .out[List[Post]] ?? Doc.p("Get a user's posts by userId and postId")
 
--- a/zio-http-example/src/main/scala/example/CliExamples.scala
+++ b/zio-http-example/src/main/scala/example/CliExamples.scala
@@ -46,9 +46,8 @@ trait TestCliEndpoints {
     Endpoint
       .get(
         "users" / int("userId") ?? Doc.p("The unique identifier of the user") /
-          "posts" / int("postId") ?? Doc.p("The unique identifier of the post"),
+          "posts" / int("postId") ?? Doc.p("The unique identifier of the post") :? paramStr("The user's name"),
       )
-      .query(query("name") ?? Doc.p("The user's name"))
       .out[List[Post]] ?? Doc.p("Get a user's posts by userId and postId")
 
   val createUser =
@@ -109,4 +108,5 @@ object TestCliClient extends zio.ZIOAppDefault with TestCliEndpoints {
       _        <- executor(getUserPosts(42, 200, "adam")).debug("result2")
       _        <- executor(createUser(User(2, "john", Some("john@test.com")))).debug("result3")
     } yield ()
+
 }
--- a/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
@@ -84,7 +84,7 @@ sealed trait HttpCodec[-AtomTypes, Value] {
     combiner: Combiner[Value, Value2],
     ev: HttpCodecType.PathQueryType <:< AtomTypes,
   ): PathQueryCodec[combiner.Out] =
-    self.asQuery ++ that
+    self.asPathQuery ++ that
 
   /**
    * Combines two route codecs into another route codec.
@@ -108,7 +108,7 @@ sealed trait HttpCodec[-AtomTypes, Value] {
    * Reinterprets this codec as a query codec assuming evidence that this
    * interpretation is sound.
    */
-  final def asQuery(implicit ev: HttpCodecType.PathQueryType <:< AtomTypes): QueryCodec[Value] =
+  final def asQuery(implicit ev: HttpCodecType.Query <:< AtomTypes): QueryCodec[Value] =
     self.asInstanceOf[QueryCodec[Value]]
 
   /**
--- a/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
@@ -75,20 +75,25 @@ sealed trait HttpCodec[-AtomTypes, Value] {
     HttpCodec.Combine[AtomTypes1, AtomTypes1, Value, Value2, combiner.Out](self, that, combiner)
 
   /**
-   * To end the route codec and begin with query codec
+   * To end the route codec and begin with query codec, and re-interprets as
+   * PathQueryCodec
    */
   final def :?[Value2](
     that: QueryCodec[Value2],
-  )(implicit combiner: Combiner[Value, Value2], ev: HttpCodecType.Path <:< AtomTypes): PathQueryCodec[combiner.Out] =
-    self.asRoute ++ that
+  )(implicit
+    combiner: Combiner[Value, Value2],
+    ev: HttpCodecType.Path <:< AtomTypes,
+  ): HttpCodec[HttpCodecType.PathQuery, combiner.Out] =
+    (self ++ that).asInstanceOf[HttpCodec[HttpCodecType.PathQuery, combiner.Out]]
 
+  // AtomType can be either query or path & query
   final def &[Value2](
     that: QueryCodec[Value2],
   )(implicit
     combiner: Combiner[Value, Value2],
-    ev: HttpCodecType.PathQueryType <:< AtomTypes,
-  ): PathQueryCodec[combiner.Out] =
-    self.asPathQuery ++ that
+    ev: HttpCodecType.PathQuery with HttpCodecType.Query <:< AtomTypes, // self can be either a query or it can be a combination of path and query
+  ): HttpCodec[AtomTypes with HttpCodecType.Query, combiner.Out] =
+    (self ++ that).asInstanceOf[HttpCodec[AtomTypes with HttpCodecType.Query, combiner.Out]]
 
   /**
    * Combines two route codecs into another route codec.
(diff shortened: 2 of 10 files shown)


zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
```diff
@@ -29,4 +29,28 @@ private[codec] trait QueryCodecs {
   def queryAs[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] =
     HttpCodec.Query(name, codec)
 
+  /**
+   * Alias to query
+   */
+  def paramStr(name: String): QueryCodec[String] =
```
afsalthaj, line 32: The motivation behind this naming was, it is confusing to encode endpoint as 

```scala

Endpoint.get(path).query(query("name"))

```

vs

```scala
Endpoint.get(path).query(paramStr("name"))

```

afsalthaj: In fact, if we agree one of the other suggestions in this PR , it will be:

```scala
Endpoint.get(path :? paramStr("name"))

```

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -103,6 +113,13 @@ sealed trait HttpCodec[-AtomTypes, Value] {
    */
   final def alternatives: Chunk[HttpCodec[AtomTypes, Value]] = HttpCodec.flattenFallbacks(self)
 
+  /**
+   * Reinterprets this codec as a query codec assuming evidence that this
+   * interpretation is sound.
+   */
+  final def asPathQuery(implicit ev: HttpCodecType.Path with HttpCodecType.Query <:< AtomTypes): PathQueryCodec[Value] =
+    self.asInstanceOf[PathQueryCodec[Value]]
```
afsalthaj: Unused. will remove

zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -154,6 +154,20 @@ final case class Endpoint[Input, Err, Output, Middleware <: EndpointMiddleware](
...
+   */
+  /**
+   * Returns a new endpoint derived from this one, whose request content must
+   * satisfy the specified schema.
+   */
+  def withRequestBody[Input2](implicit
+    schema: Schema[Input2],
+    combiner: Combiner[Input, Input2],
+  ): Endpoint[combiner.Out, Err, Output, Middleware] =
+    in[Input2]
```
afsalthaj: Will remove for the time being.

jdegoes:
@afsalthaj The CI build is failing, can you take a look?

zio-http-example/src/main/scala/example/CliExamples.scala
```diff
@@ -46,9 +46,8 @@ trait TestCliEndpoints {
     Endpoint
       .get(
         "users" / int("userId") ?? Doc.p("The unique identifier of the user") /
-          "posts" / int("postId") ?? Doc.p("The unique identifier of the post"),
+          "posts" / int("postId") ?? Doc.p("The unique identifier of the post") :? paramStr("The user's name"),
```
jdegoes:
```suggestion
          "posts" / int("postId") ?? Doc.p("The unique identifier of the post") :? paramStr("user-name"),
```

jdegoes: Be sure to run this example since I don't think you're supplying the param on the other side so it would fail.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -75,15 +75,28 @@ sealed trait HttpCodec[-AtomTypes, Value] {
...
   /**
-   * Combines two query codecs into another query codec.
+   * To end the route codec and begin with query codec, and re-interprets as
+   * PathQueryCodec
+   */
+  final def :?[Value2](
+    that: QueryCodec[Value2],
+  )(implicit
+    combiner: Combiner[Value, Value2],
+    ev: HttpCodecType.Path <:< AtomTypes,
```
jdegoes, line 87: Delete this implicit evidence, instead this function should return: `HttpCodec[AtomTypes with HttpCodecType.Query, combiner.Out]`.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -75,15 +75,28 @@ sealed trait HttpCodec[-AtomTypes, Value] {
...
    */
   final def &[Value2](
     that: QueryCodec[Value2],
   )(implicit
     combiner: Combiner[Value, Value2],
-    ev: HttpCodecType.Query <:< AtomTypes,
-  ): QueryCodec[combiner.Out] =
-    self.asQuery ++ that
+    ev: HttpCodecType.PathQuery with HttpCodecType.Query <:< AtomTypes,
+  ): HttpCodec[AtomTypes, combiner.Out] =
```
jdegoes: You need:

1. Implicit evidence that `HttpCodecType.Query with HttpCodecType.Path <:< AtomTypes`
2. Return `HttpCodec[AtomTypes with HttpCodecType.Query, combiner.Out]`

zio-http/src/main/scala/zio/http/codec/HttpCodecType.scala
```diff
@@ -19,13 +19,15 @@ package zio.http.codec
 sealed trait HttpCodecType
 
 object HttpCodecType {
-  type RequestType <: Path with Content with Query with Header with Method
+  type RequestType <: Path with Content with Query with Header with Method with PathQuery
   type ResponseType <: Content with Header with Status
 
+  type PathQuery <: HttpCodecType
```
jdegoes: Delete this, you do not need `PathQuery` (in addition to both `Query` and `Path`) and it makes reasoning about things very complex.

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
jdegoes: If these are going to change, then `queryAs` should change as well, to `paramAs`.

jdegoes: Or maybe just delete `queryAs` since you have `param` below.

zio-http/src/main/scala/zio/http/codec/package.scala
```diff
@@ -17,10 +17,11 @@
...
-  type HeaderCodec[A]  = HttpCodec[HttpCodecType.Header, A]
-  type MethodCodec[A]  = HttpCodec[HttpCodecType.Method, A]
-  type PathCodec[A]    = HttpCodec[HttpCodecType.Path, A]
-  type QueryCodec[A]   = HttpCodec[HttpCodecType.Query, A]
-  type StatusCodec[A]  = HttpCodec[HttpCodecType.Status, A]
+  type ContentCodec[A]   = HttpCodec[HttpCodecType.Content, A]
+  type HeaderCodec[A]    = HttpCodec[HttpCodecType.Header, A]
+  type MethodCodec[A]    = HttpCodec[HttpCodecType.Method, A]
+  type PathCodec[A]      = HttpCodec[HttpCodecType.Path, A]
+  type PathQueryCodec[A] = HttpCodec[HttpCodecType.PathQuery with HttpCodecType.Path, A]
```
jdegoes: Delete

zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -267,7 +267,7 @@ object Endpoint {
    * Constructs an endpoint for an HTTP DELETE method, whose path is described
    * by the specified path codec.
    */
-  def delete[Input](path: PathCodec[Input]): Endpoint[Input, ZNothing, ZNothing, EndpointMiddleware.None] = {
+  def delete[Input](path: PathQueryCodec[Input]): Endpoint[Input, ZNothing, ZNothing, EndpointMiddleware.None] = {
```
jdegoes, line 270: Rather change all of these to: `HttpCodec[HttpCodecType.Query with HttpCodecType.Path, ...]`

Or add a type alias: `type PathQueryCodec[A] = HttpCodec[HttpCodecType.Query with HttpCodecType.Path, ...]`

zio-http-example/src/main/scala/example/CliExamples.scala
afsalthaj: Yes, it was a silly mistake from my side.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
afsalthaj: @jdegoes  The problem with the evidence of `HttpCodecType.Query with HttpCodecType.Path <:< AtomTypes` is that, it will allow ` "accounts"  / "customer" & paramStr("id")` which is wrong. We should only allow, for example

```scala
GET /api/products?category=electronics&sort=price
// and not GET /api/products&category=electronics&sort=price


```

afsalthaj: The reason for that is , the `AtomTypes` of  `"accounts"  / "customer"` is `Path` and `HttpCodecType.Query with HttpCodecType.Path <:< AtomTypes` satisfies it. This is the reason for introducing a fully separate type called `PathQuery` .

`HttpCodecType.PathQuery with HttpCodecType.Query <:< AtomTypes` implies: in order to call `&`, it has to be either a `path-appended-with-atleast-one-query-parameter`, or it can be just a `Query` (example: `paramStr("id") & paramStr("age")`)

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
afsalthaj: I think we shouldn't be able to do `:?` on top of a `QueryCodec`. This is with the following pattern where users need to directly encode a REST query such as:

```scala

GET /api/products?category=electronics&sort=price

```
`/api/products` is `Path`. and we allow `:?` only on top of `Path`

zio-http/src/main/scala/zio/http/codec/HttpCodecType.scala
afsalthaj: Yes, I agree, introducing it can be mind bending to users who gets into the details of types.

However, the reason to introducing it is here:

https://github.com/zio/zio-http/pull/2154/files/d995976495ba0d24dcfc4d6108ccaecfbf556d5e#r1189264604

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
afsalthaj: I didn't want to break compatibility by deleting `queryAs`, unless we are ok with it.

I can change `param` to `paramAs`.

PR #2156, opened 2023-05-06
Fixed #2155 - Properly handle streaming body response when content length set

### LINKED ISSUE #2155: Error during streaming response write
In our project we use tapir-zio module. We built small example, everything works but on every request we get StacklessClosedChannelException error in log.

During investigation I figured out 2 scenarios related to this problem:

Scenario 1:
- a client makes a request
- zio server writing streaming body response without chunked encoding with content length header set
- after last byte received the client closes the connection
- case 1:
  - netty receives "channel close" an event and interrupts a fiber
  - ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT) in NettyBodyWriter is not called, no error appears
- case 2:
  - ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT) is called before fiber interrupt, we try to write closed channel
  - error appears in log:
```
  may 06, 2023 12:55:47 PM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException
  WARNING: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
  io.netty.channel.StacklessClosedChannelException
  	at io.netty.channel.AbstractChannel$AbstractUnsafe.write(Object, ChannelPromise)(Unknown Source)
```
Scenario 2:
- a client makes a request
- zio server writing streaming body response without chunked encoding with content length header set
- after last byte received the client makes one more request in the same connection
- it happen before ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT) is called
so zio-http tries to write second response headers
- netty channel at the moment is not in ST_INIT state so error appears, headers are sent but the handling is continues.
- the server writes request body
- the client closes the connection and shows HTTP protocol error (body is received without headers)

used StreamingResponse example to reproduce Scenario 2
```
curl localhost:8080/stream localhost:8080/stream
Hello world!
curl: (1) Received HTTP/0.9 when not allowed
```

dizinfector:


jdegoes:
@dizinfector Thank you for this fix! Can you add a test case to ensure no regression in the future?

dizinfector:
> @dizinfector Thank you for this fix! Can you add a test case to ensure no regression in the future?

Yep, sure, will do

jdegoes:
@dizinfector Thanks, ping me when you need a review!

dizinfector:
@jdegoes could you please take a look. I've added a spec for this case, checked that it fails with old code and succeds with a new.
The spec came out a bit tricky so I've added some comments

jdegoes:
Looks like scalafmt failed. Can you please format? Thanks for adding the test!

dizinfector:
Ah, sorry, forgot to run formatter. Done

dizinfector:
@jdegoes could you please take a look

vigoo (approved):

dizinfector:
I saw the test was failed on CI run so I improved it a bit

vigoo:
Could you merge the latest `main`? Some `!!` need to be changed to `Root`

dizinfector:
Done

PR #2158, opened 2023-05-07
Support for multipart/form-data in endpoint API

### LINKED ISSUE #2140: Support multi-part encoding / decoding for Endpoint API
This is a two-part feature:

 - Every `.in` and `.out` instance should capture the name of the input / output element. This name will be used in documentation (to describe what the content is), as well as multi-part encoding / decoding.
 - If the user specifies `.in` more than once, then we switch to multi-part decoding from the request, using the names of the input elements as the names of the fields in the multi-part form encoding; if the user specifies `.out` more than once, then we switch to multi-part encoding in the response, using the names of the output elements as the names of the fields in the multi-part form encoding.

This change, like the recent change to name variables in path segments, is a breaking change. However, given that we are still in RC1, and that API changes are expected at this point, and naming the individual input and output elements allows us to improve documentation as well as support multi-part encoding in the Endpoint API, I think it's a worthwhile step.

To make this change, the body codec will be modified to store a name. Then the `EncoderDecoder` object will have to count the input / output elements, and switch to multi-part encoding / decoding as appropriate.

vigoo:
Resolves #2140 
/claim #2140 

- [x] Naming body codecs
- [x] Multipart encoding for output
- [x] Multipart encoding for input

### DIFF
--- a/zio-http/src/main/scala/zio/http/FormField.scala
+++ b/zio-http/src/main/scala/zio/http/FormField.scala
@@ -20,7 +20,7 @@ import java.nio.charset._
 
 import zio._
 
-import zio.stream.{Take, ZStream}
+import zio.stream.{Take, ZPipeline, ZStream}
 
 import zio.http.FormDecodingError._
 import zio.http.Header.ContentTransferEncoding
@@ -35,10 +35,55 @@ sealed trait FormField {
   def contentType: MediaType
   def filename: Option[String]
 
-  final def valueAsString: Option[String] = this match {
+  /**
+   * Gets the value as a String, but only if it is a text or simple field. For
+   * binary fields it returns None.
+   */
+  final def stringValue: Option[String] = this match {
     case FormField.Text(_, value, _, _) => Some(value)
     case FormField.Simple(_, value)     => Some(value)
     case _                              => None
+  }
+
+  /**
+   * Gets the value of this form field as a String. If it is a binary field, the
+   * value is interpreted as an UTF-8 byte stream.
+   */
+  final def asText: ZIO[Any, CharacterCodingException, String] = this match {
+    case FormField.Text(_, value, _, _)                =>
+      ZIO.succeed(value)
+    case FormField.Binary(_, value, _, _, _)           =>
+      ZIO.succeed(new String(value.toArray, Charsets.Utf8))
+    case FormField.StreamingBinary(_, _, _, _, stream) =>
+      stream.via(ZPipeline.utf8Decode).runFold("")(_ ++ _)
+    case FormField.Simple(_, value)                    =>
+      ZIO.succeed(value)
+  }
+
+  /**
+   * Gets the value of this form field as a chunk of bytes. If it is a text
+   * field, the value gets encoded as an UTF-8 byte stream.
+   */
+  final def asChunk: ZIO[Any, Nothing, Chunk[Byte]] = this match {
+    case FormField.Text(_, value, _, _)                =>
+      ZIO.succeed(Chunk.fromArray(value.getBytes(Charsets.Utf8)))
+    case FormField.Binary(_, value, _, _, _)           =>
+      ZIO.succeed(value)
+    case FormField.StreamingBinary(_, _, _, _, stream) =>
+      stream.runCollect
+    case FormField.Simple(_, value)                    =>
+      ZIO.succeed(Chunk.fromArray(value.getBytes(Charsets.Utf8)))
+  }
+
+  def withName(newName: String): FormField = this match {
+    case FormField.Binary(_, data, contentType, transferEncoding, filename)          =>
+      FormField.Binary(newName, data, contentType, transferEncoding, filename)
+    case FormField.StreamingBinary(_, contentType, transferEncoding, filename, data) =>
+      FormField.StreamingBinary(newName, contentType, transferEncoding, filename, data)
+    case FormField.Text(_, value, contentType, filename)                             =>
+      FormField.Text(newName, value, contentType, filename)
+    case FormField.Simple(_, value)                                                  =>
+      FormField.Simple(newName, value)
   }
 }
 
--- a/zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
+++ b/zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
@@ -16,14 +16,18 @@
 
 package zio.http.codec.internal
 
+import java.nio.charset.CharacterCodingException
+import java.time._
+import java.util.UUID
+
 import scala.collection.mutable
 
 import zio._
 
-import zio.stream.ZStream
+import zio.stream.{ZPipeline, ZStream}
 
-import zio.schema.Schema
 import zio.schema.codec._
+import zio.schema.{Schema, StandardType}
 
 import zio.http._
 import zio.http.codec._
@@ -112,9 +116,6 @@ private[codec] object EncoderDecoder                   {
       {
         val erased = bodyCodec.erase
         erased.schema match {
-          case schema if schema == Schema[String] =>
-            FormField.simpleField(name, value.asInstanceOf[String])
-
           case Schema.Primitive(_, _) =>
             FormField.simpleField(name, value.toString)
           case _                      =>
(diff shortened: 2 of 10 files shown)


zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -101,15 +107,101 @@ private[codec] object EncoderDecoder                   {
...
     private val flattened: AtomizedCodecs = AtomizedCodecs.flatten(httpCodec)
 
-    private val jsonEncoders = flattened.content.map { bodyCodec =>
+    private val jsonEncoders      = flattened.content.map { bodyCodec =>
       val erased    = bodyCodec.erase
       val jsonCodec = JsonCodec.schemaBasedBinaryCodec(erased.schema)
       erased.encodeToBody(_, jsonCodec)
     }
-    private val jsonDecoders = flattened.content.map { bodyCodec =>
+    private val formFieldEncoders = flattened.content.map { bodyCodec => (name: String, value: Any) =>
```
jdegoes: Would be nice to see a type signature here (on `formFieldEncoders`) just for ease of understanding what's going on.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -101,15 +107,101 @@ private[codec] object EncoderDecoder                   {
...
+    }
+    private val jsonDecoders      = flattened.content.map { bodyCodec =>
       val jsonCodec = JsonCodec.schemaBasedBinaryCodec(bodyCodec.schema)
       bodyCodec.decodeFromBody(_, jsonCodec)
     }
+    private val formFieldDecoders = flattened.content.map { bodyCodec => (field: FormField) =>
+      {
+        def fieldAsText = field match {
+          case FormField.Text(_, value, _, _)                =>
+            ZIO.succeed(value)
```
jdegoes, line 139: This seems incorrect to decode every text field into a string. What if it's an int or other `StandardType` that was converted into a `String`?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -101,15 +107,101 @@ private[codec] object EncoderDecoder                   {
...
+        }
+      }
+    }
+    private val jsonDecoders      = flattened.content.map { bodyCodec =>
       val jsonCodec = JsonCodec.schemaBasedBinaryCodec(bodyCodec.schema)
       bodyCodec.decodeFromBody(_, jsonCodec)
     }
+    private val formFieldDecoders = flattened.content.map { bodyCodec => (field: FormField) =>
+      {
+        def fieldAsText = field match {
```
jdegoes: Maybe move this to `FormField#asText` or `FormField#asString`?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -101,15 +107,101 @@ private[codec] object EncoderDecoder                   {
...
+            ZIO.succeed(value)
+          case FormField.Binary(_, value, _, _, _)           =>
+            ZIO.succeed(new String(value.toArray, Charsets.Utf8))
+          case FormField.StreamingBinary(_, _, _, _, stream) =>
+            stream.via(ZPipeline.utf8Decode).runFold("")(_ ++ _)
+          case FormField.Simple(_, value)                    =>
+            ZIO.succeed(value)
+        }
+
+        def fieldAsChunk: ZIO[Any, Nothing, Chunk[Byte]] = field match {
```
jdegoes: Maybe move this to `FormField#asBinary` or `FormField#asChunk`?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +335,40 @@ private[codec] object EncoderDecoder                   {
       } else if (jsonDecoders.length == 1) {
         jsonDecoders(0)(body).map { result => inputs(0) = result }
       } else {
-        ZIO.foreachDiscard(jsonDecoders.zipWithIndex) { case (decoder, index) =>
-          decoder(body).map { result => inputs(index) = result }
+        val indexByName = flattened.content.zipWithIndex.map { case (codec, idx) =>
```
jdegoes: This is computation that should happen outside the hot path. Move it as a field of the class so it can be done a single time, rather than once per request.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +335,40 @@ private[codec] object EncoderDecoder                   {
...
+          codec.name.getOrElse("field" + idx.toString) -> idx
+        }.toMap
+        body.asMultipartFormStream.flatMap { form =>
+          form.fields.runForeach { field =>
```
jdegoes, line 323: If there are fewer form fields than codecs, then I think it will fail because you won't initialize `inputs(idx)` properly for every `idx`. Could be worth adding a check.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +335,40 @@ private[codec] object EncoderDecoder                   {
...
+            indexByName.get(field.name) match {
```
jdegoes, line 324: Would be nice to avoid the `Option` boxing here. Maybe `indexByName(0)` and you can catch the exception appropriately and turn it into `HttpCodecError.MalformedBody(s"Unexpected multipart/form-data field: ${field.name}")`.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +335,40 @@ private[codec] object EncoderDecoder                   {
...
+              case Some(idx) =>
+                flattened.content(idx) match {
+                  case BodyCodec.Multiple(schema, _, _) if schema == Schema[Byte] =>
+                    field match {
+                      case FormField.Binary(_, data, _, _, _)          =>
+                        ZIO.succeed {
```
jdegoes: Since you are just doing `ZIO.suceed` for the side effect, I'd suggest, instead, to do the side-effect directly, and return `ZIO.unit`. This eliminates some overhead and allocations.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -324,34 +448,62 @@ private[codec] object EncoderDecoder                   {
...
+      } else {
+        if (inputs.length > 1) {
+          Body.fromMultipartForm(encodeMultipartFormData(inputs), formBoundary)
+        } else {
+          if (jsonEncoders.length == 0) Body.empty
+          else if (jsonEncoders.length == 1) {
+            val encoder = jsonEncoders(0)
+
+            encoder(inputs(0))
+          } else throw new IllegalStateException("A request on a REST endpoint should have at most one body")
```
jdegoes: This branch should be unreachable as far as I can see, so maybe the tests should be against `inputs.length` and not `jsonEncoders.length`, so we don't have this unreachable `else`. Or am I missing something?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -324,34 +448,62 @@ private[codec] object EncoderDecoder                   {
...
 
-        encoder(inputs(0))
-      } else throw new IllegalStateException("A request on a REST endpoint should have at most one body")
+    private def encodeMultipartFormData(inputs: Array[Any]): Form = {
+      Form(
+        flattened.content.zipWithIndex.map { case (bodyCodec, idx) =>
+          val input = inputs(idx)
+          val name  = bodyCodec.name.getOrElse("field" + idx.toString)
+          bodyCodec match {
+            case BodyCodec.Multiple(schema, mediaType, _) if schema == Schema[Byte] =>
```
jdegoes, line 455: This test is repeated, it could make sense to add `isBinaryStream` (and package-private `unsafeAsBinaryStream`) to `BodyCodec`.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -324,34 +448,62 @@ private[codec] object EncoderDecoder                   {
...
-          Headers(Header.ContentType(mediaType))
-        } else throw new IllegalStateException("A request on a REST endpoint should have at most one body")
+        if (inputs.length > 1) {
+          Headers(Header.ContentType(MediaType.multipart.`form-data`))
+        } else {
+          if (jsonEncoders.length == 0) Headers.empty
+          else if (jsonEncoders.length == 1) {
+            val mediaType = flattened.content(0).mediaType.getOrElse(MediaType.application.json)
+            Headers(Header.ContentType(mediaType))
+          } else throw new IllegalStateException("A request on a REST endpoint should have at most one body")
```
jdegoes: This branch should be unreachable as far as I can see, so maybe the tests should be against inputs.length and not jsonEncoders.length, so we don't have this unreachable else. Or am I missing something?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
jdegoes: Ah it already exists I think.

zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -258,30 +303,53 @@ final case class Endpoint[Input, Err, Output, Middleware <: EndpointMiddleware](
...
+      error,
+      doc,
+      mw,
+    )
+
+  /**
+   * Returns a new endpoint derived from this one, whose output type is a stream
+   * of the specified type for the ok status code.
+   */
+  def outStream[Output2: Schema](name: String)(implicit
```
jdegoes, line 318: I think we need `inCodec` and `outCodec` variations that take a name.

zio-http/src/test/scala/zio/http/endpoint/EndpointSpec.scala
```diff
@@ -511,6 +513,143 @@ object EndpointSpec extends ZIOSpecDefault {
           )
         },
       ),
+      suite("multipart/form-data")(
```
jdegoes, line 517: This is one case where it could be useful to do a `Gen` spec:

 - generate N form fields
 - each form field generated from N types
 - form fields randomly have either a name, or no name
 - we generate round trip: decode multi-part, and encode multi-part

Otherwise the number of combinations is just staggeringly high and would be too painful to write by hand.

jdegoes (changes requested):
Excellent work! Can't wait to see this land.

A few minor changes and then a suggestion to add a powerful Gen test to cover all combinations.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
vigoo: There is a similar one that I extracted, here I did not because the pattern match itself is also used to get some of the fields of `BinaryCodec.Multiple`

zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
vigoo: I think we don't because the name is stored in the codec. So you'd write `outCodec(HttpCodec.content[X]("name"))` etc

zio-http/src/main/scala/zio/http/FormField.scala
```diff
@@ -35,11 +35,56 @@ sealed trait FormField {
...
     case FormField.Text(_, value, _, _) => Some(value)
     case FormField.Simple(_, value)     => Some(value)
     case _                              => None
   }
+
+  /**
+   * Gets the value of this form field as a String. If it is a binary field, the
+   * value is interpreted as an UTF-8 byte stream.
+   */
+  final def asText: ZIO[Any, CharacterCodingException, String] = this match {
```
jdegoes, line 52: I see now we have `stringValue`, which has some overlap with this one. Not sure if and how to unify the names.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +318,39 @@ private[codec] object EncoderDecoder                   {
...
+                    ZIO.unit
+                  case _                                                          =>
+                    formFieldDecoders(idx)(field).map { result => inputs(idx) = result }
+                }
+              case None      =>
+                ZIO.fail(HttpCodecError.MalformedBody(s"Unexpected multipart/form-data field: ${field.name}"))
+            }
+          }.zipRight {
+            ZIO.attempt {
+              for (idx <- inputs.indices) {
```
jdegoes: More performant to use a while loop with a `var idx = 0`,

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
jdegoes: No way to fix this one?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +318,39 @@ private[codec] object EncoderDecoder                   {
...
+                        inputs(idx) = ZStream.fromChunk(Chunk.fromArray(value.getBytes(Charsets.Utf8)))
+                      case FormField.Simple(_, value)                  =>
+                        inputs(idx) = ZStream.fromChunk(Chunk.fromArray(value.getBytes(Charsets.Utf8)))
+                    }
+                    ZIO.unit
+                  case _                                                          =>
+                    formFieldDecoders(idx)(field).map { result => inputs(idx) = result }
+                }
+              case None      =>
+                ZIO.fail(HttpCodecError.MalformedBody(s"Unexpected multipart/form-data field: ${field.name}"))
```
jdegoes, line 343: This will log the `Option` which contains the field name, right?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -243,8 +318,39 @@ private[codec] object EncoderDecoder                   {
...
+            }
+          }.zipRight {
+            ZIO.attempt {
+              for (idx <- inputs.indices) {
+                if (inputs(idx) == null)
+                  throw HttpCodecError.MalformedBody(
+                    s"Missing multipart/form-data field (${flattened.content(idx).name.getOrElse("field" + idx.toString)}",
```
jdegoes: This is a translation from `idx` to field name. Could be useful to precompute that, unless it's only used in the error case here.

zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
jdegoes: Ah right

zio-http/src/main/scala/zio/http/FormField.scala
vigoo: it was called `valueAsString` and I renamed it to `stringValue` :) Maybe we don't need it at all?

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
vigoo: No this is the `Form` level name (not the codec) which is `String`

jdegoes (approved):

jdegoes:
Great to see this land!

Issue #2159, opened 2023-05-07
Different behaviour between CustomAuthProviding and CustomAuth


ex0ns:
**Describe the bug**

The behavior between CustomAuth and CustomAuthProviding  is very different (and is probably related to the usage of the Context vs Simple middleware), in the first case, when not route are matching the middleware is not executed and the returned error is 404 (Http.empty). In the second case, as we first evaluate the middleware before being able to execute the route, the middleware is executed and returns a Unauthorized response.

**To Reproduce**
```scala
//> using lib "dev.zio::zio-http:3.0.0-RC1"
//> using scala "2.13"
//> using mainClass "Main"

import zio._
import zio.http._

object Main extends ZIOAppDefault {
  val priv = Http.collect[Request] { case Method.GET -> !! / "api" / "private" =>
    Response.text("private")
  }

  val auth = RequestHandlerMiddlewares.customAuth { headers =>
    headers.header(Header.Authorization).isDefined
  }

  // this auth middleware intercept ALL path, same for customAuthProvidingZIO
  val auth2 =
    RequestHandlerMiddlewares.customAuthProviding[Any, Header.Authorization] { headers =>
      headers.header(Header.Authorization)
    }

  val config = ZLayer.succeed(Server.Config.default.port(7070))
  // replace auth with auth2 in the next line
  val runner = Server.install(
    (priv.tapZIO(resp => Console.printLine(s"Handler: ${resp.status}")) @@ auth)
      .tapZIO(resp => Console.printLine(s"Middleware ${resp.status}"))
      .withDefaultErrorResponse
  )

  val serverFork = (runner *> ZIO.never).fork
  val client =
    Client.request("http://localhost:7070/test").map(_.status).debug("Should be not found")

  def run =
    serverFork.zipPar(client).provide(Client.default, Server.live, config)
}
```

**Expected behaviour**
In both case, if a route does not exist, the middleware should not be executed and the status not changed.
The error raised here is a bit more bigger than it seems because if we were to concatenate another Http application AFTER our middleware, it will never be executed

**Additional context**

I tried many other implementation of the `CustomAuthProviding` but I currently don't see a way out (mostly because I don't really understand how CustomAuth is able to determine if a route exist or not).

I don't think that we would be able to implement the following line when using a context (taken from the CustomAuth middleware):

```scala
Http.fromHttpZIO[Request](route.run(_).map(self(_))).withErrorHandler(route.errorHandler)
```
[...]

PR #2164, opened 2023-05-08
fix: delay execution of partial function in Handler

ex0ns:
This is a first try to open the discussion regarding issue [#2159](https://github.com/zio/zio-http/issues/2159)
Calling `isDefinedAt` and then `apply` is less efficient than calling `applyOrElse` but in this specific case I don't see a way of delaying the execution of the partial function while still returning a meaningfull `Http` in the `run` function.

I have included specs for both this issue and the one regarding customAuthProviding (which are closely related IMO).

I am definitely not convinced by the current implement and I feel that I must be missing something, any feedback is welcome to improve this code


vigoo:
I think we can close this in favor of the other proposed fix and ongoing discussion there

PR #2165, opened 2023-05-09
Fix authentication via middleware (#2159)

### LINKED ISSUE #2159: Different behaviour between CustomAuthProviding and CustomAuth

987Nabil:
- Only exec. code for authenticated routes
- Only authenticate matching routes

fixes #2159 

I introduced a new method `applicable` on `Http.Routes`. I think this is necessary, to be able to know outside of the constructors, if that routes object is able to handle a given request. 
Why do we care? Because if it does not, we are not allowed to check the authentication attached to this `Http`.
Else `Http.collect[Request](PartialFunction.empty) @@ myAuth` would always return unauthenticated even though it has nothing to authenticate and other apps that a user would combine via `++` would be ignored.
It also enables us, to avoid executing any code on the rhs of path/route matches in the collect methods.

One test is still red. I am not sure why. Sadly the error message is just `Assertion failed:  Exception in thread "zio-fiber-" scala.None$: None` maybe you can give me a hint?

/claim #2159

### DIFF
--- a/zio-http/src/main/scala/zio/http/Http.scala
+++ b/zio-http/src/main/scala/zio/http/Http.scala
@@ -47,6 +47,8 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
           override def run(in: In): ZIO[R1, Err1, Http[R1, Err1, In, Out1]] =
             route.run(in).map(_ >>> (handler, errorOutputMapper))
 
+          override def applicable(in: In): Boolean = route.applicable(in)
+
           override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
             route.errorHandler.map(_.andThen(_.map(errorOutputMapper)))
         }
@@ -65,6 +67,8 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
         new Route[R1, Err1, In, Out1] {
           override def run(in: In): ZIO[R1, Err1, Http[R1, Err1, In, Out1]] =
             route.run(in).map(_ >>> (handler, errorOutputMapper))
+
+          override def applicable(in: In): Boolean = route.applicable(in)
 
           override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
             route.errorHandler.map(_.andThen(_.map(errorOutputMapper)))
@@ -110,6 +114,8 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
               .map(_.catchAllZIO(f))
               .catchAll(err => f(err).map(out => Handler.succeed(out).toHttp))
 
+          override def applicable(in: In): Boolean = route.applicable(in)
+
           override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
             route.errorHandler
         }
@@ -136,6 +142,12 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
           override def run(in: In1): ZIO[R1, Err1, Http[R1, Err1, In1, Out1]] =
             route.run(in).map(_.defaultWith(that))
 
+          override def applicable(in: In1): Boolean = route.applicable(in) || (that match {
+            case Empty(_)                 => false
+            case Http.Static(_, _)        => true
+            case route: Route[_, _, _, _] => route.asInstanceOf[Route[R1, Err1, In1, Out1]].applicable(in)
+          })
+
           override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
             route.errorHandler
         }
@@ -155,6 +167,8 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
           override def run(in: In): ZIO[R, Err, Http[R, Err, In, Out1]] =
             route.run(in).map(_.map(f))
 
+          override def applicable(in: In): Boolean = route.applicable(in)
+
           override val errorHandler: Option[Cause[Nothing] => ZIO[R, Nothing, Out1]] =
             route.errorHandler.map(_.andThen(_.map(f)))
         }
@@ -171,6 +185,8 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
         new Route[R, Err1, In, Out] {
           override def run(in: In): ZIO[R, Err1, Http[R, Err1, In, Out]] =
             route.run(in).mapBoth(f, _.mapError(f))
+
+          override def applicable(in: In): Boolean = route.applicable(in)
 
           override val errorHandler: Option[Cause[Nothing] => ZIO[R, Nothing, Out]] =
             route.errorHandler
@@ -196,6 +212,8 @@ sealed trait Http[-R, +Err, -In, +Out] { self =>
           override def run(in: In): ZIO[R, Err1, Http[R1, Err1, In, Out1]] =
             route.run(in).map(_.mapZIO(f))
 
+          override def applicable(in: In): Boolean = route.applicable(in)
+
           override val errorHandler: Option[Cause[Nothing] => ZIO[R1, Nothing, Out1]] =
             route.errorHandler.map(_.andThen(_.flatMap(f).orDieWith(FailedErrorHandler.apply)))
         }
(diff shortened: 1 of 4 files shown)


ex0ns:
`Assertion failed:  Exception in thread "zio-fiber-" scala.None$: None` happened to me when trying to call `runZIO` on a path that does not have any handler, I think that in your case, it's when you are trying to call `/a` on a Http that only has `/b`, even if you have `withDefaultErrorResponse` set. 

Try to change `withDefaultErrorResponse` with `++ Handler.ok.toHttp` (or something similar), and this should work (I don't know why this is happening or why the fiber seem to die pretty hard in this case)

To create the URL, you could use `URL.empty.withPath("/a")` which does not require converting it into an option explicitly

jdegoes:
@vigoo We worked hard at fixing this problem before and now it's returned, with more besides! Appreciate a review and your thoughts when you get a chance. 

Will review too in the next day or two.

987Nabil:
Thanks @ex0ns I also fixed the last test. Though I think we need some better error message/handling than ` Exception in thread "zio-fiber-" scala.None$: None` for not available routes.

vigoo:
> @vigoo We worked hard at fixing this problem before and now it's returned, with more besides! Appreciate a review and your thoughts when you get a chance.
> 
> Will review too in the next day or two.

I did not review fully the issue and the solution yet, but what I see now is:

The guarantee we came up with is that `RequestHandlerMiddleware`s, which can only transform a `Handler` of `Request => Response` are always applied to `Http` following the rule that the route is not modified, and they get applied on the resulting handler. This is a good default behavior for most of the use cases and also makes it easy to understand what's going on.

There are also `HttpAppMiddleware`s which can transform `Http`s, affecting the route.

That said, the context providing auth middlewares are supposed to be `RequestHandlerMiddleware`s. But the current interfaces does not prevent overriding the `apply(Http)` method which supposed to be "the well defined logic of always applying request handlers middlewares in one particular way", and `customAuthProviding` and `customAuthProvidingZIO` are doing exactly that. 

The reason for them to do so is that they provide the auth context not only to the `Handler` but to the `Http` too (which is a `ZIO` producing a `Handler`. They have to do this because `Http` forces the same `R` to the routing function and the resulting handler.

So in my opinion we have a couple of problems:

- Maybe not the biggest issue, but with the current type hierarchy, a `RequestHandlerMiddleware`s can override the default behavior of how it is applied to a `Http`, causing "surprises"
- Because of the route and the handler shares the same environment, the context providing middlewares cannot really be `RequestHandlerMiddleware`s 
- And because of this, to be able to provide their context to the routing effect, they have to be evaluated early - before the routing
- Also, I think with all this, the `customAuthProviding` and `customAuthProvidingZIO` should work the same (same bad) so if they are not, that may be an additional bug :) 

I think I saw that the proposed fix basically readds the partial function introspection (but did not check in more details). I think we should think of a better solution, possibly to separate the environment of the routing and handler.

vigoo:
Checked the PR a bit more.

You cannot implement a

```scala
override def applicable(in: In): Boolean
```

on `Http`, because in case of `Http.Route` it is an effectful function:

```scala
def run(in: In): ZIO[R, Err, Http[R, Err, In, Out]]
```

Only by evaluating this effect can you know whether the given route handles the input or not (it can result in `Http.Empty`!).

Also if you would try to make `applicable` an effect too, that would not help, because then it would require `R` and you would have to evaluate the middleware to eliminate the context part of `R` again.

987Nabil:
@vigoo Yes, I actually had some kind of thought in that direction.
But, since `Http` is sealed, we know all the constructors. And as far as I can see, there is only one constructor for which we can not know if the route is applicable
```scala
  final class FromHttpZIO[In](val self: Unit) extends AnyVal {
    def apply[R, Err, Out](f: In => ZIO[R, Err, Http[R, Err, In, Out]]): Http[R, Err, In, Out] =
```
Which makes sense, because it could create `Http`s dynamically.

I am not sure I 100% understand (yet) the environment problem. Maybe we could have a little discussion about that.

PR #2173, opened 2023-05-10
Add support for error in EndpointClient

### LINKED ISSUE #2166: Endpoint error responses not handled properly by the endpoint client
Reproducing:

Have an `.outError[T]`, make the request handler fail with `T` and call it via `EndpointExecutor`. It throws an index out of bounds error.

andrzejressel:
Since stream are currently in pretty bad state, I'll update tests when they stabilise

Fixes #2166
/claim #2166

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
+++ b/zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
@@ -38,13 +38,13 @@ private[codec] trait EncoderDecoder[-AtomTypes, Value] {
   def encodeWith[Z](value: Value)(f: (URL, Option[Status], Option[Method], Headers, Body) => Z): Z
 }
 private[codec] object EncoderDecoder                   {
-  def apply[AtomTypes, Value](httpCodec: HttpCodec[AtomTypes, Value]): Option[EncoderDecoder[AtomTypes, Value]] = {
+  def apply[AtomTypes, Value](httpCodec: HttpCodec[AtomTypes, Value]): EncoderDecoder[AtomTypes, Value] = {
     val flattened = httpCodec.alternatives
 
     flattened.length match {
-      case 0 => None
-      case 1 => Some(Single(flattened.head))
-      case _ => Some(Multiple(flattened))
+      case 0 => Undefined()
+      case 1 => Single(flattened.head)
+      case _ => Multiple(flattened)
     }
   }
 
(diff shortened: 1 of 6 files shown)


zio-http/src/main/scala/zio/http/endpoint/internal/EndpointClient.scala
```diff
@@ -41,15 +41,17 @@ private[endpoint] final case class EndpointClient[I, E, O, M <: EndpointMiddlewa
...
+        } else {
           invocation.middleware.error
             .decodeResponse(response)
-            .orDie
-            .flatMap((e: invocation.middleware.Err) => ZIO.fail(alt.right(e)))
+            .map(e => alt.right(e))
+        }
 
-        leftError.orElse(rightError)
+        leftError.orElse(rightError).orDie.flip
```
vigoo: Looks nice in general! 

I think handling unexpected error bodies could be better.

It's ok to keep the original design and we fail in the error channel with the known error messages (decoded either by the endpoint's `outErr` codecs or the middleware's error codec) and _die_ on unexpected ones.

But right now:
- if either `leftError` or `rightError` manages to decode the response, we fail with that as expected
- if both `leftError` and `rightError` fails, we die with the error from the second one
- if `leftError` fails and there is no middleware, we fail with a `Middleware is none` message

as these errors can happen in production (let's say a proxy between your service and the called one starts to respond with some unexpected rate limiting response, or just somebody deployed a new version of the target service that breaks the expected error format) it's important to have some information in the exception about what happened.

So I propose to fail with an exception which 
- contains as an inner exception the decoder failure
- if both decoder fails, contains both decoder failures
- the status code from the response
- maybe even a first N characters from the reponse body (but not sure about this one)
-

jdegoes: Also we cannot simply ignore the middleware error. In fact, we need a test to make sure we properly handle middleware error.

andrzejressel: So I've added tests for middleware and rewrite the code a bit - I don't know how to create some kind of tree structure with Causes - that's the bast I could come up with

jdegoes:
@vigoo Please merge when happy! 

@andrzejressel Thank you for your work on this!

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -38,11 +38,14 @@ private[codec] trait EncoderDecoder[-AtomTypes, Value] {
...
 }
 private[codec] object EncoderDecoder                   {
-  def apply[AtomTypes, Value](httpCodec: HttpCodec[AtomTypes, Value]): EncoderDecoder[AtomTypes, Value] = {
+  def apply[AtomTypes, Value](httpCodec: HttpCodec[AtomTypes, Value]): Option[EncoderDecoder[AtomTypes, Value]] = {
     val flattened = httpCodec.alternatives
 
-    if (flattened.length == 1) Single(flattened.head)
-    else Multiple(flattened)
+    flattened.length match {
+      case 0 => None
```
vigoo: Instead of this we could introduce a third `EncoderDecoder`, let's say `Undefined`, with its encode and decode methods just failing. Then you don't have to modify `HttpCodec`s `decode` and `encodeWith` either.

zio-http/src/main/scala/zio/http/endpoint/internal/EndpointClient.scala
```diff
@@ -41,15 +41,47 @@ private[endpoint] final case class EndpointClient[I, E, O, M <: EndpointMiddlewa
...
+                success = ZIO.succeed(_),
+                failure = cause2 => {
+                  ZIO.failCause(f(cause1, cause2))
+                },
+              )
+            },
+          )
+        }
+
+        orElseCombineErrors(leftError, rightError, generateErrorMessage).orDie.flip
```
vigoo: You could just combine the two failure like this:

```scala
first.catchAllCause { firstCause =>
  second.catchAllCause { secondCause =>
    ZIO.failCause(secondCause ++ firstCause)
  }
}
```

and attach the additional information (those messages above) via `mapError` to each level (`first.mapError`, `second.mapError`)

vigoo (changes requested):
I think the behavior is now correct, I suggested two changes to make it a bit cleaner.

jdegoes:
@andrzejressel Please ping us again when you have done the requested revisions. Looking really nice!

andrzejressel:
I guess benchmark is a bit flaky ...

andrzejressel:
@vigoo I've fixed both issues

jdegoes (approved):

andrzejressel:
I don't know what are the rules regarding closing comments...

I guess the person who created them should close then.

vigoo (approved):

PR #2185, opened 2023-05-13
rework styles and classes

TomTriple:
I changed a few things I found a bit rough when using `zio.http.html`. With these changes I hope it is more friendly to add classes (`classAttr` / `classAttrs`) and styles (`styleAttr` / `styleAttrs`). 

Besides that I deprecated `css` as I think it is not entirely clear (at least for me) what it does (does it add a class attribute or a style attribute?). 

But maybe people are already satisfied with how it works, I want to see if these changes provide value - I think they do.

### DIFF
--- a/zio-http-example/src/main/scala/example/HtmlTemplating.scala
+++ b/zio-http-example/src/main/scala/example/HtmlTemplating.scala
@@ -21,11 +21,11 @@ object HtmlTemplating extends ZIOAppDefault {
         body(
           div(
             // Support for css class names
-            css := "container" :: "text-align-left" :: Nil,
+            classAttrs("container", "text-align-left"),
             h1("Hello World"),
             ul(
               // Support for inline css
-              styles := Seq("list-style" -> "none"),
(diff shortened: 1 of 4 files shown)


jdegoes:
1. The convention of using `Attrs` to suffix all the attributes seems verbose and unnecessary, useful only for `class` which is otherwise a reserved Scala keyword. But other places in ZIO HTTP use back tics to escape illegal identifiers (e.g. media types), so that is not an argument in my mind to suffix `Attrs` onto everything.
2. Probably best to keep names aligned with HTML since that's the standard, so yes, deprecating `css` seems wise.
3. It would be nice to follow great organizational patterns here: e.g. all attributes extend `Attribute`, all attribute constructors are inside companion object of `Attribute`, etc. This makes the API more discoverable and more ZIO idiomatic.

@TomTriple Any interest in pushing this forward? Sorry for the delay in review.

jdegoes:
@TomTriple Would love to get this in, please re-open if you get a chance to work on this!

PR #2187, opened 2023-05-14
`Endpoint` api en-/decodes based on accept header

### LINKED ISSUE #1506: Properly support content type headers in zio.http.api

Currently, `EndpointClient` and `EndpointServer` in `zio.http.api.internal` do not perform an adequate job supporting and respecting content type headers.

Modify the client and server to respect and generate the appropriate content type headers, and enhance the architecture to support any one of the protocols supported by ZIO Schema, which powers the current JSON support.

None.

**Additional context**

See `/src/main/scala/zio/http/api/internal/` `EndpointClient.scala` and `EndpointServer.scala`.

987Nabil:
The default codec is still json. Only added support for protobuf for now.
RFC @jdegoes @vigoo
If it fits, I can add more media types that are already supported by zio-schema (thrift and avro should be doable) before we merge.

fixes #1506 
/claim #1506

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/internal/BodyCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/internal/BodyCodec.scala
@@ -16,14 +16,13 @@
 
 package zio.http.codec.internal
 
-import zio._
-
-import zio.stream.ZStream
-
+import zio.http.{Body, MediaType}
 import zio.schema._
-import zio.schema.codec.BinaryCodec
+import zio.schema.codec.{BinaryCodec, Codec}
+import zio.stream.{ZPipeline, ZStream}
+import zio.{ZIO, _}
 
-import zio.http.{Body, FormField, MediaType}
+import java.nio.charset.Charset
 
 /**
  * A BodyCodec encapsulates the logic necessary to both encode and decode bodies
@@ -44,9 +43,19 @@ private[internal] sealed trait BodyCodec[A] { self =>
   def decodeFromBody(body: Body, codec: BinaryCodec[Element]): IO[Throwable, A]
 
   /**
+   * Attempts to decode the `A` from a body using the given codec.
+   */
+  def decodeFromBody(body: Body, codec: Codec[String, Char, Element]): IO[Throwable, A]
+
+  /**
    * Encodes the `A` to a body in the given codec.
    */
   def encodeToBody(value: A, codec: BinaryCodec[Element]): Body
+
+  /**
+   * Encodes the `A` to a body in the given codec.
+   */
+  def encodeToBody(value: A, codec: Codec[String, Char, Element]): Body
 
   /**
    * Erases the type for easier use in the internal implementation.
@@ -67,6 +76,11 @@ private[internal] sealed trait BodyCodec[A] { self =>
   def mediaType: Option[MediaType]
 
   /**
+   * Returns the media type or application/json if not specified
+   */
+  def mediaTypeOrJson: MediaType = mediaType.getOrElse(MediaType.application.`json`)
+
+  /**
    * Name of the body part
    *
    * In case of multipart/form-data encoding one request or response can consist
@@ -80,7 +94,11 @@ private[internal] object BodyCodec {
 
     def decodeFromBody(body: Body, codec: BinaryCodec[Unit]): IO[Nothing, Unit] = ZIO.unit
 
+    def decodeFromBody(body: Body, codec: Codec[String, Char, Unit]): IO[Nothing, Unit] = ZIO.unit
+
     def encodeToBody(value: Unit, codec: BinaryCodec[Unit]): Body = Body.empty
+
+    def encodeToBody(value: Unit, codec: Codec[String, Char, Unit]): Body = Body.empty
 
     def schema: Schema[Unit] = Schema[Unit]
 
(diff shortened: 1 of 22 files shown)


987Nabil:
I was not sure about it, but I guess we should also support to to encode fields of multipart form data with the codecs, or?
With what is already there, that should be easy in my opinion. If we need it, please let me know.

jdegoes:
> I was not sure about it, but I guess we should also support to to encode fields of multipart form data with the codecs, or?
With what is already there, that should be easy in my opinion. If we need it, please let me know.

Yes, for sure!

I think this is the correct direction overall and I will try to provide more feedback soon.

987Nabil:
@jdegoes I will resolve the conflicts in `EncoderDecoder` once you reviewed. I took a look at them and it is not too hard to solve them, but I'd prefer to be able to easily integrate change requests from your side into the git history. Merging the master should have no significant change to my code.

zio-http/src/main/scala/zio/http/codec/internal/BodyCodec.scala
```diff
@@ -66,6 +78,11 @@ private[internal] sealed trait BodyCodec[A] { self =>
    */
   def mediaType: Option[MediaType]
 
+  /**
+   * Returns the media type or application/json if not specified
+   */
+  def mediaTypeOrJson: MediaType = mediaType.getOrElse(MediaType.application.`json`)
```
vigoo: Having this on `BodyCodec` is a bit weird to me. Maybe it would be nicer to just have the optional `mediaType` here, and have the fallback value specified in `EncoderDecoder` (seems like that's the only place this is used anyway)

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -98,87 +102,83 @@ private[codec] object EncoderDecoder                   {
...
+        MediaTypeCodec.protobuf(flattened.content),
+        MediaTypeCodec.text(flattened.content),
+      ).flatMap(c => c.acceptedTypes.map(at => at.fullType -> c)).toMap
+      if (outputTypes.isEmpty) defaultCodecs
+      else {
+        outputTypes.collectFirst {
+          case mt if defaultCodecs.isDefinedAt(mt.fullType) => Map(mt.fullType -> defaultCodecs(mt.fullType))
+        }.getOrElse(
+          throw HttpCodecError.UnsupportedContentType(
+            s"""Non of the Accept header mime types is currently supported.
```
vigoo:
```suggestion
            s"""None of the Accept header mime types is currently supported.
```

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -98,87 +102,83 @@ private[codec] object EncoderDecoder                   {
...
     private val deconstructor = Mechanic.makeDeconstructor(httpCodec)
 
     private val flattened: AtomizedCodecs = AtomizedCodecs.flatten(httpCodec)
 
-    private val jsonEncoders: Chunk[Any => Body]                          =
-      flattened.content.map { bodyCodec =>
-        val erased    = bodyCodec.erase
-        val jsonCodec = JsonCodec.schemaBasedBinaryCodec(erased.schema)
-        erased.encodeToBody(_, jsonCodec)
+    private val codecs: Map[String, MediaTypeCodec[_]] = {
```
vigoo: If I understand correctly this means the default would be that all endpoints support json, protobuf and a plain text encoding.
I'm not sure that's the best default - in many cases you only need to support one specific (let's say json) content type in which case it seems wasteful to generate the codecs for all the other supported content types. It can also be misleading to have text codec on by default while it only works for primitive types.

So:
- maybe the default should be json only?
- only those codecs should be derived which are actually used

vigoo (commented):
Added a few comments. I think this is a nice addition and definitely worth going forward with it so please solve the merge conflicts when you have time!

One more thing I was thinking about: setting the allowed content types is something that you probably would do for a set of endpoints (often for all), but the current API only allows setting them one by one for each. We only concatenate the endpoints once they became routes so there is no place to apply "grouped settings" like this. Maybe this is out of scope for this PR - just a thought.

zio-http/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
987Nabil: This was how I understood @jdegoes in #1506. That if you would have an unconfigured instance of `Endpoint`, we accept everything that we possibly can. And that we would make it configurable later on. But in the current version, it will already narrow down based on `withOutputTypes` and the configured `ContentType`.
For this case I guess, I could make the building of the codecs lazy and reverse the filter. So only build codecs for types that are relevant.

987Nabil:
@vigoo After your comments I thought about performance and codec creation.
Therefore I changed the initialisation of the codecs. Also I realised, that the code was rebuild every time a request with an accept header was send.
Now the codecs are created lazily and are cached. The caching is currently quite naive, since it takes all body codecs as a key. One could make the key for the cache based on single codecs, which would allow the media type codecs to be shared by different endpoints. But I am not sure this is worth it.

Regarding the topic of defining content types for multiple endpoints, I think this is not only a factor for content type settings. I guess it is quite realistic, that you might have a bunch of endpoints that also require all the same header.
So I think we should fix this in a new issue and maybe in a more general way. I could imagine something like an implicit config?

@jdegoes maybe you can give some feedback too? 🙂

zio-http/src/main/scala/zio/http/endpoint/internal/EndpointClient.scala
```diff
@@ -32,9 +32,17 @@ private[endpoint] final case class EndpointClient[I, E, O, M <: EndpointMiddlewa
...
 
-    val requestPatch = invocation.middleware.input.encodeRequestPatch(mi)
+    val requestPatch            = invocation.middleware.input.encodeRequestPatch(mi)
+    val patchedRequest          = request.patch(requestPatch)
+    val withDefaultAcceptHeader =
+      if (patchedRequest.headers.exists(_.headerName == Header.Accept.name))
+        patchedRequest
+      else
+        patchedRequest.addHeader(
+          Header.Accept(MediaType.parseCustomMediaType("application/protobuf").get, MediaType.application.json),
```
987Nabil, line 42: @jdegoes If I understood your right in one of our 1on1's, you want that protopuf is the default accept header for the endpoint client, right? I am asking, because this fails the test currently. Since the error message that is tested for is different. I am not sure if this is a bug. But I wanted to make sure, that I understood you right.

987Nabil:
I am using `import scala.collection.JavaConverters.mapAsScalaConcurrentMapConverter` since it is the only thing working with Scala 2.12. But it fails the build for 2.13/3 since it is deprecated. But idk how to ignore the deprecation warning.

jdegoes:
@987Nabil Will close for now since it does not compile and needs the specified changes. Please re-open when ready for another review!

Issue #2160, opened 2023-05-07
Support multi-part forms in ZIO HTTP CLI


jdegoes:
ZIO HTTP CLI should gracefully handle multi-part requests, including file uploads, and multi-part responses, using the new functionality built into endpoint (#2140).

PR #2190, opened 2023-05-15
Support multi-part forms in ZIO HTTP CLI 

SHSongs:
/claim #2160

- [ ] get image
- [ ] upload image


jdegoes:
This does not yet compile on all platforms (see failing CI build) and is still a draft, so I will close for now, but please re-open if you want another look. Thank you!

PR #2194, opened 2023-05-16
Change naming of transform methods

987Nabil:
this aligns with methods in other codec apis

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/HeaderCodecs.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HeaderCodecs.scala
@@ -42,7 +42,7 @@ private[codec] trait HeaderCodecs {
       render,
     )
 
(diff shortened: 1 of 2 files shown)


zio-http/src/main/scala/zio/http/codec/HeaderCodecs.scala
```diff
@@ -52,6 +42,16 @@ private[codec] trait HeaderCodecs {
       render,
     )
 
+  def nameTransformOpt[A, B](name: String, parse: B => Option[A], render: A => B)(implicit
```
jdegoes:
```suggestion
  def nameTransformOption[A, B](name: String, parse: B => Option[A], render: A => B)(implicit
```

vigoo (approved):

PR #2197, opened 2023-05-17
Multipart/from file input implementation and some refactoring

pablf:
/claim #2160

It is a big refactoring, but a lot of changes are just deleting boilerplate code and implementing multipart/from file input:

- 	 Instead of having Options[A] as a parameter of CliEndpoint, we have different subclasses of HttpOptions to deal with URL, body and headers (URLOptions, BodyOptions, HeadersOptions).

- The functionality of the method embed of CliEndpoint is done now by the method transform of HttpOptions. The benefit is that instead of having to define embed for each case, we can use the same transform method for each subclass of HttpOptions. It allows also to reconstruct easily URL.
- The commands are different, we would have "method url options":
Before: "-get-users --user-id integer ..."
Now:  " get /users/user-id? --user-id integer..."
It exposes more of the HTTP API. I don't know if that is better, but it can be easily changed to the first version (or allow the possibility of choosing between both when making the CliApp). What should we do?

- Each BodyOptions would represent a FormField, implementing multipart inputs. Each FormField can be specified written in the command, by a file path or by a URL
(URL not implemented yet and might not be desired, but maybe useful). If the schema of a FormField is more complex and hides a Schema.Primitive[BinaryType] it would not be valid to specify
writing everything else and specifying the binary data as a file path. It mustn't be difficult to implement this behaviour. Is it desired?

- For the multipart output, I think that it would be better to define what it should do. Should we allow options for each FormField if we wish to save or print a FormField or
do it automatically depending on the type of data? Do we do something with the headers?

### DIFF
--- a/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliRequest.scala
+++ b/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliRequest.scala
@@ -21,7 +21,7 @@ import java.io.IOException
  */
 
 private[cli] final case class CliRequest(
-  body: Chunk[Either[Either[(String, Path, MediaType), String], FormField]],
+  body: Chunk[Retriever],
   headers: Headers,
   method: Method,
   url: URL,
@@ -29,7 +29,7 @@ private[cli] final case class CliRequest(
   saveResponse: Boolean = false
 ) { self =>
 
-  def addBody(value: Either[Either[(String, Path, MediaType), String], FormField]) =
+  def addBody(value: Retriever) =
     self.copy(body = self.body ++ Chunk(value))
 
   def addHeader(name: String, value: String): CliRequest =
@@ -49,13 +49,7 @@ private[cli] final case class CliRequest(
    * Retrieves data from files, urls or command options and construct a HTTP Request.
    */
   def toRequest(host: String, port: Int): Task[Request] = for {
-    formFields <- ZIO.foreach(body)( _ match {
-        case Left(Left((name, file, mediaType))) => for {
-          chunk <- Body.fromFile(new File(file.toUri())).asChunk
-        } yield FormField.binaryField(name, chunk, mediaType)
-        case Left(Right(url)) => ??? // TODO
-        case Right(formField) => ZIO.succeed(formField)
-      })
+    formFields <- ZIO.foreach(body)(_.retrieve())
     finalBody <- Body.fromMultipartFormUUID(Form(formFields))
   } yield Request
       .default(
--- a/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliRequest.scala
+++ b/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliRequest.scala
@@ -1,23 +1,22 @@
 package zio.http.endpoint.cli
 
-import zio._
-import zio.json.ast._
+import java.io.{File, IOException}
+import java.nio.channels.FileChannel
 import java.nio.file.Path
 
-import zio.http._
+import scala.io.Source
+
+import zio._
 import zio.cli._
-import java.nio.channels.FileChannel
+import zio.json.ast._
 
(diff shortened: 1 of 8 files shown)


zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliRequest.scala
```diff
@@ -2,27 +2,39 @@ package zio.http.endpoint.cli
...
+
+/**
+ * Represents a Request. The body parameter allows implementation of multipart form data and the retrieval of a body
+ * from a file or an URL.
+ */
 
 private[cli] final case class CliRequest(
-  url: URL,
-  method: Method,
+  body: Chunk[Either[Either[(String, Path, MediaType), String], FormField]],
```
jdegoes: At the point of `Either[Either[...]]`, it's worth introducing a `sealed trait` to ease code comprehension and maintenance. Also, it seems you moved from `Json` to `String` for the body. I don't see how that can work if the body must be generated incrementally by each option / argument.

pablf: Certainly, I would call it BodyConstructor. Are you referring to this? https://github.com/zio/zio-http/blob/76300ea8e16cb4fc603f3d0ec746028951565714/zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpOptions.scala#L59
Should it be instead of `textField` `binaryField`? `a` would be `Json` but it specifies only a FormField not the whole body. The `BodyConstructor[(String, Path, MediaType), String, FormField]` would represent  `BodyConstructor[from File, from Url, FormField from Json]` that would be converted to a FormField depending on the specification and that `Chunk[FormField]` would be used to specify the (possibly multipart) body.

jdegoes:
> The commands are different, we would have "method url options":

I think we should strive to have the most idiomatic command-line interface available. Command-line interfaces are generally verb-driven, e.g. `git commit`, so I think the older interface of `myapp get-users` is more CLI idiomatic than an HTTP-centric user-interface. I think we should view ZIO HTTP CLI as essentially encapsulating the "detail" that the command-line interface is, in fact, powered by a cloud service.

Now, that doesn't mean we can't support configurability. Users will want high control over the command-line interface, rather than just being forced to accept our defaults. But I think the default, out-of-the-box experience, should provide as idiomatic CLI experience as we can given the limitations.

jdegoes:
> f the schema of a FormField is more complex and hides a Schema.Primitive[BinaryType] it would not be valid to specify
writing everything else and specifying the binary data as a file path. It mustn't be difficult to implement this behaviour. Is it desired?

We should support things like `--name "My Image" --image-content my-image.png`, etc., where the local file is uploaded via multi-part encoding. Is that what you are asking?

pablf:
Not exactly. `CliEndpoint` decomposes an `HttpCode` using `fromInput` and `fromAtom`. `fromInput` basically breaks the `++` operators and then `fromAtom` takes an `HttpCodec.Atom` processes it depending on the type. If `fromAtom` encounters two `HttpCodec.Content` it would be possible to specify each content separately, either from file or using Json. But it would not be possible to specify one `HttpCodec.Content` using both Json and from file if the schema of the codec includes binary data. I understand that each `HttpCodec.Content` represents an unique part so it would not allow to specify different parts of the schema using different procedures. Is this so?

In other words, in the last version `fromInput` decomposed everything and included it in a Json, and now it is possible to treat separately each `HttpCodec.Content` and specify each one either using Json (if the schema does not include binary data) or from file. So `--name "My Image" --image-content my-image.png` is supported while `--data.name "My Image" --data.image-content my-image.png` is not. Although this could be changed using the infamous EitherEither.

pablf:
Added configurability for the command style and created Retriever to construct FormField from different sources.

jdegoes:
@pablf This is looking pretty good. Can you add some tests and we can try to push this through. Thank you!

(Please re-open when ready for another look.)

PR #2203, opened 2023-05-18
Docs appendable to Endpoint, no name req. for out methods in Endpoint

### LINKED ISSUE #2193: Improve API to add Docs to `Endpoint` `in`,  `out` and `outError`
The current methods `in`, `out` and `outError` on `Endpoint` do create the codec internally. Therefore the user can not add docs to them.

There should be overloaded methods that have a doc field, that is added to the internally created Codec.

### LINKED ISSUE #2195: Improve multipart vs alternation API of `Endpoint`
## Current situation
After implementing multipart form-data support for `Endpoint` #2140 / #2158, all variants of out have a mandatory field `name` (except the one without any params) which makes 100% sense for multipart, but the methods are actually all alternating the codecs. For example

```scala
  def out[Output2: Schema](
    name: String,
    status: Status,
    mediaType: MediaType,
  )(implicit alt: Alternator[Output, Output2]): Endpoint[Input, Err, alt.Out, Middleware] =
    Endpoint(
      input,
      output =
        (self.output | HttpCodec.content(name, mediaType)(implicitly[Schema[Output2]])) ++ StatusCodec.status(status),
      error,
      doc,
      mw,
    )
```

This method, having a media type and a name, seems to make only sense in the context of multipart. But the codec is added as an alternation. In fact all `out` methods are alternating. So I think you can not use them to create multipart outs at all.

## Idea

I think the `out` methods should only handle alternation. And there should be a specific method for multipart codec that takes var args. Why? Because currently, the order of the methods called on an Endpoint object does not matter. It results always in the same endpoint. If we would use a version of out that creates only a single `Http.Content` at a time, but uses `++` instead of `|` to add the codec, the order of the calls will change the resulting Endpoint and it will be harder to reason about it

I suggest something that looks like this when using it.

```scala
Endpoint.get(...).outMultipart(Status.Ok)(
	multipart[Int]("age", myMediaType),
	multipart[String]("username", myOtherMediaPart)
)
```

This offers

- a clear api that lets the user know what he creates
- keeps the guarantee, that the order of the build steps does not influence the resulting Endpoint
- removes the mandatory name from the alternation outs
- gives the bundle of multipart data exactly one status code

## Alternative

No alternation for `out`. Calling int multiple times means multipart. In this case we still need methods without name. But how could I model an API that return 200 with content and 204 without. Or another content for 201?
In that case status would also not make any sense btw. Because you can not change the status for different parts of the multipart answer, can you?

RFC @jdegoes @vigoo

987Nabil:
Also add method to append doc to `EndpointMiddleware`

fixes #2193 
/claim #2193

fixes #2195
/claim #2195

### DIFF
--- a/zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
+++ b/zio-http/src/main/scala/zio/http/endpoint/Endpoint.scala
@@ -183,13 +183,33 @@ final case class Endpoint[Input, Err, Output, Middleware <: EndpointMiddleware](
 
   /**
    * Returns a new endpoint derived from this one, whose request content must
-   * satisfy the specified schema.
+   * satisfy the specified schema and is documented.
+   */
+  def in[Input2](doc: Doc)(implicit
+    schema: Schema[Input2],
+    combiner: Combiner[Input, Input2],
+  ): Endpoint[combiner.Out, Err, Output, Middleware] =
+    copy(input = input ++ HttpCodec.content(schema) ?? doc)
+
+  /**
+   * Returns a new endpoint derived from this one, whose request content must
+   * satisfy the specified schema and is documented.
    */
   def in[Input2](name: String)(implicit
     schema: Schema[Input2],
     combiner: Combiner[Input, Input2],
   ): Endpoint[combiner.Out, Err, Output, Middleware] =
     copy(input = input ++ HttpCodec.content(name)(schema))
+
+  /**
+   * Returns a new endpoint derived from this one, whose request content must
+   * satisfy the specified schema and is documented.
+   */
+  def in[Input2](name: String, doc: Doc)(implicit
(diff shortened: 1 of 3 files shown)


987Nabil:
@jdegoes the algora bot does not seem to support two claims in one PR. Can you help out?

987Nabil:
idk why the performance test is failing. I don't think this changes should affect performance at all. Maybe just a problem in the runner?

zcesur:
> @jdegoes the algora bot does not seem to support two claims in one PR. Can you help out?

hey @987Nabil - both claims are in now :+1:

vigoo:
> idk why the performance test is failing. I don't think this changes should affect performance at all. Maybe just a problem in the runner?

restarted it

vigoo (approved):

PR #2206, opened 2023-05-19
Fail with 400 if request to endpoint is malformed

NavidJalali:
Malformed requests seem to turn into internal server errors instead of bad requests

### DIFF
--- a/zio-http/src/main/scala/zio/http/endpoint/internal/EndpointServer.scala
+++ b/zio-http/src/main/scala/zio/http/endpoint/internal/EndpointServer.scala
@@ -17,6 +17,7 @@
 package zio.http.endpoint.internal
 
 import zio._
+
 import zio.http._
 import zio.http.codec.HttpCodecError
 import zio.http.endpoint.{EndpointMiddleware, Routes}
@@ -38,7 +39,7 @@ private[endpoint] final case class EndpointServer[R, E, I, O, M <: EndpointMiddl
       .catchSomeCause {
         case Cause.Fail(underlying: HttpCodecError, _) =>
           ZIO.succeed(Response(Status.BadRequest, body = Body.fromString(underlying.getMessage())))
-        case Cause.Die(underlying: HttpCodecError, _) =>
+        case Cause.Die(underlying: HttpCodecError, _)  =>
           ZIO.succeed(Response(Status.BadRequest, body = Body.fromString(underlying.getMessage())))
       }
       .orDie
--- a/zio-http/src/main/scala/zio/http/endpoint/internal/EndpointServer.scala
+++ b/zio-http/src/main/scala/zio/http/endpoint/internal/EndpointServer.scala
@@ -19,7 +19,6 @@ package zio.http.endpoint.internal
 import zio._
 
 import zio.http._
-import zio.http.codec.HttpCodecError
 import zio.http.endpoint.{EndpointMiddleware, Routes}
 
 private[endpoint] final case class EndpointServer[R, E, I, O, M <: EndpointMiddleware](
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/endpoint/internal/EndpointServer.scala
```diff
@@ -28,10 +29,19 @@ private[endpoint] final case class EndpointServer[R, E, I, O, M <: EndpointMiddl
...
+    endpoint.input
+      .decodeRequest(request)
+      .flatMap { value =>
+        handler(value).map(endpoint.output.encodeResponse(_)).catchAll { error =>
+          ZIO.succeed(single.endpoint.error.encodeResponse(error))
+        }
       }
-    }
+      .catchSomeCause {
+        case Cause.Fail(underlying: HttpCodecError, _) =>
```
987Nabil: This breaks the alternatives handling for routes. See `Routes#dispatch`. I think `dispatch` is the place where we should turn this into a response. Dispatch currently returns a `ZIO[R, Nothing, Response]` but I think we could turn this into `ZIO[R, Response, Response]` and have a `BadRequest` response in the error channel of that zio, in case `HttpCodecError.isHttpCodecError(cause)` is true.


wdyt @jdegoes @vigoo ?

987Nabil (changes requested):

zio-http/src/main/scala/zio/http/endpoint/internal/EndpointServer.scala
jdegoes: Can't `handle` here return `ZIO[R, Response, Response]`, before it ends up as a route?

987Nabil: Not 100% sure what you mean with `before it ends up as a route?`. But I guess that happens already before?
If handle would return `ZIO[R, Response, Response]` I think it would also break `dispatch`, at least in the current form. If we would assume, that an Response in the error channel always means, we had an `HttpCodecError` and we can try the next alternative, then I guess we could change the impl. in dispatch, but it might be fragile to future changes, that actually want to return this error response instead of trying an alternative.
I actually think the best argument for moving this to dispatch is, that only after all alternatives are tried you need a error response. So responses created here, would end up garbage, if there is an alternative. But leaving this as a zio failure with a cause, makes it possible to add up all causes and give an error response of all the `HttpCodecError`s that happened, instead of just the last one.
And creating only responses that would actually be sent out to the caller, feels also more right.

NavidJalali: Yep, it makes sense, thanks Nabil, I'll move it up to dispatch

NavidJalali: In this case should list all the alternatives I tries in the response body? Or should just return a simple 400 with no body?
@jdegoes

987Nabil: I'd say added the errors should be fine. 1. You want to know the error. 2. There shouldn't be any information leaked except for what was in the request you sent anyway.

NavidJalali: Actually it would somewhat leak all the possible ways it failed so that's why I didn't add it

987Nabil: But if you only return the last error, you might get an error that is unrelated to what you actually want to do.

vigoo:
I think this is good to merge, but I would separately think about some kind of debug features for `Endpoint` that can be used to diagnose why is it returning 400/500 etc. Maybe just a (optional?) debug logging of all the tried and failed cases, while still not leaking these details in the response.

987Nabil:
@vigoo what would we be leaking anyway? For example, if there is a decoding of json failing, maybe I want to let the caller know what path is missing.

vigoo:
@987Nabil Not sure, I'm not against returning the real error in the body. Maybe it should be configurable? I would definitely prefer getting a proper error message than just "Bad Request" when trying to use an API :)

NavidJalali:
@987Nabil @vigoo 
The issue here is that since the cause accumulates all the errors possible (basically on every failed alternation). 
So there is a list of codec errors with an error message for every failed alternation. How would you decide which error to show?
The first? The last? All of them in a string? The one that "matched" the closest?

vigoo:
I'm merging this for now, let's improve the response / logging in separate PR(s).

vigoo (approved):

Issue #1996, opened 2023-02-15
Add support for Happy Eyeballs


regiskuckaertz:
**Is your feature request related to a problem? Please describe.**

For cases when a DNS resolves to multiples IP(v4/v6) addresses, this algorithm speeds up establishing the best connection.

**Describe the solution you'd like**

See https://www.rfc-editor.org/rfc/rfc8305

afaiu, right now [using the underying machinery](https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala#L93).

**Additional context**

Would be great to drive zio-http adoption at my company 😁

vigoo:
Related: #1730

grouzen:
I'm researching this one.

nicoburniske:
attempt #1996

Sukuna0007Abhi:
Hi sir @jdegoes @regiskuckaertz is this issue still open? Can I work on this...

PR #2207, opened 2023-05-23
Add support for Happy Eyeballs DNS Address Resolution

nicoburniske:
/claim #1996 

Currently facing unit test failure. NettyConnectionPoolSpec tests are not terminating


zio-http/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -114,8 +108,107 @@ object NettyConnectionPool {
...
+  private def executeHappyEyeballs[A](tasks: List[Task[A]], delay: Duration): Task[A] =
+    tasks match {
+      // What should this error be?
+      case Nil                => ZIO.fail(new IllegalStateException("No tasks left."))
+      case task :: Nil        => task
+      case task :: otherTasks =>
+        Promise.make[Nothing, Unit].flatMap { failurePromise =>
+          val taskWithFailure = task.onError(_ => failurePromise.complete(ZIO.unit).ignore)
+          val continue        = failurePromise.await.timeout(delay).ignore *> executeHappyEyeballs(otherTasks, delay)
+          taskWithFailure.race(continue)
```
nicoburniske: In the NettyConnectionPoolSpec tests `taskWithFailure` is completing, but the function does not terminate. I am not sure why the race on the right `continue` is not being interrupted. I am sure I'm doing something wrong but I don't understand what exactly

zio-http/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -114,8 +108,107 @@ object NettyConnectionPool {
       _             <- NettyFutureExecutor.executed(channelFuture)
       result        <- ZIO.attempt(channelFuture.channel())
     } yield result
+
+    for {
+      resolvedHosts <- dnsResolver.resolve(location.host)
+      sortedHosts    = sortHostsHappyEyeballs(resolvedHosts)
+      createChannels = sortedHosts.map(createChannel).toList
```
nicoburniske: If I replace this line with 

`sortedHosts.map(createChannel).map(_.debug("create channel")).toList`

the create channel log comes out as a success, but the function still fails to terminate.

jdegoes:
I know we need to take another look at this in a session, but also, it's important that we make sure the current code path is not negatively affected, performance-wise.

jdegoes:
I will close for now, please re-open when you are ready for a final review!

PR #2220, opened 2023-05-31
Shading of netty dependencies

dkarlinsky:
Netty is used by many libraries and version incompatibilities can be an issue.
For example, the version of netty brought in by aws-sdk in our projects, was incompatible with zio-http and prevented me from using it.
The purpose of this PR is to publish additional versions of zio-http jars with netty dependencies shaded.
This is WIP, since I still need to figure out the github actions `ci-release` part and I'm kind of stuck without the ability to publish and check the results...
DYT @vigoo, @scottweaver


vigoo:
The `ci-publish` plugin runs `+publishSigned` by default. I think you can do `+publishLocal` for testing, that would almost be the same (except of the signing part)

dkarlinsky:
> The `ci-publish` plugin runs `+publishSigned` by default. I think you can do `+publishLocal` for testing, that would almost be the same (except of the signing part)

`publishLocal` does work (i.e generates a jar with netty classes under `zio/http/shaded/netty/...` and a POM with netty dependencies commented out).
So if everything works right, if this is merged, the next release will be published with netty shaded as the only option.
I was thinking that it would be safer to publish the shaded jars under a different name (or maybe classifier?) at least at first

vigoo:
Yes I think it would be nice to publish both shaded and "normal"

dkarlinsky:
@vigoo any ideas on how I can test the release flow?

vigoo:
Not really, maybe we should find a time when both of us are online and just figure it out on master?

vigoo (approved):

vigoo:
Merging it to test publishing

PR #2239, opened 2023-06-07
Move Http.fromFileZIO to Handler.

landlockedsurfer:
Hi all!
There were some remarks on `Http.fromFileZIO` last week in the ZIO HTTP Discord channel.
See https://discord.com/channels/629491597070827530/819703129267372113/1113578446673678437
As per the remarks in the channel it was deemed useful to move `fromFileZIO` to `Handler`, so that the effect providing the file can access to the ZIO environment.
It was also argued that the behaviour of `Http.fromFileZIO` is a little odd, since on failure it will just delegate to the next route. Since such a behaviour could also be implemented using other primitives, it might make sense to completely remove `Http.fromFileZIO`.
Following those lines it might also make sense to move more existing methods from `Http` to `Handler` that are more concerned with generating a response vs. performing the routing.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Handler.scala
+++ b/zio-http/src/main/scala/zio/http/Handler.scala
@@ -22,8 +22,9 @@ import zio.http.Header.HeaderType
 import zio.http.internal.HeaderModifier
 import zio.stream.ZStream
 
-import java.io.File
+import java.io.{File, FileNotFoundException, IOException}
 import java.nio.charset.Charset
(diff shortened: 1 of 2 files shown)


jdegoes:
@landlockedsurfer Can you look into build failures?

landlockedsurfer:
Hi John!
I fixed an ordering issue with the imports and the build shows green now:
https://github.com/zio/zio-http/actions
Hope this branch merges fine now.

vigoo (approved):

PR #2252, opened 2023-06-14
Scope Requests

### LINKED ISSUE #2246: Behavior when client response body is not consumed
*(I'm not sure this counts as a bug, or just a misunderstanding on my part, but my guess would be this is not entirely intended to be this way)*

Consider the following minimal app example:

```
import zio._
import zio.http._

object TestClient extends ZIOAppDefault {

    val app = for {
        response <- Client.request("https://zio.dev")
        _ = println(s"Result is ${response.status}")
 //       _ <- response.body.asArray
    } yield ()
     
    def run = for {
        result <- app.provide(Client.default)
    } yield ()
}
```

*Expected behavior:*

Application prints "Result is Ok", and then exits.

*Actual behavior:*

Destruction of the `ZClient ZLayer` is delayed causing the application to appear to hang during shutdown, after printing "Result is Ok". Eventually (10-100 seconds later) the application exits - the interval seems to depend on the server (closing the connection?), but it can also just wait indefinitely.

*Additional information:*

If we consume the response body, by un-commenting the corresponding line in the example, the application exits immediately.

*Question:*

What is the correct approach to deal with requests where the response may not be consumed by the application, either through a bug or error, through no need, or some scenario where multiple requests are raced against each other with only the fast response being consumed and the rest discarded?

What is the impact of dangling requests ? (do they get cleaned up eventually?)

One can manually try to make sure the stream is always consumed, however this is a) error prone, and b) inefficient if the data is not needed (better would be to just close the connection).

Could/should requests be attached to a `Scope` or resource that provides cleanup?

adamgfraser:
Resolves #2246. /claim #2246

Making a request acquires a resource in that it opens a connection that must be closed even if the stream body is not consumed.

### DIFF
--- a/docs/dsl/body.md
+++ b/docs/dsl/body.md
@@ -30,7 +30,7 @@ On the client-side, `ZIO-HTTP` models content in `ClientRequest` as `Body` with 
 To add content while making a request using ZIO HTTP you can use the `Client.request` method:
 
 ```scala mdoc:silent
-  val actual: ZIO[Client, Throwable, Response] = 
+  val actual: ZIO[Client with Scope, Throwable, Response] = 
     Client.request("https://localhost:8073/success", content = Body.fromString("Some string"))
 ```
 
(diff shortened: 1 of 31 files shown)


jdegoes (approved):

jdegoes:
Probably the correct thing to do. 👍

PR #2258, opened 2023-06-14
feat: Allow Unknown AcceptEncoding

ex0ns:
As discussed with @vigoo  on Discord, there is an issue with the current implementation of the AcceptEncoding header.
If the header contains an invalid value or an unsupported encoding, the whole header is disregarded and None is return.

curl's --compressed returns `Accept-Encoding: gzip, br, deflate, zstd`  on my machine and `zstd` is not supported by zio-http (I was not even aware that it existed).

This is a proposal to add support for unknown accept encoding header instead of disregarding the whole header. Another implementation would have been to simply skip them, but it also meant that users of zio-http would have no way to support custom/unsupported encoding.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Header.scala
+++ b/zio-http/src/main/scala/zio/http/Header.scala
@@ -253,7 +253,14 @@ object Header {
       override val raw: String = "*"
     }
 
-    private def identifyEncodingFull(raw: String): Option[AcceptEncoding] = {
+    /**
+     * Represent an unknown encoding
+     */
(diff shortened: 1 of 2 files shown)


vigoo (approved):

jdegoes (approved):

jdegoes:
@ex0ns Looks great. Can you fix conflicts and we can get this merged?

ex0ns:
I don't know what is happening with the CI. Rebase should be ok now

PR #2261, opened 2023-06-15
Significant usability and semantic improvements to ZClient / Client

jdegoes:
This is a massive refactoring of ZClient and some associated data types, with random other cleanup thrown in for good measure.

The main goal of this refactoring was to ensure **uniform semantics** across all methods of ZClient. ZClient is reinterpreted as a simple case class that stores a "base request" (but without a body), together with a body encoder / decoder, and a driver, which is responsible for executing requests. In this new architecture, all methods on ZClient will always use every part of the bundled request, unlike the predecessor version, which would use them sometimes, but not other times.

 - Neutral or identity elements have been added to all components of a Request that did not have them, now allowing well-defined composition semantics on Request, Method, Version, Body, etc.
 - Request constructors have been standardized
 - Accessor methods on Client have been deleted
 - Variations have been reduced so there are fewer ways to do everything
 - Uniformity of conventions and method names has been improved
 - Some convenience methods have been added to things like URL to reduce boilerplate

This is currently a WIP and awaits some final polishing and consistency: in particular, we need a uniform convention for when to clobber a field versus when to add information to the field. This is currently inconsistent between Request, URL, and Client.


adamgfraser (approved):

vigoo:
I would like to challenge the number of options even a bit more. 

To me what makes sense regarding having "parts of the request" fixed in a client is to be able to create a client for a specific target which includes:
- host 
- port
- some prefix URL (defaulted to `Root`)
- settings

With this I would only have one method `request(Request)` with the following properties:
- `Request` can only have relative URLs if the client has a specific host set
- `Request` can only have absolute URLs if the client has no host set
- In case it is a relative URL, it is appended to the client's prefix url
- `Request` itself should have helper constructors to create requests for various methods (with no body for get, etc)

This makes all the `doXYZ` and `method` methods unnecessary. 

Of course having a default method for a client may make sense but I feel that's unexpected and just adds more complexity.

vigoo (approved):

jdegoes:
@vigoo I removed default method and I think I like it for the common case (e.g. `client.get("/foo")`), which is what matters most.

I left `client.get` as shorthand for `request(Request.get("path"))`, and the other ones, as it's much nicer to type.

Now there are the get/put/post/delete methods on client, and two other ones:

```scala
// The primary one if you are using body encoding:
def request(method: Method, suffix: String)(body: In)(implicit trace: Trace): ZIO[Env & Scope, Err, Out]

// A secondary one:
def request(request: Request)(implicit ev: Body <:< In): ZIO[Env & Scope, Err, Out]
```

I am thinking about URLs more and will try something.

jdegoes:
Nevermind, to try what I wanted, we would need to track relative versus absolute URLs, and this would be sticky: methods must try to preserve the type of URL, and the type of URL needs to make it into Request and Client, so we know whether or not we can allow the user to call Request given what URL has been set already and what URL is inside the Request.

All of this is possible but I am not sure would lead to a pleasant experience. It's probably more reasonable to just tell the user to do the right thing and fail fast if they do not.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -34,6 +34,12 @@ import zio.http.internal.BodyEncoding
  */
 trait Body { self =>
 
+  /**
+   * A right-biased way of combining two bodies. If either body is empty, the
+   * other will be returned. Otherwise, the right body will be returned.
+   */
+  def ++(that: Body): Body = if (that.isEmpty) self else that
```
guizmaii, line 41: What happens when both `self` and `that` are not empty here? Aren't we losing `self` content? 🤔

zio-http/src/main/scala/zio/http/Method.scala
```diff
@@ -56,4 +66,6 @@ object Method {
   object DELETE  extends Method { val name = "DELETE"  }
   object TRACE   extends Method { val name = "TRACE"   }
   object CONNECT extends Method { val name = "CONNECT" }
+
+  object Default extends Method { val name = "GET" }
```
guizmaii, line 70: Why implement this with another `object` and not just with a `def default = GET`? 🤔

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -23,14 +23,29 @@ import zio.ZIO
...
+   */
+  def ++(that: Request): Request =
+    Request(
+      self.version ++ that.version,
+      self.method ++ that.method,
+      self.url ++ that.url,
+      self.headers ++ that.headers,
+      self.body ++ that.body,
+      that.remoteAddress.orElse(self.remoteAddress),
+    )
```
guizmaii, line 47: I'm not sure to understand what's the semantical meaning nor the purpose of this method.
If I have two Requests, `GET /api/example` and `POST /api/another/example?a=b`, what would that semantically mean to combine these two requests? 
What would be the purpose? And what would be the correct result for this combination?

zio-http/src/main/scala/zio/http/Request.scala
```diff
@@ -92,18 +104,31 @@ object Request {
...
+  def options(url: URL): Request = Request(method = Method.OPTIONS, url = url)
 
-  def delete(url: URL): Request = default(Method.DELETE, url)
+  def patch(path: String, body: Body): Request = Request(method = Method.PATCH, url = URL(Path(path)), body = body)
 
-  def get(url: URL): Request = default(Method.GET, url)
+  def patch(url: URL, body: Body): Request = Request(method = Method.PATCH, url = url, body = body)
 
-  def options(url: URL): Request = default(Method.OPTIONS, url)
+  def post(path: String, body: Body): Request = Request(method = Method.POST, url = URL(Path(path)), body = body)
```
guizmaii, line 127: I think that there's an issue here. See https://github.com/zio/zio-http/issues/2308

zio-http/src/main/scala/zio/http/Version.scala
```diff
@@ -17,24 +17,40 @@
...
 
 object Version {
   val `HTTP/1.0`: Version = Http_1_0
   val `HTTP/1.1`: Version = Http_1_1
 
-  case object Http_1_0 extends Version
+  /**
+   * Indicates no preference for version. The default version will be used.
+   */
+  case object Default extends Version {
```
guizmaii, line 45: Why implement it as a `case object` you'll have to deal with in your pattern matchings and not as `def Default = Http_1_1`?

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -23,61 +23,28 @@ import zio.http.netty.NettyConfig
...
+  version: Version,
+  url: URL,
+  headers: Headers,
+  sslConfig: Option[ClientSSLConfig],
+  bodyEncoder: ZClient.BodyEncoder[Env, Err, In],
+  bodyDecoder: ZClient.BodyDecoder[Env, Err, Out],
+  driver: ZClient.Driver[Env, Err],
+) extends HeaderOps[ZClient[Env, In, Err, Out]] { self =>
+  def apply(request: Request)(implicit ev: Body <:< In, trace: Trace): ZIO[Env & Scope, Err, Out] =
+    self.request(request)
```
guizmaii, line 38: In term of API design, I'm personally not a fan of the fact that the apply method triggers the HTTP requests.
It leads to code looking like this:
```
val client: Client = ...
val request: Request = ...
val response = client(request)
```

It's kind of odd, I think. It's IMO not really idiomatic Scala.

PR #2271, opened 2023-06-19
Send HTTP requests like GET /?p=1 instead of GET ?p=1

### LINKED ISSUE #2238: URL.decode on a URL without a path or ending slash fails to "get"
**Describe the bug**
URL.decode does not produce a valid zio URL representation.

**To Reproduce**
```
val url = URL.decode("https://blah.com?m=1&k=4")
```
Results in a bad request being sent to the server e.g. `client.request(url)`.

However, `URL.decode(..).withPath("/")` then works.

**Expected behaviour**
A valid URL even if the URL string does not have a trailing slash in the string url.

**Screenshots**
NA

abcpro1:
closes #2238

<br>

An HTTP request for a URL like `http://example.com?p=1` should be
```
GET /?p=1 ...
```

not

```
GET ?p=1 ...
```
<br>

The path should start with a slash.

<br>
<br>

/claim #2238

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/client/NettyRequestEncoder.scala
+++ b/zio-http/src/main/scala/zio/http/netty/client/NettyRequestEncoder.scala
@@ -34,9 +34,11 @@ private[zio] object NettyRequestEncoder {
     val method   = Conversions.methodToNetty(req.method)
     val jVersion = Conversions.versionToNetty(req.version)
 
+    def replaceEmptyPathWithSlash(url: zio.http.URL) = if (url.path.isEmpty) url.addLeadingSlash else url
+
     // As per the spec, the path should contain only the relative path.
     // Host and port information should be in the headers.
-    val path = req.url.relative.encode
+    val path = replaceEmptyPathWithSlash(req.url).relative.encode
 
     val encodedReqHeaders = Conversions.headersToNetty(req.allHeaders)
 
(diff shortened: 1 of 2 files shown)


vigoo:
Why? I think we should send the request as it is specified.

abcpro1:
@vigoo that's a good question. It looks like the RFC for HTTP/1.1 [2616](https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2) specifies the URI part of the request as:
```
       Request-URI    = "*" | absoluteURI | abs_path | authority
```
The option which is relevant to us here is `abs_path`. The same RFC in [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc2616#section-3.2.1) defers the definition of `abs_path` to another RFC [2396](https://datatracker.ietf.org/doc/html/rfc2396#section-3) which defines `abs_path` as:
```
      abs_path      = "/"  path_segments
```

<br>

In my test lab, I have tested the behavior of `java.net.HttpURLConnection` as well as various other programming languages and programs, including `C` with `libcurl`, `python` with `requests`, `ruby` with `net/http`, `go` with `net/http`, `rust` with `reqwest`, `Node.js` with `http`, `C#` with `System.Net`, `curl`, and `wget`.  I found that there is unanimous consensus to send `/?p=1`, rather than `?p=1` given a URL like `http://example.com?p=1`.

vigoo (approved):
Thanks for the answer, if that's the case I'm ok with it

jdegoes (approved):

Issue #2191, opened 2023-05-15
ZIO HTTP CLI: Add comprehensive test case and fix bugs


jdegoes:
For every possible endpoint combination including method (GET, PUT, POST, DELETE), including query parameters of different types, and headers of different types, and body content of different types, we need a test to verify that the correct CLI representation is generated.

This implies a very comprehensive test suite, which will no doubt uncover bugs that have to be fixed.

pablf:
I would like to /attempt #2191 if none is working on it.

feliciien:
Hello we can work together if you want my discord  youknowsthevibes send me
a text message

PR #2276, opened 2023-06-23
Multipart/from file impl. in CLI and test suite

pablf:
Test suite, /claim #2191. The test suite checks the representation of the commands using `Gen` and the behaviour of the CliApp.

Continuation of #2197, /claim #2160. Addition of `CliClient` allowing the user to use its own client.

I think that it would be a great feature for **ZIO CLI** to have a test-kit to allow easy testing. It would simplify the code here and would helpful for developers using the library. I will open an issue if you think it is a good idea.


jdegoes:
@pablf The great refactoring is done and we are ready to get this looked at and merged in. Please re-open when you've had a chance to fix conflicts!

pablf:
@jdegoes Great! Changes are merged.

PR #2277, opened 2023-06-23
Routes

jdegoes:
This is an attempt to rethink the core abstractions of ZIO HTTP, in a way that allows some unification between both the endpoint-based API, and the lower-level API that is used by the majority of users, as well as overall simplification and improvement to developer experience.

Currently, the low-level API uses `PartialFunction` to "collect" partial routes, which relies on pattern matching. Not only is there no commonality between this API and the endpoint API, but there is no efficient way to perform dispatch, because each pattern must be checked, one after the other, before a matching route is found.

Instead, this pull request introduces a new `RoutePattern`, which is a reified "pattern" that matches against HTTP methods and paths. Endpoint now integrates support for producing a `RoutePattern`.

`RoutePattern` uses a syntax similar but not identical to pattern matching:

```scala
val pattern = GET / "users" / int("user-id") / "posts" / string("post-id")
```

A new type of prefix tree is introduced (`RoutePattern.Tree`), which allows efficient dispatch to both the low-level API, and the endpoint API. Both types of routes are unified in `Route`, which now describes either a route that handles its errors (by converting them to responses), or a route that does not handle its errors (by exposing them as `E`), which allows error handling to be deferred until all routes are assembled.

`RoutePattern` has a `format` method, which can print out the path to the route in a type-safe way, allowing generation of HTML that contains links to a route&mdash;a key feature in other web frameworks.

`Routes` has been introduced, which is the moral equivalent of `Http`, but simplified and specialized. Routes is merely a collection of routes. In the event of overlap, that is, two or more routes that are claiming the same method + path, then they are tried sequentially, until one handler returns a response other than NotFound.

A smart constructor for Handler has been introduced into the top-level `zio.http` package object, which is called `handler`. This facilitates a simple and unified way of constructing handlers for use in defining routes. The smart constructor can handle responses, effectful responses, functions from route input to responses, and effectful functions from route input to responses.

All forms of middleware and aspects have been deleted and replaced by two new data types:

 - `HandlerAspect` - HandlerAspect has clear, well-defined semantics that align with endpoint's needs, and which provide a bulletproof solution for cases like authorization, ensuring that failing aspects stop downstream aspects, as well as the request handler, from executing. `HandlerAspect` is built on `ProtocolStack`, which is a general-purpose way of defining a protocol that can be used to transform a handler. HandlerAspects may produce context, which allows, for example, contextual aspects that produces user identity as part of authentication. **As a result of this refactor, aspects can no longer transform handlers. HandlerAspects do not even see handlers. They are strictly about protocol-level modification to requests and responses.** HandlerAspects should permit a very efficient implementation at some point in the future, although this is not attempted in this pull request.
 - `Middleware`  - A `Middleware` has been introduced, which simply acts as a transformer of `Routes` (which is a collection of routes). Middleware may do anything that handler aspects may do, but may also modify routing, for example by adding new routes; in addition, they may transform handlers, by modifying the handlers associated with individual routes. Therefore, `Middleware` is truly the most powerful type of "middleware" in ZIO HTTP. Note that all handler aspects implement `Middleware` because they can be used to transform routes handler-by-handler.

`HttpApp` has become a concrete case class that is just a wrapper around `Routes`, used when routes have handled all their errors and converted them into responses, thus signifying that a collection of routes is ready for deployment to a server. All the type aliases for `Http` have been deleted.


zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
```diff
@@ -84,18 +84,21 @@ class EndpointBenchmark {
...
+
   // # Small Data Request
 
   // API DSL
   val usersPosts =
-    Endpoint
-      .get(PathCodec.literal("users") / PathCodec.int("userId") / "posts" / PathCodec.int("limit"))
+    Endpoint(Method.GET / "users" / int("userId") / "posts" / int("limit"))
       .query(QueryCodec.query("query"))
       .out[ExampleData]
```
afsalthaj, line 95: What's your take on naming something like below?

```scala
Endpoint(Method.GET / "users" / int("userId") / "posts" / int("limit") ^?  "query")
 .body[InputData]
 .response[ExampleData]

```

zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
```diff
@@ -84,18 +84,21 @@ class EndpointBenchmark {
...
 
+  import zio.http.codec.PathCodec.int
+
   // # Small Data Request
 
   // API DSL
   val usersPosts =
-    Endpoint
-      .get(PathCodec.literal("users") / PathCodec.int("userId") / "posts" / PathCodec.int("limit"))
+    Endpoint(Method.GET / "users" / int("userId") / "posts" / int("limit"))
```
afsalthaj, line 93: 👍  Nice

zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
vigoo: I think the `^` makes it completely undiscoverable

zio-http/src/main/scala/zio/http/HandlerConstructor.scala
```diff
@@ -0,0 +1,104 @@
+package zio.http
+
+import scala.annotation.implicitNotFound
+
+import zio.ZIO
+
+@implicitNotFound("""
```
vigoo: :heart:

zio-http/src/main/scala/zio/http/Handler.scala
```diff
@@ -737,6 +755,9 @@ object Handler {
     }
   }
 
+  def fromFile[R](makeFile: File)(implicit trace: Trace): Handler[R, Throwable, Any, Response] =
```
vigoo:
```suggestion
  def fromFile[R](makeFile: => File)(implicit trace: Trace): Handler[R, Throwable, Any, Response] =
```

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,158 @@
...
+        Route.Handled(routePattern, (pi: r.PathInput) => handler(pi).mapError(f), l)
+    }
+
+  /**
+   * Returns a route that ignores any errors produced by this one, translating
+   * them into internal server errors that have no details. In order to provide
+   * useful feedback to users of the API, it's a good idea to capture some
+   * information on the failure and embed it into the response.
+   */
+  final def ignoreErrors: Route[Env, Nothing] =
```
vigoo: I liked my original name `withDefaultErrorResponse` better, with this one it is not clear what will happen on the client side (unless you read the comment).

zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
jdegoes: I think it's difficult because the body has a response, and the response has a body.

We could be quite verbose: `requestContent`, `responseContent`, or `requestBody`, `responseBody`.

Also these changes prevent us from using `^?`, as now the route pattern needs to be separated from the other codecs to provide unification between this and the lower-level API.

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: How about something like `internalServerErrors`?

jdegoes: The thing I like about `ignoreErrors` is that it suggests you are doing a hack: just pretending like errors aren't happening. In a sense anytime a user calls this, they are throwing away valuable diagnostic information, so while it's convenient, it's usage should be discouraged in real apps.

987Nabil: When I read your arguments @jdegoes, my first thought was: "Than just throw it away!". If it is there and easy to use, but the easy way is a misuse and discouraged, I think it is bad having it in the first place. If I discover the API through autocomplete and there is no implicit and no fancy input param to this method, that makes me think twice if that is what I want to use, I won't think. I will just see, that it produces the right types and solves my problem.

zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
987Nabil: I do not like `^?` as a syntax. But I like the idea, to also have query params in the route definition.
I think it is reasonable, to have the option to define it in the routes and get the same benefit of the query param being parsed and the request rejected if a required param is missing.
It seems more user friendly than the alternative of `req.url.queryParams.get("query").flatMap(_.headOption).get` too. And more intuitive too, since the caller puts query params at the same place as a route segment.

jdegoes: With this pull request, the query parameters cannot go into the route definition, they must be separated, because endpoints and handlers are unified in `Route`, which requires a `RoutePattern`, which may not mix in query parameters.

So the whole question of `^?` is moot because if this pull request is merged, it will go away forever.

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: Well, we also have `ZIO#ignore`, which serves its purpose.

zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
987Nabil: Just for my understanding: Why can't query params be an optional part of a route pattern?

jdegoes: Because the decision to commit or not to commit occurs at the level of a route pattern: and as far as I know, no web framework allows you to have different handlers based solely on the presence or absence of query parameters.

987Nabil: I understand what you mean and guess it is true that no other framework is doing it.
But these frameworks also don't have "reified patterns" ;)

jdegoes: There is no theoretical problem doing it, but usually a query parameter is a "setting" for a specific handler, rather than an indication that a whole new handler should handle the same method + path. I think it would be very surprising if dispatch also took into consideration query parameters.

987Nabil: Really? If I see the difinition and the query params does not get parsed to an `Option`, is it surprising that when it's missing it can't be handled? And I guess optional handling is need for path segments too or?

987Nabil: Maybe it would work, if query params are when defined by default optional, except the user explicitly marks thems as required? 🤔

zio-http-example/src/main/scala/example/StaticServer.scala
```diff
@@ -3,47 +3,53 @@ package example
...
 import zio.http.html._
 
 object StaticServer extends ZIOAppDefault {
 
   // A simple app to serve static resource files from a local directory.
-  val app = Http
-    .collectHandler[Request] { case Method.GET -> "" /: "static" /: path =>
+  val app = Routes(
+    Method.GET / "static" / trailing -> handler {
+      val extractPath    = Handler.param[(Path, Request)](_._1)
```
vigoo, line 14: Idea: we could have a version of `handler` where the function's parameters are not the values, but these extractor handlers

zio-http-example/src/main/scala/example/StaticServer.scala
```diff
@@ -3,47 +3,53 @@ package example
...
+      val extractRequest = Handler.param[(Path, Request)](_._2)
+
       for {
-        file <- Handler.getResourceAsFile(path.encode)
+        path <- extractPath
+        file <- extractRequest >>> Handler.getResourceAsFile(path.encode)
```
vigoo: I think `extractRequest` is not needed for neither this nor on line 22, because both the `getResourceAsFile` and `template` handlers are independent of their input

zio-http-example/src/main/scala/example/WebSocketAdvanced.scala
```diff
@@ -42,11 +43,13 @@ object WebSocketAdvanced extends ZIOAppDefault {
...
-    Http.collectZIO[Request] {
-      case Method.GET -> Root / "greet" / name  => ZIO.succeed(Response.text(s"Greetings ${name}!"))
-      case Method.GET -> Root / "subscriptions" => socketApp.toResponse
-    }
+  val app: App[Any] =
+    Routes(
+      Method.GET / "greet" / string("name") -> handler { (name: String, req: Request) =>
+        Response.text(s"Greetings ${name}!")
+      },
+      Method.GET / "subscriptions"          -> handler(socketApp.toResponse),
```
vigoo, line 51: Although I like it this way, maybe we could have direct support to connect a route to a socket app (`-> socketApp`)?

zio-http/src/main/scala/zio/http/Handler.scala
```diff
@@ -852,6 +873,14 @@ object Handler {
   def ok: Handler[Any, Nothing, Any, Response] =
     status(Status.Ok)
 
+  def param[A]: ParamExtractorBuilder[A] =
+    new ParamExtractorBuilder[A](())
+
+  class ParamExtractorBuilder[A](val unit: Unit) extends AnyVal {
```
vigoo: Small file organization request :) 
Let's have all these helper classes either inline or in the companion object but not mixed

zio-http/src/main/scala/zio/http/HandlerConstructor.scala
```diff
@@ -0,0 +1,538 @@
...
+      type Z   = (In1, In2, In3, In4, In5, In6, In7, In8) => Response
+
+      def toHandler(z: => Z): Handler[Env, Err, In, Out] =
+        Handler.fromFunction { case (in1, in2, in3, in4, in5, in6, in7, in8) =>
+          z(in1, in2, in3, in4, in5, in6, in7, in8)
+        }
+    }
+}
+
+private[http] trait HandlerConstructorLowPriorityImplicits2 extends HandlerConstructorLowPriorityImplicits3 {
```
vigoo: We should make all the variants defined up to the same number of parameters

zio-http/src/main/scala/zio/http/HttpApp.scala
```diff
@@ -0,0 +1,21 @@
...
+import zio._
+
+/**
+ * A [[zio.http.HttpApp]] is a ready-to-execute HTTP application, which handles
+ * all requests, and always produces valid responses, translating application
+ * errors into HTTP errors.
+ */
+final case class HttpApp2[-Env](
+  requestHandler: Handler[Env, Response, Request, Response],
+  fatalHandler: Cause[Any] => ZIO[Env, Option[Nothing], Response],
```
vigoo: We can mention in the doc comment that it is not always possible to use `fatalHandler`'s result value but it will still be called for its side effects

zio-http/src/main/scala/zio/http/Route.scala
vigoo: I still don't like that is is `ignore` :) For me ignoring errors would mean that if an error happen we do nothing - but this is not the case, we _do_ respond with an internal server error. Also `ZIO#ignore`'s result is something that never produces any result while this method creates a route that always produces a response.

zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -0,0 +1,155 @@
...
+    location: Trace,
+  ) extends Route[Env, Err] {
+    type PathInput = PI
+
+    def apply(request: Request)(implicit ev: Err <:< Response): ZIO[Env, Response, Response] =
+      routePattern.decode(request.method, request.path) match {
+        case Right(pathInput) => handler(zippable.zip(pathInput, request)).mapError(ev)
+        case Left(error)      =>
+          ZIO.die(
+            new NoSuchElementException(
```
vigoo: The same error is `MatchError` in `Route.Handled`

zio-http/src/main/scala/zio/http/codec/SegmentCodec.scala
```diff
@@ -0,0 +1,217 @@
...
+        var defined = true
+        var group   = 0
+        var count   = 0
+        while (i < SegmentCodec.length) {
+          val char = SegmentCodec.charAt(i)
+          if ((char >= 48 && char <= 57) || (char >= 65 && char <= 70) || (char >= 97 && char <= 102))
+            count += 1
+          else if (char == 45) {
+            if (
+              group > 4 || (group == 0 && count != 8) || ((group == 1 || group == 2 || group == 3) && count != 4) || (group == 4 && count != 12)
```
vigoo, line 191: what kind of magic is this? :D

zio-http-example/src/main/scala/example/StaticServer.scala
jdegoes: You mean something like, `handlerParam { (param1, param2) => ... }`, where `param1`, etc., are `Handler` values that extract?

vigoo: yes

jdegoes: I think you would have to specify a lot of types, because the in and outs of the handlers would need to be known, so it would be something like: `handlerParam[(String, Int, Path), String, Int, Path]` or maybe you could do `handlerParam[(String, Int, Path)](_._1, _._2, _._3) { (param1, param2, param3) => ...}`.

Honestly I do not like this example a lot. I rewrote it in the fastest way possible, not necessarily the best way.

The difficulty is that it is a handler of handler situation: where the inner handlers want to use just the request, and the outer handlers want to use more information than the request (request + params).

It would be nice to find a good way to handle these scenarios.

zio-http-example/src/main/scala/example/StaticServer.scala
jdegoes: The `Handler.notFound` is where the problem is. It requires a request.

jdegoes: I was able to delete the other one on line 19.

zio-http-example/src/main/scala/example/WebSocketAdvanced.scala
jdegoes: 👍

jdegoes: SocketApp erases to `Handler` which makes it more difficult to support this. I will look into it in a followup.

zio-http/src/main/scala/zio/http/Handler.scala
jdegoes: Fixed

zio-http/src/main/scala/zio/http/HandlerConstructor.scala
jdegoes: Done, up to 7 for now.

zio-http/src/main/scala/zio/http/HttpApp.scala
jdegoes: Unfortunately I have to delete this because `fatalHandler` cannot compose. I think fatal errors must be handled at server level.

zio-http/src/main/scala/zio/http/Route.scala
jdegoes: Switched to `ZIO.dieMessage`.

zio-http/src/main/scala/zio/http/codec/SegmentCodec.scala
jdegoes: Doesn't look right, does it. I hope I copy/pasted it from the old TextCodec but it looks more like Copilot nonsense.

README.md
```diff
@@ -13,7 +13,8 @@ ZIO Http is a scala library for building http apps. It is powered by ZIO and [ne
 Setup via `build.sbt`:
 
 ```scala
-libraryDependencies += "dev.zio" %% "zio-http" % "3.0.0-RC2"
+libraryDependencies += "dev.zio" %% "zio-http" % "2.0.0-RC11
+"
```
sideeffffect:
```suggestion
libraryDependencies += "dev.zio" %% "zio-http" % "3.0.0-RC2"
```
probably a typo?

adamgfraser (approved):

PR #2282, opened 2023-07-04
Query params enhancement

devsprint:
updated HttpCodec definition for query that will have a NonEmptyChunk[A] instead of A

### DIFF
--- a/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
+++ b/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
@@ -95,10 +95,10 @@ sealed trait HttpCodec[-AtomTypes, Value] {
   final def &[Value2](
     that: QueryCodec[Value2],
   )(implicit
-    combiner: Combiner[Value, Value2],
+    combiner: Combiner[NonEmptyChunk[Value], NonEmptyChunk[Value2]],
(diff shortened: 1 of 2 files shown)


987Nabil:
I understand the idea of having multiple values. But why non empty chunk? What if the values are empty? `?productIds=` or you don't expect a value at all `?constant`. I think both are valid cases that would produce an empty collection.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -95,10 +95,10 @@ sealed trait HttpCodec[-AtomTypes, Value] {
   final def &[Value2](
     that: QueryCodec[Value2],
   )(implicit
-    combiner: Combiner[Value, Value2],
+    combiner: Combiner[NonEmptyChunk[Value], NonEmptyChunk[Value2]],
```
jdegoes, line 98: I think it's risky to change this combiner since you don't know what the left hand side is. Possibly you should not change `QueryCodec` so that the `NonEmptyChunk` is already reflected in the type parameters and you don't need to make any changes here.

jdegoes:
@devsprint Please re-open when you have a chance to fix conflicts and modify the encoder/decoder appropriately. I would really love to have this work finished and merged in!

PR #1107, opened 2022-02-28
Performance: Improve Scheme.decode performance

ShrutiVerma97:
Benchmarks added in the PR.

Results:

Main Branch:
```
[info] Benchmark                                     Mode  Cnt    Score    Error  Units
[info] SchemeDecodeBenchmark.benchmarkSchemeDecode  thrpt    3  148.606 ? 12.130  ops/s
```

Current Branch:
```
[info] Benchmark                               Mode  Cnt    Score      Error  Units
[info] SchemeBenchmark.benchmarkSchemeDecode  thrpt    3  550.254 ? 2245.350  ops/s
```

PR #2290, opened 2023-07-11
Change string 'scheme' size to uppercase value to match the scheme value

sallyjellyy:
thought it would be better for the safety

### DIFF
--- a/zio-http/src/main/scala/zio/http/Scheme.scala
+++ b/zio-http/src/main/scala/zio/http/Scheme.scala
@@ -18,38 +18,36 @@ package zio.http
 
 import zio.Unsafe
 
-sealed trait Scheme {
-  self =>
+sealed trait Scheme { self =>
   def encode: String = self match {
-    case Scheme.HTTP => "http"
+    case Scheme.HTTP  => "http"
     case Scheme.HTTPS => "https"
-    case Scheme.WS => "ws"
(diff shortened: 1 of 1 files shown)


guersam:
This is a rollback of #1107. 

Could you try `String#equalsIgnoreCase` to keep the intention of the original PR?

sallyjellyy:
@guersam done. thanks for the recommendation!

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -59,14 +59,12 @@ object Scheme       {
...
-      else
-        scheme.length match {
-          case 5 => Scheme.HTTPS
-          case 4 => Scheme.HTTP
-          case 3 => Scheme.WSS
-          case 2 => Scheme.WS
-          case _ => null
-        }
+      else if (scheme.equalsIgnoreCase("HTTPS")) Scheme.HTTPS
+      else if (scheme.equalsIgnoreCase("HTTPS")) Scheme.HTTPS
```
guersam:
```suggestion
```

Testing against HTTPS twice.

jdegoes:
I think this improves the safety but at the cost of performance. Since the implications of choosing the wrong scheme are just an error, I am not sure this is any kind of security issue and I would rather keep performance.

Please re-open if you would like to continue discussing!

PR #2291, opened 2023-07-12
refactor: simplify empty check

junghoon-vans:
Fixed to use a function instead of checking the length of the element directly.


jdegoes:
@junghoon-vans This is great! Please fix conflicts and we can merge. Just re-open when you're ready!

PR #2315, opened 2023-07-26
Implement Channel#sendAll

### LINKED ISSUE #2314: The ability to send WebSocket messages in batches
From version `3.0.0-RC2`, the WebSocket protocol handling API has changed (overall, the change has been very positive). 
This change resulted in the loss of the ability to send multiple messages and flush them all at once. 
In the previous versions (`0.0.4` and `0.0.5`), it was possible to optimize sending in this way 
```scala
  val channel: Channel[WebSocketFrame] = ???
  def send(messages: Chunk[WebSocketFrame]) = messages match {
    case ch if ch.isEmpty => ZIO.unit
    case ch =>
      ZIO.foreachDiscard(ch.init)(m => channel.write(m)) *> channel.writeAndFlush(chunk.last, await = true)
  }
```
In my use case, this optimization led to even a few percent increase in performance,  so my question is: would it be possible to add a feature that somehow enables way for such optimization in next versions?

adamgfraser:
Resolves #2314. /claim #2314


vigoo (approved):

jdegoes:
@adamgfraser Good to merge when conflicts fixed!

jdegoes (approved):

PR #2323, opened 2023-07-27
added mediaType parameters to Body methods

### LINKED ISSUE #2305: ZIO Hackathon Issues
- [ ] Add traces to all effectful methods and guarantee this like in the ZIO project (dapperware + Jorge + Gregor + saeed)
- [ ] Cleanup Response so that it is more like Request (Grryum + catch44)
- [ ] Make media type hint a more formal part of Body (Nabil + Adriani)
- [ ] Cleanup & improve performance of server inbound handler (user6385 + fokot + jasper)
- [ ] Hide all Netty-specific details from packages, even those claiming to be independent of Netty (Saeed)

TODO

Adriani-Furtado:
closes https://github.com/zio/zio-http/issues/2305


zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -125,10 +125,12 @@ trait Body { self =>
    */
   def isEmpty: Boolean
 
+  final def getMediaType: Option[MediaType] = this.mediaType
```
jdegoes: ZIO HTTP does not use getter-style conventions, except for `is` / `has` for boolean-returning methods. So this should be `mediaType`. Note there is already a `mediaType` that is private and serves the same function as this, so we should just make that one public.

jdegoes: In addition, we would have a 'setter' which is `def mediaType(newMediaType: MediaType): Body`.

jdegoes:
@Adriani-Furtado Can you please resolve conflicts and merge in the latest changes from master? Thank you! 🙏

Adriani-Furtado:
@jdegoes Sure, I should be able to get this ticket wrapped up by EOD

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -630,7 +637,13 @@ object ZClient {
       body: Body,
       sslConfig: Option[ClientSSLConfig],
     )(implicit trace: Trace): ZIO[Scope, Throwable, Response] = {
-      val request = Request(version, method, url, headers, body, None)
+      val requestHeaders = body.mediaType match {
```
Adriani-Furtado, line 649: Would this be the best place for this logic? I am assuming that if you have a `mediaType` attached to your `Body` and no header set with the content type, it should automatically pull the value from the `Body`

987Nabil: When @jdegoes discussed the feature in the hackerthon opening, I am 99% sure he said that the body content type should override the header. Which I guess makes sense, since this type is really bound to the data. That is also why I am not sure that we should place this logic here. Because you could manually override the content type by adding a header implicitly.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -126,10 +126,20 @@ trait Body { self =>
...
+  /**
+    * Updates the media type attached to this body, returning a new Body
+    * with the updated media type 
+    */
+  def mediaType(newMediaType: MediaType): Body
+
   private[zio] def boundary: Option[Boundary]
 
-  private[zio] def contentType(newMediaType: MediaType, newBoundary: Option[Boundary] = None): Body
+  def contentType(newMediaType: MediaType, newBoundary: Option[Boundary] = None): Body
```
987Nabil: now we have mediaType, that takes a mediaType and contentType that takes a boundary as well. This feels odd to me. I think we should just have one name, and two overloaded methods, one with only media type and one with a non optional boundary.

Adriani-Furtado: Yeah i agree it is a bit confusing. Will apply the suggestion

zio-http/src/main/scala/zio/http/ZClient.scala
Adriani-Furtado: Okay so from my understanding we should override the header if the body has a media type

jdegoes:
@Adriani-Furtado Can you resolve conflicts and format the code using the sbt command `fmt`? Thank you!

jdegoes:
@Adriani-Furtado Would love to get this merged in, however, the website is not passing which means you should run `sbt doc` and `sbt mdoc` to see what the problems are and fix them. In addition, can you please fix conflicts?

Adriani-Furtado:
Hi @jdegoes i will get that solved shortly, thanks for letting me know

jdegoes:
@Adriani-Furtado Thank you!

Adriani-Furtado:
@jdegoes PR is ready to go

jdegoes:
@Adriani-Furtado Looks like the build is passing, just a few conflicts to resolve and we can get this merged in!

jdegoes (approved):

PR #2331, opened 2023-07-27
URL string interpolation macro

vigoo:


adamgfraser (approved):

vigoo:
Note: it is not possible to check compile time that the URL is valid if there are injected string-typed variables in the interpolated string. The macro tries a best effort check, replacing the run time values with examples of the corresponding type.

When used on a static string, it is guaranteed to not throw any errors and always construct a valid URL.

jdegoes (approved):

jdegoes (approved):

PR #2336, opened 2023-07-28
issue_2212 update client proxy settings

### LINKED ISSUE #2212: Update client proxy settings
ZClientAspects are great for updating or adding behaviour to a ZClient. I want to create an Aspect that adds/modifies the proxy settings of a client.

If a program need to talk to services both with and without having to go though a proxy it needs to have multiple clients in the environment. This would be less complex if only one client was needed and an aspect could be applied when a proxy is required.

rajcspsg:
/claim #2212

closes https://github.com/zio/zio-http/issues/2212

### DIFF
--- a/zio-http/src/main/scala/zio/http/ZClient.scala
+++ b/zio-http/src/main/scala/zio/http/ZClient.scala
@@ -30,6 +30,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
   url: URL,
   headers: Headers,
   sslConfig: Option[ClientSSLConfig],
+  proxy: Option[Proxy],
   bodyEncoder: ZClient.BodyEncoder[Env, Err, In],
   bodyDecoder: ZClient.BodyDecoder[Env, Err, Out],
   driver: ZClient.Driver[Env, Err],
@@ -162,6 +163,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
             self.headers ++ request.headers,
             request.body,
             sslConfig,
+            proxy,
           ),
       )
     else
@@ -177,6 +179,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
                 self.headers ++ request.headers,
                 body,
                 sslConfig,
+                proxy,
               ),
           ),
         )
@@ -197,6 +200,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
         headers,
         body,
         sslConfig,
+        proxy,
       )
 
   def retry[Env1 <: Env](policy: Schedule[Env1, Err, Any]): ZClient[Env1, In, Err, Out] =
@@ -218,6 +222,9 @@ final case class ZClient[-Env, -In, +Err, +Out](
   def ssl(ssl: ClientSSLConfig): ZClient[Env, In, Err, Out] =
     copy(sslConfig = Some(ssl))
 
+  def proxy(proxy: Proxy): ZClient[Env, In, Err, Out] =
+    copy(proxy = Some(proxy))
+
   def transform[Env2, In2, Err2, Out2](
     bodyEncoder: ZClient.BodyEncoder[Env2, Err2, In2],
     bodyDecoder: ZClient.BodyDecoder[Env2, Err2, Out2],
@@ -228,6 +235,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
       url,
       headers,
       sslConfig,
+      proxy,
       bodyEncoder,
       bodyDecoder,
       driver,
@@ -277,6 +285,7 @@ object ZClient {
       Version.Default,
       URL.empty,
       Headers.empty,
+      None,
       None,
       BodyEncoder.identity,
       BodyDecoder.identity,
@@ -363,7 +372,7 @@ object ZClient {
 
   trait Driver[-Env, +Err] { self =>
     final def apply(request: Request)(implicit trace: Trace): ZIO[Env & Scope, Err, Response] =
-      self.request(request.version, request.method, request.url, request.headers, request.body, None)
+      self.request(request.version, request.method, request.url, request.headers, request.body, None, None)
 
     final def disableStreaming(implicit ev: Err <:< Throwable): Driver[Env, Throwable] = {
       val self0 = self.widenError[Throwable]
@@ -376,8 +385,9 @@ object ZClient {
           headers: Headers,
           body: Body,
           sslConfig: Option[ClientSSLConfig],
+          proxy: Option[Proxy],
         )(implicit trace: Trace): ZIO[Env & Scope, Throwable, Response] =
-          self0.request(version, method, url, headers, body, sslConfig).flatMap { response =>
+          self0.request(version, method, url, headers, body, sslConfig, proxy).flatMap { response =>
             response.body.asChunk.map { chunk =>
               response.copy(body = Body.fromChunk(chunk))
             }
@@ -413,8 +423,9 @@ object ZClient {
           headers: Headers,
           body: Body,
           sslConfig: Option[ClientSSLConfig],
+          proxy: Option[Proxy],
         )(implicit trace: Trace): ZIO[Env & Scope, Err2, Response] =
-          self.request(version, method, url, headers, body, sslConfig).mapError(f)
+          self.request(version, method, url, headers, body, sslConfig, proxy).mapError(f)
 
         override def socket[Env1 <: Env](
           version: Version,
@@ -443,8 +454,9 @@ object ZClient {
           headers: Headers,
           body: Body,
           sslConfig: Option[ClientSSLConfig],
+          proxy: Option[Proxy],
         )(implicit trace: Trace): ZIO[Env & Scope, Err2, Response] =
-          self.request(version, method, url, headers, body, sslConfig).refineOrDie(pf)
+          self.request(version, method, url, headers, body, sslConfig, proxy).refineOrDie(pf)
 
         override def socket[Env1 <: Env](
           version: Version,
@@ -469,10 +481,11 @@ object ZClient {
       headers: Headers,
       body: Body,
       sslConfig: Option[ClientSSLConfig],
+      proxy: Option[Proxy],
     )(implicit trace: Trace): ZIO[Env & Scope, Err, Response]
 
     final def request(req: Request)(implicit trace: Trace): ZIO[Env & Scope, Err, Response] =
-      request(req.version, req.method, req.url, req.headers, req.body, None)
+      request(req.version, req.method, req.url, req.headers, req.body, None, None)
 
     final def retry[Env1 <: Env, Err1 >: Err](policy: zio.Schedule[Env1, Err1, Any]) =
       new Driver[Env1, Err1] {
@@ -483,8 +496,9 @@ object ZClient {
           headers: Headers,
           body: Body,
           sslConfig: Option[ClientSSLConfig],
+          proxy: Option[Proxy],
         )(implicit trace: Trace): ZIO[Env1 & Scope, Err1, Response] =
-          self.request(version, method, url, headers, body, sslConfig).retry(policy)
+          self.request(version, method, url, headers, body, sslConfig, proxy).retry(policy)
 
         override def socket[Env2 <: Env1](
           version: Version,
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -33,6 +33,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
   bodyEncoder: ZClient.BodyEncoder[Env, Err, In],
   bodyDecoder: ZClient.BodyDecoder[Env, Err, Out],
   driver: ZClient.Driver[Env, Err],
+  proxy: Option[Proxy]
```
adamgfraser: Very minor but I would put this right after the `sslConfig`.

zio-http/src/main/scala/zio/http/ZClientAspect.scala
```diff
@@ -165,6 +166,50 @@ object ZClientAspect {
       }
     }
 
+  /**
+   * Client aspect that updates the proxy settings of specified client
+   */
+  final def proxyClient(proxy: Proxy) (implicit  trace: Trace): ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] =
```
adamgfraser: I think at this point you can delete the aspect and just make `proxy` an operator on `ZClient` like `ssl`.

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -629,6 +640,7 @@ object ZClient {
       headers: Headers,
       body: Body,
       sslConfig: Option[ClientSSLConfig],
+      proxy: Option[Proxy],
```
adamgfraser, line 646: You need to actually use the `proxy` in the implementation of `requestAsync`. Otherwise this isn't doing anything.

zio-http/src/test/scala/zio/http/ZClientAspectSpec.scala
```diff
@@ -45,6 +48,30 @@ object ZClientAspectSpec extends ZIOSpecDefault {
           output.head.endsWith("ms\n"),
         )
       },
+      test("proxy") {
```
adamgfraser: Take a look at `ClientProxySpec` for how to test this. I think you can probably just add a test there.

zio-http/src/main/scala/zio/http/ZClientAspect.scala
rajcspsg: deleted

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -629,10 +640,11 @@ object ZClient {
...
     )(implicit trace: Trace): ZIO[Scope, Throwable, Response] = {
       val request = Request(version, method, url, headers, body, None)
-      val cfg     = sslConfig.fold(config)(config.ssl)
-
+      val sslCfg  = sslConfig.fold(config)(config.ssl)
+      val cfg     = sslCfg.copy(proxy = config.proxy orElse proxy)
```
adamgfraser: This is somewhat confusing and I think you are doing it in the wrong order. If `proxy` is specified we want to use that, otherwise we want to fall back to the configuration. I think this would be more readable as something like:

```scala
val cgf = config.copy(ssl = sslConfig.orElse(config.ssl), proxy = proxy.orElse(config.proxy))
```

zio-http/src/test/scala/zio/http/ZClientAspectSpec.scala
```diff
@@ -19,8 +19,11 @@ package zio.http
 import zio.test.TestAspect.withLiveClock
 import zio.test._
 import zio.{Chunk, Scope, ZIO, ZLayer}
-
 import zio.http.URL.Location
+import zio.http.ZClientAspectSpec.extractStatus
```
adamgfraser: Stray changes to this file.

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -30,6 +30,7 @@ final case class ZClient[-Env, -In, +Err, +Out](
   url: URL,
   headers: Headers,
   sslConfig: Option[ClientSSLConfig],
+  proxy: Option[Proxy],
```
adamgfraser, line 33: You need to add a `def proxy` method that lets the user specify a proxy just like current exists with `def ssl`. You also need to add a test that uses this functionality.

zio-http/src/test/scala/zio/http/ZClientAspectSpec.scala
rajcspsg: removed

zio-http/src/main/scala/zio/http/ZClient.scala
rajcspsg: done

zio-http/src/test/scala/zio/http/ClientProxySpec.scala
```diff
@@ -49,6 +49,44 @@ object ClientProxySpec extends HttpRunnableSpec {
...
+          id <- DynamicServer.deploy(Handler.ok.toHttpApp)
+          proxy = Proxy.empty.url(url).headers(Headers(DynamicServer.APP_ID, id))
+
+          out <- Client
+            .request(
+              Request.get(url = url),
+            )
+            .provide(
+              ZLayer.succeed(ZClient.Config.default),
+              proxied(proxy),
```
rajcspsg: @adamgfraser I don't know how to reuse `ZClient.customized` or using `NettyClientDriver.live` to create proxy ZClient. With this I ended up with 2 `ClientDriver`s. 

Need your feedback on refactor/improve this unit test

adamgfraser: @rajcspsg I think it would be helpful to take a look at the text below `proxy respond Ok` and think about the user feature we are trying to implement and how the test could be rewritten to use our functionality.

In that test, the client is created with `Client.customized`, which requires a `Config`, and the `Config` is customized to use a proxy by using the `proxy` operator on `Config`.

The functionality we are trying to implement is to allow the user to customize the client locally. So instead of having to provide the proxy at the very top of the application the user should be able to specify it right before they call an operator on the client or at any other level.

This means that instead of providing `Client.customized` the user should be able to provide `Client.default` or leave the `Config` as `Config.default`.

Then in your test you can call `ZIO.service[Client]` to access the client, call your new `proxy` operator on `Client` to get a customized client that uses the proxy and then get the same result as in the original test.

rajcspsg: @adamgfraser I updated the code. Could you please take a look again?

rajcspsg: @adamgfraser could you please take at my code changes?

zio-http/src/test/scala/zio/http/ClientProxySpec.scala
```diff
@@ -49,6 +50,24 @@ object ClientProxySpec extends HttpRunnableSpec {
...
       assertZIO(res.either)(isLeft(isSubtype[ConnectException](anything)))
     },
+    test("ZClient proxy respond Ok") {
+      val res =
+        for {
+          port <- ZIO.environmentWithZIO[DynamicServer](_.get.port)
+          url  <- ZIO.fromEither(URL.decode(s"http://localhost:$port"))
+          id   <- DynamicServer.deploy(Handler.ok.toHttpApp)
+          proxy = Proxy.empty.url(url).headers(Headers(DynamicServer.APP_ID, id))
+          zclient <- ZIO.serviceWith[Client](_.copy(proxy = Some(proxy)))
```
adamgfraser: We should use our new `proxy` method instead of `copy`.

rajcspsg: Thanks @adamgfraser I updated the code. Could you please check again!!!

zio-http/src/test/scala/zio/http/ClientProxySpec.scala
```diff
@@ -68,6 +87,7 @@ object ClientProxySpec extends HttpRunnableSpec {
               ZLayer.succeed(NettyConfig.default),
               Scope.default,
             )
+          _ = println(out)
```
adamgfraser: Stray debug statement.

zio-http/src/test/scala/zio/http/ClientProxySpec.scala
```diff
@@ -109,6 +129,6 @@ object ClientProxySpec extends HttpRunnableSpec {
 
   override def spec: Spec[TestEnvironment with Scope, Any] = suite("ClientProxy") {
     serve.as(List(clientProxySpec))
-  }.provideShared(DynamicServer.live, severTestLayer) @@
+  }.provideShared(DynamicServer.live, ZClient.default, severTestLayer) @@
```
adamgfraser: Better to provide resources to just your test instead of to all tests.

rajcspsg: I tried to provide only to my test. But I getting below error -

```
zio.http.ZClient[Any,zio.http.Body,Throwable,zio.http.Response]

[error] /home/raj/Coding/scala/zio-http/zio-http/src/test/scala/zio/http/ClientProxySpec.scala:132:18: 
[error] 
[error] ──── ZLAYER ERROR ────────────────────────────────────────────────────
[error] 
[error]  Please provide a layer for the following type:
[error] 
[error]    1. zio.http.ZClient[Any,zio.http.Body,Throwable,zio.http.Response]
[error]       
[error] ──────────────────────────────────────────────────────────────────────
[error] 
[error]   }.provideShared(DynamicServer.live, severTestLayer) @@
[error]                  ^
    + Live
[error] one error found
```

adamgfraser: You need to provide the layer to your specific test.

rajcspsg: @adamgfraser I updated the test as below. After that I'm always getting timeout error

```scala
test("ZClient proxy respond Ok") {
      val res =
        for {
          port <- ZIO.environmentWithZIO[DynamicServer](_.get.port)
          url  <- ZIO.fromEither(URL.decode(s"http://localhost:$port"))
          id   <- DynamicServer.deploy(Handler.ok.toHttpApp)
          proxy = Proxy.empty.url(url).headers(Headers(DynamicServer.APP_ID, id))
          zclient <- ZIO.serviceWith[Client](_.proxy(proxy))
          out     <- zclient
            .request(
              Request.get(url = url),
            )
            .provide(
              Scope.default,
            )
        } yield out
      assertZIO(res.either)(isRight)
    }.provide(
        DynamicServer.live,
        ZClient.default,
      ) @@ timeout(60 seconds) @@ sequential @@ withLiveClock,
    
```

The error I'm getting is -

>   - ClientProxy - ClientProxySpec - ZClient proxy respond Ok
>   Timeout of 5 s exceeded.

When I used globally I couldn't encountered it.
Any idea?

adamgfraser:
```scala
test("ZClient proxy respond Ok") {
  val res =
    for {
      port <- ZIO.environmentWithZIO[DynamicServer](_.get.port)
      url  <- ZIO.fromEither(URL.decode(s"http://localhost:$port"))
      id   <- DynamicServer.deploy(Handler.ok.toHttpApp)
      proxy = Proxy.empty.url(url).headers(Headers(DynamicServer.APP_ID, id))
      client <- ZIO.serviceWith[Client](_.proxy(proxy))
      out    <- client.request(Request.get(url = url))
    } yield out
  assertZIO(res.either)(isRight)
}.provideSome[DynamicServer](
  Client.default,
  Scope.default,
),
```

rajcspsg: Thanks @adamgfraser That resolved my unit tests issues :)

jdegoes:
@rajcspsg Can you resolve conflicts?

@adamgfraser Is this good to merge?

adamgfraser (approved):

adamgfraser:
@jdegoes Yes!

rajcspsg:
Thanks @jdegoes @adamgfraser  I resolved the conflict. Could you please check again?

PR #2338, opened 2023-07-28
SocketApp refactoring, ability to provide custom websocket config

### LINKED ISSUE #2278: `Response.fromSocketApp` erases headers.
I tried to use WebSocket subprotocols, but they don't work. The idea of subprotocols to take list of protocols sent by client in `sec-websocket-protocol` header, chose one, and send it back with same header. Receiving protocol from client is OK, but sending back is trouble, because looks like `Response.fromSocketApp` erases the header.

See:
1. https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
2. https://datatracker.ietf.org/doc/html/rfc6455

vigoo:
Refactors `SocketApp[R]` so it also implements a `Request => Option[WebSocketConfig]` function that allows customizing the websocket connection including the handshake based on the actual request.

Resolves #2278
/claim #2278

### DIFF
--- a/zio-http/src/main/scala/zio/http/SocketApp.scala
+++ b/zio-http/src/main/scala/zio/http/SocketApp.scala
@@ -1,6 +1,6 @@
 package zio.http
 
-import zio.{Cause, Scope, Tag, Trace, ZEnvironment, ZIO, ZLayer}
+import zio._
 
 final case class SocketApp[-R](
   handler: Handler[R, Throwable, WebSocketChannel, Any],
--- a/zio-http/src/main/scala/zio/http/SocketApp.scala
+++ b/zio-http/src/main/scala/zio/http/SocketApp.scala
@@ -1,0 +1,84 @@
+package zio.http
+
+import zio.{Cause, Scope, Tag, Trace, ZEnvironment, ZIO, ZLayer}
+
+final case class SocketApp[-R](
+  handler: Handler[R, Throwable, WebSocketChannel, Any],
+  customConfig: Handler[R, Throwable, Request, Option[WebSocketConfig]],
+) { self =>
+
+  /**
+   * Creates a socket connection on the provided URL. Typically used to connect
+   * as a client.
+   */
+  def connect(
+    url: String,
+    headers: Headers = Headers.empty,
+  )(implicit
+    trace: Trace,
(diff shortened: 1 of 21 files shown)


zio-http/src/main/scala/zio/http/SocketApp.scala
```diff
@@ -0,0 +1,84 @@
+package zio.http
+
+import zio._
+
+final case class SocketApp[-R](
```
jdegoes: maybe we can rename this `WebSocketApp` while we're at it.

zio-http/src/main/scala/zio/http/SocketApp.scala
```diff
@@ -0,0 +1,84 @@
...
+  handler: Handler[R, Throwable, WebSocketChannel, Any],
+  customConfig: Handler[R, Throwable, Request, Option[WebSocketConfig]],
```
jdegoes: This seems too powerful. I'd prefer `Option[WebSocketConfig]`, or if that cannot be done, something like `<subset of Request> => WebSocketConfig`. 

But it feels like the config should be done at the route pattern level, e.g.:

```scala
val route = Method.GET / "connect" -> handler(socketApp.withConfig(...))
```

You already know the route and pattern there and can set config on a per-route basis. You indeed have access to the full request.

vigoo: The author of the issue wanted to have a way to customize the handshaking based on the request (actual contents of some headers)

kyri-petrou:
Hi @vigoo, is this PR planned to be added to the next RC release? At Caliban we're currently blocked on updating zio-http since we need to be able to do the same as https://github.com/zio/zio-http/issues/2278 for GraphQL subscriptions

vigoo:
Yes this should be in the next release

adamgfraser (approved):

zio-http/src/main/scala/zio/http/SocketApp.scala
frekw: `<subset of Request> => WebSocketConfig` would be great, especially if the result was passed into `Handler.webSocket`. As it is right now there's no way to negotiate the websocket sub-protocol for a single endpoint (such as choosing between graphql-ws and graphql-transport-ws), which would be really nice to do.

PR #2348, opened 2023-07-28
New Response model

hbibel:

PR #2353, opened 2023-07-28
#2305  refactor response class 

escafl0wne:
#2305 Refactor response class

### DIFF
--- a/zio-http/src/main/scala/zio/http/Response.scala
+++ b/zio-http/src/main/scala/zio/http/Response.scala
@@ -1,114 +1,68 @@
-/*
- * Copyright 2021 - 2023 Sporta Technologies PVT LTD & the ZIO HTTP contributors.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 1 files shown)


jdegoes:
There is overlap with #2348

escafl0wne:
> There is overlap with #2348

Sorry for that, shall i close the request?

jdegoes:
@escafl0wne It would be great to work together on a single PR! Are you available to do that?

escafl0wne:
> @escafl0wne It would be great to work together on a single PR! Are you available to do that?

@jdegoes certainly.

escafl0wne:
> @escafl0wne It would be great to work together on a single PR! Are you available to do that?

@jdegoes silly request but would you mind just have a look into the commit and let me know if the code is worth anything? just for my reassurance .

hbibel:
> @jdegoes silly request but would you mind just have a look into the commit and let me know if the code is worth anything? just for my reassurance .

The code looks great and is similar to what @grryum and I had for some time. However we went further and removed `frozen` and `socketApp` as well.

Also note that in general you don't have to implement `hashCode` and `toString` for Scala case classes.

jdegoes:
@escafl0wne I think this was the right direction but was superceded and subsumed by @hbibel and @Grryum's work. Thank you for your work on this!

PR #2355, opened 2023-07-28
Multipart/from file impl. in CLI and test suite

pablf:
#2276 reopened.

Implementation of CliApp, /claim #2160.
Test suite, /claim #2191. 

- The test suite checks the representation of the commands using Gen and the behaviour of the CliApp.
- Addition of CliClient allowing the user to use its own client for the CliApp.

I think that it would be a great feature for ZIO CLI to have a test-kit to allow easy testing. It would simplify the code here and would helpful for developers using the library. I will open an issue if you think it is a good idea.


build.sbt
```diff
@@ -202,8 +202,16 @@ lazy val zioHttpBenchmarks = (project in file("zio-http-benchmarks"))
...
+  .settings(
+    libraryDependencies ++= Seq(`zio-cli`),
+    testFrameworks += new TestFramework("zio.test.sbt.ZTestFramework"),
+    libraryDependencies ++= Seq(
+      `zio-test`,
+      `zio-test-sbt`,
+    ),
+  )
   .dependsOn(zioHttp)
+  .dependsOn(zioHttpTestkit)
```
guizmaii: We probably only want this dependency in the tests, no?

```suggestion
  .dependsOn(zioHttpTestkit % Test)
```

jdegoes:
@pablf There is one more conflict then I think this is good to merge. A test-kit for CLI would be nice!

pablf:
@jdegoes Conflicts resolved. ~~There is an error with `- Server - WebsocketSpec - Multiple websocket upgrades` that already appeared in the last commits. All the tests of the CLI are passing, though.~~

jdegoes (approved):

Issue #1520, opened 2022-09-19
Implement code generator that takes OpenAPI and generates zio.http.api.EndpointSpec code


jdegoes:
**Is your feature request related to a problem? Please describe.**

Some companies begin with OpenAPI, and then proceed to implement a server that satisfies the OpenAPI specification; then they will modify the OpenAPI specification when a change is necessary, and then re-implement the server.

**Describe the solution you'd like**

In order to support this _spec-first_ driven approach to API development (rather than _code-first_, which is the paradigm already implemented within the `zio.http.api` package), we will need a code generator, which can read in OpenAPI specification, and generate the corresponding `zio.http.EndpointSpec ` values that declaratively describe the APIs.

Then a user can implement handlers based on the code-generated `EndpointSpec ` values.

987Nabil:
I am on this

johannegger:
Hi! 
is there progress on this? I'd love to try it out!

sideeffffect:
There's been some work on that here: https://github.com/guardrail-dev/guardrail/pull/1615

987Nabil:
I am still on it. The goal is, that it will be part for the first non RC release. If it might make it into an RC release before that, I am not sure about.

kitlangton:
I'm gonna give this a shot :D If it proves to be too hairy, I'll relinquish the claim as soon as I realize that.

Update: Nabil is working on this :) I've switched over to the CI Benchmarking issue.

kitlangton:
I don't know how to unclaim—but I [unfortunately don't have time to finish this](https://github.com/zio/zio-http/pull/2357#issuecomment-1662342217)

xonx4l:
@jdegoes can you assign it to me??

987Nabil:
@xonx4l Sorry, I forgot to mention this here, but I am already on this, based on my work of the other direction. See #2470 

/attempt #1520

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clhf0v5nk000ml60f2a4wr7sl/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

sankalp142002:
/attempt #1520


<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clhf0v5nk000ml60f2a4wr7sl/cancel-attempt">
        Cancel my attempt
[...]

PR #2357, opened 2023-07-30
OpenAPI -> Endpoint code generation

kitlangton:
Here's a start! I ended up replacing the previous OpenAPI type, as it was using a custom Json encoder typeclass. It was easier to start again using `zio.json` and `zio.json.yaml` to encode/decode both formats for free! 🥳 

/claim #1520

### TODO

- [ ] Translate `description`/`summary` parameters to `.??(Doc.p(...))` on various types (Params, Endpoints, etc.)
- [ ] More tests / more examples
- [x] Fix Scala 3 compilation woes


kf-envrs:
Hi with due respect to your hard work, I am little bit afraid that this PR "reinvents" wheel in some way. There is already existing library called [guardrail](https://github.com/guardrail-dev/guardrail) which already has some "history" and generating models for example IMHO will be better there. Did you consider contributing there instead of doing basically everything from scratch? 
Again I don't want to be rude in any way, just that this will basically create yet another OpenAPI generator and as John somewhere said, we need to consolidate libraries. 

Thank you for your amazing work it looks quite promising

kitlangton:
The OpenAPI spec is rather complicated, and I won't have enough time to cover all of the schema types and edge cases—I was hoping to finish in a day or two before vacation, but now I just don't have the time. Good point @kf-envrs, building atop a tool that already deals with these details is probably the way to go. I'm not sure how to relinquish my claim, but feel free to take this from me.

PR #2371, opened 2023-08-06
By default set client idle timeout

erikvanoosten:
Having no idle timeout by default is not the right choice. Most modern environments (e.g. every cloud, every proxy) forget about idle connections fairly quickly. For example, I have worked with proxies that forgot about connections after an idle time of as low as 90 seconds.

Having to find this out in production is not a good developer experience.

In this change we set the default idle time to 50 seconds.

### DIFF
--- a/zio-http/src/main/scala/zio/http/ZClient.scala
+++ b/zio-http/src/main/scala/zio/http/ZClient.scala
@@ -608,7 +608,7 @@ object ZClient {
       localAddress = None,
       addUserAgentHeader = true,
       webSocketConfig = WebSocketConfig.default,
-      idleTimeout = None,
+      idleTimeout = Some(50.seconds),
       connectionTimeout = None,
     )
   }


zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -608,7 +608,7 @@ object ZClient {
       localAddress = None,
       addUserAgentHeader = true,
       webSocketConfig = WebSocketConfig.default,
-      idleTimeout = None,
+      idleTimeout = Some(50.seconds),
```
guizmaii, line 611: Why `50` and not `60`?

erikvanoosten: Just in case a virtual network or proxy decides to set the idle time out to 60 seconds, we need something even lower than that. On the other hand, it should be as high as possible. 50 is a random compromise I picked.

vigoo (approved):

PR #2384, opened 2023-08-13
Better `Endpoint` test coverage

### LINKED ISSUE #2204: Improve test coverage for `Endpoint`
The current test coverage of the `Endpoint` API is not very high. A lot of the overloaded methods are not used in any test.
This means as well, that some use cases the API is build for are fragile, in the sense that it might break while implementing new features. (see #2195)

Even now, the `EndpointSpec`is already quite big. Maybe we can split it into multiple files based on use case and just aggregate the specs in `EndpointSpec`.

ghost:
This is supposed to improve the overall structure of the `Endpoint` tests. It moves out most specs into separate files and aggregates them in the original `EndpointSpec` file and does various clean-ups and adds some new test coverage. I couldn't find many methods to add tests for so this is the test coverage for now.

Closes #2204

/claim #2204

### DIFF
--- a/aliases.sbt
+++ b/aliases.sbt
@@ -18,7 +18,7 @@ onLoadMessage := {
       |
       |Useful sbt tasks:
       |${item("fmt")}: Prepares source files using scalafix and scalafmt.
-      |${item("sFix")}: Fixes sources files using scalafix.
+      |${item("sFix")}: Fixes source files using scalafix.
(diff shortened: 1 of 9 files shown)


jdegoes:
@wooster0 Please resolve conflicts and re-open this pull request against #2388, which also has the same bounty as the original ticket. Thank you!

PR #2386, opened 2023-08-14
Fix stream body delay

### LINKED ISSUE #2284: Response(body.fromStream(ZStream.fromHub(hub)) has a one element delay
**Describe the bug**

A streaming Response is created from a Hub.

A sequence of elements (1,2,3..) is published to the hub.
An element N is only seen after N+1 has been published.

ZIO-HTTP 3.0.0-RC2
ZIO 2.0.15
Scala 3.3.0

**To Reproduce**

Given the code below
Connect to the server ```curl -i http://localhost:8080/sse```

Comparing the console and curl outputs demonstrates the one element delay. 

```scala
mport zio.Console.print
import zio.http.*
import zio.stream.{ZSink, ZStream}
import zio.{Chunk, Hub, Promise, ZIO}

object Simple extends zio.ZIOAppDefault {
  def run = {
    import zio.durationInt
    for {
      hub <- Hub.unbounded[String]

      _ <- ZIO.foreach(Range(1,1000))(i => hub.offer(s"$i\n").delay(3.seconds)).fork

      _ <- ZStream.fromHub(hub).run(ZSink.foreach(print(_))).fork

        ui = Http.collect[Request] { case Method.GET -> Root / "sse" =>
         Response(body = Body.fromStream(ZStream.fromHub(hub)))
      }

      _ <- Server.serve(ui).provide(Server.default)
    } yield ()
  }
}
```

guersam:
/claim #2284

This PR reverts #2156 to fix #2284.

#2156, which is an attempt to solve #2155, does not seem to be a proper solution for the problem. Especially the Scenario 2 described in #2155 requires the full HTTP 1.1 Pipelining implementation, but pipelining has several drawbacks so that it is now superseded by [HTTP 2 Multiplexing](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-http2-17#section-5).

https://developer.mozilla.org/en-US/docs/Web/HTTP/Connection_management_in_HTTP_1.x#http_pipelining
> **Note:** HTTP pipelining is not activated by default in modern browsers:
>
> - Buggy [proxies](https://en.wikipedia.org/wiki/Proxy_server) are still common and these lead to strange and erratic behaviors that Web developers cannot foresee and diagnose easily.
> - Pipelining is complex to implement correctly: the size of the resource being transferred, the effective [RTT](https://en.wikipedia.org/wiki/Round-trip_delay_time) that will be used, as well as the effective bandwidth, have a direct incidence on the improvement provided by the pipeline. Without knowing these, important messages may be delayed behind unimportant ones. The notion of important even evolves during page layout! HTTP pipelining therefore brings a marginal improvement in most cases only.
> - Pipelining is subject to the [HOL](https://en.wikipedia.org/wiki/Head-of-line_blocking) problem.
> 
> For these reasons, pipelining has been superseded by a better algorithm, _multiplexing_, that is used by HTTP/2.

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
+++ b/zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
@@ -29,7 +29,9 @@ import io.netty.channel._
 import io.netty.handler.codec.http.{DefaultHttpContent, LastHttpContent}
 object NettyBodyWriter {
 
-  def writeAndFlush(body: Body, ctx: ChannelHandlerContext)(implicit trace: Trace): Option[Task[Unit]] =
+  def writeAndFlush(body: Body, contentLength: Option[Long], ctx: ChannelHandlerContext)(implicit
+    trace: Trace,
+  ): Option[Task[Unit]] =
     body match {
       case body: ByteBufBody                  =>
         ctx.write(body.byteBuf)
@@ -66,27 +68,44 @@ object NettyBodyWriter {
         None
       case StreamBody(stream, _, _)           =>
         Some(
-          stream.chunks
-            .runFoldZIO(Option.empty[Chunk[Byte]]) {
-              case (Some(previous), current) =>
+          contentLength match {
+            case Some(length) =>
+              stream.chunks
+                .runFoldZIO(length) { (remaining, bytes) =>
+                  remaining - bytes.size match {
+                    case 0L =>
+                      NettyFutureExecutor.executed {
+                        ctx.write(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes.toArray)))
+                        ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT)
+                      }.as(0)
+
+                    case n =>
+                      NettyFutureExecutor.executed {
+                        ctx.writeAndFlush(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes.toArray)))
+                      }.as(n)
+                  }
+                }
+                .flatMap {
+                  case 0L        => ZIO.unit
+                  case remaining =>
+                    val actualLength = length - remaining
+                    ZIO.logWarning(s"Expected Content-Length of $length, but sent $actualLength bytes") *>
+                      NettyFutureExecutor.executed {
+                        ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT)
+                      }
+                }
+
+            case None =>
+              stream.chunks.mapZIO { bytes =>
                 NettyFutureExecutor.executed {
-                  ctx.writeAndFlush(new DefaultHttpContent(Unpooled.wrappedBuffer(previous.toArray)))
-                } *>
-                  ZIO.succeed(Some(current))
-              case (_, current)              =>
-                ZIO.succeed(Some(current))
-            }
-            .flatMap { maybeLastChunk =>
-              // last chunk is handled separately to avoid fiber interrupt before EMPTY_LAST_CONTENT is sent
-              ZIO.attempt(
-                maybeLastChunk.foreach { lastChunk =>
-                  ctx.write(new DefaultHttpContent(Unpooled.wrappedBuffer(lastChunk.toArray)))
-                },
-              ) *>
+                  ctx.writeAndFlush(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes.toArray)))
+                }
+              }.runDrain.zipRight {
                 NettyFutureExecutor.executed {
                   ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT)
                 }
-            },
+              }
+          },
         )
       case ChunkBody(data, _, _)              =>
         ctx.write(Unpooled.wrappedBuffer(data.toArray))
--- a/zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
+++ b/zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
@@ -75,9 +75,10 @@ object NettyBodyWriter {
                   remaining - bytes.size match {
                     case 0L =>
                       NettyFutureExecutor.executed {
+                        // Flushes the last body content and LastHttpContent together to avoid race conditions.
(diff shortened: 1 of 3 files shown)


guersam:
To fix #2155, I'd suggest enforcing chunked encoding without content-length header for stream body like [Armeria does](https://github.com/line/armeria/blob/armeria-1.24.3/core/src/main/java/com/linecorp/armeria/server/ServerHttp1ObjectEncoder.java#L136-L141).

jdegoes:
@guersam Good idea. 👍

guersam:
@jdegoes Do you want me to implement and claim for #2155 in this PR, or is it okay to do it separately?

guersam:
Added another attempt for #2155 by removing Content-Length header for streaming bodies.

jgulotta:
Perhaps I misunderstand what is happening here, but please do not tie `Content-Length` or `Transfer-Encoding: chunked` to whether or not the body content is streamed. I should be able to send and receive a large file with `Content-Length` without buffering the whole thing into memory, for both client and server.

My use case for this at work is essentially proxying content to S3, which requires `Content-Length`, where our files can exceed the memory we wish to allocate to the service.

Right now we have to use non-ZIO components to do this and it would be nice if we could consolidate

guersam:
@jgulotta Thanks for the insight. Considering your use case, it seems the issue arises from the unnecessary writing of `LastHttpContent.EMPTY_LAST_CONTENT`. Let's investigate further.

guersam:
While the spec doesn't require writing an empty buffer, Netty uses `LastHttpMessage` [for state management](https://github.com/netty/netty/blob/netty-4.1.97.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java#L272-L275), and ZIO HTTP also uses it to determine the end of the content.

~~To address both #2284 and #2155 without full-fledged request pipelining, we might consider implementing a request queue per connection or rewriting Netty's [`HttpObjectEncoder`](https://github.com/netty/netty/blob/netty-4.1.97.Final/codec-http/src/main/java/io/netty/handler/codec/http/HttpObjectEncoder.java) and ZIO HTTP's [`AsyncBodyReader`](https://github.com/zio/zio-http/blob/ca6513f7811699a6bcb49d28587cee0e2901bce9/zio-http/src/main/scala/zio/http/netty/AsyncBodyReader.scala#L66) and [`ClientResponseStreamHandler`](https://github.com/zio/zio-http/blob/ca6513f7811699a6bcb49d28587cee0e2901bce9/zio-http/src/main/scala/zio/http/netty/client/ClientResponseStreamHandler.scala#L39) to keep track of the number of the bytes written and compare it to the `Content-Length` to determine its next state, but I'm not sure how it costs yet.~~

~~For now, considering the issues with #2284 affecting web applications using long-running SSE streams, like chat applications, it might be a good idea to just revert #2156.~~

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -66,27 +68,44 @@ object NettyBodyWriter {
...
+                      NettyFutureExecutor.executed {
+                        ctx.writeAndFlush(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes.toArray)))
+                      }.as(n)
+                  }
+                }
+                .flatMap {
+                  case 0L        => ZIO.unit
+                  case remaining =>
+                    val actualLength = length - remaining
+                    ZIO.logWarning(s"Expected Content-Length of $length, but sent $actualLength bytes") *>
```
guersam, line 93: Currently it just log a warning message when the content length doesn't match. Should we abort the stream immediately when the actual stream length exceed the given Content-Length?

guersam (commented):
The new strategy handles stream body differently depending on the existence of `Content-Length` header. When it exists, `NettyBodyWriter` accumulates the number of bytes written to find the last chunk.

zio-http/src/main/scala/zio/http/netty/client/ClientInboundHandler.scala
```diff
@@ -55,7 +55,7 @@ final class ClientInboundHandler(
         ctx.writeAndFlush(fullRequest)
       case _: HttpRequest               =>
         ctx.write(jReq)
-        NettyBodyWriter.writeAndFlush(req.body, ctx).foreach { effect =>
+        NettyBodyWriter.writeAndFlush(req.body, None, ctx).foreach { effect =>
```
jgulotta, line 58: slightly tangential to the purpose of this PR, but it would be nice if the client side mirrored the server side with respect to content-length. specifically, if the request encoding path did not adjust the content-length and transfer-encoding headers so that the content-length header could be checked here and streaming bodies sent the same way for both client and server

zio-http/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -66,27 +68,45 @@ object NettyBodyWriter {
...
+          contentLength match {
+            case Some(length) =>
+              stream.chunks
+                .runFoldZIO(length) { (remaining, bytes) =>
+                  remaining - bytes.size match {
+                    case 0L =>
+                      NettyFutureExecutor.executed {
+                        // Flushes the last body content and LastHttpContent together to avoid race conditions.
+                        ctx.write(new DefaultHttpContent(Unpooled.wrappedBuffer(bytes.toArray)))
+                        ctx.writeAndFlush(LastHttpContent.EMPTY_LAST_CONTENT)
```
jgulotta, line 80: it is possible, if exceedingly unlikely, for an incorrect content-length header to be a multiple of the chunk size. this would result in writing `EMPTY_LAST_CONTENT` and then continuing to write more data as the `runFoldZIO` continues

guersam: Yes it is. It's a kind of workaround as the size of the whole stream is indeterministic. That's why I didn't apply the same for the client side yet.

guersam:
What's the expected behavior when the actual body size exceeds the content-length header?

1. Sends the entire body with a warning log. (current)
2. Discard the rest of the body after reaching the content-length.
3. Throw an exception and halt the connection.

jdegoes:
@guersam I think (1) is probably the best bet under the circumstances. But not 100% sure.

Btw please re-open this PR when you are ready for a review!

PR #2390, opened 2023-08-18
lax environment requirement when transforming a Handler

myazinn:
`Env1 <: Env` when transforming Routes / Handler seems too restrictive. Perhaps I'm missing something, but I see no reason why transformation can't change / eliminate it completely.


myazinn:
Failing build seems unrelated to my changes, I see it happens in other PRs too

jdegoes (approved):

Issue #2373, opened 2023-08-08
Handle client idle timeout to prevent abundant logging


erikvanoosten:
Applications that use the http client keep seeing logging every few minutes (using version 3.0.0-RC2+45-e3636f00-SNAPSHOT):
```
logger: "io.netty.channel.DefaultChannelPipeline"
message: "An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception."
stack_trace: "io.netty.handler.timeout.ReadTimeoutException: null"
```
They always comes in groups of 10 with just a few ms between them. There is never a request going on at that time. According to @vigoo the groups of 10 means that all connections in the connection pool time out at the same time (it has a default size of 10).

I would prefer not to have these warnings logged.

Of course, these warning could be suppressed using the logging framework, but handling this in zio-http gives a better developer experience for all users.

@vigoo suggests to handle the timeout so it does not get logged. Probably in the client inbound handler.

See [Discord discussion](https://discord.com/channels/629491597070827530/819703129267372113/1138129262973562880).

Mayhul-Jindal:
/attempt #2373

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cllewxyx7000qmk0fmyt0tqv3/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

joshuallopez:
/attempt #2373

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cllewxyx7000qmk0fmyt0tqv3/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

an0nym3sh:
/attempt #2373

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cllewxyx7000qmk0fmyt0tqv3/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

sanjulaonline:
/attempt #2373

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cllewxyx7000qmk0fmyt0tqv3/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
[...]

PR #2394, opened 2023-08-21
#2373

joshuallopez:
Whenever a ReadTimeoutException is thrown, the specific handling code within the case match will be executed. For now, I've left it to do nothing (i.e., it'll just swallow the exception). If you'd like to handle it in some other manner, like logging it at a different severity or executing some custom logic, replace the comment with the appropriate code. #2373


jdegoes:
@joshuallopez I am not really sure if this is the correct response ("do nothing" on read timeout). However, I am good merging if you can update, fix conflicts, and ensure the build is passing. Thank you!

Issue #2311, opened 2023-07-23
QueryCodec::optional should not recover from validation errors


senia-psm:
`queryCodec.optional` succeeds not only on absent query parameter, but also on any validation errors produced by `transformOrFailLeft`.

```scala
  val a: QueryCodec[String] = query("a").transformOrFailLeft[String](s => Left("error"), x => x)

  val test: Endpoint[Option[String], Nothing, String, EndpointMiddleware.None] =
    Endpoint
      .get("test")
      .query(a.optional)
      .out[String]
```

See minimal reproducer here: https://github.com/senia-psm/zio-http-2311

**Expected behaviour**
Endpoint `test` should fail if query parameter `a` is provided.

**Actual behaviour**
Endpoint `test`  never fails.

`zio-http` version - `3.0.0-RC2`
`scala` version - `3.3.0`

senia-psm:
It would be nice to have a method like this `def recover(f: HttpCodecError => Either[String, T])`

reminia:
/attempt #2311 can I have a try on this issue?

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clkl4r1sk0002l20f1yeqz747/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

reminia:
Looks like it's expected behavior from the code. No matter whether transformation on the arg is Left or Right, since it's optional, the transformation result doesn't matter.  @jdegoes  can u help verify whether it's by desgin?

https://github.com/zio/zio-http/blob/4123b171724add70d5c75afbd2ef13385291d362/zio-http/src/main/scala/zio/http/codec/HttpCodec.scala#L276

adamgfraser:
@reminia `optional` should only recover in cases where the value does not exist, not in cases where the value exists but just could not be properly parsed.

reminia:
@adamgfraser per my understanding, if this is the case, the user should write something like `query("a").optional. transformOrFailLeft` but not swapping the two, as optional at the last means arg is not important from the design of the endpoint, why not just give out the correct response?

adamgfraser:
@reminia Because it is the wrong semantics of `optional`. `optional` says "this value may or may not be there, if it's not that's fine". But it results in extremely confusing behavior if the client calls the server providing this information but mistakenly encodes it incorrectly and then the server just says `None` instead of failing with the parsing error.

[...]

PR #2396, opened 2023-08-22
BugFix: QueryCodec

aazam-gh:
claim #2311


jdegoes:
This does not compile and cannot work.

Issue #2297, opened 2023-07-14
Server streams request bodies forever when not completed normally


jgulotta:
**Describe the bug**
When a request is not completed successfully, e.g. the connection drops, the streaming body does not terminate

**To Reproduce**
1. Run this app

```scala
import zio.*
import zio.http.*

object StreamForever extends ZIOAppDefault {
  def run = {
    val forever = Http.collectZIO[Request] {
      case req @ Method.POST -> Root / "forever" =>
        req.body.asStream.runForeachChunk(c => ZIO.logInfo(s"read chunk size ${c.size}"))
          .foldCauseZIO(
            c => ZIO.logWarningCause("test stream failed", c).as(Response.status(Status.InternalServerError)),
            _ => ZIO.logInfo("test stream done").as(Response.status(Status.NoContent))
          )
    }

    val server = Server.defaultWith(_.enableRequestStreaming)

    Server.serve(forever).provide(server)
  }
}
```

2. Stream any large request, e.g. using HTTPie

```bash
yes hello | http --chunked POST localhost:8080/forever
```

The output `read chunk size 5` will print continually

3. Terminate the stream before it finishes, e.g. `Ctrl+C`

The output `read chunk size 5` will still print continually, until the buffer is cleared, and then nothing.

**Expected behaviour**
The stream from `req.body.asStream` interrupts or dies, printing the final messages in `foldCauseZIO`

Curiously, I have to make one request, cancel it, and then make a second to actually get any output at all. I also expect the initial request would trigger the streaming.

FVelasquezM:
Hi, I've done some debugging and I think I've located the place where this behavior is being caused. I'll try to fix it /attempt #2297

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clkl4xq12000ejx0fwxq8bkh5/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

FVelasquezM:
I've been working on this issue for the last couple of days, but unfortunately I've been unable to solve it, I'll leave some of my findings here so, hopefully, someone more knowledgeable can pick it up from there and solve this issue:

[...]

PR #2402, opened 2023-08-25
invoke NettyBody callback in a separate fiber to avoid deadlocks

myazinn:
When reading a request from a super-fast producer, it is possible that `zio.http.netty.AsyncBodyReader` accumulates more data in buffer than can fit into Stream buffer in `zio.http.netty.NettyBody.AsyncBody#asStream`. In that case, the request will hang because we can't consume messages until callback pushes all the data, which it can't do until we start consuming messages.
Minimized example of an issue can be found here
https://scastie.scala-lang.org/sZ8DrRKxR2qAKxrAOV52hg
For a reproduction step on a full server, you can use steps from this ticket
https://github.com/zio/zio-http/issues/2297
Sometimes server is able to read data from `yes hello | http --chunked POST localhost:8080/forever`, sometimes it's just silent.
I did try to write a test for, but couldn't find a way to reproduce it in tests. So I'd really appreciate if someone gave me a hint on how to do it.  Also, callbacks sometimes blow my mind, so I hope nothing goes wrong if we move callback invocation to a separate fiber 🤔

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/NettyBody.scala
+++ b/zio-http/src/main/scala/zio/http/netty/NettyBody.scala
@@ -25,11 +25,11 @@ import zio.stream.ZStream
 
 import zio.http.Body.{UnsafeBytes, UnsafeWriteable}
 import zio.http.internal.BodyEncoding
-import zio.http.{Body, Boundary, Header, Headers, MediaType}
+import zio.http.{Body, Boundary, Header, MediaType}
 
 import io.netty.buffer.{ByteBuf, ByteBufUtil}
-import io.netty.channel.{Channel => JChannel}
 import io.netty.util.AsciiString
+
 object NettyBody extends BodyEncoding {
 
   /**
@@ -123,21 +123,24 @@ object NettyBody extends BodyEncoding {
 
     override def asStream(implicit trace: Trace): ZStream[Any, Throwable, Byte] =
       ZStream
-        .async[Any, Throwable, Byte](emit =>
-          try {
-            unsafeAsync(new UnsafeAsync {
-              override def apply(message: Chunk[Byte], isLast: Boolean): Unit = {
-                emit(ZIO.succeed(message))
-                if (isLast) {
-                  emit(ZIO.fail(None))
+        .asyncZIO[Any, Throwable, Byte](emit =>
+          ZIO.attempt {
+            unsafeAsync {
+              new UnsafeAsync {
+                override def apply(message: Chunk[Byte], isLast: Boolean): Unit = {
+                  emit.chunk(message)
+                  if (isLast) {
+                    emit.end
+                  }
                 }
+
+                override def fail(cause: Throwable): Unit =
+                  emit.fail(cause)
               }
-              override def fail(cause: Throwable): Unit                       =
-                emit(ZIO.fail(Some(cause)))
-            })
-          } catch {
-            case e: Throwable => emit(ZIO.fail(Option(e)))
-          },
+            }
+          }.catchAll { e =>
+            ZIO.succeed(emit.fail(e))
+          }.fork,
         )
 
     override def isComplete: Boolean = false


lackhoa:
@myazinn Does this fix #2297 ?

myazinn:
@lackhoa nope 🙁

jdegoes (approved):

jdegoes:
Rather than doing ZIO stuff inside a separate fiber, we should consider another mechanism that could be higher-performance. Will take a look at the code and suggest something.

myazinn:
@jdegoes another easy option would be to break abstraction a little bit and handle it in `zio.http.netty.AsyncBodyReader#connect`. Instead of submitting chunks one by one
```scala
          buffer.result().foreach { case (chunk, isLast) =>
            callback(chunk, isLast)
          }
```
We can merge these chunks in place and submit one big chunk
```scala
          val (buffered, isLast) =
            buffer
              .result()
              .foldLeft((Chunk.empty[Byte], false)) { case ((acc, _), (chunk, isLast)) =>
                (acc ++ chunk, isLast)
              }
          callback(buffered, isLast)
```
I haven't done any benchmarks yet, but I guess it should be faster than both current version and one proposed in the MR.
Alternatively, we can go even further and just build one large chunk in a builder. E.g. replace
```scala
private val buffer: ChunkBuilder[(Chunk[Byte], Boolean)] = ChunkBuilder.make[(Chunk[Byte], Boolean)]()

...

          buffer.result().foreach { case (chunk, isLast) =>
            callback(chunk, isLast)
          }
```
with
```scala
  private val buffer: ChunkBuilder[Byte] = ChunkBuilder.make[Byte]()
  private var isLast: Boolean            = false
  
  ...
           callback(buffer.result(), isLast)
```
What do you think?

myazinn:
Looks like `main` branch doesn't compile now 🤔

vigoo:
> @jdegoes another easy option would be to break abstraction a little bit and handle it in `zio.http.netty.AsyncBodyReader#connect`. Instead of submitting chunks one by one
> 
> ```scala
>           buffer.result().foreach { case (chunk, isLast) =>
>             callback(chunk, isLast)
>           }
> ```
> 
> We can merge these chunks in place and submit one big chunk
> 
> ```scala
>           val (buffered, isLast) =
>             buffer
>               .result()
>               .foldLeft((Chunk.empty[Byte], false)) { case ((acc, _), (chunk, isLast)) =>
>                 (acc ++ chunk, isLast)
>               }
>           callback(buffered, isLast)
> ```
> 
> I haven't done any benchmarks yet, but I guess it should be faster than both current version and one proposed in the MR. Alternatively, we can go even further and just build one large chunk in a builder. E.g. replace
> 
> ```scala
> private val buffer: ChunkBuilder[(Chunk[Byte], Boolean)] = ChunkBuilder.make[(Chunk[Byte], Boolean)]()
> 
> ...
> 
>           buffer.result().foreach { case (chunk, isLast) =>
>             callback(chunk, isLast)
>           }
> ```
> 
> with
> 
> ```scala
>   private val buffer: ChunkBuilder[Byte] = ChunkBuilder.make[Byte]()
>   private var isLast: Boolean            = false
>   
>   ...
>            callback(buffer.result(), isLast)
> ```
> 
> What do you think?

The purpose of the async body reader is to be able to process the body in a streaming way - when you don't want to buffer it in memory. So changing that would basically disable the streaming feature.

myazinn:
@vigoo not sure how it disables streaming feature. This buffer is used only until we connect some consumer to a channel. Once consumer is connected, this buffer will not be used and chunks received from netty will be passed directly to it.
The only thing that changes is what happens when a consumer connects to a body reader, that has _already_ buffered a lot of data. Currently it'll feed chunk of chunks to a consumer, while I propose to flatten them and pass as one large chunk

vigoo:
Ah ok sorry, misunderstood.

jdegoes:
@myazinn Let's do this in the most performant way possible, please! 🙏

jdegoes:
@myazinn Master is fixed, btw. 👍

987Nabil:
@myazinn could you update your branch?

jdegoes:
@kyri-petrou Do you have any thoughts on how to do this better?

kyri-petrou:
> @kyri-petrou Do you have any thoughts on how to do this better?

@jdegoes I've given this some thought; I think the main issue here is that we're using a bounded queue in `ZStream.async`. To be honest I don't think using a bounded queue makes much sense because in the case where we have an ultra-fast producer and a slow consumer, the messages will still end up piling in Netty's queues, so backpressure doesn't really help with anything (unless we were to limit netty's queue sizes, which we're not doing).

AFAICT there are 2 possible solutions:
1. Limit Netty's inbound queue size to the same size as the ZStream buffer size. Frankly, I'm not even sure how would this behave in cases the queue becomes full.
2. Use an unbounded queue as the source of the ZStream. I think this makes the most sense in this situation. In case we do want to add backpressure at some point, it can be done on Netty's side

Let me know if you're happy with (2) and I'll start working on it; should be relatively easy to implement

PR #2404, opened 2023-08-26
Fixed #2202 doc interpolation

### LINKED ISSUE #2202: `doc` string interpolation for `Endpoint` documentation
This would be purely to support syntax highlighting for the interpolator in IntelliJ.
There are some pre configured. I guess we could open a PR in the IntelliJ to support this.

pranjalg13:
Added A Doc Interpolation based on understanding

/claim #2202


jdegoes:
This does not compile and cannot work.

PR #2419, opened 2023-09-01
simplify query param access

TomTriple:
Simplifies retrieving the **first** query parameter value with `queryParams.get("id")` or `queryParams.getOrElse("id", "")` . 

If one wants to operate on **all** query parameter values (like before this commit) one should instead of `get` / `getOrElse` now use `queryParams.getAll("id")` / `queryParams.getAllOrElse("id", Nil)`.


987Nabil:
I like the idea, but I would suggest, that get should look like this `def get[T: TextCodec](name: String):T`. Or maybe `getAs` that offers to get typed data based on text codec.

jdegoes (approved):

PR #2443, opened 2023-09-12
chore: support mtls settings

### LINKED ISSUE #2211: Client proxy mTLS support
I need to send HTTP traffic though a proxy that requires mutual TLS authentication. There currently is no way to provide the required certificates to the client to do this.

rajcspsg:
closes https://github.com/zio/zio-http/issues/2211

### DIFF
--- a/zio-http/src/main/scala/zio/http/SSLConfig.scala
+++ b/zio-http/src/main/scala/zio/http/SSLConfig.scala
@@ -17,16 +17,32 @@
 package zio.http
 
 import zio.Config
-import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import zio.http.SSLConfig._
 
-final case class SSLConfig(behaviour: HttpBehaviour, data: Data, provider: Provider)
+sealed trait ClientAuth
+
+object ClientAuth {
+  case object Required       extends ClientAuth
+  case object NoneClientAuth extends ClientAuth
+  case object Optional       extends ClientAuth
+
+}
+
+final case class SSLConfig(
+  behaviour: HttpBehaviour,
+  data: Data,
+  provider: Provider,
+  clientAuth: Option[ClientAuth] = None,
+)
 
 object SSLConfig {
 
   def apply(data: Data): SSLConfig =
-    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK)
+    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK, None)
+
+  def apply(data: Data, clientAuth: ClientAuth): SSLConfig =
+    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK, Some(clientAuth))
 
   val config: Config[SSLConfig] =
     (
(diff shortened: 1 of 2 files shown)


rajcspsg:
@adamgfraser still I need to write test. could you have quick glance whether the approach is correct.

jdegoes (approved):

jdegoes:
@danestig Can you review and ensure this meets your needs?

danestig:
Hello! Thanks for taking a stab at this @rajcspsg.

I have 2 concerns: 
1. How do I provide both client keytstore/key/certificate and a truststore/ca-certificate? Looking at the `ClientSSLConfig` they look mutually exclusive to me. 
2. Since I could not figure out 1 I have not tried this yet. Will mTLS be used to connect to both the proxy and the server, or only one of them? #2211 Is specifically about authenticating with mTLS with the proxy, I expected that to be configured in the proxy setting.

rajcspsg:
@danestig Sorry I missed your comment. I did mistake I updating the ClientSSLConfig config instead of ServerSSLConfig. 
I going to configure only the proxy server verifies the client certificates, does that work for you?

danestig:
@rajcspsg Yes, I believe that would solve my issue. Thank you.

zio-http/src/main/scala/zio/http/ClientSSLConfig.scala
```diff
@@ -42,9 +52,22 @@ object ClientSSLConfig {
     )
   }
 
-  case object Default                                                                         extends ClientSSLConfig
-  final case class FromCertFile(certPath: String)                                             extends ClientSSLConfig
-  final case class FromCertResource(certPath: String)                                         extends ClientSSLConfig
-  final case class FromTrustStoreResource(trustStorePath: String, trustStorePassword: String) extends ClientSSLConfig
-  final case class FromTrustStoreFile(trustStorePath: String, trustStorePassword: String)     extends ClientSSLConfig
+  sealed trait ClientAuth
+  case object RequireClientAuth  extends ClientAuth
```
jdegoes: @rajcspsg Can you put these in the companion object of `ClientAuth`?

jdegoes: Also, once you put them there, they can have shorter names, e.g.:

```scala
object ClientAuth {
  case object Required extends ClientAuth
  case object None extends ClientAuth
  case object Optional extends ClientAuth
}
```

jdegoes (changes requested):
Just one minor suggestion, then looks good to merge. Thankk you!

zio-http/src/main/scala/zio/http/ClientSSLConfig.scala
rajcspsg: @jdegoes sure. I've to set this config on ServerSslContext not on ClientSSLContext

rajcspsg: I'll update my PR shortly

rajcspsg:
@jdegoes @danestig @adamgfraser I updated the PR to add clientAuth config on SSLServerContext.
Please review again and let me know if you have any questions.

zio-http/src/main/scala/zio/http/SSLConfig.scala
```diff
@@ -17,16 +17,32 @@
...
+)
 
 object SSLConfig {
 
   def apply(data: Data): SSLConfig =
-    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK)
+    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK, None)
+
+  def apply(data: Data, clientAuth: ClientAuth): SSLConfig =
+    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK, Some(clientAuth))
```
rajcspsg, line 46: @jdegoes @adamgfraser  I tried to combine both the apply (and other duplicate functions below)
into single function like 

```scala
def apply(data: Data, clientAuth: Option[ClientAuth] = None): SSLConfig =
    new SSLConfig(HttpBehaviour.Redirect, data, Provider.JDK, clientAuth)
```

But I'm getting below error -

 > /home/raj/Coding/scala/zio-http/zio-http/src/main/scala/zio/http/SSLConfig.scala:39:8: in object SSLConfig, multiple overloaded alternatives of method fromFile define default arguments.


I did workaround to make it compile. Not sure why the first option didn't work

jdegoes (approved):

PR #2450, opened 2023-09-18
add static fileserver

TomTriple:
I have a simple static fileserver as middleware, which provides an API like that: 

```scala
package example

import zio._
import zio.http._

object AppMain extends ZIOAppDefault {

  val staticServeMiddleware = StaticServe.middleware(
      Path.root / "images", 
      StaticServe
        .fromDirectory("/opt/repos/os/aaa/images")
        .orElse(StaticServe.fromDirectory("/opt/repos/os/aaa/assets"))
        .orElse(StaticServe.fromResource)
        .orElse(StaticServe.fromFileZIO(ZIO.succeed(new java.io.File("./fallback.png"))))    
  )

  val appDashboard = new AppDashboard()
  val appLogin = new AppLogin(appDashboard)
  val appRoutes = (
    appLogin.routes ++ appDashboard.routes
  )

  override val run =
    Server.serve(appRoutes.toHttpApp @@ staticServeMiddleware).provide(Server.default)

} 
```
I also added checks to anticipate (and prevent) traversal attacks (same as in akka http): 

- prevent fishy requests with ".." in path segments
- prevent serving files outside of the configured document root directory


zio-http/src/main/scala/zio/http/StaticServe.scala
```diff
@@ -0,0 +1,88 @@
...
+                val path   = request.path.unnest(unnest).addLeadingSlash
+                staticServe.run(path).sandbox
+              }
+            },
+        )
+        routes ++ other
+      }
+    }
+
+  def middleware[R, E](path: Path, staticServe: StaticServe[R, E, Any, Response]): Middleware[R] =
```
jdegoes: This should be `toMiddleware`. 

Also, I think there is value in **only** having a public `Middleware.staticServer` interface, and hiding all these types. Thoughts?

TomTriple: Yes, I think it's a good idea to stay consistent with `Middleware` in general. 

If I see that correctly we don´t need anything composable for the user when configuring a static server. It could be enough to only support the two main use-cases:

1. publish a given directory at a specified path: `Middleware.staticServer.fromDirectory(atPath:Path, docRoot: File):Middleware`

2. publish resources at a specified path: `Middleware.staticServer.fromResources(atPath:Path):Middleware`

Is that what you mean? If yes, I think we should do it that way.

TomTriple: One drawback would be that we could not add caching of static assets transparently, no (assuming we want to do that)? With `StaticServe` we could add a new operator `cached:StaticServe` that can cache results. Maybe it's best to go with 1. and 2. from above and wait for feedback...

jdegoes: I would go for:

```scala
object Middleware {
  ...
  def serveDirectory(path: Path, docRoot: File, cached: Boolean = false): Middleware
}
```

Then it has minimal impact on the surface area, and all of these helper classes / functions can be made private inside the companion object of `Middleware`.

TomTriple: I discarded adding an option for `cache` as that should be more carefully thought through - if I find time to do that we can still add that option without any problems. Besides that I changed the API according to your suggestion!

jdegoes (approved):

jdegoes:
One more suggestion as a followup PR: add this as either an HttpApp or as a Route (e.g. `Route.serveDirectory(...)`). Since not everyone might look for it inside middleware.

Issue #2399, opened 2023-08-23
Client hangs when attempting reading response body out of Scope


myazinn:
**Describe the bug**
When HTTP client reads a response out of response Scope, it can't receive any new data from upstream due to a closed connection. However, the data stream itself is not closed and just waits for a new data, causing consumption to hang indefinitely.

**To Reproduce**
Steps to reproduce the behaviour:
1) Run simple HTTP server with infinite streaming response
```scala
package example

import zio.http._
import zio.stream.ZStream
import zio._

object StreamingResponse extends ZIOAppDefault {
  def run: Task[Nothing]        = Server.serve(app).provide(Server.default)
  private def app: HttpApp[Any] = Routes(
    Method.GET / "stream" ->
      handler(
        Response(
          status = Status.Ok,
          body = Body.fromStream {
            ZStream
              .iterate(1)(_ + 1)
              .tap { i =>
                ZIO.when(i % 10000 == 0)(ZIO.debug("sent 10k numbers"))
              }
              .map(_.toString)
          },
        ),
      ),
  ).toHttpApp
}

```
2) Run a client, that will read data from server
```scala
package example

import zio._
import zio.http._

import java.util.concurrent.atomic.AtomicInteger

object SimpleClient extends ZIOAppDefault {
  def run: RIO[Scope, Any]               = program.provideSome[Scope](Client.default)
  def program: RIO[Scope & Client, Unit] = for {
    client <- ZIO.service[Client]
    request = client.request(Request.get("http://localhost:8080/stream"))
   //    res <- request // this one works as expected. It consumes data and prints it to console infinitely
    res <- ZIO.scoped(request)
    counter = new AtomicInteger(0)
    _ <- res.body.asStream.runForeach { _ => // this line should've fail with an error, but it does nothing
      ZIO.when(counter.incrementAndGet() % 10000 == 0)(ZIO.debug("Received 10k bytes"))
    }
  } yield ()

}
```

**Expected behaviour**
Since I'm reading data out of scope, I'd expect stream to fail with any meaningful error. 

**Additional context**
I get that no one should ever do it. However, it might be a common mistake if a person doesn't know what Scope is doing in this particular case / is new to ZIO and doesn't know how to work with Scopes. Handing indefinitely is an unexpected result, _especially_ since it might actually work sometimes because all data was read into internal buffer before the Scope was closed, as shown [here](https://scastie.scala-lang.org/XDhkGSnwSiGzrKWNGV88pA).
[...]

PR #2454, opened 2023-09-22
Throw exception when reading from AsyncBody after the channel closed (#2399)

lackhoa:
Currently, when we call `Response.body.asStream`, we don't check to see if the corresponding response channel is open or not. If the channel has been closed and the user attempts to read the response body, it's possible for the body stream to hang (i.e never finish or throw an error), leading to the code hanging as illustrated in #2399. This behavior is undesirable since the user isn't aware of the channel closing (because we take care of that behind the scene), leading to the issue being hard to debug.

This commit adds a check to `AsyncBodyReader.connect` so that we will throw an exception when the user attempts to read an unfinished response body from a closed channel.

PR note: I'm new to the code base, so let me know if throwing an exception is bad style (I'm doing it because there's already an exception in that same function :)

/claim #2399

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/AsyncBodyReader.scala
+++ b/zio-http/src/main/scala/zio/http/netty/AsyncBodyReader.scala
@@ -38,11 +38,22 @@ abstract class AsyncBodyReader(implicit trace: Trace) extends SimpleChannelInbou
     this.synchronized {
       state match {
         case State.Buffering =>
-          state = State.Direct(callback)
-          buffer.result().foreach { case (chunk, isLast) =>
-            callback(chunk, isLast)
+          val result: Chunk[(Chunk[Byte], Boolean)] = buffer.result()
+          val readingDone: Boolean                  = result.lastOption match {
+            case None              => false
+            case Some((_, isLast)) => isLast
           }
-          ctx.read()
+
+          if (ctx.channel.isOpen || readingDone) {
+            state = State.Direct(callback)
+            result.foreach { case (chunk, isLast) =>
+              callback(chunk, isLast)
+            }
+            ctx.read()
+          } else {
+            throw new IllegalStateException("Attempting to read from a closed channel, which will never finish")
+          }
+
         case State.Direct(_) =>
           throw new IllegalStateException("Cannot connect twice")
       }
--- a/zio-http/src/main/scala/zio/http/netty/AsyncBodyReader.scala
+++ b/zio-http/src/main/scala/zio/http/netty/AsyncBodyReader.scala
@@ -16,6 +16,8 @@
 
 package zio.http.netty
 
+import java.io.IOException
+
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 import zio.{Chunk, ChunkBuilder, Trace, Unsafe}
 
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/netty/AsyncBodyReader.scala
```diff
@@ -38,11 +38,22 @@ abstract class AsyncBodyReader(implicit trace: Trace) extends SimpleChannelInbou
...
-          buffer.result().foreach { case (chunk, isLast) =>
-            callback(chunk, isLast)
+          val result: Chunk[(Chunk[Byte], Boolean)] = buffer.result()
+          val readingDone: Boolean                  = result.lastOption match {
+            case None              => false
+            case Some((_, isLast)) => isLast
           }
-          ctx.read()
+
+          if (ctx.channel.isOpen || readingDone) {
```
myazinn, line 49: It looks like a feature, but just want to highlight that it'll lead to cases when we can read from a closed channel. It is possible that reading will be done before we try to connect to a body, thus leading to confusing flaky behaviour when request sometimes succeeds, sometimes doesn't.
E.g.
```scala
package example

import zio._
import zio.http._

object SimpleClient extends ZIOAppDefault {
  def run: RIO[Scope, Any]               = program.provideSome[Scope](Client.default)
  def program: RIO[Scope & Client, Unit] = for {
    client <- ZIO.service[Client]
    request = client.request(Request.get("https://jsonplaceholder.typicode.com/todos/1"))
    res <- ZIO.scoped(request)
    _   <- res.body.asString.debug
  } yield ()
}
```
It should fail IMO, but on my machine it succeeds most of the time.
BTW latest version of bio-http sometimes fails with mysterious 
```
timestamp=2023-09-23T14:16:20.575404Z level=ERROR thread=#zio-fiber-1 message="" cause="Exception in thread "zio-fiber-19,4" io.netty.channel.AbstractChannel$AnnotatedNoRouteToHostException: No route to host: jsonplaceholder.typicode.com/[2606:4700:e4:0:0:0:ac40:a206]:443
```
Which is not related to your changes (looks like another bug, lol)

lackhoa: I'm aware of the flakiness you suggested. But the check I added there is not a feature so much as a requirement. Like, is it *required* to keep the channel open while we read from the buffer? (ie Is it the case that in the future, our code will never close the channel before `AsyncBodyReader.connect` is called?)

If that's the case, I'm all for failing whenever the channel is closed. If not, then the check is required.

And I agree that your example should fail, but a fix for that can be proposed in another PR.
(For future context, the mentioned flakiness refers to *broken* code, not *correct* ones)

jdegoes:
@lackhoa Thanks for your work on this! Please add a regression test based on the original issue to ensure the issue is succcessfully fixed and won't regress in the future. 🙏

lackhoa:
> @lackhoa Thanks for your work on this! Please add a regression test based on the original issue to ensure the issue is succcessfully fixed and won't regress in the future. 🙏

It hath been done!
Edit: Ok I spoke too soon, CI was angry at me?

lackhoa:
@jdegoes Ok test issue is fixed now, turned out there was another bug lurking in the code.

adamgfraser (changes requested):
I don't think this works. We are checking whether the channel is open immediately before reading from it, but there is a concurrency problem here. The channel may be open when we check it but then closed before we read from it.

lackhoa:
@adamgfraser I’ve added a channelInactive to AsyncBodyReader as well. I don’t remember exactly, but that should cover all the cases.

adamgfraser (approved):

PR #2460, opened 2023-09-26
fix: case-insensitive FormField headers

jgranstrom:
Some clients send FormField headers in lower case. However we seem to require them to be capitalized. I can't find a clear spec for headers in this context, but from headers in general we can probably expect it to vary. One example where we're incompatible is [react-native](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/Network/FormData.js#L77).

### DIFF
--- a/zio-http/src/main/scala/zio/http/FormField.scala
+++ b/zio-http/src/main/scala/zio/http/FormField.scala
@@ -25,9 +25,6 @@ import zio.stream.{Take, ZPipeline, ZStream}
 import zio.http.FormDecodingError._
 import zio.http.Header.ContentTransferEncoding
 import zio.http.internal.FormAST
-import zio.stream.{Take, ZPipeline, ZStream}
-
-import java.nio.charset._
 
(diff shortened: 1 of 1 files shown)


zio-http/src/main/scala/zio/http/FormField.scala
```diff
@@ -155,15 +155,15 @@ object FormField {
           Chunk.empty[FormAST.Content],
         ),
       ) {
-        case (accum, header: FormAST.Header) if header.name == "Content-Disposition"       =>
+        case (accum, header: FormAST.Header) if header.name.toLowerCase == "content-disposition"       =>
```
ccerbusca: I would recommend using the `equalsIgnoreCase` method on String instead. Same for the other cases
```suggestion
        case (accum, header: FormAST.Header) if header.name.equalsIgnoreCase("Content-Disposition")        =>
```

jdegoes (approved):

Issue #2280, opened 2023-06-29
NoRouteToHostException when making a request to https url from a zio-http Client using Scala 3


lachezar:
I am getting weird error when making a request towards any https website when executing unit test (`sbt test`), however I don't get the same error when running my application (`sbt run`). I use zio 2.0.15, scala 3.3.0, zio-http 3.0.0-rc2, sbt 1.9.0 and Java 11.0.19.

Here is the code that you can use to reproduce the issue:
```
import zio.*
import zio.http.*
import zio.test.*

object ATest extends ZIOSpecDefault:

  val url = URL.decode("https://google.com/").toOption.get

  def spec = suite("ATest")(
    test("test") {
      ZIO
        .serviceWithZIO[Client](_.request(Request.get(url)))
        .map(res => assert(res)(Assertion.anything))
    }
  ).provideShared(Client.default)
```

The error I get is 
```
Exception in thread "zio-fiber-113,110" io.netty.channel.AbstractChannel$AnnotatedNoRouteToHostException: null: google.com/2a00:1450:400f:802:0:0:0:200e:443
    	at ATest.spec(ATest.scala:14)
    	at ATest.spec(ATest.scala:15)
    	at ATest.spec(ATest.scala:16)
    	Suppressed: java.net.NoRouteToHostException: null
    		at io.netty.channel.unix.Errors.newConnectException0(Errors.java:158)
    		at io.netty.channel.unix.Errors.handleConnectErrno(Errors.java:131)
    		at io.netty.channel.unix.Socket.connect(Socket.java:351)
    		at io.netty.channel.kqueue.AbstractKQueueChannel.doConnect0(AbstractKQueueChannel.java:727)
    		at io.netty.channel.kqueue.KQueueSocketChannel.doConnect0(KQueueSocketChannel.java:102)
    		at io.netty.channel.kqueue.AbstractKQueueChannel.doConnect(AbstractKQueueChannel.java:712)
    		at io.netty.channel.kqueue.AbstractKQueueChannel$AbstractKQueueUnsafe.connect(AbstractKQueueChannel.java:558)
    		at io.netty.channel.DefaultChannelPipeline$HeadContext.connect(DefaultChannelPipeline.java:1342)
    		at io.netty.channel.AbstractChannelHandlerContext.invokeConnect(AbstractChannelHandlerContext.java:653)
```

[...]

PR #2464, opened 2023-09-30
fix: Avoid IPv6 in system DNS queries by default (#2280)

lackhoa:
Currently, there are cases where users cannot use IPv6 due to their network
infrastructure not supporting it. Therefore, we should not resolve hostnames to
IPv6 addresses unless told otherwise.

This commit defaults DNS queries to IPv4-only. It also allows users to resolve
IPv6 addresses by setting the system property `java.net.preferIPv6Addresses`.

/claim https://github.com/zio/zio-http/issues/2280

Reference: https://docs.oracle.com/javase/8/docs/technotes/guides/net/ipv6_guide/

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -3,7 +3,7 @@ import sbt.Keys.scalaVersion
 
 object Dependencies {
   val JwtCoreVersion                = "9.1.1"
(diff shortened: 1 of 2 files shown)


lackhoa:
I wanna change the branch name, so closing this (will open a new one).

PR #2465, opened 2023-09-30
Adding Support for chrome extension 'scheme'

leandrocruz:
ZIO Http fails to handle requests from chrome extensions, when the origin header is chrome-extension://EXTENSION_ID.


zio-http/src/main/scala/zio/http/Header.scala
```diff
@@ -3039,8 +3039,8 @@ object Header {
       if (value == "null") Right(Null)
       else
         URL.decode(value) match {
-          case Left(_)                                              => Left("Invalid Origin header")
-          case Right(url) if url.host.isEmpty || url.scheme.isEmpty => Left("Invalid Origin header")
+          case Left(_)                                              => Left(s"Invalid Origin header '$value'")
```
leandrocruz, line 3042: Better error message for clarity

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -60,14 +62,16 @@ object Scheme       {
...
     def decode(scheme: String)(implicit unsafe: Unsafe): Scheme = {
       if (scheme == null) null
-      else
-        scheme.length match {
-          case 5 => Scheme.HTTPS
-          case 4 => Scheme.HTTP
-          case 3 => Scheme.WSS
-          case 2 => Scheme.WS
-          case _ => null
+      else {
```
leandrocruz, line 65: Decided for clarity over "speed"

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -21,10 +21,11 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
...
   def encode: String = self match {
-    case Scheme.HTTP  => "http"
-    case Scheme.HTTPS => "https"
-    case Scheme.WS    => "ws"
-    case Scheme.WSS   => "wss"
+    case Scheme.HTTP            => "http"
+    case Scheme.HTTPS           => "https"
+    case Scheme.WS              => "ws"
+    case Scheme.WSS             => "wss"
+    case Scheme.ChromeExtension => "chrome-extension"
```
987Nabil, line 28: I am not sure, that having this explicitly is a good idea. Maybe we should have a `Custom` case, like for http method?
@vigoo @jdegoes

leandrocruz: Not sure if this is the right solution also, but, as it is today, zio http fails to accept requests from chrome extensions

987Nabil:
I think #2489 confirms my hinch. We should have a custom type that stores non predefined schema values.

FabioPinheiro:
I'm just concerned about efficiency.
Is there benchmark that I can run on https://github.com/zio/zio-http/pull/2490?

987Nabil:
What kind of efficiency are we talking about?
I mean we can match against the ones that have a fixed type and return the fixed type. For all others, we would have 2 options I guess. 
* just put the string as is in a custom wrapper
* validate via regex the validity and only if valid put it in the wrapper

So, where do you see an issue?

vigoo:
I agree to have a `Custom(String)` case instead

987Nabil:
@vigoo there is https://github.com/zio/zio-http/pull/2490 which is already further in development.
@leandrocruz I am closing this PR in favor of the other.

PR #2466, opened 2023-10-01
add new flash type and changed flash api

TomTriple:
The `flash scope` now is based on a serialized, url-encoded json object (with `zio-schema`) that is transported in a cookie named `zio-http-flash`. I removed the super simple `flashMessage` operator in favor of a new compositional domain type `zio.http.Flash` that has proper operators for `setting` and `getting` typed values from the flash scope. 

**Setting a value with a Response**: `Response.ok.addFlash(Flash.set("name", "mr. miller") ++ Flash.set("email", Email("hello@domain.com")))`

**Getting a value from a Request**: `req.flash(Flash.get[Email])` or `req.flash(Flash.getString("namePrimary") <> Flash.getString("nameSecondary") <*> Flash.getFloat("priceSum"))`

I think that we could mirror that same API in a type for (cookie-based) session handling.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Flash.scala
+++ b/zio-http/src/main/scala/zio/http/Flash.scala
@@ -1,9 +1,9 @@
 package zio.http
 
+import java.net.{URLDecoder, URLEncoder}
+
 import zio.schema.Schema
 import zio.schema.codec.JsonCodec
-
-import java.net.{URLDecoder, URLEncoder}
 
 /**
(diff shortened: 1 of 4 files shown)


jdegoes (approved):

TomTriple:
I'll reopen when this is more feature-complete - the idea stays the same (thanks for reviewing @jdegoes) but I want to add a few more useful things. I'm actively working on it and it should not take long.

PR #2467, opened 2023-10-01
fix: Avoid IPv6 in system DNS queries by default (#2280)

lackhoa:
Currently, there are cases where users cannot use IPv6 due to their network
infrastructure not supporting it. Therefore, we should not resolve hostnames to
IPv6 addresses unless told otherwise.

This commit defaults DNS queries to IPv4-only. It also allows users to resolve
IPv6 addresses by setting the DnsResolver config.

/claim https://github.com/zio/zio-http/issues/2280

Reference: https://docs.oracle.com/javase/8/docs/technotes/guides/net/ipv6_guide/

### DIFF
--- a/zio-http/src/main/scala/zio/http/DnsResolver.scala
+++ b/zio-http/src/main/scala/zio/http/DnsResolver.scala
@@ -16,7 +16,8 @@
 
 package zio.http
 
-import java.net.{InetAddress, UnknownHostException}
+import java.lang.{System => JSystem}
+import java.net.{Inet6Address, InetAddress, UnknownHostException}
 import java.time.Instant
 
 import zio._
@@ -30,12 +31,19 @@ object DnsResolver {
   def resolve(host: String)(implicit trace: Trace): ZIO[DnsResolver, UnknownHostException, Chunk[InetAddress]] =
     ZIO.serviceWithZIO(_.resolve(host))
 
-  private final case class SystemResolver() extends DnsResolver {
-    override def resolve(host: String)(implicit trace: Trace): ZIO[Any, UnknownHostException, Chunk[InetAddress]] =
+  private final case class SystemResolver(useIPv6: Boolean) extends DnsResolver {
+    override def resolve(host: String)(implicit trace: Trace): ZIO[Any, UnknownHostException, Chunk[InetAddress]] = {
       ZIO
         .attemptBlocking(InetAddress.getAllByName(host))
         .refineToOrDie[UnknownHostException]
-        .map(Chunk.fromArray)
+        .map(result => {
+          val allHosts           = Chunk.fromArray(result)
+          val preferredAddresses =
+            if (useIPv6) allHosts.filter(_.isInstanceOf[Inet6Address])
+            else allHosts.filter(!_.isInstanceOf[Inet6Address])
+          if (preferredAddresses.isEmpty) allHosts else preferredAddresses
+        })
+    }
   }
 
   private[http] final case class CacheEntry(
@@ -233,7 +241,12 @@ object DnsResolver {
     maxConcurrentResolutions: Int,
     expireAction: ExpireAction,
     refreshRate: Duration,
-  )
+    useIPv6: Boolean,
+  ) {
+    self =>
+
+    def useIPv6(value: Boolean) = self.copy(useIPv6 = value)
+  }
 
   object Config {
     lazy val config: zio.Config[Config] =
@@ -242,9 +255,10 @@ object DnsResolver {
         zio.Config.int("max-count").withDefault(Config.default.maxCount) ++
         zio.Config.int("max-concurrent-resolutions").withDefault(Config.default.maxConcurrentResolutions) ++
         ExpireAction.config.nested("expire-action").withDefault(Config.default.expireAction) ++
-        zio.Config.duration("refresh-rate").withDefault(Config.default.refreshRate)).map {
-        case (ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate) =>
-          Config(ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate)
+        zio.Config.duration("refresh-rate").withDefault(Config.default.refreshRate) ++
+        zio.Config.boolean("use-ipv6").withDefault(Config.default.useIPv6)).map {
+        case (ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate, useIPv6) =>
+          Config(ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate, useIPv6)
       }
 
     lazy val default: Config = Config(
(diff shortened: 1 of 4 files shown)


vigoo:
Try `sbt fmt`

lackhoa:
@vigoo `sbt fmt ` did some fix but CI didn't like it either. What a giant waste of time this formatting business ☹️ .
But anw I fixed the error manually.

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -27,15 +28,23 @@ trait DnsResolver {
 }
 
 object DnsResolver {
+  // NOTE: This system property is the only way for users to opt-in to IPv6.
+  // Even if the flag is set, we use both IPv4 and IPv6 and don't "prefer" IPv6
+  // in any way.
+  private final val useIPv6 = JSystem.getProperty("java.net.preferIPv6Addresses") == "true"
```
ccerbusca: can this perhaps be done via zio.Config? Not a fan of the arbitrary system props. It is already used in the same file. See https://github.com/zio/zio-http/blob/94a6fa04746e2a40244a98de1ae889476f85d254/zio-http/src/main/scala/zio/http/DnsResolver.scala#L308

lackhoa: Yeah, at first I thought it'd be a good idea to allow configuration from the command line.
But using ZIO config is more flexible, plus no need to explain the system props.
Done!

ccerbusca (approved):

ex0ns:
I would argue that it should be the other way around, ipv6 should be enabled by default and people that require disabling it should be taking care of that.
More and more infrastructure/provider/ISP do provide/use (because they don't have the choice) IPv6 and IMO every piece of software that can, should support IPv6 by default

lackhoa:
@ex0ns This is my personal circumstance: I live in the second largest city in Finland, one of the richest country in the world. My ISP is the most popular in Finland, and yet they don't support IPv6.
The default must work out of the box for most users, IPv4 works for *all* users, currently the default fails for some users. Therefore this change is for the better.

Browsers do prefer IPv6 by default, but they also have a fallback mechanism to IPv4. If you have a way to detect if IPv6 would work without compromising performance, please tell me about it.

ex0ns:
I understand, but still, I feel like disabling a feature by default is not the best way of going forward. I do understand that I might not be the way everywhere, as you are describing, but from my personal experience, ISP in both France and Switzerland tend to heavily rely on IPv6 because of the availability/price and hide the IPv4 behind an horrible carrier grade nat/translation layer and I would argue that in those cases, the IPv6 is notably faster that using IPv4.

It's a bit of a side talk but my main point here is that I think that we should not hide this feature behind a feature flag that will be forgotten/not used in most of the case where it could. That is only my point of view and someone could defend the exact opposite, I'm only stating that this decision can seem pretty small but they will impact people using the library.

edit:
As a side note, I wonder how InetAddress.getAllByName would behave when using the JVM flags 
```
-Djava.net.preferIPv4Stack=false -Djava.net.preferIPv6Addresses=true
```

lackhoa:
@ex0ns Well I'm not exactly arguing for my use case since I'm not using this library right now: but if you have to weigh the options:
1. Default client breaks for some people
2. Default client works for everyone, but some (most?) people might get a suboptimal experience, which they have an option to fix (I'm not sure what you mean by *hide*, I mean it's the DNS configuration)
3. ... *is there some other option*?

What'd be the answer?

ex0ns:
I agree  with your second point, let the default work for everyone :ok_hand:

lackhoa:
@ex0ns Well you know what, I now learned that ZLayers are shared. Maybe we can just do this test once when creating systemResolver to check for IPv6 interface: https://stackoverflow.com/a/51904361
If have time can you check if that work on your machine? I don’t trust SO and don’t have IPv6 to test.

ex0ns:
Good idea, I try to run the following code (scala 3) on my machine:
```scala
import java.net.NetworkInterface
import java.net.InetAddress
import java.net.Inet6Address
import collection.JavaConverters._

@main
def main =
  val interfaces = 
    NetworkInterface.getNetworkInterfaces().asScala.toList
  val ipv6Support = interfaces.exists: interface =>
    val addresses = interface.getInterfaceAddresses().asScala.toList
    addresses.exists: address =>
      address.getAddress match 
        case add: Inet6Address if !add.isLoopbackAddress => true
        case _ => false
  println(ipv6Support)

```

And it returned the  correct result, would you mind running it on your system ?

lackhoa:
@ex0ns Hm, the main interface on my laptop does have IPv6 address:
```
en0: flags=8863<UP,BROADCAST,SMART,RUNNING,SIMPLEX,MULTICAST> mtu 1500
	options=6460<TSO4,TSO6,CHANNEL_IO,PARTIAL_CSUM,ZEROINVERT_CSUM>
	ether 38:f9:d3:be:61:01
	inet6 fe80::1019:1efe:4f18:3f90%en0 prefixlen 64 secured scopeid 0x6 
	inet 10.192.64.101 netmask 0xffffff00 broadcast 10.192.64.255
	nd6 options=201<PERFORMNUD,DAD>
	media: autoselect
	status: active
```

So unfortunately that didn't really help much :(

ex0ns:
Too bad, thanks for trying anyway !

lackhoa:
Ok I've fixed it so that if you choose to enable IPv6, we'd only resolve to IPv6 addresses. That should improve performance (if IPv6 is indeed faster).

zio-http/src/main/scala/zio/http/DnsResolver.scala
```diff
@@ -242,9 +257,10 @@ object DnsResolver {
         zio.Config.int("max-count").withDefault(Config.default.maxCount) ++
         zio.Config.int("max-concurrent-resolutions").withDefault(Config.default.maxConcurrentResolutions) ++
         ExpireAction.config.nested("expire-action").withDefault(Config.default.expireAction) ++
-        zio.Config.duration("refresh-rate").withDefault(Config.default.refreshRate)).map {
-        case (ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate) =>
-          Config(ttl, unknownHostTtl, maxCount, maxConcurrentResolutions, expireAction, refreshRate)
+        zio.Config.duration("refresh-rate").withDefault(Config.default.refreshRate) ++
+        zio.Config.boolean("use-IPv6").withDefault(Config.default.useIPv6)).map {
```
jdegoes:
```suggestion
        zio.Config.boolean("use-ipv6").withDefault(Config.default.useIPv6)).map {
```

lackhoa: Ok I've included your fix in the commit

jdegoes:
@lackhoa Is this ready for final review? Can you please fix conflicts? 🙏

lackhoa:
> @lackhoa Is this ready for final review? Can you please fix conflicts? 🙏

No actually I wanna close it, because it's not a good way to do things. We should try querying both IPv6 and IPv4, preferring IPv6 if it works. It'd take more work and I don't wanna do it.

PR #2470, opened 2023-10-05
Improve OpenAPI model; Add OpenAPI generator for EndpointAPI (#1498)

### LINKED ISSUE #1498: Generate OpenAPI documentation from EndpointSpec data type

The `zio.http.api.EndpointSpec` contains a declarative description of an API endpoint, while the `ServiceSpec` data type contains a declarative description of more than one API endpoints. 

These declarative descriptions are wholly sufficient to generate extremely good OpenAPI documentation for the endpoints. Yet, this is a feature that is wholly missing in the current version of `zio.http.api`.

Add a method `EndpointSpec#toOpenAPI`, which translates the route, query parameter, header, input and output, into the `zio.http.api.openapi.*` family of data types, incorporating the custom documentation that may be included into the `EndpointSpec` or into one of the `HttpCodec` values that make up its description of the input.

N/A

**Additional context**

This ticket does **not** include translating the OpenAPI values into HTML or JSON, which is considered outside the scope of this ticket. This ticket is merely to translate the `EndpointSpec` data type into typed OpenAPI values.

A separate ticket will cover the generation of JSON (etc.) from the OpenAPI data types.

987Nabil:
Todos before this is no longer draft:

- [x] Add more tests
- [x] More methods to call the OpenAPI generator that allows to hand over more meta infos and config params
- [x] tidy up the code

fixes #1498 
/claim #1498


jdegoes:
@987Nabil Can you fix build failures? 🙏

987Nabil:
Scala 3 build is failing due to a bug in the Schema derivation macro. I'll check this. There is already a known bug that might be related.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -0,0 +1,711 @@
...
+          case SegmentCodec.UUID(_)         => Json.Str(value.asInstanceOf[java.util.UUID].toString)
+          case SegmentCodec.Trailing        =>
+            throw new IllegalArgumentException("Trailing segment is not supported.")
+          case SegmentCodec.Annotated(_, _) =>
+            throw new IllegalStateException("Annotated SegmentCodec should never be nested.")
+        }
+      },
+    )
+
+  def fromZSchema(schema: Schema[_], refType: SchemaStyle = SchemaStyle.Inline): JsonSchema =
```
guersam, line 443: Would it be beneficial to support new `zio.schema.annotation.description` here like https://github.com/zio/zio-http/pull/2499/ ?

987Nabil: I'd leave this PR as is and add it later.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -0,0 +1,711 @@
...
+    schema match {
+      case enum0: Schema.Enum[_] if refType != SchemaStyle.Inline && nominal(enum0).isDefined     =>
+        JsonSchema.RefSchema(nominal(enum0, refType).get)
+      case enum0: Schema.Enum[_]                                                                  =>
+        JsonSchema.Enum(enum0.cases.map(c => EnumValue.Str(c.id)))
```
runtologist: ZIO Schema expresses ADTs as Enums, not just lists of possible values. If one of `enum0` cases is not a literal / case object, then this needs to be translated into `oneOf`.

The id of the case should respect any `caseName` annotations on the schema. 

```suggestion
    private def allPrimitive(enum: Schema.Enum[_]): Boolean =
    enum.cases.forall(case_ => allPrimitive_(case_.schema))

  @tailrec
  private def allPrimitive_(schema: Schema[_]): Boolean =
    schema match {
      case _: Schema.Primitive[_]               => true
      case _: Schema.CaseClass0[_]              => true
      case Schema.Transform(schema, _, _, _, _) => allPrimitive_(schema)
      case _                                    => false
    }

  private def enumCaseId(schema: Schema.Case[_, _]): String =
    schema.annotations.collectFirst { case caseName(name) => name }.getOrElse(schema.id)

  def fromZSchema(schema: Schema[_], refType: SchemaStyle = SchemaStyle.Inline): JsonSchema =
    schema match {
      case enum0: Schema.Enum[_] if refType != SchemaStyle.Inline && nominal(enum0).isDefined =>
        JsonSchema.RefSchema(nominal(enum0, refType).get)
      case enum0: Schema.Enum[_] if allPrimitive(enum0)                                       =>
        JsonSchema.Enum(enum0.cases.map { c => EnumValue.Str(enumCaseId(c)) })
      case enum0: Schema.Enum[_]                                                              =>
        JsonSchema
          .OneOfSchema(enum0.cases.flatMap { case_ =>
            case_.schema match {
              case Schema.Lazy(schema0) => ; None
              case schema               => Some(fromZSchema(schema, refType))
            }
          })
          .deprecated(deprecated(enum0))
          // .discriminator(enum0.annotations.collectFirst { case discriminatorName(discriminator) =>
          //   OpenAPI.Discriminator(discriminator)
          // })

```

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -0,0 +1,711 @@
...
+          .Object(
+            Map.empty,
+            additionalProperties,
+            Chunk.empty,
+          )
+          .addAll(record.fields.map { field =>
+            field.name ->
+              fromZSchema(field.schema, refType).deprecated(deprecated(field.schema))
+          })
+          .required(record.fields.filterNot(_.schema.isInstanceOf[Schema.Optional[_]]).map(_.name))
```
runtologist: Fields annotated as optional are still modeled as required in the openapi.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
runtologist, line 443: IMO `fromZSchema` should be able to accumulate component schemas. For example `oneOf`, `allOf` cases should be named. If they are not named, this prohibits proper use of discriminators. Also schemas that are used in multiple places should not be expanded multiple times. 

One possible approach would be to allow `fromZSchema` to return multiple  `JsonSchemas`, but I'm sure there are other approaches.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: You are right. I remember I had this on my list, but must have missed it. Also I missed a test for this. Idk if I will take your your code, but I'll tackle this.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: True. I missed that one. Also transient fields should actually be filtered out. Thanks.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: The case of expanding multiple times was intentionally ignored. As it does not change meaning, but is just an optimization. And also right now, we build schemas per endpoint and then accumulate. In this case, we can't be very efficient, since we don't know all Endpoints. Also I am not sure how efficient this needs to be, since I don't expect that building OpenAPI is happening more than once per application.
Actually, we could probably generate it at compile time 🤔 
Your argument with the naming I don't get. Could you please give me more details? Or an example?

runtologist: I am referring to this section https://swagger.io/specification/#discriminator-object, where I am not sure the example for `oneOf` with discriminator can be modelled without naming cases. One can probably get pretty far without, but I have seen it in the past.

My concern with duplication is not about efficiency, but ergonomics. Usually the oeverall openapi spec is not for a single endpoint, but for a more complex API, where lots of types appear in multiple places. A human reader of the API will need to read a very long documents with lots of repetitions. A client that is generated from the spec will have basically the same datatype repeated lots of times under different names. If everything was expanded, such types could be generated exactly once.

987Nabil: I think I get what you mean. You want the schemas not to be inlined, but part of components, so you can use a discriminator to reference the schema. I'll impl. this

runtologist: Exactly. Wonderful. :)

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: @runtologist Can you tell me your reasoning behind `case _: Schema.Primitive[_]  => true` ?
I think this should be `false`, since primitive schemas still accept a set of values and a fixed value.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: @runtologist Thanks for the helpful review. Maybe you could check the changes 🙂

runtologist: Sure. This is really a lot better! Using discriminator + references works really well. I love it. There are a few things I've noticed:

After the change `caseName`s on enums are not applied any more. All enum values are the name of the class/object. 

Optional fields and fields with defaults are no longer listed as required, but their type is 
```yaml
type:
  - OriginalType
  - 'null'
```
AFAIU it should just be `OriginalType`.

Enums are still inlined instead of referenced.

Element types of arrays are still inlined instead of referenced.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -0,0 +1,910 @@
...
+      JsonSchemas(
+        JsonSchema.ArrayType(Some(nested.root)),
+        ref,
+        nested.children,
+      )
+    } else {
+      JsonSchemas(
+        JsonSchema.ArrayType(Some(nested.root)),
+        ref,
+        nested.children + (nested.rootRef.get -> nested.root),
```
runtologist, line 438: This crashes with a `None.get` for me.

```suggestion
        nested.children ++ nested.rootRef.map(_ -> nested.root),
```

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
runtologist: I think you are right.

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: Can you give me a reproducer?

zio-http/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: > Enums are still inlined instead of referenced
> 
> Element types of arrays are still inlined instead of referenced.

In the test I see something else. Maybe  matter of what SchemaStyle is used. Do you have an example?

> AFAIU it should just be OriginalType.

I am not so sure about that. I'll try to read specs again and think about what implications that has. Anyhow I do not see this as a show stopper, since it is not wrong. Option[Value] in json is null | Value.

> After the change caseNames on enums are not applied any more. All enum values are the name of the class/object.

I'll check this

jdegoes (approved):

jdegoes:
@987Nabil Please submit followups in a separate PR. Thanks for all your hard work on this PR!

Issue #2321, opened 2023-07-27
Support query parameters with multiple values in Endpoint API


jdegoes:
Currently every query parameter is supposed to yield an `A`, but this should probably be changed to yield a `NonEmptyChunk[A]`.

The main work is modifying `Query` which extends `HttpCodec` (inside the companion object of `HttpCodec`, located in file `HttpCodec.scala`) from `extends HttpCodec[A]` to `extends HttpCodec[NonEmptyChunk[A]]`. The secondary task is to modify the `EncoderDecoder.scala` so that it properly encodes and decodes non-empty chunks for query parameters.

NavidJalali:
I can give this a go

ghost:
/attempt #2321

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clkl5jq9y0002l30f88914e48/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

ghost:
/attempt #2321

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clkl5jq9y0002l30f88914e48/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

eshu:
> Currently every query parameter is supposed to yield an `A`, but this should probably be changed to yield a `NonEmptyChunk[A]`.
> 
> The main work is modifying `Query` which extends `HttpCodec` (inside the companion object of `HttpCodec`, located in file `HttpCodec.scala`) from `extends HttpCodec[A]` to `extends HttpCodec[NonEmptyChunk[A]]`. The secondary task is to modify the `EncoderDecoder.scala` so that it properly encodes and decodes non-empty chunks for query parameters.

@jdegoes 

It does not look so simple. It does not look possible to make `Query[A]` just `extends HttpCodec[NonEmptyChunk[A]]`, it breaks a lot of things inside, such as operator `++`, `Query.optional`, I think there are many other things. So Query should stay as `HttpCodec[A]`.

In other hand when I try to change just `EncoderDecoder`, it breaks `Mechanic` object in `makeConstructorLoop` method (`coerce` function).

Maybe there should be a special case `Query` or `TextCodec`.

Sorry, I saw this code just yesterday and I do not have a good understanding what is going on under the hood, most probably my thoughts are amateurish.

[...]

PR #2472, opened 2023-10-07
Issue #2321: Accept multiple query parameter values

eshu:
/claim #2321

Also please check this PR: it could be a good addition to current one: https://github.com/eshu/zio-http/pull/1


zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -574,14 +574,49 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
 
     def index(index: Int): ContentStream[A] = copy(index = index)
   }
-  private[http] final case class Query[A](name: String, textCodec: TextCodec[A], index: Int = 0)
-      extends Atom[HttpCodecType.Query, A]  {
-    self =>
-    def erase: Query[Any] = self.asInstanceOf[Query[Any]]
+
+  private[http] sealed trait Query[A, I] extends Atom[HttpCodecType.Query, A] {
```
987Nabil: Mono and Multi are not different structure wise. Do we need an ADT for it?
Can't we just annotate it as metadata? Or just add a boolean flag?

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
```diff
@@ -15,30 +15,36 @@
...
 
   def queryInt(name: String): QueryCodec[Int] =
-    HttpCodec.Query(name, TextCodec.int)
+    HttpCodec.MonoQuery(name, TextCodec.int)
 
   def queryAs[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] =
-    HttpCodec.Query(name, codec)
+    HttpCodec.MonoQuery(name, codec)
+
+  def queries[I](name: String)(implicit codec: TextCodec[I]): QueryCodec[Chunk[I]] =
```
987Nabil: `queries` sounds strange. `queryAll`?

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
```diff
@@ -15,30 +15,36 @@
...
+    HttpCodec.MultiQuery(name, codec)
 
   def paramStr(name: String): QueryCodec[String] =
-    HttpCodec.Query(name, TextCodec.string)
+    HttpCodec.MonoQuery(name, TextCodec.string)
```
987Nabil: @jdegoes I think we should delete one group of constructors. Either query or param not both.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
eshu: @987Nabil They are different in functionality, but yes, MultiQuery could replace MonoQuery. Please check the approach: https://github.com/eshu/zio-http/pull/1. If it is ok, I can merge it and then remove monoquery.

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
eshu: @987Nabil I think so. I just thought one of groups needed for backward compatibility.

987Nabil: @eshu we are in RC versions. There is no compat. guarantees. This is the time to remove one of them. Now or never basically ;)

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
987Nabil: That one looks too complicated I think. Also it duplicates features that already exist on `HttpCodec` level like optional.

eshu: @987Nabil Actually they have a different structures if you check the `textCodec`. Usage of it is completely different in `MonoQuery` and `MultiQuery`: for MonoQuery (which returns a scalar) it returns the result of query itself. And for `MultiQuery` the role of textCodec is different: it transforms each element of the collection.

So to merge it I need to return a collection every time, For the single element query the good candidate is `Id`, but if I drop `MonoQuery`, it requires to change all cases where scalar should be returned, so for backward compatibility it could be good to keep `MonoQuery`.

In the PR I mentioned before I added the parameter `QueryCardinality`. Maybe it could help to drop MonoQuery in future.

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
eshu: @987Nabil I think `param` should be better, but in this case it will be correct to change object `Query` name too

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
eshu: @987Nabil Anyway I think cardinality is a good idea. What do you think?

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
jdegoes: `queryAll` would be more idiomatic.

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
jdegoes: Yes, we should be consistent with naming.

eshu: @987Nabil @jdegoes So what should I drop - queries or params? If I drop queries, I need to rename Query objects to Param, correct?

Actually I prefer to drop queries and rename, but you are owners of the project, the decision should be yours.

987Nabil: I think param is to general. A header can also be seens as a param. I prefer query. And it is consistent with the name query codec.

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
987Nabil: We do not need backwards compatible, since we are still in RC. Clean design is what we want.
I think it might be okay, to store the information if you expect single or multiple values for example for docs, but keep the multi API on the codec. Where we internally then use the codec and in user facing API, we can still behave in a differt way.
Or we try to build something like for headers, which also supports both, but has only one codec, iirc

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
eshu: Ok

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
eshu: @987Nabil I will try.

Just curious: if you do not care about backward compatibility, why do you need Scala 2.12 support?

eshu: @987Nabil @jdegoes 

I can't reproduce locally errors that I see in CI:
```
[error] Failed tests:
[error] 	zio.http.codec.HttpCodecSpec
[error] 	zio.http.endpoint.RoundtripSpec
[error] 	zio.http.endpoint.QueryParameterSpec
```
Could you please help?

And another problem: test `zio.http.ClientStreamingSpec` is working extremely slow (8-10 minutes) and fails even for the main branch, and this issue is not reproducible in CI.

eshu: Also here
```
+ HttpCodecSpec
...
  + optional
    + no fallback for decoding errors - 59 ms
    - fallback for missing values
      Exception in thread "zio-fiber-30195" zio.http.codec.HttpCodecError$InvalidQueryParamCardinality: Wrong query parameter name cardinality 0, exactly one expected
      	at zio.http.codec.HttpCodecSpec.spec(HttpCodecSpec.scala:138)
      	at zio.http.codec.HttpCodecSpec.spec(HttpCodecSpec.scala:130)
```
in lines 130 and 138 of HttpCodecSpec in the local copy I see dummy tests only, they should not fail...

eshu: Found the problem...

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
```diff
@@ -15,30 +15,23 @@
...
+  def queryBool(name: String): QueryCodec[Boolean] = queryAs[Boolean](name)
 
-  def queryAs[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] =
-    HttpCodec.Query(name, codec)
+  def queryInt(name: String): QueryCodec[Int] = queryAs[Int](name)
 
-  def paramStr(name: String): QueryCodec[String] =
-    HttpCodec.Query(name, TextCodec.string)
+  def queryOpt[I](name: String)(implicit codec: TextCodec[I]): QueryCodec[Option[I]] =
+    HttpCodec.Query(name, TextChunkCodec.optional(codec))
```
987Nabil, line 30: I am pretty sure, that this is not necessary. We have already `HttpCodec#optional` that should cover this case.
Please try to replace `QueryCodec.queryOpt[X](y)` in the tests with `QueryCodec.query[X](y).optional`.
If it works, I think you can remove this constructor and `TextChunkCodec.optional`.
If not, please report back why it does not work.

zio-http/src/main/scala/zio/http/codec/TextChunkCodec.scala
```diff
@@ -0,0 +1,79 @@
+package zio.http.codec
+
+import scala.annotation.tailrec
+
+import zio.prelude._
```
987Nabil: I am not sure what you use prelude for, but this is currently only a transitive dependency. And I think we should not use it, but rather keep dependencies low.

zio-http/src/main/scala/zio/http/codec/TextChunkCodec.scala
```diff
@@ -0,0 +1,79 @@
...
+import zio.{Chunk, ChunkBuilder, NonEmptyChunk}
+
+sealed trait TextChunkCodec[A, I] {
```
987Nabil, line 7: I like your idea, of having more detailed errors. But I still think the whole `TextChunkCodec` is not necessary.
I think you could just fit it in `TextCodec`. You could add two new methods to `TextCodec`, `def decodeAll(values: Chunk[String]): Chunk[A]` and `def encodeAll(values: Chunk[A]): Chunk[String]`.
For `Query`, add a parameter `cardinality: Cardinality`
```scala
sealed trait Cardinality
object Cardinality {
  case object ExactyOne extends Cardinality
  case object AtLeastOne extends Cardinality
  case object Many extends Cardinality
}
```
And set this according to the constructors.
Then just use the two new methods in `EncoderDecoder` and throw an exception if cardinality and the size of the chunk do not match.

This has imho the following benefits:
- no extra class
- every query codec is just one object instead of 2
- still having better error messages, without wrapping each result of decoding a query in `DecodeResult`

This seems to me better in terms of maintainability and allocations during decoding.

zio-http/src/main/scala/zio/http/codec/TextChunkCodec.scala
eshu: prelude is a part of ZIO ecosystem.

zio-http/src/main/scala/zio/http/codec/QueryCodecs.scala
eshu: Yes, but it would be better to solve all cardinality cases in the same class and do not spread it over all code base.

zio-http/src/main/scala/zio/http/codec/TextChunkCodec.scala
eshu: You ask me to drop one class and replace with another one. :)
It won't decrease the number of classes in code and definitely increase complexity because we put more functionality in TextCodec unrelated to its main purpose.

TextCodec is good for encoding/decoding scalar values, this is it's speciality. TextChunkCodec do not have an intersection with it, and have another specialization. I believe it is good to follow to the single responcibility principle and do not pull all functionality in one class.

987Nabil: You are right, that my idea still have one new data structure. The difference is, there are only singletons and it is analysable. What you have right now, can't be translated into docs. I can't take a look at an instance of `TextChunkCodec` and know the cardinality.
Also I don't see any problem with extending `TextCodc` by these methods, since they can be added to the trait and would be the same for all implementations.
As well I'd argue that it is still single responsibility. The responsibility to transform text to values and values to text.
@vigoo @jdegoes do you have some opinion / other idea(s)?

eshu: For documentation it is possible to add a method "description" like in TextCodec or add another way of tagging. This is not an issue.

eshu: @987Nabil I added the method `cardinality` to `TextChunkCodec`.

I can't agree that `TextChunkCodec` and `TextCodec` have the same responsibility. `TextCodec` solves the scalar transformation. One string - one value. `TextChunkCodec` solves the issue of cardinality: check amount of values and coerce it to expected cardinality. It uses `TextCodec` to encode each string scalar to a typed value.

987Nabil: I offered my opinion. And I don't feel like this is my field of expertise for zio-http (yet). I'll neither ask you to change nor approve this PR but rather summon my fellow zio-http team members ༼ つ ◕◕ ༽つ

jdegoes:
@eshu @987Nabil 

I really do not like the idea of `TextChunkCodec`. Rather, I feel like every query parameter may have zero or more values (according to spec), so the base primitive for `QueryCodec` should in fact generate `Chunk[A]` rather than `A`.

Essentially, `HttpCodec.Query` should be changed from this:

```scala
private[http] final case class Query[A](name: String, textCodec: TextCodec[A], index: Int = 0)
      extends Atom[HttpCodecType.Query, A]  {
```

to this:

```scala
private[http] final case class Query[A](name: String, textCodec: TextCodec[A], index: Int = 0)
      extends Atom[HttpCodecType.Query, Chunk[A]]  {
```

and all types updated. Now, we shouldn't break backward compatibility: and, indeed, most people using query parameters will just be looking for one value, rather than zero to many. So we should expose the same set of `QueryCodecs` (QueryCodecs.scala) that we do today, but we can expose a more foundational one that provides access to 0 to many query parameter values.

How does this approach sound?

Please re-open if you get a chance to simplify this solution, and I will be happy to review and merge quickly.

987Nabil:
Sounds good to me.

PR #2476, opened 2023-10-11
Add test that uses authentication with context

987Nabil:


vigoo (approved):
Thanks!
Maybe we could have another with `customAuthProvidingZIO` (I hope that still exists ;) ) so we have an example of that too

PR #2481, opened 2023-10-13
Add Flash, Flash.Message and Flash.Backend

TomTriple:
Same idea as in https://github.com/zio/zio-http/pull/2466. Summary: 

- Add typed and composable access to the flash-scope with a new type `Flash`. For that there exists various `Flash#get` as well as `Flash#set` operators. 
- Add a new type `Flash.Message` that adds an explicit type for typical use cases of flash messages representing a `notice` or `alert` (or both!). Inspired by https://guides.rubyonrails.org/action_controller_overview.html#the-flash. 
- Add **cookie-based flash-scope** with the typical size limitation that cookies have. The entire payload will be transported in the content of a cookie. Using this flash-scope involves no effects, only `Response#addFlash` / `Request#flash`. 
- Add **backend-based (`Flash.Backend`) flash-scope** that has no limitation in size as the payload is not transported in a cookie but kept in an internal structure. Still, it uses a cookie to transport an identifier for the lookup into the internal structure. Using this flash-scope involves effects as we need to access an internal `Ref` for the payload.


guizmaii:
@TomTriple To overcome the cookie content size limitation we could split the content of the flash into several cookies. WDYT?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+
+  def optional: Flash[Option[A]] = self.map(Option(_)) <> Flash.succeed(None)
+
+  def foldHtml[A1 >: A, B](f: Html => B, g: Html => B)(h: (B, B) => B)(implicit
+    ev: A1 =:= Flash.Message[Html, Html],
+  ): Flash[B] =
+    self.map(a => a.asInstanceOf[A1].fold(f, g)(h))
+
+  def toHtml[A1 >: A](implicit ev: A1 =:= String): Flash[Html] =
+    self.map(Html.fromString(_))
```
guizmaii: Don't we need these methods to be `final`? Is there a need for them to be overridable?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+      override def addFlash[A](response: Response, setter: Setter[A]): UIO[Response] = {
+        val map = Flash.Setter.run(setter, Map.empty)
+        for {
+          flashId       <- zio.Random.nextUUID
+          setterFlashId <- ref.update(in => in + (flashId -> map)).as(Flash.setValue(flashIdName, flashId))
+        } yield response.addFlash(setterFlashId)
+      }
+    }
+
+    val layer: ULayer[Backend] = ZLayer(Ref.make(Map.empty[UUID, Map[String, String]]).map(Impl.apply))
```
guizmaii: I do not see the point of providing a flash backend based on a Map as most probably users' apps will be horizontally scaled. For dev purposes? 🤔

What if we rename this layer to `inMemory` then?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+    case class SetValue[A](schema: Schema[A], key: String, a: A) extends Flash.Setter[A]
+
+    case class Concat[A, B](left: Setter[A], right: Setter[B]) extends Flash.Setter[(A, B)]
+
+    def run[A](setter: Setter[A]): Cookie.Response =
+      Cookie.Response(
+        Flash.COOKIE_NAME,
+        URLEncoder.encode(
+          JsonCodec.jsonEncoder(Schema[Map[String, String]]).encodeJson(run(setter, Map.empty)).toString,
+          java.nio.charset.Charset.defaultCharset.toString.toLowerCase,
```
guizmaii: Why not UTF8? 🤔

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+        ),
+      )
+
+    def run[A](setter: Setter[A], map: Map[String, String]): Map[String, String] = {
```
guizmaii, line 207: We could inline this `run` function into the previous one, no?

```scala
def run[A](setter: Setter[A]): Cookie.Response = {
    def loop(setter: Setter[A], map: mutable.Map[String, String]): mutable.Map[String, String] = 
      setter match {
        case SetValue(schema, key, a) =>
          map += (key -> JsonCodec.jsonEncoder(schema).encodeJson(a).toString)
        case Concat(left, right)      =>
          loop(right, loop(left, map))
        case Empty                    => map
      }

      Cookie.Response(
        Flash.COOKIE_NAME,
        URLEncoder.encode(
          JsonCodec.jsonEncoder(Schema[Map[String, String]]).encodeJson(loop(setter, mutable.Map.empty).toMap).toString,
          java.nio.charset.Charset.defaultCharset.toString.toLowerCase,
      )
    )         
}
```

Or you want to expose both function maybe?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+    /**
+     * Gets an `A` from the backend-based flash-scope or fails with a
+     * `Throwable`.
+     */
+    def flash[A](request: Request, flash: Flash[A]): IO[Throwable, A]
+
+    /**
+     * Gets an `A` from the backend-based flash-scope and provides a fallback.
+     */
+    def flashOrElse[A](request: Request, flash: Flash[A])(orElse: => A): UIO[A] =
```
guizmaii: `final`?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+     * Adds flash values to the backend-based flash-scope and returns a workflow
+     * with an updated `Response`.
+     */
+    def addFlash[A](response: Response, setter: Flash.Setter[A]): UIO[Response]
+
+    /**
+     * Optionally adds flash values to the backend-based flash-scope and returns
+     * a workflow with an updated `Response`.
+     */
+    def addFlash[A](response: Response, setterOpt: Option[Flash.Setter[A]]): UIO[Response] =
```
guizmaii: `final`?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+    private val flashIdName = "flashId"
+
+  }
+
+  sealed trait Setter[A] { self =>
+
+    /**
+     * Combines setting this flash value with another setter `that`.
+     */
+    def ++[B](that: => Setter[B]): Setter[(A, B)] = Setter.Concat(self, that)
```
guizmaii: `final`?

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+  def getBoolean(key: String): Flash[Boolean] = get[Boolean](key)
+
+  /**
+   * Gets the first flash value of type `A` regardless of any key.
+   */
+  def get[A: Schema]: Flash[A] = withInput { map =>
+    map.keys.map(a => Flash.get(a)(Schema[A])).reduce(_ <> _)
+  }
+
+  private def loop[A](flash: Flash[A], map: Map[String, String]): Either[Throwable, A] =
```
guizmaii: Why using `Throwable` here in the Left part and not just String? 🤔
Using String would simplify the code a bit, no?

🟢 Also, as this function is only used in the `run` function below (I'm maybe wrong here), I'd put it as a inner function of the `run` function:
```scala
private[http] def run[A](flash: Flash[A], sourceMap: Map[String, String]): Either[Throwable, A] = {
   def loop[A](flash: Flash[A], map: Map[String, String]): Either[Throwable, A] = ...
  
  loop(flash, sourceMap)
}
```

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+        loop(f(map), map)
+      case OrElse(self, that) =>
+        (loop(self, map) match {
+          case Left(_)      => loop(that, map)
+          case Right(value) => Right(value)
+        }).asInstanceOf[Either[Throwable, A]]
+      case FlatMap(self, f)   =>
+        loop(self, map) match {
+          case Right(value) => loop(f(value), map)
+          case Left(e)      => Left(e)
```
guizmaii:
```suggestion
          case l @ Left(_)  => l.asInstanceOf[Either[Throwable, A]]
```

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+      case Get(schema, key)   =>
+        map.get(key).toRight(new Throwable(s"no flash key: $key")).flatMap { value =>
+          JsonCodec.jsonDecoder(schema).decodeJson(value).left.map(e => new Throwable(e))
+        }
+      case WithInput(f)       =>
+        loop(f(map), map)
+      case OrElse(self, that) =>
+        (loop(self, map) match {
+          case Left(_)      => loop(that, map)
+          case Right(value) => Right(value)
```
guizmaii:
```suggestion
          case r @ Right(_) => r.asInstanceOf[Either[Throwable, A]]
```

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+          case Left(e)      => Left(e)
+        }
+      case Succeed(a)         => Right(a)
+      case Fail(message)      => Left(new Throwable(message))
+    }
+
+  private[http] def run[A](flash: Flash[A], sourceRequest: Request): Either[Throwable, A] = {
+    sourceRequest
+      .cookie(COOKIE_NAME)
+      .toRight(new Throwable("flash cookie doesn't exist"))
```
guizmaii: 🟢 I always prefer to use the Java exceptions are they're supposed to be used. `RuntimeException` is supposed to be used in this case, not `Throwable`, as it's a runtime exception
```suggestion
      .toRight(new RuntimeException("flash cookie doesn't exist"))
```

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,390 @@
...
+      .flatMap { cookie =>
+        try Right(URLDecoder.decode(cookie.content, java.nio.charset.Charset.defaultCharset.toString.toLowerCase))
+        catch {
+          case e: Exception => Left(e)
+        }
+      }
+      .flatMap { cookieContent =>
+        JsonCodec.jsonDecoder(Schema.map[String, String]).decodeJson(cookieContent).left.map(e => new Throwable(e))
+      }
```
guizmaii: 🟢 For performances, I'd inline these 2 `.flatMap`:
```scala
      .flatMap { cookie =>
        try {
          val content = URLDecoder.decode(cookie.content, java.nio.charset.Charset.defaultCharset.toString.toLowerCase))
          JsonCodec.jsonDecoder(Schema.map[String, String]).decodeJson(content).left.map(e => new Throwable(e))
        } catch {
          case e: Exception => Left(e)
        }
      }

```

TomTriple:
> @TomTriple To overcome the cookie content size limitation we could split the content of the flash into several cookies. WDYT?

I think that's a interesting proposal - even though the limit then is at about 300 cookies (which... should be enough for most / all? things regarding payload data in the flash). I like your idea but right now I´m not sure about any other side effects that come up with it.

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: I can't hurt to make them final!

guizmaii:
@TomTriple We could keep this "split into several cookies" idea for another PR, I just wanted to mention it 🙂

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: Yes, you're right - when you're scaling the app that does not work anymore. In this case one needs to use an external data service. I think it should be possible to add another `Backend` with the given interface. 

Yes, calling it `inMemory` is a good idea!

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: yes

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: yes

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: yes

TomTriple:
> @TomTriple We could keep this "split into several cookies" idea for another PR, I just wanted to mention it 🙂

Yes of course - we will see what other reviewers think about it. I'll send another PR with your changes from the review. Thanks a lot for reviewing :+1:

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: Internally there both `run` operations are used

zio-http/src/main/scala/zio/http/Flash.scala
TomTriple: Can not hurt to make that explicit!

zio-http/src/main/scala/zio/http/Flash.scala
```diff
@@ -0,0 +1,392 @@
...
+  }
+
+  private[http] def run[A](flash: Flash[A], sourceRequest: Request): Either[Throwable, A] =
+    sourceRequest
+      .cookie(COOKIE_NAME)
+      .toRight(new RuntimeException("flash cookie doesn't exist"))
+      .flatMap { cookie =>
+        try {
+          val content =
+            URLDecoder.decode(cookie.content, java.nio.charset.Charset.forName("UTF-8").toString.toLowerCase)
```
guizmaii: Can't you just use `StandardCharsets.UTF_8`? 🤔 Same for the `.encode` usage
```suggestion
            URLDecoder.decode(cookie.content, StandardCharsets.UTF_8)
```

TomTriple: Thanks, somehow I did not know about `StandardCharsets`

zio-http/src/main/scala/zio/http/Flash.scala
guizmaii: @TomTriple You should probably do the same with all your `Throwable` usages

jdegoes (approved):

Issue #1535, opened 2022-09-20
Security audit of ZIO HTTP


jdegoes:
**Is your feature request related to a problem? Please describe.**

ZIO HTTP relies mostly on Netty, and therefore (largely) shares the security considerations of this project, while also benefiting from the large amount of work that has been done for Netty to identify and fix vulnerabilities and exploits.

That said, ZIO HTTP builds on and extends Netty in several ways that may introduce new vulnerabilities.

**Describe the solution you'd like**

A comprehensive security audit of ZIO HTTP in preparation for 1.0, including but not limited to:

1. Use of ZIO Schema for decoding and encoding JSON (zio.http.api).
2. Locally serving file content.
3. Any middleware that interacts with in-memory structures (including metrics) or local file system
4. Source for any and all configuration data
5. All local disk access, environment variables, properties, and resources
6. How any 'user data' coming from the request is used or incorporated into any response

**Describe alternatives you've considered**

NA

**Additional context**

Would be great to find a third-party firm to conduct an audit if the cost of such can be sponsored.

jdegoes:
I'm going to define additional details and process for performing the security audit, which should be considered the "bare minimum" necessary to close this ticket:

1. Creating a test suite for static file serving. The test suite should verify that it is impossible to escape the sandbox in which local files are being served, even in the presence of OS-specific symlinks (etc.).
2. Creating various tests for any case in which potential user-data is turned into HTTP headers, cookies, content, etc. For example, `HttpError` turns (potentially user-defined) strings into content that is ultimately embedded into the response. This allows various security exploits such as XSS, which must be prevented through appropriate escaping. The tests should attempt to embed malicious data into the construction of errors, headers, responses, etc., and should demonstrate appropriate escape mechanisms are utilized to prevent these exploits.
3. Creating a test suite that verifies any authentication code that ships with ZIO HTTP is not vulnerable to [timing attacks](https://en.wikipedia.org/wiki/Timing_attack). ZIO 2 has a data type called `zio.Config.Secret` that can be used to hold passwords, keys, etc., which can prevent timing attacks.
[...]

PR #2483, opened 2023-10-14
ZIO-HTTP Security Audit

Mayhul-Jindal:
wip issue #1535

## Coverage
1. DotDotSlash attack (passed)
2. Symlink attack (passed)
3. Encoded url attack (passed)

## Improvements
1. `java.nio` can be used instead of `java.io` in staticServe middleware. You can find the key differences [here](https://www.baeldung.com/java-io-vs-nio)

## Sources
1. [owasp](https://owasp.org/www-community/attacks/Path_Traversal)
2. [Burp Academy](https://portswigger.net/web-security/file-path-traversal)
3. [DotDotSlash](https://github.com/jcesarstef/dotdotslash/blob/master/match.py)

## Todo
- [X]  Creating a test suite for static file serving. The test suite should verify that it is impossible to escape the sandbox in which local files are being served, even in the presence of OS-specific symlinks (etc.).
- [ ]  Creating various tests for any case in which potential user-data is 
turned into HTTP headers, cookies, content, etc. For example, `HttpError` turns (potentially user-defined) strings into content that is ultimately embedded into the response. This allows various security  exploits such as XSS, which must be prevented through appropriate escaping. The tests should attempt to embed malicious data into the construction of errors, headers, responses, etc., and should demonstrate appropriate escape mechanisms are utilized to prevent these exploits.
- [ ]  Creating a test suite that verifies any authentication code that ships with ZIO HTTP is not vulnerable to [[timing attacks](https://en.wikipedia.org/wiki/Timing_attack)](https://en.wikipedia.org/wiki/Timing_attack). ZIO 2 has a data type called `zio.Config.Secret` that can be used to hold passwords, keys, etc., which can prevent timing attacks.
- [ ]  Creating a test suite that attempts to crash the server (OOME) by  generating extremely large: infinite URLs, infinite individual headers (e.g. `Content-type: application/sjkldjfklsjdflkjsdkfljsdfljsdfk....`), infinite total headers (each of which is small), infinite request bodies; infinite multi-part forms, etc.; and demonstrating that user-configurable limits on all of these are exposed and available and have reasonable defaults such that an out-of-the-box server is reasonably secure.
- [ ]  Creating a test suite that demonstrates any built-in ZIO HTTP exception types do not leak stack trace or other sensitive information in HTTP responses.
- [ ]  Creating a test suite designed to exercise any metrics that are built into ZIO HTTP, and verifying that it is not possible to run out of memory by, e.g., generating infinite URLs, infinite content types, etc.


Mayhul-Jindal:
Currently did the first one only to get your reviews as to what more should be added or it is fine like this. I will proceed after your comments.
Thnx 🙂

jdegoes:
@Mayhul-Jindal Please re-open if you get a chance to finish this. I think @987Nabil might be interested in it too.

PR #2487, opened 2023-10-18
Add a test of a middleware providing a context to a `Routes`

guizmaii:
It gives an example on how to use a "providing middleware" when using `Routes`
I struggled to find how

### DIFF
--- a/zio-http/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
+++ b/zio-http/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
@@ -80,13 +80,13 @@ object AuthSpec extends ZIOHttpSpec with HttpAppTestExtensions {
         val app = {
           Routes(
             Method.GET / "context" -> basicAuthContextM ->
-              Handler.fromFunction[(AuthContext, Request)] { case (c: AuthContext, _) => Response.text(c.value) }
+              Handler.fromFunction[(AuthContext, Request)] { case (c: AuthContext, _) => Response.text(c.value) },
           )
         }.toHttpApp
         assertZIO(
           app
             .runZIO(Request.get(URL.root / "context").copy(headers = successBasicHeader))
-            .flatMap(_.body.asString)
+            .flatMap(_.body.asString),
         )(equalTo("user"))
       },
     ),
--- a/zio-http/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
+++ b/zio-http/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
@@ -79,11 +79,15 @@ object AuthSpec extends ZIOHttpSpec with HttpAppTestExtensions {
       test("Extract username via context with Routes") {
         val app = {
           Routes(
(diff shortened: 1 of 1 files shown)


zio-http/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
```diff
@@ -77,6 +76,15 @@ object AuthSpec extends ZIOHttpSpec with HttpAppTestExtensions {
...
       },
+      test("Extract username via context with Routes") {
+        val app = {
+          Routes(
+            Method.GET / "context" ->
+              Handler.fromFunction[(AuthContext, Request)] { case (c, _) => Response.text(c.value) },
+          ) @@ basicAuthContextM
+        }.toHttpApp
+        assertZIO(app.runZIO(Request.get(URL.root / "context").copy(headers = successBasicHeader)))(equalTo("user"))
+      },
```
guizmaii: I'm adding this test because I try to do exactly the same thing in my code (as it's documented here: https://github.com/zio/zio-http/blob/main/docs/dsl/middleware.md#need-for-middlewares-and-handling-aspects) but the code doesn't compile

As we can see in the CI of this PR ([here](https://github.com/zio/zio-http/actions/runs/6560671157/job/17818793815)), here's the compilation error:
```scala
[error] -- [E172] Type Error: /home/runner/work/zio-http/zio-http/zio-http/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala:83:101 
[error] 83 |              Handler.fromFunction[(AuthContext, Request)] { case (c, _) => Response.text(c.value) },
[error]    |                                                                                                     ^
[error]    |
[error]    |Your request handler is required to accept both parameters Unit, as well as the incoming [[zio.http.Request]].
[error]    |This is true even if you wish to ignore some parameters or the request itself. Try to add missing parameters 
[error]    |until you no longer receive this error message. If all else fails, you can construct a handler manually using 
[error]    |the constructors in the companion object of [[zio.http.Handler]] using the precise types..
[error]    |I found:
[error]    |
[error]    |    zio.http.RequestHandlerInput.apply[Unit,
[error]    |      (zio.http.internal.middlewares.AuthSpec.AuthContext, zio.http.Request)](
[error]    |      zio.Zippable.Zippable2[A, B])
[error]    |
[error]    |But method Zippable2 in trait ZippableLowPriority3 does not match type zio.Zippable.Out[Unit, zio.http.Request,
[error]    |  (zio.http.internal.middlewares.AuthSpec.AuthContext, zio.http.Request)].
```

FYI, this test is a slightly adapted version of the test just above, see: https://github.com/zio/zio-http/pull/2487/files#diff-10ab181b979038396ef70fb62642122d4788336ab39ccadbe3595d7ed741a128R73-R78

guizmaii: Solved thanks to https://github.com/guizmaii/zio-http/pull/1

jdegoes (approved):

Issue #2489, opened 2023-10-18
Type `Scheme` is too restrictive and the parser is wrong


FabioPinheiro:
**Describe the bug**
CORS middleware not working with `origin: chrome-extension://appid`

**To Reproduce**
Steps to reproduce the behaviour:
1. Make any request that call CORS middleware 'curl 'http://localhost:8080/' -v -X 'OPTIONS' -H 'access-control-request-method: POST'  -H 'origin: chrome-extension://appid''

**Expected behaviour**
To return `HTTP/1.1 204` instead of a `HTTP/1.1 404`

---

**Additional context**

The ZIO type `Scheme` is too restrictive...
It must support
`scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`
[URI -> RFC from IETF](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1)
[URL -> RFC from IETF](https://datatracker.ietf.org/doc/html/rfc1738#section-2.1)

In ZIO:
```scala
scala> zio.http.Scheme.decode(new java.net.URI("chrome-extension://appid").getScheme)
val res1: Option[zio.http.Scheme] = None

zio.http.Header.Origin.parse("chrome-extension://appid")
val res2: scala.util.Either[String,zio.http.Header.Origin] = Left(Invalid Origin header
```
Because in `zio.http.Header.Origin` we have
`case Right(url) if url.host.isEmpty || url.scheme.isEmpty => Left("Invalid Origin header")`

Also this is clear a bug...

```scala
zio.http.Header.Origin.parse("abc://appid")
val res4: scala.util.Either[String,zio.http.Header.Origin] = Right(Value(wss,appid,None))
```

FabioPinheiro:
The parsing is done by the string size for efficient I suppose
```scala
scheme.length match {
  case 5 => Scheme.HTTPS
  case 4 => Scheme.HTTP
  case 3 => Scheme.WSS
  case 2 => Scheme.WS
  case _ => null
}
```

jdegoes:
/bounty $75

For fix and test case, but only if not already fixed.

987Nabil:
This is already fixed

PR #2490, opened 2023-10-18
Make type Scheme follow the specs

FabioPinheiro:
Make type Scheme more flexible
Fix for https://github.com/zio/zio-http/issues/2489 Follow the specs:
- URI RFC - https://datatracker.ietf.org/doc/html/rfc3986#section-3.1
- URL RFC - https://datatracker.ietf.org/doc/html/rfc1738#section-2.1

I'm just concerned about efficiency and keep the source code interface as similar as possible.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Scheme.scala
+++ b/zio-http/src/main/scala/zio/http/Scheme.scala
@@ -21,13 +21,17 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 sealed trait Scheme { self =>
   def encode: String = self match {
-    case Scheme.HTTP  => "http"
-    case Scheme.HTTPS => "https"
-    case Scheme.WS    => "ws"
-    case Scheme.WSS   => "wss"
+    case Scheme.HTTP           => "http"
+    case Scheme.HTTPS          => "https"
+    case Scheme.WS             => "ws"
+    case Scheme.WSS            => "wss"
+    case Scheme.Custom(scheme) => scheme
   }
 
-  def isHttp: Boolean = !isWebSocket
+  def isHttp: Boolean = self match {
+    case Scheme.HTTP | Scheme.HTTPS => true
+    case _                          => false
+  }
 
   def isWebSocket: Boolean = self match {
     case Scheme.WS  => true
@@ -35,38 +39,45 @@ sealed trait Scheme { self =>
     case _          => false
   }
 
-  def isSecure: Boolean = self match {
-    case Scheme.HTTPS => true
-    case Scheme.WSS   => true
-    case _            => false
+  def isSecure: Option[Boolean] = self match {
+    case Scheme.HTTPS | Scheme.WSS => Some(true)
+    case Scheme.HTTP | Scheme.WS   => Some(false)
+    case _                         => None
   }
 
-  def defaultPort: Int = self match {
-    case Scheme.HTTP  => 80
-    case Scheme.HTTPS => 443
-    case Scheme.WS    => 80
-    case Scheme.WSS   => 443
+  /** default ports is only define for the Schemes: http, https, ws, wss */
+  def defaultPort: Option[Int] = self match {
+    case Scheme.HTTP      => Some(Scheme.defaultPortForHTTP)
+    case Scheme.HTTPS     => Some(Scheme.defaultPortForHTTPS)
+    case Scheme.WS        => Some(Scheme.defaultPortForWS)
+    case Scheme.WSS       => Some(Scheme.defaultPortForWSS)
+    case Scheme.Custom(_) => None
   }
+
 }
-object Scheme       {
+
+object Scheme {
 
   /**
    * Decodes a string to an Option of Scheme. Returns None in case of
    * null/non-valid Scheme
+   *
+   * The should be lowercase and follow this syntax:
+   *   - Scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
    */
   def decode(scheme: String): Option[Scheme] =
     Option(unsafe.decode(scheme)(Unsafe.unsafe))
 
   private[zio] object unsafe {
     def decode(scheme: String)(implicit unsafe: Unsafe): Scheme = {
-      if (scheme == null) null
+      if (scheme == null || scheme.isEmpty) null
       else
-        scheme.length match {
-          case 5 => Scheme.HTTPS
-          case 4 => Scheme.HTTP
-          case 3 => Scheme.WSS
-          case 2 => Scheme.WS
-          case _ => null
+        scheme match {
+          case "http"  => HTTP
+          case "https" => HTTPS
+          case "ws"    => WS
+          case "wss"   => WSS
+          case custom  => Custom(custom.toLowerCase)
         }
     }
   }
@@ -78,4 +89,15 @@ object Scheme       {
   case object WS extends Scheme
 
   case object WSS extends Scheme
+
+  /**
+   * @param scheme
+   *   value MUST not be "http" "https" "ws" "wss"
+   */
+  final case class Custom(scheme: String) extends Scheme
+
+  def defaultPortForHTTP  = 80
+  def defaultPortForHTTPS = 443
+  def defaultPortForWS    = 80
+  def defaultPortForWSS   = 443
 }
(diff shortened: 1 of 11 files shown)


FabioPinheiro:
Is there benchmark that I can run here?

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -41,41 +40,46 @@ sealed trait Scheme { self =>
...
 }
-object Scheme       {
+
+object Scheme {
 
   /**
    * Decodes a string to an Option of Scheme. Returns None in case of
    * null/non-valid Scheme
+   *
+   * Scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
```
FabioPinheiro: I was thinking about add a regex for this.
Also make sure everything is lowercase.

But I was concerned about efficient. Should we create a new method for that

zio-http/src/main/scala/zio/http/ZClient.scala
```diff
@@ -675,7 +675,7 @@ object ZClient {
             case Some(Scheme.HTTPS) => Scheme.WSS
             case Some(Scheme.WS)    => Scheme.WS
             case Some(Scheme.WSS)   => Scheme.WSS
-            case None               => Scheme.WS
+            case _                  => Scheme.WS
```
987Nabil: I am not sure, that this is correct. Having no `Schema`, it is reasonable to assume `WS`, but having an explicit non websocket schema, is an error I'd say. So I would keep `None => Schema.WS` and fail with an exception else.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -101,7 +101,7 @@ final case class URL(
 
   def host(host: String): URL = {
     val location = kind match {
-      case URL.Location.Relative      => URL.Location.Absolute(Scheme.HTTP, host, URL.portFromScheme(Scheme.HTTP))
+      case URL.Location.Relative      => URL.Location.Absolute(Scheme.HTTP, host, Scheme.defaultPortForHTTP)
```
987Nabil: The port in `Location` is publicly accessible. Therefore I think we should not have a `-1` here. I'd rather make port an `Option` or have two variants of `Absolute`, one with and one without port.

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -19,15 +19,14 @@ package zio.http
...
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
-sealed trait Scheme { self =>
-  def encode: String = self match {
-    case Scheme.HTTP  => "http"
-    case Scheme.HTTPS => "https"
-    case Scheme.WS    => "ws"
-    case Scheme.WSS   => "wss"
-  }
+case class Scheme(value: String) { self =>
```
987Nabil: We do never use `case class` as super types in Scala. That leads to unwanted behaviour. Please keep the sealed trait structure and add a new `case class Custom(schema: String)` in the companion object.

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -41,41 +40,52 @@ sealed trait Scheme { self =>
...
-          case 4 => Scheme.HTTP
-          case 3 => Scheme.WSS
-          case 2 => Scheme.WS
-          case _ => null
+        scheme match {
+          case "http"  => Scheme.HTTP
+          case "https" => Scheme.HTTPS
+          case "ws"    => Scheme.WS
+          case "wss"   => Scheme.WSS
+          case any     => Scheme(any)
```
987Nabil: I'd normalize the string content in custom schemas by calling `.toLowerCase`

zio-http/src/main/scala/zio/http/Scheme.scala
987Nabil: I am not sure if we should check the validity here. But I think we should ignore casing

987Nabil (changes requested):

zio-http/src/main/scala/zio/http/ZClient.scala
FabioPinheiro: So -+ like ?
```suggestion
            case None               => ZIO.succeed(Scheme.WS)
            case _                  => ZIO.fail
```

zio-http/src/main/scala/zio/http/URL.scala
FabioPinheiro: I did consider making the port an Option.
ok I would start changing some APIs (methods signature)

zio-http/src/main/scala/zio/http/Scheme.scala
FabioPinheiro: I will do
Can I ask what type of unwanted behaviors are we talking? Is there discussions about that anywhere that I can read about?

987Nabil: this one for example
https://nrinaudo.github.io/scala-best-practices/tricky_behaviours/final_case_classes.html

zio-http/src/main/scala/zio/http/ZClient.scala
987Nabil: You can not use zio here. The method expects a pure value to be returned. I am not sure, where this method is used internally. Maybe it is okay to throw an exception. Maybe we need to change the return type to option.

987Nabil: Yes, I think returning ZIOs is correct here.

FabioPinheiro: In your opinion what Throwable should I return in `ZIO.fail`?

zio-http/src/main/scala/zio/http/URL.scala
FabioPinheiro: Done

zio-http/src/main/scala/zio/http/Scheme.scala
FabioPinheiro: @987Nabil what if we use `final case class` is that ok? I did some more improvements

I experiment have `case class Custom(schema: String)` as an addition to the original code. But after a hour of experimenting I was not liking the result.

987Nabil: Please use this structure

```scala
sealed trait Scheme

object Scheme {
  case object HTTP extends Scheme

  case object HTTPS extends Scheme

  case object WS extends Scheme

  case object WSS extends Scheme
  
  case class Custom(scheme: String) extends Scheme
}
```
This is the way we model sum types in Scala. It is the type safe way, that ensures exhaustive pattern match.

If you would model it like you have it right now, you will not see in a pattern match, that the default types are HTTP(S)/WS(S). You will only the a string wrapper.

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -140,25 +145,26 @@ final case class URL(
...
   }
 
   def port: Option[Int] = kind match {
     case URL.Location.Relative      => None
-    case abs: URL.Location.Absolute => Option(abs.port)
+    case abs: URL.Location.Absolute => abs.port
   }
 
-  def portOrDefault: Int = port.getOrElse(portFromScheme(scheme.getOrElse(Scheme.HTTP)))
+  def portOrDefault: Option[Int] = kind match {
```
FabioPinheiro, line 160: Note the old logic was the default port of the default Scheme

zio-http/src/main/scala/zio/http/Scheme.scala
FabioPinheiro: Can you have another look?

zio-http/src/main/scala/zio/http/Scheme.scala
```diff
@@ -78,4 +89,15 @@ object Scheme       {
   case object WS extends Scheme
 
   case object WSS extends Scheme
+
+  /**
+   * @param scheme
```
987Nabil, line 94: What would actually happen, if this is http(s)/ws(s)?

987Nabil (approved):

zio-http/src/main/scala/zio/http/Scheme.scala
FabioPinheiro: It's just a question of pattern matching.

We are allowing two representations for the same value. (`Scheme.HTTP` and `Scheme.Custom("http")`)
Since the construction method is public.
But I see this pattern is being used everywhere in zio-http so it's a more general problem

One solution would be to make it private and the only way to create Custom is with the Scheme's method `def decode(scheme: String)`.
This would make it a bit more type-safe. Instead of being on the documentation

PR #2493, opened 2023-10-24
Allow HandlerAspect.customAuthProvidingZIO To Fail With User Specified Response

adamgfraser:


jdegoes (approved):

jdegoes:
`zio.http.gen.scala.CodeGenSpec` is failing (but may be unrelated, cc @987Nabil ).

987Nabil:
The scalafmt support comes with downloading the scalafmt version defined in the config file. This downloading seems to be a little flakey from the runners. I'll try to add the scalafmt core to the dependencies of the module, so that it should be found in the cash when the test runs

PR #2504, opened 2023-10-31
Resolve Routes environment via context providing middleware

987Nabil:


987Nabil:
It was running locally, but is having a problem with some missing tag in the pipeline. Might take a look later.

davidlar:
This would be very nice to have!

987Nabil:
This approche does not work. I am working on something else.

PR #2505, opened 2023-11-01
Added getAs for QueryParams

### LINKED ISSUE #2423: Add `getAs[T]` to `QueryParams` for typed query parameter access
Should use an implicit `TextCodec[T]` to return an instance of `[T]`

BanyMaciej:
Add getAs[A] for QueryParams

Closes https://github.com/zio/zio-http/issues/2423

### DIFF
--- a/zio-http/src/main/scala/zio/http/QueryParams.scala
+++ b/zio-http/src/main/scala/zio/http/QueryParams.scala
@@ -18,10 +18,10 @@ package zio.http
 
 import java.nio.charset.Charset
 
-import zio.Chunk
+import zio.{Chunk, NonEmptyChunk}
 
 import zio.http.Charsets
-import zio.http.codec.TextCodec
+import zio.http.codec.{HttpCodecError, TextCodec}
 import zio.http.internal.QueryParamEncoding
 
 /**
@@ -92,8 +92,17 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
   /**
    * Retrieves all typed query parameter values having the specified name.
    */
-  def getAllAs[A](key: String)(implicit codec: TextCodec[A]): Option[Chunk[A]] =
-    map.get(key).map(_.flatMap(codec.decode))
+  def getAllAs[A](key: String)(implicit codec: TextCodec[A]): Either[QueryParamsError, Chunk[A]] =
+    map.get(key) match {
+      case Some(params) =>
+        params
+          .map(param => codec.decode(param).toRight(QueryParamsError.MalformedQueryParam(key, param, codec)))
+          .partitionMap(identity) match {
+          case (errors, _) if errors.nonEmpty => Left(QueryParamsError.MultiMalformedQueryParam(errors))
+          case (_, typedParams)               => Right(typedParams)
+        }
+      case None         => Left(QueryParamsError.MissingQueryParam(key))
+    }
 
   /**
    * Retrieves the first query parameter value having the specified name.
@@ -103,7 +112,10 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
   /**
    * Retrieves the first typed query parameter value having the specified name.
    */
-  def getAs[A](key: String)(implicit codec: TextCodec[A]): Option[A] = get(key).flatMap(codec.decode)
+  def getAs[A](key: String)(implicit codec: TextCodec[A]): Either[QueryParamsError, A] = for {
+    param      <- get(key).toRight(QueryParamsError.MissingQueryParam(key))
+    typedParam <- codec.decode(param).toRight(QueryParamsError.MalformedQueryParam(key, param, codec))
+  } yield typedParam
 
   /**
    * Retrieves all query parameter values having the specified name, or else
--- a/zio-http/src/main/scala/zio/http/QueryParams.scala
+++ b/zio-http/src/main/scala/zio/http/QueryParams.scala
@@ -21,6 +21,7 @@ import java.nio.charset.Charset
 import zio.Chunk
 
 import zio.http.Charsets
+import zio.http.codec.TextCodec
 import zio.http.internal.QueryParamEncoding
 
 /**
@@ -89,9 +90,20 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
   def getAll(key: String): Option[Chunk[String]] = map.get(key)
 
   /**
+   * Retrieves all typed query parameter values having the specified name.
+   */
+  def getAllAs[A](key: String)(implicit codec: TextCodec[A]): Option[Chunk[A]] =
+    map.get(key).map(_.flatMap(codec.decode))
+
+  /**
    * Retrieves the first query parameter value having the specified name.
    */
   def get(key: String): Option[String] = getAll(key).flatMap(_.headOption)
+
+  /**
+   * Retrieves the first typed query parameter value having the specified name.
+   */
+  def getAs[A](key: String)(implicit codec: TextCodec[A]): Option[A] = get(key).flatMap(codec.decode)
 
   /**
    * Retrieves all query parameter values having the specified name, or else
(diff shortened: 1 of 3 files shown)


zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -88,25 +89,50 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
...
+
   /**
    * Retrieves the first query parameter value having the specified name.
    */
   def get(key: String): Option[String] = getAll(key).flatMap(_.headOption)
 
+  /**
+   * Retrieves the first typed query parameter value having the specified name.
+   */
+  def getAs[A](key: String)(implicit codec: TextCodec[A]): Option[A] = get(key).flatMap(codec.decode)
```
guersam: I think there could be needs to distinguish between the absence of the parameter and decoding failure. What do you think?

BanyMaciej: It's good point, but I'm afraid that if we have output type like `Either[String, Option[A]]` or `Option[Either[String, A]]`, it may be even harder to use than separating it to `get` and `decode` on the call site. Maybe I should define separate typed error for this, something like `DecodingError` and `NotFoundError` and then result would be `Either[QueryParamError, A]`?

guersam: I agree with your concern. Personally I'd prefer the latter.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -88,25 +89,62 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
...
   /**
    * Retrieves the first query parameter value having the specified name.
    */
   def get(key: String): Option[String] = getAll(key).flatMap(_.headOption)
 
+  /**
+   * Retrieves the first typed query parameter value having the specified name.
+   */
+  def getAs[A](key: String)(implicit codec: TextCodec[A]): Either[QueryParamsError, A] = for {
+    param      <- get(key).toRight(QueryParamsError.MissingQueryParam(key))
```
guersam: Great! `QueryParamsError.Missing` seems more readable to me. What do you think?

BanyMaciej: Seems logical :)

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -18,9 +18,10 @@ package zio.http
 
 import java.nio.charset.Charset
 
-import zio.Chunk
+import zio.{Chunk, NonEmptyChunk}
```
987Nabil: `NonEmptyChunk` is unused or?

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -88,25 +89,62 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
...
+
   /**
    * Retrieves the first query parameter value having the specified name.
    */
   def get(key: String): Option[String] = getAll(key).flatMap(_.headOption)
 
+  /**
+   * Retrieves the first typed query parameter value having the specified name.
+   */
+  def getAs[A](key: String)(implicit codec: TextCodec[A]): Either[QueryParamsError, A] = for {
```
987Nabil, line 121: I am wondering, if a `def getAsZIO[A](key: String)(implicit codec: TextCodec[A]): IO[QueryParamsError, A]` makes sense too.

zio-http/src/main/scala/zio/http/QueryParams.scala
BanyMaciej: You are right - I left some unused imports, I would fix it

zio-http/src/main/scala/zio/http/QueryParams.scala
BanyMaciej: Isn't wrapping it in ZIO to heavy machinery for this task? I believe calling it like `ZIO.fromEither(queryParams.getAs ...)` is not much trouble, and current version could be possibly used in some lower level implementations :)

guersam: I believe there could be a significant gain in readability at the cost of a minor performance overhead. Keeping `getAs` as is and adding `getAsZIO` as a variant seems consistent with existing ZIO functions like `ZIO.serviceWith` and `ZIO.serviceWithZIO`.

BanyMaciej: Thanks for explaining :) You are right, I will add additional functions

987Nabil (approved):

zio-http/src/main/scala/zio/http/QueryParamsError.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import zio.http.codec.TextCodec
+import zio.http.internal.QueryParamEncoding
+
+sealed trait QueryParamsError extends Exception with NoStackTrace {
+  override def getMessage(): String = message
+  def message: String
+}
+object QueryParamsError {
+  final case class Missing(queryParamName: String) extends QueryParamsError {
```
guersam, line 33: `Missing#queryParamName` seems inconsistent with `Malformed#name`.

zio-http/src/main/scala/zio/http/QueryParamsError.scala
```diff
@@ -0,0 +1,44 @@
...
+    def message = s"Missing query parameter with name $queryParamName"
+  }
+
+  final case class Malformed(name: String, value: String, codec: TextCodec[_]) extends QueryParamsError {
+    def message = s"Unable to decode query parameter with name $name and value $value using $codec"
+  }
+
+  final case class MultiMalformed(chunk: Chunk[Malformed]) extends QueryParamsError {
```
guersam, line 41: Aren't `name` and `codec` the same across the entire chunk in a `MultiMalformed`? If so, I'd suggest an another encoding:

```scala
final case class MultiMalformed(name: String, codec: `TextCodec[_]`, values: Chunk[String]) extends QueryParamsError
```

zio-http/src/main/scala/zio/http/QueryParamsError.scala
```diff
@@ -0,0 +1,44 @@
...
+  override def getMessage(): String = message
+  def message: String
+}
+object QueryParamsError {
+  final case class Missing(queryParamName: String) extends QueryParamsError {
+    def message = s"Missing query parameter with name $queryParamName"
+  }
+
+  final case class Malformed(name: String, value: String, codec: TextCodec[_]) extends QueryParamsError {
+    def message = s"Unable to decode query parameter with name $name and value $value using $codec"
```
guersam, line 38: I doubt that `TextCodec` always provides stable `toString`. [`codec.describe`](https://github.com/zio/zio-http/blob/v3.0.0-RC3/zio-http/src/main/scala/zio/http/codec/TextCodec.scala#L40) might be more reliable choice.

guersam (commented):
Sorry for late, added my two cents over the error ADT.

zio-http/src/main/scala/zio/http/QueryParamsError.scala
BanyMaciej: Actually, then we could get rid of the "single' `Malformed` - it could be represented as `MultiMalformed` (then I will rename it to `Malformed`) with one element chunk. What do you think?

987Nabil: Like it

zio-http/src/main/scala/zio/http/QueryParamsError.scala
987Nabil: Yes, that is true.

zio-http/src/main/scala/zio/http/QueryParamsError.scala
guersam: :+1: `NonEmptyChunk` could give more guarantee here.

PR #2512, opened 2023-11-09
Improve `RichTextCodec` decoding performance

987Nabil:
Error messages are the same per instance of `CharIn`. So we compute them
only once.


kyri-petrou:
@987Nabil just leaving this comment here for visibility.

Thanks for the fix but just want to raise a concern with this approach. It "fixes" the latency the issue in situations that an instance of the `RichTextCodec` is reused, but my worry is that at some point in a few months / years will use `RichTextCodec.filter(_ => true)` in a method and the same issue will arise again

I'm sure you already thought of different approaches, but does the error message _needs_ to contain every single character in the BitSet? Perhaps it can be limited to the first 5 or 10 or even 50? If I'm understanding the code correctly, it currently contains all 65535 chars, and I'm not too sure whether that's really useful.

zio-http/src/main/scala/zio/http/codec/RichTextCodec.scala
```diff
@@ -154,15 +154,17 @@ sealed trait RichTextCodec[A] { self =>
 }
 object RichTextCodec {
   private[codec] case object Empty                                        extends RichTextCodec[Unit]
-  private[codec] final case class CharIn(set: BitSet)                     extends RichTextCodec[Char]
+  private[codec] final case class CharIn(set: BitSet)                     extends RichTextCodec[Char] {
+    lazy val errorMessage = s"Not found: ${set.toArray.map(_.toChar).mkString}"
```
jdegoes: You can do one better by removing `lazy`, and adding:

```scala
private[codec] val leftErrorMessage = Left(errorMessage)
```

Then using `leftErrorMessage` below.

987Nabil:
@kyri-petrou I see your point. I must say, the need for going with `RichTextCodec.filter(_ => true)` is, that one can not provide error messages. I think this could maybe already be better handled right now. I could also choose some max for adding the char to the error message.

jdegoes:
We discussed using `describe` to provide compact, nice descriptions of the codec on failure.

zio-http/src/test/scala/zio/http/codec/RichTextCodecSpec.scala
```diff
@@ -250,6 +250,10 @@ object RichTextCodecSpec extends ZIOHttpSpec {
         assertTrue(success(123) == codec.decode("123--")) &&
         assertTrue(codec.decode("4123").isLeft)
       },
+      test("With error message") {
+        val codec = RichTextCodec.literal("123").withError("Not 123")
+        assertTrue(codec.decode("678") == Left("(Expected, but did not find: Paragraph(Code(“1”,Inline)), Not 123)"))
```
jdegoes, line 255: This is rendering poorly, perhaps because `describe` returns `Doc`, that should first be properly converted into a string.

jdegoes (approved):

zio-http/src/test/scala/zio/http/codec/RichTextCodecSpec.scala
987Nabil: The problem is actually the `|` operator, that chains error messages this way explicitly.

PR #2513, opened 2023-11-09
Implement handleErrorCauseZIO for executing ZIO effects on error

SHSongs:
- **Error Handling in Routes**: Implemented the `handleErrorCauseZIO` function that is triggered upon encountering an error, ensuring that side effects are performed.

- **Testing**: Added tests to check that `handleErrorCauseZIO` correctly executes ZIO effects using TestConsole, when handling errors.

cc: @guersam

### DIFF
--- a/zio-http/src/main/scala/zio/http/Handler.scala
+++ b/zio-http/src/main/scala/zio/http/Handler.scala
@@ -368,14 +368,15 @@ sealed trait Handler[-R, +Err, -In, +Out] { self =>
     self.foldHandler(err => Handler.fromZIO(f(err)), Handler.succeed(_))
 
   /**
-   * Transforms all failures except effectfull interruption.
+   * Transforms all failures of the handler effectfully except pure
+   * interruption.
    */
   final def mapErrorCauseZIO[R1 <: R, Err1, Out1 >: Out](
     f: Cause[Err] => ZIO[R1, Err1, Out1],
   )(implicit trace: Trace): Handler[R1, Err1, In, Out1] =
     self.foldCauseHandler(
       err =>
-        if (err.isInterruptedOnly) Handler.fromZIO(f(err.asInstanceOf[Cause[Nothing]])) else Handler.fromZIO(f(err)),
+        if (err.isInterruptedOnly) Handler.failCause(err.asInstanceOf[Cause[Nothing]]) else Handler.fromZIO(f(err)),
       Handler.succeed(_),
     )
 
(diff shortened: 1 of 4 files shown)


zio-http/src/main/scala/zio/http/Handler.scala
```diff
@@ -367,6 +367,18 @@ sealed trait Handler[-R, +Err, -In, +Out] { self =>
...
 
+  /**
+   * Transforms all failures except effectfull interruption.
+   */
+  final def mapErrorCauseZIO[R1 <: R, Err1, Out1 >: Out](
+    f: Cause[Err] => ZIO[R1, Err1, Out1],
+  )(implicit trace: Trace): Handler[R1, Err1, In, Out1] =
+    self.foldCauseHandler(
+      err =>
+        if (err.isInterruptedOnly) Handler.fromZIO(f(err.asInstanceOf[Cause[Nothing]])) else Handler.fromZIO(f(err)),
```
jdegoes: These code paths do not differ, they both create the handler from `f(err)`. You shouldn't need the cast or the branch.

SHSongs:
```scala
  /**
   * Transforms all failures except pure interruption.
   */
  final def mapErrorCause[Err2](f: Cause[Err] => Err2)(implicit trace: Trace): Handler[R, Err2, In, Out] =
    self.foldCauseHandler(
      err => if (err.isInterruptedOnly) Handler.failCause(err.asInstanceOf[Cause[Nothing]]) else Handler.fail(f(err)),
      Handler.succeed(_),
    )
```

I have referred to mapErrorCause. mapErrorCause casts `Cause[Nothing]` if `isInterruptedOnly` is true. Do you have more information about its origin?
Additionally, do you recommend any modifications to the `mapErrorCause`?

guersam: @SHSongs `f` should not be applied when it's interrupt only. Please double check the usage of `f` in the original code.

SHSongs: I reviewed the usage of `f`. The `f` used transforms all failures of the handler effectfully, except for pure interruption. I also made changes to the code, following the original.

SHSongs: I reviewed the usage of f. The f used transforms all failures, except for pure interruption. I also made changes to the code, following the original.

jdegoes (approved):

PR #2520, opened 2023-11-14
Maintain query parameter ordering in QueryParams

easel:
The upstream Netty request uses a java.util.LinkedHashMap to store the query string parameters. By using an ordered collection, it is possible to exactly reconstruct the request url, which is sometimes necessary to implement request signing and other security measures.

This change uses a scala.collection.immutable.ListMap inside the QueryParams to allow this ordering to be maintained while keeping the convenience of the Map interface. It also replaces the various ordered constructors to assembly the ListMap one piece at a time, maintaining ordering, instead of using groupBy.


easel:
As it turns out SeqMap isn't available on 2.12 but plain ListMap is, it just hasn't got the nicely optimized `.from` constructor. Adjusted this to use `ListMap` as the underlying.

easel:
Ok, this implementation "works for me" on Scala 2.12 and 2.13 and has a meaningful test that passes. Modulo CI I think it's ready for review if anybody has time to do so.

987Nabil:
The build is still failing.

jdegoes:
The build looks to be failing for unrelated reasons, it's failing in master.

easel:
> The build looks to be failing for unrelated reasons, it's failing in master.

Thanks for confirming! I'll go ahead and mark this as ready for review -- happy for feedback on the approach I landed on, or to discuss whether or not query parameters need to be ordered.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -191,15 +193,34 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
...
+    QueryParams(map = result)
+  }
 
-  def apply(tuple1: (String, String), tuples: (String, String)*): QueryParams =
-    QueryParams(map = Chunk.fromIterable(tuple1 +: tuples.toVector).groupBy(_._1).map { case (key, values) =>
-      key -> values.map(_._2)
-    })
+  def apply(tuple1: (String, String), tuples: (String, String)*): QueryParams = {
+    var result = ListMap.empty[String, Chunk[String]]
+    Chunk.fromIterable(tuple1 +: tuples.toVector).foreach { case (key, value) =>
```
987Nabil, line 214: I think the `toVector` call is not needed.

987Nabil: Did not try it, but should something like this not work?
```scala
QueryParams(map = ListMap.from(Chunk.fromIterable(tuple1 +: tuples).groupBy(_._1).map { case (key, values) =>
      key -> values.map(_._2))
```

zio-http/src/main/scala/zio/http/URL.scala
```diff
@@ -170,11 +171,11 @@ final case class URL(
   def queryParams(queryParams: QueryParams): URL =
     copy(queryParams = queryParams)
 
-  def queryParams(queryParams: Map[String, Chunk[String]]): URL =
+  def queryParams(queryParams: ListMap[String, Chunk[String]]): URL =
```
987Nabil, line 174: I think this should just be `Map` not `ListMap`.
If the user don't care about the order, we don't need to force him.
```suggestion
  def queryParams(queryParams: Map[String, Chunk[String]]): URL =
     copy(queryParams = QueryParams(ListMap.from(queryParams)))
```

This will still keep the order if the `Map` is in fact a `ListMap`

987Nabil (changes requested):

jdegoes:
@easel I would love to keep query parameter ordering. However, `LinkedMap` is more than twice as slow as `Map` for many operations.

Could I talk you into switching to the following representation:

```scala
trait QueryParams {
  def keys: Iterable[String]

  def get(key: String): Option[Chunk[String]]

  def getOrElse(key: String, default: Chunk[String]): Chunk[String]
}
```

This is a more flexible solution for the future, because we can then have one implementation backed by ListMap but others which are not backed by ListMap.

jdegoes:
@easel Would really love a fix here. Please re-open if you get a chance. I will increase the bounty on this ticket!

jdegoes:
See #2586

PR #2531, opened 2023-11-23
Deactivate broken SSL test

987Nabil:


vigoo (approved):

987Nabil:
@vigoo I set the test to ignore for now, since it blocks all work on the repo. I tried out some stuff that did not work.
It might has to do something with the `truststore.jks`? But I am not sure. I think I better focus on my open PRs.
I opened #2532 to keep track of this issue.

PR #2533, opened 2023-11-24
Limit Scheme.Custom construction to calling decode

987Nabil:


987Nabil:
This is to avoid having something like `Custom("https")`

jdegoes (approved):

PR #2540, opened 2023-11-29
added all htmx attributes

THM068:
This PR is to add htmx attributes to the zio http. This will enable users to create interactive application using the backend.
You will need to pull the htmx javascript to make the attributes functional.

### DIFF
--- a/zio-http/src/main/scala/zio/http/template/HtmxAttributes.scala
+++ b/zio-http/src/main/scala/zio/http/template/HtmxAttributes.scala
@@ -1,48 +1,78 @@
 package zio.http.template
 
 import zio.http.template.Attributes.PartialAttribute
+
 /*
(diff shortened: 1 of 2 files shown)


guersam:
This update introduces htmx attributes for every user who imports `zio.http.template._`, even when htmx is not used.

I believe it would be more appropriate to use `zio.http.template.htmx`, ideally in a separate (sub)project, to provide enhanced convenience for htmx usage.

jdegoes:
@THM068 I think it makes sense to put into a separate project but can live inside this repo if you like (`zio-http-htmx`?). What do you think?

THM068:
> @THM068 I think it makes sense to put into a separate project but can live inside this repo if you like (`zio-http-htmx`?). What do you think?

Make sense, I will do that

jdegoes:
@THM068 I will close for now, please re-open when it's in a separate submodule of this repository, linked into root project. Thanks for your work on this!

PR #2543, opened 2023-12-05
feat: Allow dynamic schema in openapi

runtologist:
Openapi just constrains the possible values, but they are always JSON. In case of dynamic schema we do not know anything about the possible values, so we simply remove all constraints.


jdegoes (approved):

jdegoes:
@runtologist Good to merge when build is passing. 👍

PR #2546, opened 2023-12-06
Error handling methods on `Route`/`Routes` that can access `Request`

987Nabil:
Bug fix on Warning header

### DIFF
--- a/zio-http/src/main/scala/zio/http/Route.scala
+++ b/zio-http/src/main/scala/zio/http/Route.scala
@@ -16,9 +16,6 @@
 package zio.http
 
 import zio._
-import zio.stacktracer.TracingImplicits.disableAutoTrace
-
-import zio.http.Route.Provided
 
 /*
  * Represents a single route, which has either handled its errors by converting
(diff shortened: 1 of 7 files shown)


zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -47,14 +44,25 @@ sealed trait Route[-Env, +Err] { self =>
...
     self.handleErrorCause(Response.fromCauseWith(_)(f))
 
   /**
-   * Handles the error of the route. This method can be used to convert a route
+   * Handles all typed errors in the route by converting them into responses,
+   * taking into account the request that caused the error. This method can be
+   * used to convert a route that does not handle its errors into one that does
+   * handle its errors.
+   */
+  final def handleErrorRequest(f: (Err, Request) => Response)(implicit trace: Trace): Route[Env, Nothing] =
```
jdegoes: One can argue that handleError with Request could be useful in non-cause scenarios, which would argue for the naming convention of `handleError[Request][Cause]`.

jdegoes: `handleError[Request][Cause][ZIO]`

987Nabil: @jdegoes done

jdegoes (approved):

jdegoes (approved):

PR #2547, opened 2023-12-08
Add prefix method to `Route`/`Routes`

987Nabil:


zio-http/src/main/scala/zio/http/Route.scala
```diff
@@ -120,6 +119,16 @@ sealed trait Route[-Env, +Err] { self =>
    */
   def location: Trace
 
+  def prefix(prefix: PathCodec[Unit])(implicit ev: Err <:< Response): Route[Env, Err] =
```
jdegoes: Shouldn't this be called `nest`, because it nests routes under a path prefix?

jdegoes (changes requested):
One tiny comment on naming, otherwise looks great!

zio-http/src/main/scala/zio/http/Route.scala
987Nabil: I think what you say is not wrong, but I find prefix more intuitive. Akka http calls it `pathPrefix` in their routing api, play has `SimpleRouter#withPrefix`. I think prefix is quite right.

PR #2550, opened 2023-12-08
Implement Routes#apply

### LINKED ISSUE #2548: Add Route#apply and Routes#apply
We can add `apply` methods that return `ZIO[R, Either[E, Response], Response]`.

In a common case, `E =:= Response`, so we can add a variant (with an implicit `E <:< Response`) that merges the either, to get the simplified return type `ZIO[R, Response, Response]`.

adamgfraser:
Resolves #2548. /claim #2548

`Route#apply` is already implemented and returns a `ZIO[Env, Response, Response]` so I think it makes sense for `Routes#apply` to do so as well. Also, I don't think the more polymorphic variant is well defined since a `Route` may already have a middleware applied to it that is a function from handler to handler where the error type is already specialized to `Response` and converting a `Cause` to a `Response` is not information preserving.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Routes.scala
+++ b/zio-http/src/main/scala/zio/http/Routes.scala
@@ -57,6 +57,9 @@ final class Routes[-Env, +Err] private (val routes: Chunk[zio.http.Route[Env, Er
   def @@[Env1 <: Env](aspect: Middleware[Env1]): Routes[Env1, Err] =
     aspect(self)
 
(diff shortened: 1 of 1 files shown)


vigoo (approved):

jdegoes:
I was pairing with a developer and we specifically needed an apply with no implicit evidence (i.e. polymorphic E). I thought it might be possible using an error type of Either[E, Response]. Namely by stuffing an E into a hidden defect Throwable and recovering at the end.

Issue #2488, opened 2023-10-18
There's no way to use a "providing middleware" on a `Routes` nor an `HttpApp`


guizmaii:
**Is your feature request related to a problem? Please describe.**

For now, it seems that there's no way to use a "providing middleware" on a `Routes` nor on an `HttpApp`.
It can only be used on a `Handler`, which means that to have some "private routes" I need to repeat the "providing middleware" in each route declaration:
```scala
final case class LoggedUser(email: String)
val authProviderMiddleware: HandlerAspect[Any, LoggedUser] = ...
val sharedMiddlewares: Middleware[Any] = ... ++ ... ++ ...

val publicRoute = 
  Routes(
    Method.GET / "example" -> ...
    ...
  )

val privateRoutes =
  Routes(
    Method.GET / "private" / "example" -> authProviderMiddleware -> handler { case (user: LoggedUser, req: Request) => ... } 
    Method.POST / "private" / "example" -> authProviderMiddleware -> handler { case (user: LoggedUser, req: Request) => ... } 
    ...
  )
    
Server.serve(app @ sharedMiddleware)
```

**Describe the solution you'd like**

It would be nicer to be able to do something like this:
```scala
final case class LoggedUser(email: String)
val authProviderMiddleware: HandlerAspect[Any, LoggedUser] = ...
val sharedMiddlewares: Middleware[Any] = ... ++ ... ++ ...

val publicRoute = 
  Routes(
    Method.GET / "example" -> ...
    ...
  )

val privateRoutes =
  Routes(
    Method.GET / "private" / "example" -> handler { case (user: LoggedUser, req: Request) => ... } 
    Method.POST / "private" / "example" -> handler { case (user: LoggedUser, req: Request) => ... } 
    ...
  ) @@ authProviderMiddleware
    
Server.serve(app @ sharedMiddleware)
```

jdegoes:
/bounty $100

@987Nabil Working on a fix for this.

guizmaii:
Thanks @987Nabil! ❤️

PR #2553, opened 2023-12-08
Provide context to multiple routes (#2488)

987Nabil:
This PR is not good to go, but shows that I found a way how to provide a context to multiple routes without any more explicit typing or extremely strange syntax.

RFC and I happily accept suggestions on the syntax.

/claim #2488

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/RouteSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/RouteSpec.scala
@@ -74,7 +74,9 @@ object RouteSpec extends ZIOHttpSpec {
     ),
     suite("error handle")(
       test("handleErrorCauseZIO should execute a ZIO effect") {
-        val route = Method.GET / "endpoint" -> handler { (_: Request) => ZIO.fail(new Exception("hmm...")) }
+        val route: Route[Any, Exception] = Method.GET / "endpoint" -> handler { (_: Request) =>
+          ZIO.fail(new Exception("hmm..."))
+        }
         for {
           p <- zio.Promise.make[Exception, String]
 
@@ -88,7 +90,9 @@ object RouteSpec extends ZIOHttpSpec {
         } yield assertTrue(extractStatus(response) == Status.InternalServerError, result.contains("hmm..."))
       },
       test("handleErrorCauseRequestZIO should produce an error based on the request") {
-        val route = Method.GET / "endpoint" -> handler { (_: Request) => ZIO.fail(new Exception("hmm...")) }
+        val route: Route[Any, Exception] = Method.GET / "endpoint" -> handler { (_: Request) =>
+          ZIO.fail(new Exception("hmm..."))
+        }
         for {
           p <- zio.Promise.make[Exception, String]
 
@@ -109,12 +113,14 @@ object RouteSpec extends ZIOHttpSpec {
         )
       },
       test("handleErrorCauseRequest should produce an error based on the request") {
-        val route        = Method.GET / "endpoint" -> handler { (_: Request) => ZIO.fail(new Exception("hmm...")) }
-        val errorHandled =
+        val route: Route[Any, Exception] = Method.GET / "endpoint" -> handler { (_: Request) =>
+          ZIO.fail(new Exception("hmm..."))
+        }
+        val errorHandled                 =
           route.handleErrorRequest((e, req) =>
             Response.internalServerError(s"error accessing ${req.path.encode}: ${e.getMessage}"),
           )
-        val request      = Request.get(URL.decode("/endpoint").toOption.get)
+        val request                      = Request.get(URL.decode("/endpoint").toOption.get)
         for {
           response      <- errorHandled.toHttpApp.runZIO(request)
           resultWarning <- ZIO.fromOption(response.headers.get(Header.Warning).map(_.text))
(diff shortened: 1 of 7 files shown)


vigoo:
Some random thoughts, only by looking at the code, did not play with it:

- it can be confusing what middleware you have to pass to this `Routes` constructor an what not (or what happens if you pass something that is not providing any context there? etc)
- also introducing a new route type with a different operator (partial route / `-->`) although I understand why, I'm afraid that it makes the API hard - when to use `->` and when `-->`? What if I already have a `Route` value and want to mix it with partial routes? etc. I'm sure most of these have valid answers and maybe we can make it feel seamless, but it is not clear to me yet

On the other hand, I this proposal solves one special case when you want one context-providing middleware to provide context to a set of routes - which I think is a quite common case so maybe worth to introduce something "special" like this to help that use case.

987Nabil:
@vigoo The middleware at the beginning is already the case if you define it per route. If you provide something without a context, nothing happens. You define just the routes like it would not be there.

To the second point, I agree. That is why this is a POC. What I try to proof is, that you can construct routes without a lot of type annotations and still provide a context. The API design / syntax is something different. But I can't imagine there is no way to find a solution for that. How beautiful it will be, I can't tell. But I think we are a step closer to something usable.

987Nabil:
@jdegoes ping! You wanted to give me some feedback.

zio-http/src/main/scala/zio/http/RoutePattern.scala
```diff
@@ -64,6 +64,13 @@ final case class RoutePattern[A](method: Method, pathCodec: PathCodec[A]) { self
   ): Route[Env, Err] =
     Route.route(Route.Builder(self, HandlerAspect.identity))(handler)(zippable.zippable, trace)
 
+  def -->[Env, Err, I, Ctx, In](handler: Handler[Env, Err, I, Response])(implicit
```
jdegoes: If we are going to use `-->` for this, then we should make sure to be consistent across all of ZIO HTTP.

In addition: We should give this a named variant too, rather than just the symbolic operator.

987Nabil: Can you be more specific on being consistent with what? I don't even say we should use it, but it was not possible to simply overload `->`

jdegoes (approved):

987Nabil:
@jdegoes I took a look again after letting intentionally some time pass. Which was good, since I managed to remove the new `Route` type.

I did this by restructuring the code for the `Route.Builder`. Instead of taking `RoutePattern` and `HandlerAspect` to build a route later on with a `Handler`, it takes now a `RoutePattern` and a `Handler` to build a route with a `HandlerAspect`.
This leads to the fact that the `pattern -> aspect -> handler` syntax is gone and is now instead `pattern -> handler @@ aspect`.

I provided a method currently called `Routes.build`

```scala
def build[Env, Err, Ctx](
    route: Route.Builder[Env, _, Ctx, _, Err],
    routes: Route.Builder[Env, _, Ctx, _, Err]*,
  ): Routes.Builder[Env, Ctx, Err] =
    Routes.Builder[Env, Ctx, Err](route +: routes)
```
 And added a builder
 
 ```scala
   case class Builder[Env, Ctx, Err](routes: Seq[Route.Builder[Env, _, Ctx, _, Err]]){
    def @@(aspect: HandlerAspect[Env, Ctx]): Routes[Env, Err] =
      new Routes(Chunk.fromIterable(routes.map(_ @@ aspect)))
  }
 ```
 
 Which leads to the syntax
 
 ```scala
 Routes.build(
          Method.GET / "a"                  -> handler((ctx: AuthContext, _: Request) => Response.text(ctx.value)),
          Method.GET / "b" / int("id")      -> handler((id: Int, ctx: AuthContext, _: Request) =>
            Response.text(s"for id: $id: ${ctx.value}"),
          ),
          Method.GET / "c" / string("name") -> handler((name: String, ctx: AuthContext, _: Request) =>
            Response.text(s"for name: $name: ${ctx.value}"),
          ),
        ) @@ basicAuthContextM
```

Now we have basically two choices:

**We leave the code structure wise as it is right now** 

Which means constructing routes inside of the two constructors `Routes.apply` and `Routes.build` works just fine. You can use for both `pattern -> handler` with the same requirements as on the main branch, which means explicit types for the handler input.
What does not work anymore is
```scala
val route = Method.GET / "endpoint" -> handler { (_: Request) =>
  ZIO.fail(new Exception("hmm..."))
}
```
  This would need an explicit type to resolve to the right version of `->`
  
```scala
val route: Route[Any, Exception] = Method.GET / "endpoint" -> handler { (_: Request) =>
  ZIO.fail(new Exception("hmm..."))
}
val route: Route.Builder[Any, (Int, AuthContext), AuthContext, (Int, AuthContext, Request), Nothing] = Method.GET / "b" / int("id") ->
  handler((id: Int, ctx: AuthContext, _: Request) =>  Response.text(s"for id: $id: ${ctx.value}"),
)
```

**Have two different operators**
Constructing routes will use `->` for no context and another operator (maybe `-->` like in the prev version) for handlers with context.

Maybe the first solution could be more attractive, if we would have a way of simplifying the type parameters of `Route.Builder[-Env, PC, Ctx, In, +Err]` ?

Anyhow, I am sure this is a step forward overall.

@adamgfraser @vigoo Maybe you have some thoughts on this too.

Issue #2521, opened 2023-11-15
rendering of script html template is broken


987Nabil:
For code like
```scala
script(s"""myCoolJsMethod("myStringInput")""")
```

rendering this will cause the quotes to be escaped. This is not the correct behaviour for script.
We might need to special case script, since I think it is the only tag for which the content should not be escaped.

ezhil56x:
Hello @987Nabil 👋🏻
I'm interested to work on this 🙋🏻
I couldn't find [Contributor's guide](https://zio.dev/about/contributing) for this repo, link is broken on [README.md](https://github.com/zio/zio-http#readme)

987Nabil:
Hey @ezhil56x I think this is the right link
https://zio.dev/contributor-guidelines

Anyhow, just fork, clone and open a PR. I am sure you'll just be doing fine 🙂

Issue #2494, opened 2023-10-24
Create Middleware to provide SwaggerUI OpenAPI endpoint


987Nabil:
A middleware should be added, that creates an endpoint that returns an html document that renders a given `OpenAPI` instance.
We should not add any library dependencies for that, but rather use [unpkg](https://swagger.io/docs/open-source-tools/swagger-ui/usage/installation/).
We can give a default value for the swagger ui version, but should make it configurable.
API for the middleware should look somewhat like this

```scala
def swaggerUI(api: OpenAPI, path: String, version: String = "4.5.0")
```

This is blocked by #2470

PR #2556, opened 2023-12-12
Middleware for adding swagger ui endpoint (#2494)

987Nabil:
Html template script rendering fix (#2521)

I also removed 
```scala
  /**
   * Returns a new pattern that is extended with the specified segment pattern.
   */
  final def /[B](segment: SegmentCodec[B])(implicit combiner: Combiner[A, B]): PathCodec[combiner.Out] =
    self ++ Segment[B](segment)
```
Since it lead to code like `Middleware.swaggerUI("docs" / "openapi", openAPI)` being invalid.
Here `"docs" / "openapi"` was ambiguous, since there is a implicit conversion from string to `PathCodec` and to `SegmentCodec` and a `/` method for each of them. I saw that the  `SegmentCodec` method was internally used only in one place. And I don't think that ppl will explicitly create `SegmentCodec` in there user code.
It seems reasonable to only have one impl of `def /...`

### DIFF
--- a/zio-http-example/src/main/scala/example/EndpointExamples.scala
+++ b/zio-http-example/src/main/scala/example/EndpointExamples.scala
@@ -3,12 +3,13 @@ package example
 import zio._
 
 import zio.http.Header.Authorization
+import zio.http._
 import zio.http.codec.{HttpCodec, PathCodec}
+import zio.http.endpoint.openapi.{OpenAPIGen, SwaggerUI}
 import zio.http.endpoint.{Endpoint, EndpointExecutor, EndpointLocator, EndpointMiddleware}
-import zio.http.{int => _, _}
 
 object EndpointExamples extends ZIOAppDefault {
-  import HttpCodec._
+  import HttpCodec.query
   import PathCodec._
 
   val auth = EndpointMiddleware.auth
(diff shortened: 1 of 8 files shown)


zio-http/src/main/scala/zio/http/Middleware.scala
```diff
@@ -398,6 +399,89 @@ object Middleware extends HandlerAspects {
...
+   * Middleware.swaggerUI("docs" / "openapi", openAPIv1, openAPIv2) val
+   * routesWithSwagger = routes @@ swaggerUI }}}
+   *
+   * With this middleware in place, a request to `https://www.domain.com/[path]`
+   * would serve the Swagger UI. The different OpenAPI specifications are served
+   * at `https://www.domain.com/[path]/[title].json`. Where `title` is the title
+   * of the OpenAPI specification and is url encoded.
+   */
+  def swaggerUI(path: PathCodec[Unit], api: OpenAPI, apis: OpenAPI*): Middleware[Any] =
+    swaggerUI(path, "5.10.3", api, apis: _*)
```
vigoo: What is this hard-coded version here?

987Nabil: It is just a default version, that is the latest for now. I thought it might be nice to not have to look up a swagger ui version for a simple use case.

zio-http-example/src/main/scala/example/EndpointExamples.scala
```diff
@@ -36,7 +37,9 @@ object EndpointExamples extends ZIOAppDefault {
       }
     }
 
-  val routes = Routes(getUserRoute, getUserPostsRoute)
+  val openAPI = OpenAPIGen.fromEndpoints(title = "Endpoint Example", version = "1.0", getUser, getUserPosts)
+
+  val routes = Routes(getUserRoute, getUserPostsRoute) @@ Middleware.swaggerUI("docs" / "openapi", openAPI)
```
vigoo: I would like to ask the question why is this functionality a middleware? It is not modifying the behavior of existing routes just generates one or more new ones. Would not it be more logical if it is just a function returning `Routes` which can be appended to an existing `Routes`?

vigoo (approved):

PR #2558, opened 2023-12-13
Don't flatten nested case classes for JsonSchema

987Nabil:
Correctly generate references instead of inlining schemas


987Nabil:
@runtologist I think this were some of the problems you spotted on the initial PR

adamgfraser (approved):

PR #2562, opened 2023-12-20
OpenAPI to Endpoint code gen (#1520)

### LINKED ISSUE #1520: Implement code generator that takes OpenAPI and generates zio.http.api.EndpointSpec code

987Nabil:
Add general data structure and logic for any endpoint code gen

This PR solves the code gen, but does not offer any CLI, sbt or similar support.
This does not need to be done at once I think. Follow up tickets for that should be fine.

fixes #1520
/claim #1520

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
         run: sbt '++ ${{ matrix.scala }}' zioHttpShadedTests/test
 
       - name: Compress target directories
-        run: tar cf targets.tar zio-http-cli/target target zio-http/target zio-http-docs/target zio-http-benchmarks/target zio-http-example/target zio-http-testkit/target project/target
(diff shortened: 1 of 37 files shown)


jdegoes (approved):

jdegoes:
@987Nabil Good to merge when build is passing.

Issue #2559, opened 2023-12-15
Update the test kit so that it supports the current development style as supported by 3.0.0.RC4 or later


gmixa:
**What is not easy to do right now?**
Currently the http test kit does not support the latest Http Development Standard implemented in zio-http
**Describe the solution you'd like**
Supporting the use of Routes, Route definitions and the current handler implementation so that the test kit fits to the current implementation

**Why is it important?**
A Testkit should support testing to the most recent coding standards. Importance of tests should be clear

PR #2565, opened 2023-12-22
Partial solution to https://github.com/zio/zio-http/issues/2559

gmixa:
Hi,
I think I managed to create a partial solution for https://github.com/zio/zio-http/issues/2559.
Adding the same feature to the client side is too tricky for me but anyway I think its better than nothing


gmixa:
gmixa 
and Gerald Mixa is both me. I have no idea why I am listened twice

987Nabil:
This feature is already merged.

PR #2569, opened 2023-12-26
Schema based ops for en- and decoding bodies

987Nabil:
Currently, we leave transforming data from and to bodies more or less to the user.
We only understand basic types like `String` or `Chunk[Byte]`.

But since we have the endpoint API and a strong relationship with zio-schema, I think it makes sense, to have schema based operators, that can create and decode bodies.

* We don't need to add additional libs
* It will ease the fallback to the low-level API, when the user fails to do their task with the Endpoint API

This implementation is agnostic of the encoding (not json specific) and just requires an implicit `BinaryCodec`.

### DIFF
--- a/zio-http/src/main/scala/zio/http/Body.scala
+++ b/zio-http/src/main/scala/zio/http/Body.scala
@@ -25,6 +25,8 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import zio.stream.ZStream
 
+import zio.schema.codec.BinaryCodec
+
 import zio.http.internal.BodyEncoding
 
 /**
@@ -40,6 +42,23 @@ trait Body { self =>
    * other will be returned. Otherwise, the right body will be returned.
    */
   def ++(that: Body): Body = if (that.isEmpty) self else that
+
+  /**
+   * Decodes the content of the body as a value based on a zio-schema
+   * [[zio.schema.codec.BinaryCodec]].<br>
+   *
+   * Example for json:
+   * {{{
+   * import zio.schema.json.codec._
+   * case class Person(name: String, age: Int)
+   * implicit val schema: Schema[Person] = DeriveSchema.gen[Person]
+   * val person = Person("John", 42)
+   * val body = Body.from(person)
+   * val decodedPerson = body.as[Person]
+   * }}}
+   */
+  def as[A](implicit codec: BinaryCodec[A], trace: Trace): Task[A] =
+    asChunk.flatMap(bytes => ZIO.fromEither(codec.decode(bytes)))
 
   /**
    * Returns an effect that decodes the content of the body as array of bytes.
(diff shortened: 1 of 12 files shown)


zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -41,6 +43,23 @@ trait Body { self =>
...
+   * {{{
+   * import zio.schema.json.codec._
+   * case class Person(name: String, age: Int)
+   * implicit val schema: Schema[Person] = DeriveSchema.gen[Person]
+   * val person = Person("John", 42)
+   * val body = Body.from(person)
+   * val decodedPerson = body.as[Person]
+   * }}}
+   */
+  def as[A](implicit codec: BinaryCodec[A], trace: Trace): Task[A] =
```
jdegoes: Should this be `to` or `into`? `as` has a meaning in some ZIO contexts as `map(_ => a)`.

zio-http/src/main/scala/zio/http/Body.scala
```diff
@@ -150,6 +169,20 @@ object Body {
...
+   * [[zio.schema.codec.BinaryCodec]].<br> Example for json:
+   * {{{
+   * import zio.schema.codec.JsonCodec._
+   * case class Person(name: String, age: Int)
+   * implicit val schema: Schema[Person] = DeriveSchema.gen[Person]
+   * val person = Person("John", 42)
+   * val body = Body.from(person)
+   * }}}
+   */
+  def from[A](a: A)(implicit codec: BinaryCodec[A], trace: Trace): Body =
```
jdegoes, line 188: The fact that this is called `from` almost does imply `to` or `into`.

zio-http/src/main/scala/zio/http/Body.scala
987Nabil: I chose the name because it is called `queryAs` if you want to get a typed query param

987Nabil: Actually I am fine with both. I can change it next time a PC is in reach

jdegoes: Maybe that should be `queryTo` or `queryInto` for consistency?

jdegoes (changes requested):
Good to merge with a name change, including one to `queryAs` (queryTo / queryInto). Thanks for your work on this!

zio-http/src/main/scala/zio/http/Body.scala
987Nabil: Should it then also be `body.toString` and not `body.asString`? But this has an obvious conflict.

987Nabil:
@jdegoes please take a look again. I changed too much to feel go to merge without an additional review

zio-http/src/main/scala/zio/http/Body.scala
jdegoes: Agreed, please update any other cases like this that you find.

jdegoes (approved):

PR #2574, opened 2023-12-31
Sandbox without absolving Throwables

yisraelU:
adds an orDie method to remove typed errors from Route and routes where the error is <:< Throwable

### DIFF
--- a/zio-http/src/main/scala/zio/http/Route.scala
+++ b/zio-http/src/main/scala/zio/http/Route.scala
@@ -270,6 +270,18 @@ sealed trait Route[-Env, +Err] { self =>
   final def sandbox(implicit trace: Trace): Route[Env, Nothing] =
     handleErrorCause(Response.fromCause(_))
 
+  /**
+   * Transforms a Route[Env, Throwable] into a Route[Env, Nothing] by shifting
+   * all typed errors into defects
+   */
+  final def orDie(implicit ev1: Err <:< Throwable, ev2: CanFail[Err], trace: Trace): Route[Env, Nothing] =
+    self match {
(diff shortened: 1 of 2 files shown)


987Nabil:
I am not sure, this is a good idea. It seems confusing for me. The contract of zio-http error handling is, that we always return a response. One might not like the default behavior of sandbox, but there is iirc an even more hidden way, that I am not sure a user can always influence, if there is a defect.

Can you elaborate on your use case and why the already existing ways to handle errors don't fit?

yisraelU:
Hi @987Nabil , sure. I totally get that this is not the standard usecase.
These methods can be of benefit when interoping with libraries that I don't control and don't expose enough capabilities.
The idea is that many libraries rely on Exception throwing as a means of passing information. I want my throwables to be interpreted to Responses , however I do not want to use the default ```Response.fromCause(_)``` , nor do I have the local context to  provide such logic 
Essentially what I am currently doing is a 
```handleError(throw _).toHttpApp```

and allowing the library to convert the throwable to a Response

jdegoes:
@yisraelU I think that @adamgfraser recently added a method that may help you out here.

987Nabil:
@yisraelU did you build this changes locally and check that they do want you want? Because seeing the code, I am not sure they do. I think we might do some defact to response logic deep in the guts of zio-http.

That said, I am not sure, that map error throw is not good enough for your use case and we should rather keep a misleading or misbehaving orDie away from the user. Btw I can't test around my self, I am just checking from the phone 😉

yisraelU:
@jdegoes nice. yes thats great

yisraelU:
@987Nabil No I didnt test locally.

PR #2581, opened 2024-01-03
Allow creating a body from `Array[Byte]`

kyri-petrou:
I was doing some profiling on a simple "Hello world" app for large byte payloads (~65kb), and one of the things that stood out was that there was a very big memory allocation due to the copying of the `Chunk[Byte]` into an `Array[Byte]` (see screenshot at the bottom).

I think that this allocation can be avoided in 1 of 2 ways:

1. Extract the underlying `Array[Byte]` from `Chunk.Arr[Byte]` in the `UnsafeBytes#unsafeAsArray` method
2. Expose a new method allowing users to create a body directly via an array

In this PR, I went with (2) mostly because it eliminates the possibility of the array wrapped in a Chunk to be mutated. Having said that, (1) would not require changes in user's code to see the performance improvement so I'm happy to change this PR and go with option 1 instead.

I also added some extra benchmarks to showcase that besides the memory allocations, this change also improves the throughput of responses with large payloads. In the case of a 100kb payload, that's ~5% improvement in throughput. Note that the copying of the array happens _prior_ to any compression taking place, so it's likely that many users will see some benefit from this change

```
[info] Benchmark                                           Mode  Cnt      Score    Error  Units
[info] ServerInboundHandlerBenchmark.benchmarkLargeArray  thrpt    5  12104.157 ± 78.598  ops/s
[info] ServerInboundHandlerBenchmark.benchmarkLargeChunk  thrpt    5  11519.093 ± 99.231  ops/s
```

Flamegraph:

image

### DIFF
--- a/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/ServerInboundHandlerBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zhttp.benchmarks/ServerInboundHandlerBenchmark.scala
@@ -13,10 +13,39 @@ import sttp.client3.{HttpURLConnectionBackend, UriContext, basicRequest}
 @BenchmarkMode(Array(Mode.Throughput))
 @OutputTimeUnit(TimeUnit.SECONDS)
 class ServerInboundHandlerBenchmark {
+  private val random      = scala.util.Random
+  random.setSeed(42)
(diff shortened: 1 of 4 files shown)


jdegoes:
@kyri-petrou Good to merge when CI Is passing!

987Nabil (approved):

Issue #2039, opened 2023-03-14
Adding more options to the SSL configuration


user5100:
**Is your feature request related to a problem? Please describe.**
Hi all, I would need to have more parameterization of the SSL context than it is currently available.  Basically I would need the server to also use a trustStore, to enable/disable mutual authentication between client and server, and to define other additional options, including setting a custom cipher suite.

**Describe the solution you'd like**
The solution I would need should allow the following options to be added to the ssl configuration:
- Add an option to directly pass objects of type javax.net.ssl.keyManagerFactory and javax.net.ssl.trustManagerFactory (both for both client and server contexts)
- Add an option to allow the creation and parameterization of a javax.net.ssl.SSLEngine, so as to indicate enabled cipher suites, enabled ssl protocols, client authentication, etc...
- Allow the SSL context to be changed during application execution (although I do not know if it is already possible to do this)

**Describe alternatives you've considered**
none

**Additional context**
Thank you for the support

jdegoes:
We need to think of a way to do this in such a fashion as to minimize the imprint of the JVM in the public API, for Scala Native, etc.

LannyRipple:
Could also use this feature.  My company has implemented a custom KeyStore to be update-aware of short-term credentials injected into K8S pods.  No telling what else they are doing under the hood but we're supposed to use it to initialize a KeyManagerFactory for SSLContext initialization.  We'd like zio-http instead of our current akka-http but this is a showstopper short of forking zio-http for an in-house workaround.

jdegoes:
@LannyRipple How would you like to specify this? `KeyManagerFactory` and `SSLContext` are not Netty-specific so could be used in ZIO HTTP.

LannyRipple:
I'm a bit at a loss there.  We aren't using Scala Native and such so for us it's just adding to ClientSSLConfig and modifying ClientSSLConverter to do the needful and not worrying that our zio-http would be leaking JVM types.  Sadly there's no way my team will agree to us maintaining a fork (I don't blame them).

Looking at my company's example code (SSL/PKI newbie here) it looks like we need
```scala
keyStoreInstanceName: String
keyStoreInputStreamtoLoad: => InputStream
// keyManagerFactoryAlgorithm -- going with default for now
// sslContextInstanceVersion: String -- going with latest.  suspect this doesn't change much
```
[...]

PR #2583, opened 2024-01-04
Adding more options to the SSL configuration (#2039)

LannyRipple:
* Provides ClientSSLConfig FromJavaxNetSsl to allow configuring client SSLContext using a javax.net.ssl.KeyManagerFactory and/or a javax.net.ssl.TrustManagerFactory.
* Only partially addresses #2039.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/ClientSSLConverter.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/ClientSSLConverter.scala
@@ -27,6 +27,7 @@ import zio.http.ClientSSLConfig
 
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory
 import io.netty.handler.ssl.{SslContext, SslContextBuilder}
+
 object ClientSSLConverter {
   private def trustStoreToSslContext(trustStoreStream: InputStream, trustStorePassword: Secret): SslContext = {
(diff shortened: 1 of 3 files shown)


LannyRipple:
Attempt to address some of Client SSL Configuration using javax.net.ssl.  Leaving this as draft for guidance and criticism.

Things I don't like or am not sure of:
* My naming of variables and types.
* ClientSSLConfig.FromJavaxNetSsl has odd builder() entry points.  Needed to ensure a valid configuration.  Any better way?
* Implemented tests but ClientHttpsSpec (and my extension) ignores all tests.

987Nabil:
@LannyRipple I changed the secrets from `String` to `Secret` so you have to solve some conflicts please 🙂

987Nabil:
And the `ClientHttpsSpec` is broken, we had before a config for a website that is no longer available via https. Therefore we deactivated the tests to fix CI. You can remove the `@@ ignore` to test it. If you can fix the test too, we would be very grateful 🙏

LannyRipple:
@987Nabil Thanks for eyeballing and the feedback.  Will fix these up.

jdegoes:
@LannyRipple Looks good to merge! Please re-open when conflicts are fixed and the build is passing. Thank you!

LannyRipple:
Thanks.  Will do when I can find the time for tests (even if just local).

Issue #2576, opened 2023-12-31
depricated links in readme


devansh-m12:
Changes Overview:
Update the link on the readme of contributing page to direct to the correct resource.

Context:
The existing link in the readme page had become outdated, causing confusion for contributors seeking essential information. 

Previous Link:
https://zio.dev/about/contributing

Updated Link: (v2 for contributor guidelines)
https://zio.dev/contributor-guidelines/

PR #2587, opened 2024-01-06
Update README.md

neo773:
/claim #2576


jdegoes:
README cannot be updated directly, but index.md must be updated, whose content is fed into README.md. Please re-open if fixed.

PR #2588, opened 2024-01-06
updated readme link

guptadeepak8:
/claim #2576


jdegoes:
README cannot be updated directly, but index.md must be updated, whose content is fed into README.md. Please re-open if fixed.

PR #2591, opened 2024-01-07
Implement QueryParam ordering

### LINKED ISSUE #2586: Maintain query parameter ordering in QueryParams
```scala
trait QueryParams {
  def keys: Iterable[String]

  def get(key: String): Option[Chunk[String]]

  def getOrElse(key: String, default: Chunk[String]): Chunk[String]
}
```

With a `ListMap` implementation and use of it on ingest from Netty to preserve ordering.

easel:
Extracts QueryParams as a trait, with a concrete ListMapQueryParams implementation by default.

/claim #2586


easel:
@jdegoes I lifted out most of QueryParams to a trait, with a concrete ListMapQueryParams implementation that is used in the QueryParams companion object for the various .apply methods. I also folded in the suggestions from the previous PR. 

Is this what you had in mind?

kyri-petrou:
I know this PR is still in draft, but I was curious to see the changes as I was thinking of attempting the ticket but was too late😅. I would highly recommend to use a different underlying implementation than a ListMap for this. From `ListMap`s scaladoc:

>This object provides a set of operations to create values.
Note that each element insertion takes O(n) time, which means that creating a list map with n elements will take O(n2) time. This makes the builder suitable only for a small number of elements.

I understand that query params are usually very few, but that's not guaranteed. Instead of a ListMap, I would perhaps consider using a `java.util.LinkedHashMap` instead

easel:
> I know this PR is still in draft, but I was curious to see the changes as I was thinking of attempting the ticket but was too late😅. I would highly recommend to use a different underlying implementation than a ListMap for this. From `ListMap`s scaladoc:
> 
> > This object provides a set of operations to create values.
> > Note that each element insertion takes O(n) time, which means that creating a list map with n elements will take O(n2) time. This makes the builder suitable only for a small number of elements.
> 
> I understand that query params are usually very few, but that's not guaranteed. Instead of a ListMap, I would perhaps consider using a `java.util.LinkedHashMap` instead

It's a possibility. The underlying Netty decoder creates a LinkedHashMap, so using the java type inside QueryParams would avoid any construction overhead for some use cases. I went with the ListMap because the existing implementation exposed a scala.collections.immutable.Map` in a couple places outside QueryParams. My thinking was there would be multiple accesses of the immutable collections throughout the lifecycle of the QueryParams and therefore it would be best to pay the conversion price once on construction, sticking as close to the existing implementation while maintaining ordering.

If that's not the case, we could drive down the cost of construction and only convert to an immutable collection on use. Lifting out QueryParams as a trait is a first step towards that. As things stand, there are only a few uses of the `.map` member, so we could likely scrub them out of the trait entirely and work towards an optimized implementation with appropriate performance tests.

My hunch is that we need some robust benchmarks to know the best balance between construction overhead, usage overhead, and memory overhead.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -169,37 +159,111 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
...
    */
   def toForm: Form = Form.fromQueryParams(self)
+
 }
 
-object QueryParams {
+/**
+ * A ordered collection of query parameters.
+ */
+final case class ListMapQueryParams(map: ListMap[String, Chunk[String]]) extends QueryParams {
```
jdegoes: I would move this into `QueryParams` companion object and make it package private to zio.http (e.g. `private[http]`). 

Why? 

To avoid polluting the top-level package with something no one will ever use directly, or even know about.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -169,37 +159,111 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
...
 
-  def apply(tuples: (String, Chunk[String])*): QueryParams =
-    QueryParams(map = Chunk.fromIterable(tuples).groupBy(_._1).map { case (key, values) =>
-      key -> values.flatMap(_._2)
+  /**
+   * Combines two collections of query parameters together. If there are
+   * duplicate keys, the values from both sides are preserved, in order from
+   * left-to-right.
+   */
+  override def ++(that: QueryParams): QueryParams =
```
jdegoes: Would be good to add a test that `++` is order-preserving.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -26,41 +29,31 @@ import zio.http.internal.QueryParamEncoding
...
    */
-  def addAll(key: String, value: Chunk[String]): QueryParams = {
-    val previousValue = map.get(key)
-    val newValue      = previousValue match {
-      case Some(prev) => prev ++ value
-      case None       => value
-    }
-    QueryParams(map.updated(key, newValue))
-  }
+  def addAll(key: String, value: Chunk[String]): QueryParams
```
jdegoes: I would not leave this abstract. Provide a default implementation in terms of `++` and the `QueryParam.apply` constructor.

zio-http/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -72,16 +65,12 @@ final case class QueryParams(map: Map[String, Chunk[String]]) {
    */
   def encode(charset: Charset): String = QueryParamEncoding.default.encode("", self, charset)
 
-  override def equals(that: Any): Boolean = that match {
-    case that: QueryParams => self.normalize.map == that.normalize.map
-    case _                 => false
-  }
+  def equals(that: Any): Boolean
```
jdegoes: Leaving this abstract seems to imply that subclasses are free to make their own definition of equality. I would say that's not true. We should implement this via `this.map == that.map`, i.e. rely on `Map` equality, which is order-insensitive. Maybe add a test for that.

jdegoes (changes requested):
Overall, looks fantastic! Just a few minor tweaks to clean things up for merging.

jdegoes:
I agree with the comment about using `java.util.LinkedHashMap`.

zio-http/src/main/scala/zio/http/QueryParams.scala
easel: Makes sense. Implementation coming.

easel:
> I agree with the comment about using `java.util.LinkedHashMap`.

Ok, I'm going to move the concrete class into the companion object and make the underlying map a private j.u.LinkedHashMap. The `.map` method will implement the conversion to a scala map. I'm leaning towards keeping the implementation of `.map` as a ListMap so we don't throw out ordering, but could be convinced otherwise. At first inspection, it looks like there's very few usages of `.map`, so it might be worth considering excising it from the interface entirely. @jdegoes any opinion there? The current usages I'm seeing are these:

https://github.com/zio/zio-http/blob/e8dd3517e0ad8b46bad0b2d0ebe906e0be1e2185/zio-http/src/main/scala/zio/http/Form.scala#L214

https://github.com/zio/zio-http/blob/cff63d8150fc3aee9d363859985c1b19966f397e/zio-http/src/main/scala/zio/http/netty/NettyQueryParamEncoding.scala#L50

In both cases, the map is just a handle to iterate over the keys, so exposing a sequence of kv pairs ought to suffice for iteration, and leave us room in the future for how to implement the map interface via the various `get` methods. 

```
def seq: Seq[(String, Seq[String])]
```

or 
```
def toChunk: Chunk[(String, Chunk[String])]
```

My inclination would be to expose the `seq` above to avoid conversion to chunk.

In an ideal scenario, the rest of the concrete methods will be implemented directly against the underlying java map. I'll see how that all works out.

zio-http/src/main/scala/zio/http/QueryParams.scala
easel: This bears some discussion. Current implementation (now) implements equals as `this.toChunk == that.toChunk` and `.hashcode` as `this.toChunk.hashCode`. My initial inspection indicates that toChunk should be quite efficient,   given `Chunk.fromIterable` creates "Iterable backed chunks" rather than recreating the collection.

After getting all the wiring in place, I was able to scrub out `.map` entirely in favor of implemented `.getAll` directly against the underlying container. All of the mutations are now implemented in the trait as immutable transformations based on `.toChunk` -- recreating the `LinkedHashMap` for every change is necessary to work around it's mutability anyway, and it makes for a nice pure scala implementation.

This is a long way of saying what's the true and correct meaning of `.equals`? Given ordering is sometimes necessary, it seems an ordering-sensitive equality check is still appropriate, but I think the counter-argument can be made.

zio-http/src/main/scala/zio/http/QueryParams.scala
easel: Done.

zio-http/src/main/scala/zio/http/QueryParams.scala
easel: Good call, done.

zio-http/src/main/scala/zio/http/QueryParams.scala
easel: Done, but fully private to the companion object. The apply methods are all there so it works out.

easel:
Ok. I believe all the review feedback has been accounted for. I'm fairly happy with where the implementation landed. There are now implementations for *every* method except `.toChunk` on the QueryParams trait, with just a few performance overrides on `JavaLinkedHashMapQueryParams`. As long as `.toChunk` is pretty fast, so too should be everything else. 

There are probably some usage scenarios where a `ChunkQueryParams` implementation could be faster, but without a bunch of benchmarking to determine the cutover(s) and given that Netty creates a `java.util.LinkedHashMap`, I think this implementation should be quite optimal for most use cases, especially given the common case is likely an empty query params with no mutation and an emptiness check or reading a single parameter.

zio-http/src/main/scala/zio/http/QueryParams.scala
easel: Both `.equals` and `.hashCode` are implemented against `.normalize.toChunk` now.

jdegoes:
@easel 

>  At first inspection, it looks like there's very few usages of .map, so it might be worth considering excising it from the interface entirely. @jdegoes any opinion there?

I agree with this. Note, however, that if you are backing with j.u.c, then you can avoid the overhead of boxing the tuples by exposing `Seq[j.u.c.Map.Entry[String, Chunk[String]]]`. Unfortunately that's mutable, but you could work around that issue by making the `def seq` package-private (e.g. `private[http] def seq: ...`). 

What do you think?

easel:
> @easel
> 
> > At first inspection, it looks like there's very few usages of .map, so it might be worth considering excising it from the interface entirely. @jdegoes any opinion there?
> 
> I agree with this. Note, however, that if you are backing with j.u.c, then you can avoid the overhead of boxing the tuples by exposing `Seq[j.u.c.Map.Entry[String, Chunk[String]]]`. Unfortunately that's mutable, but you could work around that issue by making the `def seq` package-private (e.g. `private[http] def seq: ...`).
> 
> What do you think?

Good idea. I just pushed an implementation that now relies on `private[http] def seq: Seq[util.Map.Entry[String, util.List[String]]]` for the immutable transforms and seems to pass tests. It ought to be faster as it avoids the tuple boxing and wrapping the values in Chunk for most of the intermediate stages.

The remaining issue, in my mind, is exposing some sort of interface to finding *all* the parameters. The previous implementation exposed it's `map`, and given ListMap can maintain ordering while exposing a map interface and the `.get` methods on the QueryParams should be the preferred implementation for performance, I've created a new `.map` as the following. WDYT?

```
  /**
   * All query parameters as a map. Note that by default this method constructs the map on
   * access from the underlying storage implementation, so should be used with care. Prefer 
   * to use `getAll` and friends if all you need is to access the values.
   */
  def map: Map[String, Chunk[String]] = ListMap.from(seq.map { entry =>
    (entry.getKey, Chunk.fromIterable(entry.getValue.asScala))
  })
```

jdegoes:
@easel Looks good to me!

jdegoes (approved):

easel:
Thanks for the feedback and merge @jdegoes !

Issue #2584, opened 2024-01-05
Client hangs on a streamed response if an exception is raised  


dacr:
**Describe the bug**
no response from the server if an exception is raised when the stream is created (from a java stream in my particular context)  

**To Reproduce**

Once this script started :
```scala
// ---------------------
//> using scala  "3.3.1"
//> using dep "dev.zio::zio-http:3.0.0-RC4"
// ---------------------

import zio.*, zio.http.*, zio.stream.*
import java.util.stream.{Stream => JStream}

object Main extends ZIOAppDefault {
  def getJavaStream: Task[JStream[String]] = ZIO.attempt {
    throw RuntimeException("failure")
    //JStream.of("1", "2", "3", "4")
  }

  override def run =
    for {
      _     <- ZIO.logInfo("started")
      stream = ZStream.fromJavaStreamZIO(getJavaStream).flatMap(s => ZStream.fromIterable(s.getBytes))
      app    = Routes(Method.GET / "stream" -> handler(Response(body = Body.fromStream(stream)))).toHttpApp
      _     <- Server.serve(app).provide(Server.default)
    } yield ()
}

Main.main(args)
```

Then if you run `curl -v http://127.0.0.1:8080/stream`, you'll get : 
```
$ curl -v http://127.0.0.1:8080/stream
* Uses proxy env variable no_proxy == '*.orange-labs.fr,localhost,127.0.0.1,127.0.0.*,10.100.*,*.intraorange,*.francetelecom.fr,*.itn.ftgroup,www.forge.orange-labs.fr,repos.tech.orange,*.tech.orange,welcome.com.intraorange'
*   Trying 127.0.0.1:8080...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET /stream HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< transfer-encoding: chunked
< 


HANG NO ERROR RESPONSE RECEIVED
```

**Expected behaviour**

I expect to get an http error code such as error 500.

**Additional context**

The following exception is raised : 
```
20:09 $ scl zio-learning-zhttp-server-streaming-2.sc 
[...]

PR #2599, opened 2024-01-07
Fix hanging on stream response errors

kyri-petrou:
/claim #2584

The reported behaviour was due to the channel not being closed when the stream errored. There are 2 ways that we can recover from such issues without causing clients to hang:

1. Handle the error, potentially logging it (not added to this PR but happy to add it) and closing the netty channel
2. Don't handle the error and let netty deal with it. Downside to this is that it causes netty to log a lot of errors which can get very noisy

With both approaches, what the client receives is the same response. Note that we cannot change the initial 200 OK response that was sent to the client.

```
curl -v http://127.0.0.1:8080/stream

*   Trying 127.0.0.1:8080...
* Connected to 127.0.0.1 (127.0.0.1) port 8080
> GET /stream HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/8.4.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< transfer-encoding: chunked
< 
* transfer closed with outstanding read data remaining
* Closing connection
curl: (18) transfer closed with outstanding read data remaining
```

### Alternative approach (requires user-facing breaking changes)

We could potentially change the `Body.fromStream` / `Body.fromStreamX` methods to force users to handle streaming errors (i.e., `ZStream[R, Nothing, A]`) and in cases of defects, leave it up to Netty to handle them. This would very likely require changes to users' code, but this option is less likely to result to unexpected behaviour due to ZStream failures

### DIFF
--- a/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -323,11 +323,13 @@ private[zio] final case class ServerInboundHandler(
             }
             None
           }
-        }.flatMap(_.getOrElse(ZIO.unit)).catchSomeCause { case cause =>
-          ZIO.attempt(
-            attemptFastWrite(ctx, withDefaultErrorResponse(cause.squash)),
-          )
-        }
+        }.foldCauseZIO(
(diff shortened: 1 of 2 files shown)


jdegoes (approved):

jdegoes:
The change to body to prevent users from using typed errors may be a good one.

PR #2602, opened 2024-01-07
Server doesn't print unexpected error logs (#2491)

### LINKED ISSUE #2491: Server doesn't print unexpected error logs
**Describe the bug**
The server doesn't print error logs when exceptions were thrown.

**To Reproduce**
Scala 3:
```Scala
import zio.http.*
import zio.{Scope, ZIO, ZIOAppArgs, ZIOAppDefault, ZLayer}

object Test extends ZIOAppDefault {
  val app = Http.collectZIO[Request] { case Method.GET -> Root / "test" => throw RuntimeException("test error") }

  override def run: ZIO[Any with ZIOAppArgs with Scope, Any, Any] = {
    Server
      .serve(app)
      .provide(
        ZLayer.succeed(Server.Config.default.enableRequestStreaming.port(8080)),
        Server.live
      )
  }
}
```
When send a request to `http://localhost:8080/test`, it responds empty body with 500 status code, and no error logs in server.

**Expected behaviour**
It should print error logs in server.

cheikhsimsol:
fixes #2491 
/claim #2491

The approach here is to update the fromCause

**[Update]** Managed to add some form of testing.

Capture d’écran (11)


vigoo:
I think this should be done through ZIO's logging instead of directly printing to the stderr, so it gets integrated with all the logs of a server.

cheikhsimsol:
Ok, I'll update it to use it.

cheikhsimsol:
To confirm @vigoo, you're referring to zio.Console or zio.logging?

vigoo:
No, to `ZIO.logError` etc (https://zio.dev/reference/observability/logging/)

cheikhsimsol:
@vigoo I've updated it to use logError, please check and confirm.

vigoo:
Sorry previously I did not properly check where the logging is added and only reacted to how is it logging. 

I would not add any logging into constructors of a `Response` - those are just data structures and creating a response should not do any side effect. .

Beside that there are two other problems as well:
- Using `Runtime.default` is not correct, we need to use the runtime the server layer was created with. That way the logging effect will use the configured loggers.
- There is a `printStackTrace` as well which should also be logged through ZIO logging. 
 
Basically both the `Throwable` and the `Cause` case should be logged with `ZIO.logErrorCause`  using the runtime the server layer was created with.

I will take a look at the codebase and recommend a better place for this in my next comment.

vigoo (changes requested):
One place is https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala#L91 which happens when we were not even able to decode the request. Here we have `runtime: NettyRuntime` available to run the ZIO effect (`runtime.run`).

The other one is more interesting, as the `sandbox` (https://github.com/zio/zio-http/blob/main/zio-http/src/main/scala/zio/http/Handler.scala#L543) transforms a handler in a way that any error (`Cause`) gets transformed into a `Response`. This is still not the place to actually log an error though, as this just transforms the `Handler`, which is a _description_ of how to handle a request.

But this could be changed to `mapErrorCauseZIO` and do the logging there.

An additional question could be whether we always want to do this logging or not. I'm not sure about it - maybe there could be a fiber ref to configure it? @jdegoes what do you think?

cheikhsimsol:
@vigoo I requested the review by mistake, no need to check.

vigoo:
> An additional question could be whether we always want to do this logging or not. I'm not sure about it - maybe there could be a fiber ref to configure it? @jdegoes what do you think?

Or maybe it could be always logged in the invalid request case (or configured via server config) and there could be a separate `sandboxLogged` or similar variant that does the logging.

cheikhsimsol:
With one approach, I tried to throw an exception, although it logs the error, all of the tests break. I updates the test itself to catch and read the exception. However, I'm using printStackTrace to get around needing the runtime.. I'm new to Scala so unsure if this will be ideal.

jdegoes:
@vigoo If there will be performance cost, it's important that it be optional.

987Nabil:
@cheikhsimsol can you update your PR please?

cheikhsimsol:
> @cheikhsimsol can you update your PR please?

Please give me some additional time, I'll try to re-visit this during the weekend.

jdegoes:
@cheikhsimsol Please re-open if you get a chance to fix this up!

PR #2619, opened 2024-01-14
Fixed #2373: don't propagate ReadTimeoutExceptions

### LINKED ISSUE #2373: Handle client idle timeout to prevent abundant logging

ekhov:
On idle timeout for NettyConnectionPool handle ReadTimeoutExceptions- just log them with debug level, so they don't propagate as unhandled errors further

#2373 
/claim #2373


project/Dependencies.scala
```diff
@@ -16,6 +16,8 @@ object Dependencies {
 
   val scalafmt = "org.scalameta" %% "scalafmt-dynamic" % "3.7.17"
 
+  val `scala-logging` = "com.typesafe.scala-logging" %% "scala-logging"   % "3.9.5"
```
vigoo: We should not depend on this library, ZIO has it's own built-in logging API

ekhov: Used built-in logging

zio-http/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -127,9 +125,24 @@ object NettyConnectionPool {
     } yield result
   }
 
+  private final class ReadTimeoutErrorHandler(nettyRuntime: NettyRuntime)(implicit trace: Trace)
+      extends ChannelInboundHandlerAdapter {
```
ekhov, line 129: Also I had an idea of implementing `ReadTimeoutHandler` itself and rewrite `readTimedOut` method, so instead of raising an exception, it would just write a log message. But I'm not sure which way is more correct/idiomatic

jdegoes (approved):

PR #2624, opened 2024-01-15
Speedup test suite execution

kyri-petrou:
While working on a different PR, I realised that many of the test suites were using the default Netty config, which has a long `shutdownQuietPeriodDuration`. This means that tests take a long time to execute due to waiting for the server / client to shut down.

In this PR, I changed the test suites (as many as I could find) to use `NettyConfig.defaultWithFastShutdown`. Running locally, the `zioHttp` module tests now execute in 1 minute (down from 3 minutes) which should make other devs a bit happier as they won't have to wait as long for CI to finish & potentially help with CI costs


jdegoes (approved):

jdegoes:
🙏

PR #2631, opened 2024-01-17
Intoroduce query parameters with multiple values to Endpoint API

ekhov:
Introduce query codec for multi value query parameters. Now `HttpCodec.Query[A]` returns `Chunk[A]` as a result value. Add new query codecs to describe these multi value parameters to `QueryCodecs` preserving old ones(`query` etc), which are now a transformed versions of multi value codecs.

Now single value codecs fail if endpoint receives multiple values for this query

/claim #2321

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
@@ -21,24 +21,16 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
 private[codec] trait QueryCodecs {
 
   def query(name: String): QueryCodec[String] =
-    HttpCodec
-      .Query(name, TextCodec.string)
-      .transform[String] { (c: NonEmptyChunk[String]) => c.head }(s => NonEmptyChunk(s))
+    toSingleValue(HttpCodec.Query(name, TextCodec.string))
 
   def queryBool(name: String): QueryCodec[Boolean] =
-    HttpCodec
-      .Query(name, TextCodec.boolean)
-      .transform { (c: NonEmptyChunk[Boolean]) => c.head }(s => NonEmptyChunk(s))
+    toSingleValue(HttpCodec.Query(name, TextCodec.boolean))
 
   def queryInt(name: String): QueryCodec[Int] =
-    HttpCodec
-      .Query(name, TextCodec.int)
-      .transform { (c: NonEmptyChunk[Int]) => c.head }(s => NonEmptyChunk(s))
+    toSingleValue(HttpCodec.Query(name, TextCodec.int))
 
   def queryTo[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] =
-    HttpCodec
-      .Query(name, codec)
-      .transform { (c: NonEmptyChunk[A]) => c.head }(s => NonEmptyChunk(s))
+    toSingleValue(HttpCodec.Query(name, codec))
 
   def queryMultiValue(name: String): QueryCodec[NonEmptyChunk[String]] =
     HttpCodec.Query(name, TextCodec.string)
@@ -51,5 +43,8 @@ private[codec] trait QueryCodecs {
 
   def queryMultiValueTo[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[NonEmptyChunk[A]] =
     HttpCodec.Query(name, codec)
+
+  private def toSingleValue[A](queryCodec: QueryCodec[NonEmptyChunk[A]]): QueryCodec[A] =
+    queryCodec.transform { (c: NonEmptyChunk[A]) => c.head }(s => NonEmptyChunk(s))
 
 }
--- a/zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
@@ -23,22 +23,22 @@ private[codec] trait QueryCodecs {
   def query(name: String): QueryCodec[String] =
     HttpCodec
       .Query(name, TextCodec.string)
-      .transform[String] { c: NonEmptyChunk[String] => c.head }(s => NonEmptyChunk(s))
+      .transform[String] { (c: NonEmptyChunk[String]) => c.head }(s => NonEmptyChunk(s))
 
   def queryBool(name: String): QueryCodec[Boolean] =
     HttpCodec
       .Query(name, TextCodec.boolean)
-      .transform { c: NonEmptyChunk[Boolean] => c.head }(s => NonEmptyChunk(s))
+      .transform { (c: NonEmptyChunk[Boolean]) => c.head }(s => NonEmptyChunk(s))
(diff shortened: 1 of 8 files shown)


zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -591,16 +591,37 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
 
     def index(index: Int): ContentStream[A] = copy(index = index)
   }
-  private[http] final case class Query[A](name: String, textCodec: TextCodec[A], index: Int = 0)
-      extends Atom[HttpCodecType.Query, A]  {
-    self =>
+
+  private[http] sealed case class Query[A](name: String, textCodec: TextCodec[_], index: Int = 0)
+      extends Atom[HttpCodecType.Query, A] { self =>
```
jdegoes: I don't think you should add `QuerySingleValue` or `QueryMultipleValue`. Rather, I would just modify `Query` as follows:

```suggestion
      extends Atom[HttpCodecType.Query, Chunk[A]] { self =>
```

This way the base case produces zero or more values. Then you can derive all the others from this one.

ekhov: Unfortunately, with this approach I couldn't find a way to preserve single value codecs. All `QueryCodecs` will start to produce `Chunk[A]` as a result value, it will break compatibility as well. 

`Chunk[A]` could internally mean 0 to many query params- yes, but how on the level of Endpoint API describe that this codec should expect only 1, or for many params. How to preserve a way to describe that `Endpoint`'s `Input` type is `A` and not `Chunk[A]` without adding new `HttpCodec` for query or creating wrappers around `HttpCodec.Query[A]`?

> Then you can derive all the others from this one.

Could you please explain a little bit more your vision on how we can resolve single value codecs type wise?

ekhov: Looked at `HttpCodec` once more and now I see that it has `transform` method which could do exactly what I need.

jdegoes: Yes, no need to break backward compatibility, the API can remain exactly the same, because `transform` lets you go from `Chunk[A]` to `A`, and then fail if there is zero (or more than one).

jdegoes:
There is a [Scala 3](https://github.com/zio/zio-http/actions/runs/7567427765/job/20606560288?pr=2631#step:6:987) build failure that is failing the build (maybe among other things).

zio-http/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -592,7 +592,7 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
     def index(index: Int): ContentStream[A] = copy(index = index)
   }
   private[http] final case class Query[A](name: String, textCodec: TextCodec[A], index: Int = 0)
-      extends Atom[HttpCodecType.Query, A]  {
+      extends Atom[HttpCodecType.Query, NonEmptyChunk[A]] {
```
jdegoes: I would not do `NonEmptyChunk`, because it is valid query string: `foo&bar&baz`. In such case, the chunk will be empty. The RFC explicitly allows this.

ekhov: If we're talking about query codec for `String` values, It seems that currently it work a little bit differently. If query string contains expected query parameter field which doesn't have associated value, this query parameter will be parsed as an empty string. So even with `NonEmptyChunk` we get non empty chunk with empty strings.

For query parameters of other types such query(`foo&bar&baz`) string doesn't work.

ekhov: Do we want add an optionality by default to multi value parameters? I mean if no multi value query parameters with given name are found, should this case produce an empty chunk or should it produce an error? Because it seems like only in this case of having default optionality it makes sense to change `NonEmptyChunk` to `Chunk`

jdegoes: - If there is a name, but no value, then it should produce empty chunk
 - If there is no name, it should not produce an entry
 - If there is a name and one or more values, it should produce a non empty chunk

This allows us to (1) fail when it's missing, (2) succeed when it's present but no values, (3) fail on more than one values, etc.

ekhov: @jdegoes Please note that failing when single value parameter actually receives multiple query values will break compatibility. Endpoints at the moment successfully handle requests with multiple query values- they just get the first value

987Nabil: @ekhov it is okay to break compatibility at this point. We explicitly do not guarantee it before 3.0.0 final. And to get a cleaner API, that is okay at this point.

zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
```diff
@@ -16,18 +16,37 @@
...
+    toSingleValue(name, HttpCodec.Query(name, TextCodec.boolean))
 
   def queryInt(name: String): QueryCodec[Int] =
-    HttpCodec.Query(name, TextCodec.int)
+    toSingleValue(name, HttpCodec.Query(name, TextCodec.int))
 
   def queryTo[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] =
+    toSingleValue(name, HttpCodec.Query(name, codec))
+
+  def queryMultiValue(name: String): QueryCodec[Chunk[String]] =
```
987Nabil: please call these methods `queryAllXXX` so it maches the logical counterparts in `QueryParams`

987Nabil:
@jdegoes The way this is modeled rn makes it impossible to to analyse if it is single or multi value. This might make it hard to generate docs etc. from endpoints. What do you think about adding a field `queryStyle` to the codec or a new annotation? 
Both could also contain information about how multi values should be rendered. `?a=1&a=2` vs `?a=1,2` vs `?a=1|2` or maybe even others.
That would likely be more important for clients, to be compliant with their requirements.

jdegoes:
@987Nabil The way this is modeled rn makes it impossible to to analyse if it is single or multi value. 

How about `QueryParamHint` which could be `QueryParamHint.Zero`, `QueryParamHint.One`, `QueryParamHint.Many`, `QueryParamHint.Any`. Then the constructors can feed the hint in, and the documentation generation can use the hint.

Most code doesn't need to know about that, it's only really for doc or client generation.

cc @ekhov

987Nabil:
@ekhov I think you get the idea. Try what @jdegoes suggested and adjust the idea if needed. What is important is, that we can find intend in the data.

ekhov:
@jdegoes Do I get it correctly that you want this hint to be added as a `Query` parameter? And at the moment we will use only `One`(single value codecs) and `Many`(for multi value)?

jdegoes (approved):

PR #2635, opened 2024-01-19
Fix OpenAPI example extraction

### LINKED ISSUE #2609: Tuple Error in OpenAPI examples that use query params and POST bodies
**Describe the bug**
Cannot provide OpenAPI examples that combine query parameters and post bodies.

This could be known and simply because I'm using an incomplete feature, but just wanted to make sure.
It might also be a bug in zio-schema, rather than zio-http

#2572 

**To Reproduce**
```
import zio.http._
import zio.http.codec.PathCodec
import zio.http.codec.PathCodec.string
import zio.http.endpoint.Endpoint
import zio.http.endpoint.openapi.{OpenAPIGen, SwaggerUI}
import zio.schema.{DeriveSchema, Schema}

case class Payload(content: String)

object ExampleData extends App {
  implicit val schema: Schema[Payload] =
    DeriveSchema.gen[Payload]

  val endpoint =
    Endpoint(Method.GET / "root"/ string("name"))
      .in[Payload]
      .out[String]
      .examplesIn("hi" -> ("name_value", Payload("input")))

  val openApi =
    OpenAPIGen.fromEndpoints(
      title = "Burner",
      version = "1.0",
      endpoint,
    )

  import PathCodec._

  SwaggerUI.routes("docs" / "openapi", openApi)
}
```

Error:
```
Exception in thread "main" java.lang.ClassCastException: class scala.Tuple2 cannot be cast to class example.Payload (scala.Tuple2 and example.Payload are in unnamed module of loader 'app')
	at zio.schema.codec.JsonCodec$ProductEncoder$.$anonfun$caseClassEncoder$4(JsonCodec.scala:810)
	at zio.schema.codec.JsonCodec$ProductEncoder$.$anonfun$caseClassEncoder$4$adapted(JsonCodec.scala:802)
```

This example was written in the `example` directory of `zio-http` alongside the other examples, using the latest branch.

987Nabil:
fixes #2609
/claim #2609

- [x] fix example tuple deconstruction
- [x] fix example either deconstruction
- [x] handover examples correctly to path codecs

@jdegoes I see how I can deconstruct tuple examples for combined codecs. But for alternation, we don't have a type class that helps us to deconstruct.
I would add in the docs, that we don't support this for open api gen and that instead of calling `inExamples` on the endpoint, ppl should call `ContentCodec.content[A].examples(...) | ContentCodec.content[B].examples(...)`.
wdyt?

### DIFF
--- a/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliEndpoint.scala
+++ b/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliEndpoint.scala
@@ -7,9 +7,7 @@ import zio.cli._
 import zio.schema._
 
 import zio.http._
-import zio.http.codec.HttpCodec.Metadata
 import zio.http.codec._
-import zio.http.codec.internal._
 import zio.http.endpoint._
 
 /**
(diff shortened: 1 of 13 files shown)


jdegoes:
@987Nabil `Alternator` does have destruction:

See [here](https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/codec/Alternator.scala#L34).

987Nabil:
@jdegoes I missed it, because it was missing from the parameters of `Fallback`. I will add it, so it can be used later. Then it should be possible to make it work 👍

jdegoes (approved):

PR #2637, opened 2024-01-19
feat: added server log

### LINKED ISSUE #2632: `Server.serve` to log basic server info when mounted
Currently `Server.serve` runs silently and it's fine in simple cases. However, when there are dependencies that take time to initialize before server runs, it's difficult to determine whether the server is running or not, so I usually replace `Server.serve` with this snippet: 

```scala
Server.install(httpApp).flatMap(port => ZIO.logInfo(s"ZIO HTTP server is running at port $port") *> ZIO.never)
```

But doing this every time is cumbersome, and it's easy to forget to add `*> ZIO.never` at the and so that the server stops immediately.

It would be convenient to add a default log message to `Server.serve` which can easily be turned off.

ibishal:
/claim #2632
fixes #2632

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Server.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Server.scala
@@ -58,6 +58,7 @@ object Server extends ServerPlatformSpecific {
     gracefulShutdownTimeout: Duration,
     webSocketConfig: WebSocketConfig,
     idleTimeout: Option[Duration],
+    logStartingServer: Boolean,
   ) {
     self =>
 
@@ -158,6 +159,9 @@ object Server extends ServerPlatformSpecific {
 
     def webSocketConfig(webSocketConfig: WebSocketConfig): Config =
       self.copy(webSocketConfig = webSocketConfig)
+
+    def logStartingServer(log: Boolean): Config = 
+      self.copy(logStartingServer = log)
   }
 
   object Config {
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/Server.scala
```diff
@@ -331,11 +336,18 @@ object Server extends ServerPlatformSpecific {
 
   def serve[R](
     httpApp: HttpApp[R],
+    logStartingServer: Boolean = true,
   )(implicit trace: Trace): URIO[R with Server, Nothing] =
     install(httpApp) *> ZIO.never
```
ekhov, line 341: Maybe it would be better to just add logging in this method, one log with something like "starting the server..." before call to `install(httpApp)` and other with "server started" after that call? I think it's not reasonable to add it to `install` itself. What do you think?

zio-http/shared/src/main/scala/zio/http/Server.scala
```diff
@@ -331,11 +336,18 @@ object Server extends ServerPlatformSpecific {
...
 
-  def install[R](httpApp: HttpApp[R])(implicit trace: Trace): URIO[R with Server, Int] = {
-    ZIO.serviceWithZIO[Server](_.install(httpApp)) *> ZIO.service[Server].map(_.port)
+  def install[R](httpApp: HttpApp[R], logStartingServer: Boolean = true)(implicit
+    trace: Trace,
+  ): URIO[R with Server, Int] = {
+    ZIO.serviceWithZIO[Server] { server =>
+      val logMessage = ZIO.when(logStartingServer)(ZIO.logInfo(s"ZIO HTTP server is running at port ${server.port}"))
+
+      logMessage *> server.install(httpApp).tap(_ => logMessage).as(0)
```
ekhov, line 349: `install` should return port on which server is running, but in your case it will always return 0, that's not right. Plus see my comment about putting log into `serve`

PR #2638, opened 2024-01-19
Micro-optimizations and improved `IOException` error handling

kyri-petrou:
This PR contains some micro-optimizations and small fixes I run across while working on other issues:

1. Use Netty's `headers.get(key)` in `Headers.Native` to obtain a header value. Note that this method uses a case-insensitive key, so the behaviour should be the same as previously. The performant benefit of this depends largely on the number of headers in the request. In the added benchmark, that's ~x5 faster than the current implementation
2. Avoid pattern for converting between `zio.http.Status` <-> `io.netty.handler.codec.http.HttpResponseStatus`. Using the status code to do the conversion is both less boilerplate-y but also more performant as the pattern matching can be compiled to a tableswitch.
3. When handling `IOException` in `ServerInboundHandler#exceptionCaught`, check if the message is not null and that it **contains** the `Connection reset` string. This is because when using native transports (e.g., EPoll, KQueue), the `Connection reset` message is prefixed by the native method's name
4. Small micro-optimizations to `CharSequenceExtensions.compare` by checking that the length of the CharSequence is the same prior to iterating the chars, and by avoiding calls to `toLower` when not necessary. Performance benefit is highly dependent on different factors (in some cases, up to 1 order of magnitude), and is now en par with Java's `equalsIgnoreCase` method

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/model/Conversions.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/model/Conversions.scala
@@ -22,7 +22,6 @@ import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import zio.http.Server.Config.CompressionOptions
 import zio.http._
-import zio.http.internal.{CaseMode, CharSequenceExtensions}
 
 import io.netty.handler.codec.compression.{DeflateOptions, StandardCompressionOptions}
 import io.netty.handler.codec.http._
@@ -60,9 +59,9 @@ private[netty] object Conversions {
 
   def headersToNetty(headers: Headers): HttpHeaders =
     headers match {
-      case Headers.FromIterable(_)     => encodeHeaderListToNetty(headers.toList)
+      case Headers.FromIterable(_)     => encodeHeaderListToNetty(headers)
       case Headers.Native(value, _, _) => value.asInstanceOf[HttpHeaders]
-      case Headers.Concat(_, _)        => encodeHeaderListToNetty(headers.toList)
+      case Headers.Concat(_, _)        => encodeHeaderListToNetty(headers)
       case Headers.Empty               => new DefaultHttpHeaders()
     }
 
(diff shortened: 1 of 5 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/model/Conversions.scala
```diff
@@ -60,9 +59,9 @@ private[netty] object Conversions {
 
   def headersToNetty(headers: Headers): HttpHeaders =
     headers match {
-      case Headers.FromIterable(_)     => encodeHeaderListToNetty(headers.toList)
+      case Headers.FromIterable(_)     => encodeHeaderListToNetty(headers)
```
987Nabil, line 62: Love this change 👍

zio-http/shared/src/main/scala/zio/http/Status.scala
```diff
@@ -176,7 +176,7 @@ object Status {
     Try(code.toInt).toOption.map(fromInt)
 
   def fromInt(code: Int): Status = {
-    code match {
+    (code: @annotation.switch) match {
```
987Nabil, line 179: Just for me: Is this an optional hint? Might the compiler create a switch without it? Might the compiler **not** create a switch with it?

987Nabil (approved):
lgtm
But you have to fix the build 🙂

zio-http/shared/src/main/scala/zio/http/Status.scala
kyri-petrou: > Might the compiler create a switch without it

Yes
> Might the compiler not create a switch with it

No

This is very similar to `tailrec`; if the compiler can create a `lookupswitch` or `tableswitch`, it will do it even without the annotation. However, if the annotation is present, then it'll raise an error if a `lookupswitch` / `tableswitch` cannot be created from the expression.

kyri-petrou:
> But you have to fix the build

@987Nabil CI is failing due to a 502 error from sonatype when fetching dependencies. I think the failing task just needs to be retriggered.

zio-http/shared/src/main/scala/zio/http/Status.scala
987Nabil: Thx! 👍

PR #2644, opened 2024-01-21
Avoid unnecessary parsing of `content-type` header in `ServerInboundHandler`

kyri-petrou:
While #2597 improved the performance of parsing the content-type header significantly, it is still a _relatively_ expensive computation (w.r.t everything else) so if we can avoid it, why not? While most applications won't see any measurable difference with this change, I'm certain it'll make a difference on benchmarks that include the content-type header as part of the request.

To demonstrate the difference, I added a benchmark in `ServerInboundHandlerBenchmark` that includes the content type header:

Before:
```
[info] Benchmark                                                  Mode  Cnt      Score     Error  Units
[info] ServerInboundHandlerBenchmark.benchmarkSimpleContentType  thrpt   10  30047.175 ± 245.613  ops/s
```

After:
```
[info] Benchmark                                                  Mode  Cnt      Score     Error  Units
[info] ServerInboundHandlerBenchmark.benchmarkSimpleContentType  thrpt   10  32728.912 ± 182.722  ops/s
```
 
I rerun the benchmark 5-10 times and it seems that avoiding the parsing of the content-type header is consistently 5-10% faster

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
@@ -40,30 +40,45 @@ object NettyBody extends BodyEncoding {
   private[zio] def fromAsync(
     unsafeAsync: UnsafeAsync => Unit,
     knownContentLength: Option[Long],
-    contentTypeHeader: Option[Header.ContentType] = None,
-  ): Body = AsyncBody(
-    unsafeAsync,
-    knownContentLength,
-    contentTypeHeader.map(_.mediaType),
-    contentTypeHeader.flatMap(_.boundary),
-  )
+    contentTypeHeader: () => Option[Header.ContentType] = () => None,
+  ): Body = {
+    lazy val cachedCt = contentTypeHeader()
+    AsyncBody(
+      unsafeAsync,
+      knownContentLength,
+      () => cachedCt.map(_.mediaType),
+      () => cachedCt.flatMap(_.boundary),
+    )
+  }
 
   /**
    * Helper to create Body from ByteBuf
    */
   def fromByteBuf(byteBuf: ByteBuf, contentTypeHeader: Option[Header.ContentType] = None): Body =
-    ByteBufBody(byteBuf, contentTypeHeader.map(_.mediaType), contentTypeHeader.flatMap(_.boundary))
+    fromByteBufLazy(byteBuf, () => contentTypeHeader)
+
+  private[zio] def fromByteBufLazy(
+    byteBuf: ByteBuf,
+    contentTypeHeader: () => Option[Header.ContentType] = () => None,
+  ): Body = {
+    lazy val cachedCt = contentTypeHeader()
+    ByteBufBody(byteBuf, () => cachedCt.map(_.mediaType), () => cachedCt.flatMap(_.boundary))
+  }
 
   override def fromCharSequence(charSequence: CharSequence, charset: Charset): Body =
     fromAsciiString(new AsciiString(charSequence, charset))
 
-  private[zio] final case class AsciiStringBody(
+  private[zio] final case class AsciiStringBody private[NettyBody] (
     asciiString: AsciiString,
-    override val mediaType: Option[MediaType] = None,
-    override val boundary: Option[Boundary] = None,
+    private val mediaType0: () => Option[MediaType] = () => None,
+    private val boundary0: () => Option[Boundary] = () => None,
   ) extends Body
       with UnsafeWriteable
       with UnsafeBytes {
+
+    override def mediaType: Option[MediaType] = mediaType0()
+
+    override def boundary: Option[Boundary] = boundary0()
 
     override def asArray(implicit trace: Trace): Task[Array[Byte]] = ZIO.succeed(asciiString.array())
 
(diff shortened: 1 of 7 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
```diff
@@ -113,21 +132,26 @@ object NettyBody extends BodyEncoding {
     override private[zio] def unsafeAsArray(implicit unsafe: Unsafe): Array[Byte] =
       ByteBufUtil.getBytes(byteBuf)
 
-    override def contentType(newMediaType: MediaType): Body = copy(mediaType = Some(newMediaType))
+    override def contentType(newMediaType: MediaType): Body = copy(mediaType0 = () => Some(newMediaType))
 
     override def contentType(newMediaType: MediaType, newBoundary: Boundary): Body =
-      copy(mediaType = Some(newMediaType), boundary = boundary.orElse(Some(newBoundary)))
+      copy(mediaType0 = () => Some(newMediaType), boundary0 = () => boundary.orElse(Some(newBoundary)))
```
kyri-petrou: By the way, is this the correct behaviour? I'm a bit confused on why we would keep the old `boundary` if it was defined. Shouldn't this be `boundary = Some(newBoundary)` instead?

kyri-petrou:
The test failure is due to a flaky test ScalaJS. [I opened an issue for it](https://github.com/zio/zio-http/issues/2645); I tried rerunning the entire test suite via an empty commit but the test failed again so it's perhaps better to just restart the failed run

zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
```diff
@@ -40,31 +40,46 @@ object NettyBody extends BodyEncoding {
...
     unsafeAsync: UnsafeAsync => Unit,
     knownContentLength: Option[Long],
-    contentTypeHeader: Option[Header.ContentType] = None,
-  ): Body = AsyncBody(
-    unsafeAsync,
-    knownContentLength,
-    contentTypeHeader.map(_.mediaType),
-    contentTypeHeader.flatMap(_.boundary),
-  )
+    contentTypeHeader: () => Option[Header.ContentType] = () => None,
```
jdegoes: Note that allocation of `Function0` can be measured in benchmarks.

I would suggest the following:

1. Statically make `Header.ContentType` for common content types (e.g. `Header.ContentType.`application/json`)
2. Use fast comparison (or *maybe* hash map lookup) to decide which `Header.ContentType` to map to which standard Netty header. If Netty reuses the same char sequence, it could be even faster (`eq`).

If done properly this can as fast as `Function0` and will also be less disruptive to the code.

zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
jdegoes: Probably so

zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
jdegoes: Oh also we can do:

```scala
Header.ContentType.`some-application/json` 
```

to pre-allocate the option.

zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
```diff
@@ -40,31 +40,46 @@ object NettyBody extends BodyEncoding {
...
 
   override def fromCharSequence(charSequence: CharSequence, charset: Charset): Body =
     fromAsciiString(new AsciiString(charSequence, charset))
 
-  private[zio] final case class AsciiStringBody(
+  private[zio] final case class AsciiStringBody private[NettyBody] (
     asciiString: AsciiString,
-    override val mediaType: Option[MediaType] = None,
-    override val boundary: Option[Boundary] = None,
+    private val mediaType0: () => Option[MediaType] = () => None,
```
jdegoes: Same story here for `MediaType`. It can also eliminate the `Function0` allocation.

zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
```diff
@@ -40,31 +40,46 @@ object NettyBody extends BodyEncoding {
...
+    private val boundary0: () => Option[Boundary] = () => None,
```
jdegoes: Boundary type is tricky because it comes in dynamically. However, it would be nice to eliminate the more complex API here and the `Function0` allocation. Did you measure to see if this has any impact?

kyri-petrou:
@jdegoes thanks for the feedback. While reading though your comments, I realised that we already do cache media types (via a hash map). I reverted all the `Function0` changes and utilized `MediaType.forContentType` to parse the header in `ServerInboundHandler` (which is 1 or 2 orders of magnitude faster than `ContentType.parse`, depending whether the content-type header contains parameters (2 orders if there aren't any parameters).

With the changes in the latest commit, the performance is _slightly_ improved (compared to the previous commit), but the biggest benefit is that the API is much cleaner

```
[info] Benchmark                                                  Mode  Cnt      Score     Error  Units
[info] ServerInboundHandlerBenchmark.benchmarkSimpleContentType  thrpt   10  33409.354 ± 155.748  ops/s
```

The only "downside" to these changes is the use of `.toLowerCase`. To make sure there aren't any performance implications associated with it, I added a benchmark to compare whether it makes any difference and it was negllible (~10 nanoseconds slower 🤷‍♂️ )

In addition, I shamelessly broke binary / source compatibility of the `NettyBody.fromByteBuf` method by changing the signature and making it private. I can revert this change and expose a different method; the main reason I did it was that I'm not sure whether the method is meant to be public at all. If that's not the case, please let me know and I'll revert the change

kyri-petrou:
PS: Build is failed due to Sonatype 502 error. Should work after a restart

jdegoes (approved):

PR #2648, opened 2024-01-23
fix: ZioSchemaVersion version

### LINKED ISSUE #2628: Fix broken build
CI is failing and needs to be fixed.

jkcs:
/claim #2628
fix #2628


987Nabil:
this is an invalid change. The project will not build without the snapshot version. And the download is only flaky. We will update once zio-schema is released

PR #2649, opened 2024-01-23
feature:  added Server Logging 

### LINKED ISSUE #2632: `Server.serve` to log basic server info when mounted

guptadeepak8:
fixes #2632
/claim #2632


guptadeepak8:
Hey @jdegoes can review this

guptadeepak8:
@jdegoes can you review this

PR #2655, opened 2024-01-27
HttpContentCodec - Customizable encoding for Endpoint API

### LINKED ISSUE #2625: Support custom encoding and decoding for Endpoint parameters (in, out, error)
1. Introduce `HttpContentCodec`, a new data type that builds on ZIO Schema's notion of `BinaryCodec`, but also adds crucial information on the mime-type for a given binary codec.
    
    ```scala
    package zio.http.codec
    
    final case class HttpContentCodec[A](choices: ListMap[MediaType, BinaryCodec[A]] /* order is priority during encoding */ )
      def ++ (that: HttpContentCodec[A]): HttpContentCodec[A] = ??? // Right one wins (overrides left) if conflict
    
      def lookup(contentType: Header.ContentType): Option[BinaryCodec[A]] = ???
    
      def decodeRequest(request: Request): Either[String, A] = ???
      def decodeResponse(response: Response): Either[String, A] = ???
      def encodeRequest(value: A): Either[String, Request] = ???
      def encode(value: A): Either[String, Response] = ???
    }
    object HttpContentCodec{
      // Derive http codec from schema:
      implicit def fromSchema[A](implicit schema: Schema[A]): HttpContentCodec[A] = ??? // Add all ZIO Schema media types
    
      // Codecs for all the standard types?
      
    }
    ```
2.  Modify `HttpCodec.Content` so that it no longer embeds `Schema`, but rather, so that it uses `HttpContentCodec` directly, which is much more flexible.
    
    ```scala
    private[http] final case class Content[A](
        codec: HttpContentCodec[A],
        name: Option[String],
        index: Int = 0,
      ) extends Atom[HttpCodecType.Content, A] {
    ```
3. Modify all constructors of `Content` (mostly, I think, they appear in `ContentCodecs.scala`) to account for the refactoring of `HttpCodec.Content`.
4. Update all downstream code, particular the encoding/decoding part of `Endpoint`.
5. Change the constructors of `HttpCodec` to require `HttpContentCodec` for the `Err`, `In`, and `Out` types, rather than `Schema`. For convenience and backward compatibility, we could provide a Schema-based variant of all these constructors. Or we could go "all-in" on `HttpContentCodec`.

These changes, taken collectively, will ensure that users can choose whatever way they want to map Scala data types to and from requests and responses.

987Nabil:
@jdegoes this gives freedom to support all kinds of encodings, but it comes currently with the price of forced explicit media types for multipart. Maybe we need a different way to create multipart codecs.

fixes #2625
/claim #2625


zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+) { self =>
+
+  /**
+   * A right biased merge of two HttpContentCodecs.
+   */
+  def ++(that: HttpContentCodec[A]): HttpContentCodec[A] =
+    HttpContentCodec(choices ++ that.choices, schema)
+
+  def decodeRequest(request: Request): Task[A]    = {
+    val contentType =
```
jdegoes: All this code is in the hot path, which means that it should be heavily optimized to avoid allocations.

One example: lookup the codec using `contains` and `get` instead of allocating an `Option` with `getOrElse`. Also, do not continuously create a new `Header.ContentType`, but rather, define `Header.ContentType.application/json` so that you can reuse the same one again and again.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+      case Some(codec) =>
+        request.body.asChunk.flatMap { bytes =>
+          ZIO.fromEither(codec.decode(bytes))
+        }
+      case None        =>
+        ZIO.fail(throw new IllegalArgumentException(s"No codec found for content type $contentType"))
+    }
+  }
+  def decodeResponse(response: Response): Task[A] = {
+    val contentType =
```
jdegoes: Same here. Hot path code.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+      case Some(codec) =>
+        response.body.asChunk.flatMap { bytes =>
+          ZIO.fromEither(codec.decode(bytes))
+        }
+      case None        =>
+        ZIO.fail(throw new IllegalArgumentException(s"No codec found for content type $contentType"))
+    }
+  }
+
+  def encode(value: A): Either[String, Body] =
```
jdegoes: This is hot path, too. Look at allocations:

1. headOption (1)
2. Tuple2 (1)
3. Body (2)
4. Either (1)

Some are unavoidable, but some can be avoided. For example, on `choices` use `nonEmpty` and `head`. Do not pattern match on the tuple, rather extract out the components individually. And construct `Body` with both the byte **and** the media type at the same time.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+    choices.headOption match {
+      case Some((mediaType, codec)) =>
+        Right(Body.fromChunk(codec.encode(value)).contentType(mediaType))
+      case None                     =>
+        Left("No codec found")
+    }
+
+  def only(mediaType: MediaType): HttpContentCodec[A] =
```
jdegoes, line 74: I like the elegance of this, although we should avoid using it internally in ZIO HTTP due to overhead, unless we are precomputing an HttpCodec and it won't matter (i.e. non-hot-path).

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+              ),
+          ),
+          schema,
+        )
+      case None            =>
+        self
+    }
+  }
+
+  def onlyOrNone(mediaType: MediaType): Option[HttpContentCodec[A]] =
```
jdegoes: This is awkward because `HttpContentCodec` already has a `None`, namely, when the `ListMap` is empty. It may be a better idea to simply have `isEmpty` and `nonEmpty` methods on `HttpContentCodec`.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+    if (choices.contains(mediaType)) Some(only(mediaType))
+    else None
+
+  def lookup(mediaType: MediaType): Option[BinaryCodec[A]] =
```
jdegoes: Due to extreme overhead, these `lookup` methods cannot be executed on a hot path.

I would suggest to precompute a map, stored as a private member of `HttpContentCodec`, which allows you to quickly perform the `matches` comparison against a given `MediaType`, and which already returns an `Option` (preallocated `Some`), to avoid all this overhead.

Then, of course, we must say: no dynamic `HttpContentCodec`, because it will be more expensive to create them. But that's fine, we would want to do that anyway.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+    choices.collectFirst { case (mt, codec) if mt.matches(mediaType) => codec }
+
+  def lookup(contentType: Header.ContentType): Option[BinaryCodec[A]] =
+    lookup(contentType.mediaType)
+
+  private[http] lazy val defaultMediaType: MediaType =
```
jdegoes: `lazy` is slow compared to non-lazy, at least in a world where we only have static `HttpContentCodec` and are not creating them dynamically during request/response handling. So make this a `val` and not a `lazy val`.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+    choices.headOption.map(_._1).getOrElse {
+      throw new IllegalArgumentException(s"No codec defined")
+    }
+
+  lazy val defaultCodec: BinaryCodec[A] = choices.headOption.map(_._2).getOrElse {
```
jdegoes: Ditto here.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -0,0 +1,195 @@
...
+}
+
+object HttpContentCodec {
+  implicit def fromSchema[A](implicit schema: Schema[A]): HttpContentCodec[A] = {
+    json[A] ++ protobuf[A] ++ text[A]
+  }
+
+  object json {
+    def apply[A](implicit schema: Schema[A]): HttpContentCodec[A]               = fromSchema[A]
+    implicit def fromSchema[A](implicit schema: Schema[A]): HttpContentCodec[A] = {
```
jdegoes: 😱  I don't like multiple implicit conversions. Well, I don't like implicit conversions, and I REALLY don't like multiple implicit conversions. 😄 

Let's remove all of this and just say if you want that much control, you are on your own.

Then maybe we can turn `HttpContentCodec.json.apply` into `HttpContentCodec.jsonOnly`, etc.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
987Nabil: I know that there needs to be optimizations. I made it work first. And wanted some feedback for the implications

zio-http/shared/src/main/scala/zio/http/codec/internal/TextCodec.scala
```diff
@@ -0,0 +1,102 @@
...
+          }
+
+        override def decode(c: Chunk[Byte]): Either[DecodeError, A] = {
+          val s = c.asString
+          schema match {
+            case Schema.Primitive(standardType, _) =>
+              (standardType match {
+                case StandardType.UnitType           => Right("")
+                case StandardType.StringType         => Right(s)
+                case StandardType.BoolType           => Try(s.toBoolean).toEither
```
jdegoes: Overhead: `Try(...).toEither`

zio-http/shared/src/main/scala/zio/http/codec/ContentCodecs.scala
```diff
@@ -16,40 +16,48 @@
...
+    HttpCodec.Content(codec, Some(name))
 
-  def content[A](implicit schema: Schema[A]): ContentCodec[A] =
-    HttpCodec.Content(schema, mediaType = None, None)
+  def content[A](implicit codec: HttpContentCodec[A]): ContentCodec[A] =
+    HttpCodec.Content(codec, None)
 
-  def content[A](name: String, mediaType: MediaType)(implicit schema: Schema[A]): ContentCodec[A] =
-    HttpCodec.Content(schema, mediaType = Some(mediaType), Some(name))
+  def content[A](name: String, mediaType: MediaType)(implicit codec: HttpContentCodec[A]): ContentCodec[A] =
```
jdegoes, line 30: This makes me want to add a covariant phantom type to `HttpCodecContent`, some `MediaTypes` (`<: MediaType`), such that this could be written:

```suggestion
  def content[A, M <: MediaType](name: String, mediaType: M)(implicit codec: HttpContentCodec[M, A]): ContentCodec[A] =
```

i.e. proof that you can handle that media type. Intersection for composition (e.g. `HttpContentCodec[`application/json` with `text/html`]`).

It'd be a bit more complicated, however, because `M` could be inferred but `A` cannot be, so would need partial application trick to ease type inference.

jdegoes:
@987Nabil Overall, looks good, just a bit of optimization and polish and good to go!

Curious: why did this change multi-part at all?? Can you link?

987Nabil:
@jdegoes it is not that I changed the multipart, but that my change has implications on how media types are resolved. With the multipart, there was never an explicit media type. We did analyse the input in `EncoderDecoder` and derived the media type.
But my change makes media types non optional. Every content has an explicit (set of) media type. Therefore the need to make media types explicit when creating multipart content codecs.
I think this is overall the better design, because we have the information what it is where it belongs. In the codce. Not in the execution part (namely `EncoderDecoder`).
`EncoderDecoder` got much dumber with my changes.

jdegoes:
@987Nabil unless I misunderstand, sounds like a good thing!

zio-http/shared/src/main/scala/zio/http/codec/ContentCodecs.scala
987Nabil: Currently the media type is also not reflected in the type of the val. So I would need to change the whole media type impl and make it look similar to `Header` where each heder impl. has its own type.

And I am understanding, that you could see the media type in the type signature, but I don't see how you could utilize it besides a human reading it. And from my experience, I don't have type annotations on my endpoint vals.

Maybe if one would explicitly define the content/accept header we could check that  you can only use valid MediaTypes. But then these headers need to be parameterized too.

Sounds like a) a lot of work and maybe too much for this pr? and b) like something that is not obviously a great change.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
987Nabil: I think precomputing all is too expensive. Imagine you have to create all `text/<whatever>` keys in the map.  And then never use it. There are almost 2k predefined media types (all lazy val) in zio-http.
I would implement a cache via `ConcurrentHashMap`?

zio-http/shared/src/main/scala/zio/http/codec/ContentCodecs.scala
jdegoes: Yeah, I think I would skip it. It just occurred to me it was possible.

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
jdegoes: Sure!

jdegoes: Btw you don't need concurrent hash map. Just a `var cache: Map[...]`. Faster

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
987Nabil: Actually, the encode and decode methods are not used at all. The magic is happening in `BodyCodec`

987Nabil:
@jdegoes There is one Cli spec that is failing. Idk why. It is hard for me to follow the code. There are many abstractions only for the tests. And I am not sure how it relates to my changes. I deactivated the test for now. Maybe we open an issue to fix it later? I think we should try to get these changes in.

jdegoes (approved):

Issue #2652, opened 2024-01-24
OpenAPI gen throws errors on lazy schemas


987Nabil:
No explicit schema definition in a companion object of a type `B`  that is used by a type `A`, that derives its schema, leads to the schema of `B` being lazy. This leads to an exception during OpenAPI generation.

```scala
case class A(b:B)

object A:
  given Schema[A] = DeriveSchema.gen

case class B(i: Int)
```

sroopsai:
/attempt #2652

<div id="algora" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clrt18rif0004la0ffspkxoqs/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

35C4n0r:
@jdegoes Is this issue still open, if not I would like to attempt it.

jdegoes:
cc @987Nabil

hochgi:
/attempt #2652

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clrt18rif0004la0ffspkxoqs/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

Issue has been duplicated by #2933 and has already been fixed in #2935 
The opened PR ( #3004 ) simply adds a test with the example from above to show it has already been fixed.

PR #2658, opened 2024-01-28
Added Missing Companion Objects in OpenAPI to avoid OpenAPIGen Lazy Scheme Exceptions

sroopsai:


sroopsai:
fixes #2652 
/claim #2652

PR #2668, opened 2024-02-08
Fix :Add raw in Html object

### LINKED ISSUE #2651: Html.fromString Behavior Change
**Describe the bug**
I'm not sure if this is a bug or if it was intentional, but `Html.fromString` now encodes the string, so the browser just shows the escaped text instead of the rendered html.

I believe it was [this commit](https://github.com/zio/zio-http/commit/61fa20b15e8e3210c37561c8e005df423dcc3d48#diff-eaeb9cdca05316230a5f8a646e0e77b0f55204715fd479c95878b3dead921c3cR37) that caused the behavior change.

If it was not intentional, then I think the easiest fix would be to change `Html.fromString` to:

```
implicit def fromString(string: CharSequence): Html = Html.Single(Dom.raw(string))
```

I wasn't able to test this as sbt is giving me some trouble.

I did figure out I could use `Html.fromDomElement(Dom.raw(myString))` instead so as long as that doesn't change, I'm good to go. Just thought you would want to know about the behavior change.

**To Reproduce**
Create a Route that uses `Html.fromString`:
```
Method.GET / "" -> handler(
  Response.html(
    Html.fromString(
      """<div class="container1 container2"><a href="http://zio.dev">ZIO Homepage</a></div>"""
    )
  )
)
```

I created a [git repo](https://github.com/eltimn/zio-http-string) that has a route like that, it also includes a route using `formDomElement`:

guptadeepak8:
fix: #2651
/claim #2651

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/template/HtmlSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/template/HtmlSpec.scala
@@ -81,10 +81,10 @@ case object HtmlSpec extends ZIOHttpSpec {
         assert(view.encode)(equalTo(expected.stripMargin))
       },
       test("raw HTML content is not escaped") {
-        val view = Html.raw("<div>Hello, world!</div>")
+        val view     = Html.raw("<div>Hello, world!</div>")
         val expected = "<div>Hello, world!</div>"
         assert(view.encode)(equalTo(expected))
-     },
+      },
     )
   }
 }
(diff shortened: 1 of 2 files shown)


guptadeepak8:
@jdegoes can you review this

jdegoes:
We need a test case, then looks good to merge!

guptadeepak8:
@jdegoes @987Nabil made changes can review this

jdegoes (approved):

PR #2673, opened 2024-02-08
Add brotli compression (#2646)

### LINKED ISSUE #2646: Adding Brotli as a compression option

Most browsers support brotli compression, which is [a faster alternative to gzip](https://medium.com/@bansal.suneet/brotli-vs-gzip-compression-surprising-compression-result-brotli-power-782aac2ee29f). Netty already supports compressing payloads using brotli compression. It would be nice if zio-http exposed that as a compression option.

The main "issue" with adding this as an option is that this requires an additional dependency on `brotli4j`. While this could be added as an optional dependency (which is what Netty does), it might require adding some checks during the server initialization to ensure that the user has the dependency installed if they choose it

Having `Brotli` as an additional `CompressionType` option

N/A

**Additional context**
N/A

987Nabil:
@jdegoes I think the brotli4j dependency is okay. It would go to the netty backend module once we split zio-http core and netty backend.

fixes #2646
/claim #2646

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -1,5 +1,4 @@
-import sbt._
-import sbt.Keys.scalaVersion
+import sbt.*
 
 object Dependencies {
   val JwtCoreVersion                = "9.1.1"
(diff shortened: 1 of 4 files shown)


987Nabil:
@jdegoes I am not sure what the problem with the shading is. Never had to configure it before. Can you give me a hint? The build error is not very helpful.

jdegoes:
@987Nabil I have not used that setting before, I just know why it's used and how it works.

jdegoes:
@987Nabil Please re-open when build is passing. 🙏

PR #2674, opened 2024-02-09
Improve performance & memory usage of `asMultipartFormStream`

### LINKED ISSUE #2660: asMultipartForm method consuming a lot of heap memory
**Describe the bug**
When uploading a file using `multipart/form-data` we have to get the `Form` using `asMultipartForm` method from request body. And using debugging and live memory stats I understood that when `asMultipartForm` function is called:
If I'm uploading ~ 500MB of file => the heap consumption is more than 4GB, which is a lot in comparison to file size.

**Screenshots**
Screenshot 2024-02-01 at 12 10 19 PM

**Additional context**
 - OS: macOS
 - Java 17
 - `zio-http` version: `3.0.0-RC4`

kyri-petrou:
/claim #2660
/fixes #2660

I managed to reproduce the issue using [this file](https://upload.wikimedia.org/wikipedia/commons/e/ea/Van_Gogh_-_Starry_Night_-_Google_Art_Project.jpg) (warning, it's a ~350MB image) and the code below

<details close>
  <summary>Code (click to expand)</summary>
  
```scala
package zio.http.netty

import zio.{Scope, ZIO, ZIOAppArgs, ZIOAppDefault}

import zio.http.Server.RequestStreaming
import zio.http._

object TestMultipartFormData extends ZIOAppDefault {

  private val app: HttpApp[Any] =
    Routes(
      Method.POST / "upload" ->
        handler { (req: Request) =>
          if (req.header(Header.ContentType).exists(_.mediaType == MediaType.multipart.`form-data`))
            for {
              _        <- ZIO.unit
              form     <- req.body.asMultipartForm
                .mapError(ex =>
                  Response(
                    Status.InternalServerError,
                    body = Body.fromString(s"Failed to decode body as multipart/form-data (${ex.getMessage}"),
                  ),
                )
              response <- form.get("file") match {
                case Some(file) =>
                  file match {
                    case FormField.Binary(_, data, contentType, transferEncoding, filename) =>
                      ZIO.succeed(
                        Response.text(
                          s"Received ${data.length} bytes of $contentType filename $filename and transfer encoding $transferEncoding",
                        ),
                      )
                    case _                                                                  =>
                      ZIO.fail(
                        Response(Status.BadRequest, body = Body.fromString("Parameter 'file' must be a binary file")),
                      )
                  }
                case None       =>
                  ZIO.fail(Response(Status.BadRequest, body = Body.fromString("Missing 'file' from body")))
              }
            } yield response
          else ZIO.succeed(Response(status = Status.NotFound))
        },
    ).sandbox.toHttpApp

  private def program: ZIO[Client with Server with Scope, Throwable, Unit] =
    for {
      port <- Server.install(app)
      _    <- ZIO.logInfo(s"Server started on port $port")
      _    <- ZIO.never
    } yield ()

  override def run: ZIO[Any with ZIOAppArgs with Scope, Any, Any] =
    program
      .provide(
        Server.defaultWith(config => config.port(8080).requestStreaming(RequestStreaming.Enabled)),
        Client.default,
        Scope.default,
      )
}
```
</details>

After initializing the server with a profiler attached (I used the IntelliJ profiler for this), one can upload the file using curl:

```shell
curl -F "file=@Van_Gogh_-_Starry_Night_-_Google_Art_Project.jpg" http://localhost:8080/upload
```

The changes in this PR drastically improve both the execution time and memory usage of the program:

- Execution time: ~30s -> ~7s
- Memory usage: (whatever -Xmx would be set to) -> ~500-600mb

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/StreamingForm.scala
+++ b/zio-http/shared/src/main/scala/zio/http/StreamingForm.scala
@@ -69,7 +69,7 @@ final case class StreamingForm(source: ZStream[Any, Throwable, Byte], boundary: 
                     case newFormState: FormState.FormStateBuffer =>
                       if (
                         state.currentQueue.isEmpty &&
-                        newFormState.phase == FormState.Phase.Part2 &&
+                        (newFormState.phase eq FormState.Phase.Part2) &&
                         !state.inNonStreamingPart
                       ) {
                         val contentType = FormField.getContentType(newFormState.tree)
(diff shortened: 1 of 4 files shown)


987Nabil:
JS test was failing. I restarted the build

zio-http/shared/src/main/scala/zio/http/StreamingForm.scala
```diff
@@ -164,24 +160,31 @@ final case class StreamingForm(source: ZStream[Any, Throwable, Byte], boundary:
 }
 
 object StreamingForm {
-  private final case class State(
```
987Nabil, line 167: Why is this now a class instead of case class? Is there a benefit that I am missing?
But if we keep this a class:
This is actually some kind of iteration state, right? Would it makes sense, since we have methods like rest and so on, to have mutable fields and reassign them to avoid allocation of new `State` instances? I did not check in detail, but it seems that one state is only progressed in a linear fashion. So I don't see that this would lead to race conditions.

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -25,69 +25,88 @@ private[http] sealed trait FormState
...
-    buffer: Chunk[Byte],
-    lastByte: Option[Byte],
+  final class FormStateBuffer(
+    private var lastByte: OptionalByte,
     boundary: Boundary,
-    dropContents: Boolean,
   ) extends FormState { self =>
 
-    def append(byte: Byte): FormState = {
+    private val tree0: ChunkBuilder[FormAST] = ChunkBuilder.make[FormAST]()
```
987Nabil, line 32: I see you did for `FormState` what I suggested for `State`.
Did you try it for `State`? Did it fail?

zio-http/shared/src/main/scala/zio/http/StreamingForm.scala
kyri-petrou: @987Nabil The main reason for making it a class was so that the `copy` method is not created. The only "real" benefit from this is that prevents someone in the future assigning the old `formState` without knowing that the class mutates.

I like your suggestion though, I think if we'll use a mutable class we might as well use internal variables and avoid the extra allocations

jdegoes (approved):

PR #2675, opened 2024-02-09
Simplify style attributes

johnspade:
In the age of Tailwind and long class chains in HTML, it's a bit cumbersome to wrap every class name in quotes, especially when copying styles and classes from tutorials or documentation. This PR changes the value type to `String` for `class` and `style` attributes in the HTML templates.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/template/Attributes.scala
+++ b/zio-http/shared/src/main/scala/zio/http/template/Attributes.scala
@@ -49,7 +49,7 @@ trait Attributes {
 
   final def citeAttr: PartialAttribute[String] = PartialAttribute("cite")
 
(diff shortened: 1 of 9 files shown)


zio-http/shared/src/main/scala/zio/http/template/Attributes.scala
```diff
@@ -49,7 +49,7 @@ trait Attributes {
 
   final def citeAttr: PartialAttribute[String] = PartialAttribute("cite")
 
-  final def classAttr: PartialAttribute[List[String]] = PartialAttribute("class")
+  final def classAttr: PartialAttribute[String] = PartialAttribute("class")
```
johnspade, line 52: The main change

jdegoes (approved):

PR #2687, opened 2024-02-19
Snippet Utils: Handle Exceptions When Running Docs Commands on The Root Project

khajavi:
This PR makes the snippet utils work even when running docs commands on the root project. This is required when we need to publish docs on ci.


khajavi:
@adamgfraser This is ready for your review.

adamgfraser (approved):

PR #2718, opened 2024-03-08
feat: Proper imports for generated openapi code 

runtologist:
Extend the code generator, such that each generation step
can add imports, which are then rendered at the file or object level.

This fixes a few issues where imports were missing for generated case
classes, that had `Chunk` fields. It also unifies the handling of UUIDs
and opens the door for additional future configuration options. For
example the collection type can be made configurable, or user defined
types can be added to the imports.

`writeFiles` now returns a list of written files. This allows for use
in sbt code generation tasks. I will add an sbt plugin in a later PR.

### DIFF
--- a/zio-http-gen/src/main/scala/zio/http/gen/scala/CodeGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/scala/CodeGen.scala
@@ -13,23 +13,23 @@ object CodeGen {
       Code.Import("zio.http.codec._"),
     )
 
-  def format(config: Path)(file: Path, content: String): String = {
+  def makeFormat(config: Path): (Path, String) => String = {
     import org.scalafmt.interfaces.Scalafmt
-
-    val scalafmt: Scalafmt = Scalafmt.create(this.getClass.getClassLoader)
-    scalafmt.format(config, file, content)
+    val scalafmt = Scalafmt.create(this.getClass.getClassLoader).createSession(config)
+    (file: Path, content: String) => scalafmt.format(file, content)
   }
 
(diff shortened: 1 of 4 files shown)


zio-http-gen/src/main/scala/zio/http/gen/scala/CodeGen.scala
```diff
@@ -250,17 +282,17 @@ object CodeGen {
...
+  def renderQueryCode(queryCode: Code.QueryParamCode): (List[Code.Import], String) = queryCode match {
     case Code.QueryParamCode(name, queryType) =>
-      val tpe = queryType match {
-        case Code.CodecType.Boolean => "Boolean"
-        case Code.CodecType.Int     => "Int"
-        case Code.CodecType.Long    => "Long"
-        case Code.CodecType.String  => "String"
-        case Code.CodecType.UUID    => "UUID"
+      val (imports, tpe) = queryType match {
+        case Code.CodecType.Boolean => Nil                                 -> "Boolean"
```
987Nabil, line 288: Since we don't return a map here, can you change it to (a,b) instead of a -> b

987Nabil:
@runtologist sbt plugin was also on my list. But I intentionally did not add it yet, since I wanted at least one more code gen. So that I can be more sure to have something generic enough to be extensible with new code gens.

runtologist:
@987Nabil I've got an sbt plugin running in production, it is working well. :)

zio-http-gen/src/main/scala/zio/http/gen/scala/CodeGen.scala
runtologist: Do you mean in this place, or everywhere?

jdegoes (approved):

PR #2729, opened 2024-03-15
Document HandlerAspect in middleware.md

SmedbergM:
Middlewares that return values for consumption by handlers (i.e. `HandlerAspect`s) need documenting.

### DIFF
--- a/docs/dsl/middleware.md
+++ b/docs/dsl/middleware.md
@@ -287,10 +287,11 @@ object Example extends ZIOAppDefault {
 
 ## Handler Aspects
 
-Ordinary Middlewares are intended to bracket a request's execution by intercepting the request, possibly modifying it or short-circuiting its execution (for example if it fails authentication).
+Ordinary Middlewares are intended to bracket a request's execution by intercepting the request, possibly modifying it or short-circuiting its execution
+(for example if it fails authentication), and then performing some post-processing on the response.
 However, we sometimes want to gather some contextual information about a request and pass it alongside to the request's handler.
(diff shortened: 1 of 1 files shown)


docs/dsl/middleware.md
```diff
@@ -285,3 +285,37 @@ object Example extends ZIOAppDefault {
...
 
+## Handler Aspects
+
+Ordinary Middlewares are intended to bracket a request's execution by intercepting the request, possibly modifying it or short-circuiting its execution
+(for example if it fails authentication), and then performing some post-processing on the response.
+However, we sometimes want to gather some contextual information about a request and pass it alongside to the request's handler.
+This can be achieved with the `HandlerAspect[Env, CtxOut]` type, which extends `Middleware[Env]`.
+This middleware produces a value of type `CtxOut` on each request, which the routing DSL will accept just like a path component:
+
+```scala
```
khajavi: Please try to use the `mdoc` as much as possible for snippet codes.

SmedbergM: Done, though I don't know how much that adds here, since these snippets won't compile as written.

khajavi: The goal is to make the code snippets maintainable by ensuring they can be compiled against the source code. So the "crash" modifier isn't helpful in this case.

SmedbergM: I didn't use `crash` but rather `fail` (the former is for code that compiles but fails at runtime, the latter for code that won't compile in the first place).

In any case, I had to resort to the hack of a `mdoc:invisible` block to declare some names.

jdegoes (approved):

jdegoes (approved):

PR #2730, opened 2024-03-15
Bump zio-schema to 1.0.1 and remove snapshot resolver

### LINKED ISSUE #2726: 3.0.0-RC5 References -SNAPSHOT Deps
The 3.0.0-RC5 release includes references to `-SNAPSHOT` deps, resulting in:
```
[error] (update) sbt.librarymanagement.ResolveException: Error downloading dev.zio:zio-schema-protobuf_3:0.4.17+4-0acf5af7-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/jw/.ivy2/local/dev.zio/zio-schema-protobuf_3/0.4.17+4-0acf5af7-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/dev/zio/zio-schema-protobuf_3/0.4.17+4-0acf5af7-SNAPSHOT/zio-schema-protobuf_3-0.4.17+4-0acf5af7-SNAPSHOT.pom
[error] Error downloading dev.zio:zio-schema_3:0.4.17+4-0acf5af7-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/jw/.ivy2/local/dev.zio/zio-schema_3/0.4.17+4-0acf5af7-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/dev/zio/zio-schema_3/0.4.17+4-0acf5af7-SNAPSHOT/zio-schema_3-0.4.17+4-0acf5af7-SNAPSHOT.pom
[error] Error downloading dev.zio:zio-schema-json_3:0.4.17+4-0acf5af7-SNAPSHOT
[error]   Not found
[error]   Not found
[error]   not found: /home/jw/.ivy2/local/dev.zio/zio-schema-json_3/0.4.17+4-0acf5af7-SNAPSHOT/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/dev/zio/zio-schema-json_3/0.4.17+4-0acf5af7-SNAPSHOT/zio-schema-json_3-0.4.17+4-0acf5af7-SNAPSHOT.pom
```

POM for reference:
https://repo1.maven.org/maven2/dev/zio/zio-http_3/3.0.0-RC5/zio-http_3-3.0.0-RC5.pom

Workaround is to manually specify offending deps:
```
  "dev.zio" %% "zio-schema-json"     % "1.0.1",
  "dev.zio" %% "zio-schema-protobuf" % "1.0.1",
  "dev.zio" %% "zio-schema"          % "1.0.1",
```

guersam:
Fix #2726

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -402,10 +402,10 @@ object JsonSchema {
           ref,
           leftSchema.children ++ rightSchema.children,
         )
-      case Schema.Fallback(left, right, fullDecode, _)                                                     =>
+      case Schema.Fallback(left, right, fullDecode, _)                                       =>
         val leftSchema  = fromZSchemaMulti(left, refType)
         val rightSchema = fromZSchemaMulti(right, refType)
-        val innerSchema = 
+        val innerSchema =
           if (fullDecode)
             AnyOfSchema(Chunk(leftSchema.root, rightSchema.root))
           else
@@ -566,10 +566,12 @@ object JsonSchema {
       case Schema.Fail(_, _)             => throw new IllegalArgumentException("Fail schema is not supported.")
       case Schema.Tuple2(left, right, _) => AllOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
       case Schema.Either(left, right, _) => OneOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
-      case Schema.Fallback(left, right, true, _) => AnyOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
-      case Schema.Fallback(left, right, _, _) => OneOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
-      case Schema.Lazy(schema0)          => fromZSchema(schema0(), refType)
-      case Schema.Dynamic(_)             => AnyJson
+      case Schema.Fallback(left, right, true, _) =>
+        AnyOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
+      case Schema.Fallback(left, right, _, _)    =>
+        OneOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
+      case Schema.Lazy(schema0)                  => fromZSchema(schema0(), refType)
+      case Schema.Dynamic(_)                     => AnyJson
 
     }
 
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -405,25 +405,21 @@ object JsonSchema {
       case Schema.Fallback(left, right, fullDecode, _)                                       =>
         val leftSchema  = fromZSchemaMulti(left, refType)
         val rightSchema = fromZSchemaMulti(right, refType)
-        val innerSchema =
+        val candidates  =
           if (fullDecode)
-            OneOfSchema(
-              Chunk(
-                AllOfSchema(Chunk(leftSchema.root, rightSchema.root)),
-                leftSchema.root,
-                rightSchema.root,
-              ),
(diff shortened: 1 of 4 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -402,6 +402,20 @@ object JsonSchema {
...
           leftSchema.children ++ rightSchema.children,
         )
+      case Schema.Fallback(left, right, fullDecode, _)                                       =>
+        val leftSchema  = fromZSchemaMulti(left, refType)
+        val rightSchema = fromZSchemaMulti(right, refType)
+        val innerSchema =
+          if (fullDecode)
+            AnyOfSchema(Chunk(leftSchema.root, rightSchema.root))
+          else
+            OneOfSchema(Chunk(leftSchema.root, rightSchema.root))
```
guersam: I wonder if I understood `Schema.Fallback.fullDecode` correctly here.

987Nabil: I think this is wrong. I took a look how it is handled in schema and `fullDecode` will return a left and a right.
If not set, it will return either a left or right. But in both cases, we can't say if left contais right in their schema or the other way around. And one of says they don't, but any of they say may. Since I don't see a reason to enforce a possible "one of" by checking the schema structure of the two schemas, I think we should just always use any of.

guersam: @987Nabil Thanks for clarifying. I assumed `Schema.Fallback` was equivalent to [`Schema.Either`](https://github.com/zio/zio-http/pull/2730/files/c2a49d6cf3afc22582219242b4f22a113c02b626#diff-1dc286225d0eee11fb5608307e6eef3097d66c5b49426a2d76fc78ba0f05536aR401) when `fullDecode` is off, using `oneOf`.

Just double-checking: is `anyOf(left, right)` sufficient, or should we consider adding a variant for `fullDecode`, like `oneOf(left, right, allOf(left, right))`?

987Nabil: That's a good question. Thinking about it the idea of `oneOf(left, right, allOf(left, right))` is more precise. And I don't see a reason not to do it.

guersam:
Forgot to ping after update: @987Nabil

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -552,8 +577,18 @@ object JsonSchema {
...
+      case Schema.Fallback(left, right, true, _) =>
+        OneOfSchema(
+          Chunk(
+            AllOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType))),
+            fromZSchema(left, refType),
+            fromZSchema(right, refType),
+          ),
+        )
+      case Schema.Fallback(left, right, _, _)    =>
+        AnyOfSchema(Chunk(fromZSchema(left, refType), fromZSchema(right, refType)))
```
987Nabil: I am not sure this should be any of. Since the flag for full decode is false here, we expect exactly one value.
That would fit more to one of. It is true, that we don't know, if left contains right or the other way around.
But this would also make this API unusable.
Imagine we have a `case class User(name: String)` and `case class Admin(name: String, roles:List[String]`, if you hand over a json with `name` and `roles`, which one will be deserialized (additional fields are allowed by default)? We have here no discriminator. Only structure. If we use one of and one contains the other, at least the error is visible.

guersam: Ok, I'll change this.

Off the topic, but one thing I'm worried of is the spec doesn't guarantee that the order of elements in `oneOf` or `anyOf` matters, so we can't be 100% sure that it exactly expresses the purpose of `Fallback`.

987Nabil (approved):

yisraelU:
@987Nabil can a new RC be published, current RC5 is broken because of this

987Nabil:
There is more that needs to be fixed. But I'll do it soonish

PR #2755, opened 2024-03-30
Add tapErrorZIO and tapErrorCauseZIO to Route and Routes

tjarvstrand:
Useful for things like error logging or reporting to external services such as Sentry.


zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -118,6 +118,33 @@ sealed trait Route[-Env, +Err] { self =>
...
+  /**
+   * Effectfully peeks at the unhandled failure cause of this Route.
+   */
+  final def tapErrorCauseZIO[Err1 >: Err](
+    f: Cause[Err] => ZIO[Any, Err1, Any],
+  )(implicit trace: Trace): Route[Env, Err1] =
+    self match {
+      case Provided(route, env)                        => Provided(route.tapErrorCauseZIO(f), env)
+      case Augmented(route, aspect)                    => Augmented(route.tapErrorCauseZIO(f), aspect)
+      case Handled(routePattern, handler, location)    => Handled(routePattern, handler, location)
```
987Nabil: iirc, ensures handled that the typed error is handled. But a cause (defect) might still happen. So I think there should be a tap on the handler here.

tjarvstrand: That doesn't seem to work as the `Handled.handler`'s failure type is `Response` whereas `f`'s failure type is `Err` which is not necessarily a `Response`.

jdegoes: You can catch the `Cause[Err]`, and if it contains an `Err`, then you rethrow it. But if it's really a `Cause[Nothing]`, i.e. pure defect, without an `Err` inside it, then you can allow that to be caught by `f`, since `f` can handle those (`Cause[Nothing]` is a subtype of `Cause[Err]`).

jdegoes: @tjarvstrand This does need to be fixed, and then will be good to merge!

tjarvstrand: Sorry, busy week! :/

I've been struggling a bit with the failure case of `f`. Does something like this make sense?

```
/**
   * Effectfully peeks at the unhandled failure cause of this Route.
   */
  final def tapErrorCauseZIO[Err1 >: Err](
    f: Cause[Err] => ZIO[Any, Err1, Any],
  )(implicit trace: Trace): Route[Env, Err1] =
    self match {
      case Provided(route, env)                        => Provided(route.tapErrorCauseZIO(f), env)
      case Augmented(route, aspect)                    => Augmented(route.tapErrorCauseZIO(f), aspect)
      case Handled(routePattern, handler, location)    =>
        Handled(
          routePattern,
          handler.tapErrorCauseZIO {
            case cause0: Cause[Err] => f(cause0).catchAllCause(cause => ZIO.succeed(Response.fromCause(cause)))
            case _                  => ZIO.unit
          },
          location,
        )
      case Unhandled(rpm, handler, zippable, location) =>
        Unhandled(rpm, handler.tapErrorCauseZIO(f), zippable, location)
    }
```

tjarvstrand: Sorry, busy week! :/

I've been struggling a bit with the error case of `f` which, unless I'm mistaken, must be a `Response`. Does something like this make sense?
```
/**
   * Effectfully peeks at the unhandled failure cause of this Route.
   */
  final def tapErrorCauseZIO[Err1 >: Err](
    f: Cause[Err] => ZIO[Any, Err1, Any],
  )(implicit trace: Trace): Route[Env, Err1] =
    self match {
      case Provided(route, env)                        => Provided(route.tapErrorCauseZIO(f), env)
      case Augmented(route, aspect)                    => Augmented(route.tapErrorCauseZIO(f), aspect)
      case Handled(routePattern, handler, location)    =>
        Handled(
          routePattern,
          handler.tapErrorCauseZIO {
            case cause0: Cause[Err] => f(cause0).catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))
            case _                  => ZIO.unit
          },
          location,
        )
      case Unhandled(rpm, handler, zippable, location) =>
        Unhandled(rpm, handler.tapErrorCauseZIO(f), zippable, location)
    }
```

jdegoes: When you get `cause0`, do `errorOrCause`, and if it's an error, you have to refail that, but if it's a `Cause[Nothing]`, you can feed it to `f`.

tjarvstrand: Maybe I'm slow, but shouldn't `handler.tapErrorCauseZIO` leave the error channel of the original effect intact so that I don't have to refail it?

Not trying to be difficult. I just want to make sure I understand what's going on.

jdegoes: Ah yes, of course! Forgot the point of this function, which is tapping not recovering. 🤦‍♂️

tjarvstrand: 👍 

Tried to move too fast and now the build is failing. I'll have another look at this as soon as I can.

tjarvstrand: I'm afraid I think I'm in need of some assistance.

There are warnings in Scala 3 (which I have actually not used yet):
```
[warn] -- Unchecked Warning: /Users/tjarvstrand/src/zio-http/zio-http/shared/src/main/scala/zio/http/Route.scala:134:17
[warn] 134 |            case err: Fail[Err] => f(err.value).catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))
[warn]     |                 ^
[warn]     |the type test for zio.Cause.Fail[Err] cannot be checked at runtime because its type arguments can't be determined from zio.Cause[zio.http.Response]
```

I'm assuming this is related to type erasure somehow, but I'm not sure why there's no warning in Scala 2. Any tips on how to proceed?

guersam: @tjarvstrand You can use the own methods of `Cause` to determine whether it's a failure or a defect.

You can refer to [the implementations of `ZIO#tapDeffect` and `ZIO#tapError`](https://github.com/zio/zio/blob/v2.1.3/core/shared/src/main/scala/zio/ZIO.scala#L2034-L2065), and there might be a more convenient method to fold defect and failure separately.

tjarvstrand: @guersam Thanks, AFAICT the issue here is checking whether the `f`s input (`Cause[Err]`) is a subtype of `Cause[Response]` or not (since `handler` is a `Handler[Any, Response, Request, Response]`) and only call it then

guersam: @tjarvstrand What I meant originally was we don't need pattern matching to determine if it's `Cause[Nothing]` or not, according to John's previous comment.

> You can catch the Cause[Err], and if it contains an Err, then you rethrow it. But if it's really a Cause[Nothing], i.e. pure defect, without an Err inside it, then you can allow that to be caught by f, since f can handle those (Cause[Nothing] is a subtype of Cause[Err]).

But now we want to let `f` handle failure cases as well when `Err =:= Response`, right?

I think there are two options:
1. ~~Extract the possible failure with something like `Cause#failureOrCause`, and do runtime type check for the failure case using `.isInstanceOf[Response]`. It's the simplest solution.~~ EDIT: It seems irrelevant.
2. Require an implicit `zio.Tag[Err]` or `scala.reflect.ClassTag[Err]` and check its runtime class.
3. Add an implicit evidence like this:

    ```scala
    sealed trait IsResponse[+A] { def isResponse: Boolean }
    object IsResponse {
      implicit val response: IsResponse[Response] = new IsResponse[Response] { isResponse = true }
    
      // to avoid unnecessary allocation
      private val otherInstance: IsResponse[Nothing] = new IsResponse[Nothing] { isResponse = false }
      implicit def other: IsResponse[A] = otherInstance
    }
    ```

    The name `IsResponse` might cause confusion with `Handler.IsRequest`, which has the similar purpose but it's for compile time check only.

guersam: By the way, `Handled.handler` is a handler that produces another handler ([code](https://github.com/zio/zio-http/blob/v3.0.0-RC9/zio-http/shared/src/main/scala/zio/http/Route.scala#L460)), so checking `handler`'s failure won't achieve what you want. You need `handler.map(_.tapErrorCauseZIO(...))` to catch the `Response` failure.

tjarvstrand: Thanks a lot for your comments!

If I understand it correctly, both alternative 2 and 3 would require a call to `asInstanceOf`, since neither of them provide a type level proof that `Err <: Response`. Is that correct?

> By the way, Handled.handler is a handler that produces another handler ([code](https://github.com/zio/zio-http/blob/v3.0.0-RC9/zio-http/shared/src/main/scala/zio/http/Route.scala#L460)), so checking handler's failure won't achieve what you want. You need handler.map(_.tapErrorCauseZIO(...)) to catch the Response failure.

Aha, I missed that! Thanks!

tjarvstrand: Alternative 2 sadly doesn't work since `ClassTag` is invariant: `Covariant type Err occurs in invariant position in type ClassTag[Err] of value ct`.

I gave alternative 3 a shot instead, but I had to make it contravariant rather than covariant.

tjarvstrand: BTW, in other parts of `Route.scala`, I see things like this:
```
handler.map { handler =>
  Handler.fromFunctionHandler[Request] { (req: Request) =>
    handler.mapErrorCause(c => f(req, c.asInstanceOf[Cause[Nothing]]))
  }
}
```

Is this really safe? To me it looks like there has perhaps been a mixup, since the outer `handler` is infallible, but the inner one isn't.

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,52 @@ sealed trait Route[-Env, +Err] { self =>
...
+    f: Err => ZIO[Any, Err1, Any],
+  )(implicit trace: Trace, ev: IsResponseCauseHandler[Err]): Route[Env, Err1] =
+    self match {
+      case Provided(route, env)                        => Provided(route.tapErrorZIO(f), env)
+      case Augmented(route, aspect)                    => Augmented(route.tapErrorZIO(f), aspect)
+      case Handled(routePattern, handler, location)    =>
+        Handled(
+          routePattern,
+          handler.map(_.tapErrorCauseZIO {
+            case err: Fail[_] if ev.isResponse =>
```
guersam: If `ev.IsResponse == false` we don't need `handler.map(...)` at all.

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,52 @@ sealed trait Route[-Env, +Err] { self =>
...
+  )(implicit trace: Trace, ev: IsResponseCauseHandler[Err]): Route[Env, Err1] =
+    self match {
+      case Provided(route, env)                        => Provided(route.tapErrorCauseZIO(f), env)
+      case Augmented(route, aspect)                    => Augmented(route.tapErrorCauseZIO(f), aspect)
+      case Handled(routePattern, handler, location)    =>
+        Handled(
+          routePattern,
+          handler.map(_.tapErrorCauseZIO {
+            case cause0 if ev.isResponse =>
+              f(cause0.asInstanceOf[Cause[Err]]).catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))
```
guersam: If the cause is not a failure, it can be passed to `f` even if `Err =!= Response` as John pointed out in the previous comment I quoted.

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -489,4 +537,16 @@ object Route                   {
     }
   }
 
+  sealed trait IsResponseCauseHandler[-A] { def isResponse: Boolean }
```
guersam: The most type evidences starting with `Is` tend provide the implicit instance only and if only the type.

As it always provides the implicit value, maybe we need another name to avoid confusion.

Possible alternatives:
- `DetermineResponse`
- `CheckResponse`
- `ResponseTypeChecker`
- ...

I think `CauseHandler` is somewhat misleading as this type is not directly related to handler.

zio-http/shared/src/main/scala/zio/http/Route.scala
guersam: You can't directly summon `ClassTag[Err]` because it's invariant. Instead, introduce a new type variable like `Err1`.

Check out [ZIO#refinedToOrDie](https://github.com/zio/zio/blob/v2.1.5/core/shared/src/main/scala/zio/ZIO.scala#L5320-L5321) for an example:

```scala
def refineToOrDie[E1 <: E: ClassTag](implicit ev: CanFail[E], trace: Trace): ZIO[R, E1, A] =
  self.refineOrDie { case e: E1 => e }
```

zio-http/shared/src/main/scala/zio/http/Route.scala
tjarvstrand: True!

zio-http/shared/src/main/scala/zio/http/Route.scala
tjarvstrand: Ah, yeah, sorry I misread that one!

zio-http/shared/src/main/scala/zio/http/Route.scala
tjarvstrand: Good point. Changing it to CheckResponse

tjarvstrand:
Thanks for the feedback!

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,57 @@ sealed trait Route[-Env, +Err] { self =>
...
+  )(implicit trace: Trace, ev: CheckResponse[Err]): Route[Env, Err1] =
+    self match {
+      case Provided(route, env)                        => Provided(route.tapErrorZIO(f), env)
+      case Augmented(route, aspect)                    => Augmented(route.tapErrorZIO(f), aspect)
+      case Handled(routePattern, handler, location)    =>
+        Handled(
+          routePattern,
+          if (ev.isResponse) {
+            handler.map(_.tapErrorCauseZIO {
+              case err: Fail[_] if ev.isResponse =>
```
guersam: Now we don't need this pattern guard :)

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,57 @@ sealed trait Route[-Env, +Err] { self =>
...
+                f(err.value.asInstanceOf[Err]).catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))
```
guersam: Now I doubt this `.catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))` is necessary, because it may cause unexpected behavior when there is a global error handler, say catching all defects and produces a formatted error response.

As we're sure that `Err =:= Response`, just '.asInstance[...]' can be enough.

zio-http/shared/src/main/scala/zio/http/Route.scala
guersam: Same here about the `.catchAllCause`

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,57 @@ sealed trait Route[-Env, +Err] { self =>
...
+      case Handled(routePattern, handler, location)    =>
+        Handled(
+          routePattern,
+          handler.map(_.tapErrorCauseZIO {
+            case cause0 if ev.isResponse =>
+              f(cause0.asInstanceOf[Cause[Err]]).catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))
+            case _: Fail[_]              =>
+              ZIO.unit
+            case cause0                  =>
+              f(cause0.asInstanceOf[Cause[Nothing]]).catchAllCause(cause => ZIO.fail(Response.fromCause(cause)))
```
guersam: Same here about the `.catchAllCause`.

987Nabil:
@tjarvstrand Could you handle the review comments? I'd like to get this in for 3.0

zio-http/shared/src/main/scala/zio/http/Route.scala
tjarvstrand: I think we do need the `catchAllCause` since we don't know that `Err1 <: Response`

zio-http/shared/src/main/scala/zio/http/Route.scala
tjarvstrand: Removed!

tjarvstrand:
The tests seem to fail on something unrelated to my changes. Not sure if it's a fluke but I don't seem to be allowed to restart the workflow.

987Nabil:
I updated your branch. Let's see if it builds

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,57 @@ sealed trait Route[-Env, +Err] { self =>
...
+  final def tapErrorZIO[Err1 >: Err](
+    f: Err => ZIO[Any, Err1, Any],
+  )(implicit trace: Trace, ev: CheckResponse[Err]): Route[Env, Err1] =
+    self match {
+      case Provided(route, env)                        => Provided(route.tapErrorZIO(f), env)
+      case Augmented(route, aspect)                    => Augmented(route.tapErrorZIO(f), aspect)
+      case Handled(routePattern, handler, location)    =>
+        Handled(
+          routePattern,
+          if (ev.isResponse) {
```
987Nabil: This is always true. `Handled` has a fixed error type of `Nothing` and will pick up the `implicit val response: CheckResponse[Response]`

Handled Routes are by definition routes without error. This should just returned the route without any modification.
The becomes clear when substituting the types. Because f is `Nothing => ZIO[Any, Err1, Any]`

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -131,6 +133,57 @@ sealed trait Route[-Env, +Err] { self =>
...
+    f: Cause[Err] => ZIO[Any, Err1, Any],
+  )(implicit trace: Trace, ev: CheckResponse[Err]): Route[Env, Err1] =
+    self match {
+      case Provided(route, env)                        => Provided(route.tapErrorCauseZIO(f), env)
+      case Augmented(route, aspect)                    => Augmented(route.tapErrorCauseZIO(f), aspect)
+      case Handled(routePattern, handler, location)    =>
+        Handled(
+          routePattern,
+          handler.map(_.tapErrorCauseZIO {
+            case cause0 if ev.isResponse =>
```
987Nabil: `ev.isResponse` is always true. Also the concrete type of `f` here is `Cause[Nothing] => ZIO[Any, Err1, Any]`

987Nabil (changes requested):

987Nabil:
Please add some tests. For example adding a log output and use the test logger to find the message you logged. Or use a promise. Or whatever comes to your mind :)

987Nabil:
@tjarvstrand could you give this PR the finishing touch? :)

tjarvstrand:
Looking into writing some tests ATM. In the meantime, I think I'm lacking the context needed to make wise decisions on my own here 😄 

Just to make sure I don't misunderstand, you are saying that I should basically just return the same `Handled` value on both what is currently line 172 and 197? E.g.:
```
  final def tapErrorCauseZIO[Err1 >: Err](
    f: Cause[Err] => ZIO[Any, Err1, Any],
  )(implicit trace: Trace, ev: CheckResponse[Err]): Route[Env, Err1] =
    self match {
      case Provided(route, env)                        => Provided(route.tapErrorCauseZIO(f), env)
      case Augmented(route, aspect)                    => Augmented(route.tapErrorCauseZIO(f), aspect)
      case handled @ Handled(_, _, _)                  => handled
      case Unhandled(rpm, handler, zippable, location) =>
        Unhandled(rpm, handler.tapErrorCauseZIO(f), zippable, location)
    }
```

EDIT: I think I figured most of it out. Everything became clearer once my mind got back into "scala mode". Please have another look.

However, I noticed that the handler of an `Unhandled` doesn't seem to be evaluated. Should that case also just return the same value?

zio-http/shared/src/main/scala/zio/http/Route.scala
tjarvstrand: Having `f` as `Cause[Nothing] => ZIO[Any, Err1, Any]` doesn't seem to work as it makes the call to `handler.tapErrorCauseZIO` in the `Unhandled`-case sad.

987Nabil (approved):

PR #2773, opened 2024-04-17
Add the `Date` header to responses by default

kyri-petrou:
Accoding to the [HTTP spec](https://www.rfc-editor.org/rfc/rfc7231#section-7.1.1.2), a server MUST generate a `Date` header unless it can't provide a reasonable approximation of the current instance in Coordinated Universal Time.

Since this doesn't apply to us, `zio-http` should be including the `Date` header by default to all responses (with the exception of 1xx and 5xx responses where the Date header can be omitted)

The main issue with providing a `Date` header, and why it was omitted in the past, is that it relies on retrieving the current date-time and rendering it. Since getting the date-time requires a system call, this is a somewhat expensive operation. In order to avoid it for every single request, this PR implements a date header value cache, which is refreshed by an single thread every 1 second (which is the granularity of the header value).

After testing this implementation, I found that in some very rare cases, the header value will be lagging 1 second from the actual time, and this is usually when the time is retrieved exactly at at the same time as the second changes. The scheduled thread will then refresh the cached value within less than 1ms in most cases

As the spec allows for a small error in the header value, I think that this error is acceptable given the performance benefit of this approach.

Based on the benchmark I added, this approach is ~250x faster than rendering the Date fresh each time, and adds virtually no overhead.

```
[info] Benchmark                                  Mode  Cnt    Score   Error  Units
[info] CachedDateHeaderBenchmark.benchmarkCached  avgt    5    1.576 ± 0.013  ns/op
[info] CachedDateHeaderBenchmark.benchmarkFresh   avgt    5  268.054 ± 4.272  ns/op
```

Happy to hear your thoughts on this!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,6 +209,40 @@ jobs:
         id: push_codecov
         run: 'bash <(curl -s https://codecov.io/bash)'
 
+  Jmh_CachedDateHeaderBenchmark:
+    name: Jmh CachedDateHeaderBenchmark
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        scala: [2.13.12]
+        java: [temurin@8]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: zio-http
+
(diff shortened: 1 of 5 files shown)


tusharmath:
Can you update the workflows. There is a sbt command for it.

tusharmath (approved):

PR #2775, opened 2024-04-17
Showcase for Both Client and Server

khajavi:
Let's introduce the simplicity of ZIO HTTP for both Client and Server APIs on the landing page.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -18,37 +18,60 @@ libraryDependencies += "dev.zio" %% "zio-http" % "@VERSION@"
 
 **NOTES ON VERSIONING:**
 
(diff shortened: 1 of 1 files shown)


khajavi:
@jdegoes This is ready for review!

jdegoes (approved):

erikvanoosten:
@khajavi In this PR section `Steps to run an example` was removed but section `Watch Mode` was retained. I think the second doesn't make sense without the first. Shall we restore the `Steps to run an example` section?

khajavi:
@erikvanoosten That section contained something magical that was not suitable for end users, but for contributors who accessed the source code! But you are right that the 'watch mode' does not make sense without the 'steps to run' section.

PR #2778, opened 2024-04-17
Provide HandlerAspect context via env instead of function param (#2488)

### LINKED ISSUE #2488: There's no way to use a "providing middleware" on a `Routes` nor an `HttpApp`

987Nabil:
This PR is a different way of solving #2488 that seems to me to be superior to #2553 

With this PR, the context provided by a `HandlerAspec` is no longer injected via the function param if a `handler`/`Handler.fromFunction` but instead via the environment. So users would just require context via `ZIO.serviceX`.
The implementation is such, that `h: Handler[Env with Ctx, _, _, _] @@ aspect: HandlerAspect[_, Ctx]` would eliminate the  `Ctx` from the env of the handler. The aspect may add new `Env` requirements as well. Only for partial `Env` providing in Scala 2, an explicit param needs to be handed over to `@@`. For all other cases inference works fine.

This change also enables auth context to be handed over to handlers of endpoint API implementations, since it does not require the context to be known while creating the endpoint. There is an example in the added tests.

This seems to be the most consistent way to provide context via an aspect that works for all use cases
 * single handler
 * Routes
 * Endpoint API

fixes #2488
/claim #2488

### DIFF
--- a/zio-http-example/src/main/scala/example/AuthenticationServer.scala
+++ b/zio-http-example/src/main/scala/example/AuthenticationServer.scala
@@ -46,8 +46,8 @@ object AuthenticationServer extends ZIOAppDefault {
   def app: HttpApp[Any] =
     Routes(
       // A route that is accessible only via a jwt token
-      Method.GET / "profile" / "me" -> handler { (name: String, _: Request) =>
-        ZIO.succeed(Response.text(s"Welcome $name!"))
+      Method.GET / "profile" / "me" -> handler { (_: Request) =>
+        ZIO.serviceWith[String](name => Response.text(s"Welcome $name!"))
       } @@ bearerAuthWithContext,
 
       // A login route that is successful only if the password is the reverse of the username
--- a/zio-http/shared/src/main/scala-2/zio/http/HandlerVersionSpecific.scala
+++ b/zio-http/shared/src/main/scala-2/zio/http/HandlerVersionSpecific.scala
@@ -1,0 +1,22 @@
+package zio.http
+
+import zio._
+
+trait HandlerVersionSpecific {
+  private[http] class ApplyContextAspect[-Env, +Err, -In, +Out, Env0](private val self: Handler[Env, Err, In, Out]) {
+    def apply[Env1, Ctx: Tag, In1 <: In](aspect: HandlerAspect[Env1, Ctx])(implicit
+      in: Handler.IsRequest[In1],
+      ev: Env0 with Ctx <:< Env,
+      out: Out <:< Response,
+      err: Err <:< Response,
+      trace: Trace,
+    ): Handler[Env0 with Env1, Response, Request, Response] =
+      aspect.applyHandlerContext {
+        handler { (ctx: Ctx, req: Request) =>
+          val handler: ZIO[Env, Response, Response] = self.asInstanceOf[Handler[Env, Response, Request, Response]](req)
+          handler.provideSomeEnvironment[Env0](_.add[Ctx](ctx).asInstanceOf[ZEnvironment[Env]])
+        }
+      }
+  }
+
+}
--- a/zio-http/shared/src/main/scala-3/zio/http/HandlerVersionSpecific.scala
+++ b/zio-http/shared/src/main/scala-3/zio/http/HandlerVersionSpecific.scala
@@ -1,0 +1,22 @@
+package zio.http
+
+import zio.*
+
+trait HandlerVersionSpecific {
+  private[http] class ApplyContextAspect[-Env, +Err, -In, +Out, Env0](self: Handler[Env, Err, In, Out]) {
+    transparent inline def apply[Env1, Ctx, In1 <: In](aspect: HandlerAspect[Env1, Ctx])(implicit
+      in: Handler.IsRequest[In1],
+      ev: Env0 with Ctx <:< Env,
+      out: Out <:< Response,
+      err: Err <:< Response,
+      trace: Trace,
+    ): Handler[Env0 with Env1, Response, Request, Response] =
+      aspect.applyHandlerContext {
+        handler { (ctx: Ctx, req: Request) =>
+          val handler: ZIO[Env, Response, Response] = self.asInstanceOf[Handler[Env, Response, Request, Response]](req)
+          handler.provideSomeEnvironment[Env0](_.add(ctx).asInstanceOf[ZEnvironment[Env]])
+        }
+      }
+  }
+
+}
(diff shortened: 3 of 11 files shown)


zio-http/shared/src/main/scala-3/zio/http/HandlerVersionSpecific.scala
```diff
@@ -0,0 +1,22 @@
...
+      in: Handler.IsRequest[In1],
+      ev: Env0 with Ctx <:< Env,
+      out: Out <:< Response,
+      err: Err <:< Response,
+      trace: Trace,
+    ): Handler[Env0 with Env1, Response, Request, Response] =
+      aspect.applyHandlerContext {
+        handler { (ctx: Ctx, req: Request) =>
+          val handler: ZIO[Env, Response, Response] = self.asInstanceOf[Handler[Env, Response, Request, Response]](req)
+          handler.provideSomeLayer[Env0 with Env1](ZLayer.environment[Env0] ++ ZLayer.environment[Env1] ++ ZLayer.succeed(ctx))
```
guersam: Might be simplified:

```suggestion
          handler.provideSomeEnvironment[Env0 with Env1](_.add(ctx))
```

~Not sure if it'll work for Scala 2 either.~

zio-http/shared/src/main/scala-2/zio/http/HandlerVersionSpecific.scala
```diff
@@ -0,0 +1,25 @@
...
+      err: Err <:< Response,
+      trace: Trace,
+      tag: Tag[Env0],
+    ): Handler[Env0 with Env1, Response, Request, Response] =
+      aspect.applyHandlerContext {
+        handler { (ctx: Ctx, req: Request) =>
+          val handler: ZIO[Env, Response, Response] = self.asInstanceOf[Handler[Env, Response, Request, Response]](req)
+          handler.provideSomeLayer[Env0 with Env1](
+            ZLayer.environment[Env0] ++ ZLayer.environment[Env1] ++ ZLayer.succeed(ctx),
+          )
```
guersam: Might be simplified:

```suggestion
          handler.provideSomeEnvironment[Env0 with Env1](_.add[Ctx](ctx).asInstanceOf[ZEnvironment[Env]])
```

~~`asInstanceOf` was needed in Scala 3. Not sure if Scala 2 can just infer this.~~

zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -42,18 +48,23 @@ sealed trait Handler[-R, +Err, -In, +Out] { self =>
...
-  ): Handler[Env1, Response, Request, Response] = {
-    def convert(handler: Handler[R, Err, In, Out]): Handler[R, Response, In1, Response] =
-      handler.asInstanceOf[Handler[R, Response, In1, Response]]
+    tag: Tag[Ctx],
+    tag1: Tag[Env0],
+  ): Handler[Env0, Response, Request, Response] =
+    aspect.applyHandlerContext {
+      handler { (ctx: Ctx, req: Request) =>
+        val handler: ZIO[Ctx, Response, Response] = self.asInstanceOf[Handler[Ctx, Response, Request, Response]](req)
+        handler.provideSomeLayer[Env0](ZLayer.environment[Env0] ++ ZLayer.succeed(ctx))
```
guersam:
```suggestion
        handler.provideSomeEnvironment[Env0](_.add[Ctx](ctx))
```

`Ctx` type should be annotated for Scala 2.

zio-http/shared/src/main/scala-2/zio/http/HandlerVersionSpecific.scala
```diff
@@ -0,0 +1,25 @@
+package zio.http
+
+import zio.{Tag, Trace, ZIO, ZLayer}
```
guersam:
```suggestion
import zio.{Tag, Trace, ZIO, ZEnvironment, ZLayer}
```

For `asInstanceOf` at line 18

zio-http-example/src/main/scala/example/AuthenticationServer.scala
```diff
@@ -46,8 +46,8 @@ object AuthenticationServer extends ZIOAppDefault {
   def app: HttpApp[Any] =
     Routes(
       // A route that is accessible only via a jwt token
-      Method.GET / "profile" / "me" -> handler { (name: String, _: Request) =>
-        ZIO.succeed(Response.text(s"Welcome $name!"))
+      Method.GET / "profile" / "me" -> handler { (_: Request) =>
+        ZIO.serviceWith[String](name => Response.text(s"Welcome $name!"))
       } @@ bearerAuthWithContext,
```
khajavi, line 51: By the changes in this PR, we can move `bearerAuthWithContext` to apply it to the whole routes on line 72 :100:

guersam:
It's just what I have been wanted, thanks for the work, @987Nabil  :D 

Providing context via environment has a great DX benefit that attaching a middleware in the future doesn't break existing handlers' interface.

Also, as one of the primary use case for this is authentication, there are many endpoints that don't care of the exact user identity but as long as the user is authenticated. This implementation enables mixing two types of endpoints seamlessly .

jdegoes (approved):
Nice work! You're going to make a lot of people happy!

PR #2779, opened 2024-04-18
Dependency updates & CI improvements

kyri-petrou:
Bumped the versions on quite a few dependencies that haven't been updated for a while. The netty update required fixing some deprecations. As I was resolving the deprecations, the one thing that was confusing to me is why we have request header validation disabled by default?

I'm guessing that this is for performance reasons - as such I think it's still probably a good idea to expose that as an option to users via the config so I added it.

Finally, I realised that in some cases, SBT would run out of code profiling memory, which would make compiling much much slower. I increased the `NonProfiledCodeHeapSize` to 256m (default is 128m) to prevent this from happening


kyri-petrou:
Aghh `NonProfiledCodeHeapSize` is not supported on JDK8. I reverted that change - although I'm wondering if testing on JDK8 is still needed since it's almost at EOL; but I guess that's a question for some other time

EDIT: Nevermind, I just realised that EOL for JDK8 was moved back from 2024 to 2030 🙃

guizmaii:
> although I'm wondering if testing on JDK8 is still needed since it's almost at EOL

Java8 support has been dropped in ZIO ecosystem, no need to support it in zio-http. Minimum Java version we officially support is Java11+ (we're more conservative than the Java env itself which has started to support Java17+ only)
FYI, same for Scala 2.11 (haven't check if it used)

project/BuildHelper.scala
```diff
@@ -6,12 +6,12 @@ import xerial.sbt.Sonatype.autoImport.*
...
-  val Scala3           = "3.3.1"
-  val ScoverageVersion = "2.0.9"
-  val JmhVersion       = "0.4.3"
-  val SilencerVersion  = "1.7.14"
+  val Scala212         = "2.12.19"
+  val Scala213         = "2.13.13"
+  val Scala3           = "3.3.3"
+  val ScoverageVersion = "2.0.11"
+  val JmhVersion       = "0.4.7"
+  val SilencerVersion  = "1.7.16"
```
guizmaii: FYI, Silencer can be removed as it's only needed for Scala 2.11. The `@nowarn` annotation has been added in Scala 2.12.

kyri-petrou:
Hey @guizmaii do you know why all of a sudden there were compiling errors (I'm guessing from warnings?) all over the codebase? With some of them being actual bugs that were not caught before like this one: https://github.com/zio/zio-http/pull/2779/files#diff-70952fe5aaf159adfde720d4292f9912a26a6c6d0cdf7bc4fc2c80348d866415L71-L72 👀

kyri-petrou:
Superseded by #2853

PR #2782, opened 2024-04-18
Use `fork` instead of `runOrFork` in NettyRuntime

### LINKED ISSUE #2577: Requests are incorrectly(?) executed on Netty's event loop executor
**Describe the bug**

I'm not sure if this a bug or a feature, but I thought to open this issue just to discuss this behaviour.

Currently, effectful requests are being executed on Netty's executor until a yield operation (or similar) occurs (see reproducer below). I'm not sure whether this is by design (minimize executor context switching) or whether this is a bug. The main issue with this (assuming it's by design) is that with ZIO 2.1-RC1 which uses virtual threads for JDK21+, blocking effects might end up being executed on Netty's event loop group executor which uses platform threads, potentially leading to thread starvation.

Having a look through the code, a simple solution I see would be to "force" the program to be run on ZIO's executor by changing [this line](https://github.com/zio/zio-http/blob/badb929383cdb41d5f88558e60b5e49878c250f4/zio-http/src/main/scala/zio/http/netty/NettyRuntime.scala#L52C5-L52C42) to something like:

```scala
    rtm.unsafe.runOrFork(ZIO.descriptorWith(d => program.onExecutor(d.executor)))
```

Having said that, I don't fully understand the heuristics behind `runOrFork` to know whether in cases that the effect can be run without forking should be executed on ZIOs or Netty's executor. I'm not too sure on the performance implications of doing so, but perhaps in JDK21+ & ZIO 2.1+ all effects should be run with `rtm.unsafe.fork`?

**To Reproduce**

```scala
import zio.*
import zio.http.*

  private def threadName =
    ZIO.succeed(Response.text(Thread.currentThread().getName))
  def run =
    Server.serve {
        HttpApp(Routes(
          Method.GET / "foo" -> handler(threadName),
          Method.GET / "bar" -> handler(ZIO.yieldNow *> threadName),
        ))
      }
      .provide(Server.defaultWithPort(8080))
}
```

And then:

```shell
curl http://localhost:8080/foo
# KQueueEventLoopGroup-2-2

curl http://localhost:8080/bar 
# ZScheduler-Worker-0
```

kyri-petrou:
/claim #2577
/fixes #2577

I'll also need to backport https://github.com/zio/zio/pull/8752 to 2.0.x since that bug is preventing us from getting the benefit of this change

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
@@ -49,6 +49,7 @@ private[zio] trait NettyRuntime { self =>
     // When connection closes, interrupt the program
     var close: GenericFutureListener[Future[_ >: Void]] = null
 
+    // See https://github.com/zio/zio-http/pull/2782 on why forking is preferable over runOrFork
     val fiber = rtm.unsafe.fork(program)
 
     if (interruptOnClose) {
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
@@ -49,28 +49,21 @@ private[zio] trait NettyRuntime { self =>
     // When connection closes, interrupt the program
     var close: GenericFutureListener[Future[_ >: Void]] = null
 
-    rtm.unsafe.runOrFork(program) match {
-      case Left(fiber) =>
-        if (interruptOnClose) {
-          close = closeListener(rtm, fiber)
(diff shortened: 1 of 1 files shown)


kyri-petrou:
Upstream PR merged: https://github.com/zio/zio/pull/8764

987Nabil (approved):

987Nabil:
@kyri-petrou can we merge this before the zio release?

kyri-petrou:
Yeah I don't think a release is needed to merge it. We just won't get the benefit until a new zio release is out

987Nabil:
Hick up in pipeline. I restarted it and activated auto merge

PR #2783, opened 2024-04-19
Implement client SSL certificates

WesselVS:
Thanks for this great library! I have a use case where I have a server and need to verify the identity of the client. I want to use SSL client certificates for this, but zio-http has no support for configuring this with netty. I added support for the client and the server. Additionally, the `Request` type also optionally contains the certificate provided by the remote client.

One thing I noticed is that it seems hard to properly extend the SSL config types (`ClientSSLConfig`, `SSLConfig`) without breaking existing configurations. Please let me know how to improve here.


zio-http/shared/src/main/scala/zio/http/ClientSSLConfig.scala
```diff
@@ -22,31 +22,61 @@ import zio.Config.Secret
 sealed trait ClientSSLConfig
 
 object ClientSSLConfig {
+  val clientSSLCertConfig: Config[ClientSSLCertConfig] = {
```
jdegoes: It's already namespaced inside `ClientSSLConfig`, so it can have a shorter name: `config`.

zio-http/shared/src/main/scala/zio/http/ClientSSLConfig.scala
```diff
@@ -22,31 +22,61 @@ import zio.Config.Secret
...
   case object Default                                                                         extends ClientSSLConfig
   final case class FromCertFile(certPath: String)                                             extends ClientSSLConfig
   final case class FromCertResource(certPath: String)                                         extends ClientSSLConfig
   final case class FromTrustStoreResource(trustStorePath: String, trustStorePassword: Secret) extends ClientSSLConfig
+  final case class FromClientAndServerCert(
+    serverCertConfig: ClientSSLConfig,
+    clientCertConfig: ClientSSLCertConfig,
+  ) extends ClientSSLConfig
+
+  final case class FromClientCertFile(certPath: String, keyPath: String)     extends ClientSSLCertConfig
```
jdegoes: By convention and to aid organization, these constructors should be located in the companion object of `ClientSSLCertConfig`.

jdegoes (changes requested):
Two minor comments but looks good! Thanks for your work on this.

zio-http/shared/src/main/scala/zio/http/ClientSSLConfig.scala
WesselVS: Done!

zio-http/shared/src/main/scala/zio/http/ClientSSLConfig.scala
WesselVS: Moved to companion object

WesselVS:
> Two minor comments but looks good! Thanks for your work on this.

No problem, thanks for your hard work!

jdegoes (approved):

PR #2787, opened 2024-04-21
rename enumValues to enum to comply with OSA 3

lachezar:
Rename enumValues to enum to comply with OSA 3 as described here - https://swagger.io/docs/specification/data-models/enums/

The current generated code for enums look like
```
"ProductType":{"type":"string","enumValues":["Electronics","Clothes","VideoGames","Toys"]}
```
which is not rendering correct in Swagger UI - there are not enum values listed in the UI, just the type "string". 

However if I change "enumValues" to "enum", then the Sagger UI correctly shows the possible values and tags the type as enumeration.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -18,7 +18,7 @@ private[openapi] case class SerializableJsonSchema(
   oneOf: Option[Chunk[SerializableJsonSchema]] = None,
   allOf: Option[Chunk[SerializableJsonSchema]] = None,
   anyOf: Option[Chunk[SerializableJsonSchema]] = None,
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -18,7 +18,7 @@ private[openapi] case class SerializableJsonSchema(
   oneOf: Option[Chunk[SerializableJsonSchema]] = None,
   allOf: Option[Chunk[SerializableJsonSchema]] = None,
   anyOf: Option[Chunk[SerializableJsonSchema]] = None,
-  enumValues: Option[Chunk[Json]] = None,
+  `enum`: Option[Chunk[Json]] = None,
```
987Nabil:
```suggestion
  @fieldName("enum") enumValues: Option[Chunk[Json]] = None,
```

jdegoes (approved):

PR #2796, opened 2024-04-23
breaking: Make `Root` a `PathCodec`, remove `Empty`

erikvanoosten:
As suggested by @khajavi, we're making the `Root` constant an alias to `PathCodec.empty` so that it can be used in routes as follows:
```scala
Method.GET / Root -> handler(Response.text("Hello world!"))
```
For consistency the `Empty` constant is removed completely.

### Migration
⚠️ This is a breaking change.
All current usages of `Root` have to be replaced with `Path.root`,
all current usages of `Empty` have to be replaced with `Path.empty`.

Also:
- add examples on how to use root routes in the very first examples so that it is easy to find for new users,
- cleanup documentation start page.


khajavi:
Good catch! But I think we should change it to better DSL. The `GET / ""` is somewhat unattractive!
Maybe it is better to change the `Root` definition to an alias for `PatchCodec.empty` instead of `Path.root`.

jdegoes (approved):

erikvanoosten:
Alright @khajavi , I like that. I have changed the `Root` constant as you suggested.

@jdegoes this is a breaking change, so I have requested a new review.

erikvanoosten:
@khajavi, what is the purpose of `Empty`? (Currently defined as `val Empty: Path = Path.empty`.) Should that be a `PathCodec` as well?

erikvanoosten:
> @khajavi, what is the purpose of `Empty`? (Currently defined as `val Empty: Path = Path.empty`.) Should that be a `PathCodec` as well?

Since I had to replace all occurrences of `Root` with `Path.root`, I went ahead and replaced `Empty` with `Path.empty` as well. Because I think `Empty` is no longer needed I removed it.

erikvanoosten:
Since this is a breaking change, I understand if the others think this going to far. If you ask me, `Method.GET / "" -> handler(Response.text("Hello world!"))` is still a valid option.

docs/migration/RC6-to-xx.md
```diff
@@ -0,0 +1,3 @@
+**`Root` and `Empty`**
+- replace `Root` with `Path.root`
+- replace `Empty` with `Path.empty`
```
khajavi, line 3: :heart: for documenting the migration guide

jdegoes (approved):

jdegoes:
@erikvanoosten Just need a merge conflict fix. Thanks!

erikvanoosten:
> @erikvanoosten Just need a merge conflict fix. Thanks!

That will be the third...

erikvanoosten:
> @erikvanoosten Just need a merge conflict fix. Thanks!

@jdegoes Ready to merge!

jdegoes (approved):

PR #2799, opened 2024-04-24
Add `context` method to request context

987Nabil:
@jdegoes rfc


guersam:
It seems like a DX improvement with a function that can replace `ZIO.service[A]` with a shorthand, revealing the intention more clearly. Nice to see this addition!

My two cents: `context` tends to be used much more frequently in user land code than `handler`, so exposing `zio.http.context` via `import zio.http.*` might cause unwanted shadowing. I'd prefer `useContext` or `withContext`.

jdegoes:
@987Nabil Looks good! Maybe we can call the function `withContext` instead of `context`? Matches the name of the type class too.

987Nabil:
@jdegoes done

jdegoes (approved):

PR #2800, opened 2024-04-25
Avoid providing the ZEnvironment on very request

kyri-petrou:
Currently, we provide the `ZEnvironment` to every request, even though that in almost all cases it'll only change during app wiring. Since the ZEnvironment is part of `Runtime` used by `NettyRuntime`, we can avoid doing this on every request and save ourselves modifying `FiberRefs`

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
@@ -16,15 +16,17 @@
 
 package zio.http.netty
 
+import scala.annotation.unused
+
 import zio._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import io.netty.channel._
 import io.netty.util.concurrent.{Future, GenericFutureListener}
 
-private[zio] trait NettyRuntime { self =>
+private[zio] final class NettyRuntime(zRuntime: Runtime[Any]) {
 
-  def runtime(ctx: ChannelHandlerContext): Runtime[Any]
+  def runtime(@unused ctx: ChannelHandlerContext): Runtime[Any] = zRuntime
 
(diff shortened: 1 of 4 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
```diff
@@ -16,15 +16,17 @@
...
+import scala.annotation.unused
+
 import zio._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import io.netty.channel._
 import io.netty.util.concurrent.{Future, GenericFutureListener}
 
-private[zio] trait NettyRuntime { self =>
+private[zio] final class NettyRuntime(zRuntime: Runtime[Any]) {
```
kyri-petrou, line 27: I was struggling a bit to see how this could be beneficial as a trait, but I'm happy to revert this change if we think there's a reason for it

zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
```diff
@@ -16,15 +16,17 @@
...
 
-  def runtime(ctx: ChannelHandlerContext): Runtime[Any]
+  def runtime(@unused ctx: ChannelHandlerContext): Runtime[Any] = zRuntime
```
kyri-petrou, line 29: Similarly, I was tempted to remove the `ctx: ChannelHandlerContext` argument from here. Do we know if there was some use for it, like somehow incorporating it into the zio Runtime?

jdegoes: Remove 👍

jdegoes (approved):

PR #2807, opened 2024-04-26
Use current runtime when initializing AppRef

kyri-petrou:
I accidentally introduced a bug in #2800 where a custom bootstrapped runtime wasn't propagated to the request execution 🤦. We need to make sure that `AppRef` is initialized with the current Runtime, not the default one.

PS: I tried to add a unit test for this but I'm struggling a little bit with writing one. I'll give it another go with adding a test, but in the meantime opening this PR as it's a pretty severe bug


987Nabil (approved):

987Nabil:
Merged, but please add the test @kyri-petrou 🙂

PR #2808, opened 2024-04-26
Add middlewares to forward header and ensure the existence of a header

987Nabil:


987Nabil:
@jdegoes this PR has mainly microservice architecture in mind. But I am not sure about the header forwarding. This solution uses fiber refs, but it could be as well using `Env` and the context providing middleware. The trade off is, does one like the requirement of forwarded headers be explicit in the type or not? I am not sure what I prefer.

jdegoes:
Hiding it would be better. Let me know when build is passing.

jdegoes (approved):

Issue #1521, opened 2022-09-19
Implement a code generator that takes protobufs and generates zio.http.api.EndpointSpec values


jdegoes:
**Is your feature request related to a problem? Please describe.**

Some companies begin with protobufs for gRPC, and then proceed to implement a server that satisfies the protobuf specification; then they will modify the protobuf specification when a change is necessary, and then re-implement the server.

**Describe the solution you'd like**

In order to support this _spec-first_ driven approach to API development (rather than _code-first_, which is the paradigm already implemented within the `zio.http.api` package), we will need a code generator, which can read in protobuf specification, and generate the corresponding `zio.http.EndpointSpec ` values that declaratively describe the APIs.

Each gRPC service definition should translate into a single `EndpointSpec ` definition.

Then a user can implement handlers based on the code-generated `EndpointSpec ` values.

987Nabil:
@jdegoes @adamgfraser Is this not what zio-grpc is doing, just adjusted to the new API design?
Should this be added to zio-http? Or should this live in zio-grpc?
If it would be added to zio-http, would we use scalapb to parse the messages need for the service?
Or would we just create an interface/trait for the messages? I think messages to case classes makes more sense.

jdegoes:
I think we generate both the EndpointSpec and the case classes from the protobufs, and use ZIO Schema to do the encoding / decoding.

dejvid:
This would be great to add. Is anyone working in this?

987Nabil:
@dejvid I don't think so. I am focusing on OpenAPI currently and that is also planned for the 3.0 release afaik. I might work on protobuf afterwards, but maybe this would not be part of the next release. If you like to work on it, feel free ofc ;)

gstrainovic:
I will work on it.

digital-phoenix:
@jdegoes What's the current status of this issue?

adamgfraser:
@digital-phoenix Looks like someone was working on it but that was quite a while ago so if you are interested you should feel free to pick it up.

pablf:
/attempt #1521

<div id="algora" />

| [Algora profile](https://console.algora.io/@/pablf) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @pablf | 23 ZIO bounties | <div align="center">Scala, Rust</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/clhf0u3ly000al60f8dm7o5ln/cancel-attempt) |

PR #2814, opened 2024-04-30
Code generator for protobuf specification

pablf:
/claim #1521

Generates ZIO Http endpoints from services with the Protobuf codec from ZIO Schema. Messages can be generated with scalapb. Technically, it could be already published as a sbt plugin using the project `sbt-zio-http-grpc`, but I think that it could make more sense to have a common sbt plugin with the OpenAPI generator.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -12,3 +12,6 @@ addSbtPlugin("io.get-coursier"    % "sbt-shading"               % "2.1.4")
 addSbtPlugin("com.github.cb372"   % "sbt-explicit-dependencies" % "0.3.1")
 addSbtPlugin("org.scala-js"       % "sbt-scalajs"               % "1.16.0")
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject"  % "1.3.2")
+addSbtPlugin("com.thesamet"       % "sbt-protoc"                % "1.0.7")
+addSbtPlugin("com.thesamet"       % "sbt-protoc-gen-project"    % "0.1.8")
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.11.15"
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
@@ -238,6 +238,7 @@ final case class EndpointGen() {
               status = status,
               mediaType = Some("application/json"),
               doc = None,
+              false,
             )
         case (OpenAPI.StatusOrDefault.StatusValue(status), OpenAPI.ReferenceOr.Or(response: OpenAPI.Response))    =>
           val (imports, code) =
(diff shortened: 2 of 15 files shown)


zio-http-gen/src/main/scala/zio/http/gen/scala/Code.scala
```diff
@@ -133,23 +133,25 @@ object Code {
...
+  object HeadersCode     { val empty: HeadersCode = HeadersCode(Nil)                             }
   final case class HeaderCode(name: String)
   final case class InCode(
     inType: String,
     name: Option[String],
     doc: Option[String],
+    streaming: Boolean,
   )
-  object InCode          { def apply(inType: String): InCode = InCode(inType, None, None) }
+  object InCode          { def apply(inType: String): InCode = InCode(inType, None, None, false) }
```
987Nabil: Please name the boolean params

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -283,6 +284,7 @@ final case class EndpointGen() {
             status = status,
             mediaType = Some("application/json"),
             doc = None,
+            false,
```
987Nabil: Please name the boolean params

project/plugins.sbt
```diff
@@ -12,3 +12,6 @@ addSbtPlugin("io.get-coursier"    % "sbt-shading"               % "2.1.4")
 addSbtPlugin("com.github.cb372"   % "sbt-explicit-dependencies" % "0.3.1")
 addSbtPlugin("org.scala-js"       % "sbt-scalajs"               % "1.16.0")
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject"  % "1.3.2")
+addSbtPlugin("com.thesamet"       % "sbt-protoc"                % "1.0.7")
+addSbtPlugin("com.thesamet"       % "sbt-protoc-gen-project"    % "0.1.8")
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.11.15"
```
987Nabil, line 17: I am not sure about this. Is this dependency to a compiler plugin? Or just a dependency to the sources of it?

987Nabil (commented):
I am not sure, I like the dependencies. I think the format of the protobuf files is not so complicated and we could maybe use RichTextCodec to parse it ourselves? We have already means to generate files anyway.
@jdegoes what is your opinion?

project/plugins.sbt
pablf: It's a dependency for the build to be able to define the project `sbtZioHttpGrpcTests`.

pablf:
We could convert the protobuf descriptor class to a `zio-http`-made descriptor within `sbt-zio-http-grpc` and move the protobuf dependencies to that project so they are not in `zio-http-gen`. Then, the final plugin could either parse itself the protobuf files or be published using scalapb.

jdegoes (approved):

Issue #2812, opened 2024-04-30
Codecov: Rate Limit Reached and CI Is Failing


khajavi:
The CI is failing because the Codecov has [reached](https://github.com/zio/zio-http/actions/runs/8873905605/job/24410498522?pr=2810#step:6:55) its rate limit:

```js
{
    'detail': ErrorDetail(
        string='Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected available in 1631 seconds.', 
        code='throttled'
    )
}
```

PR #2816, opened 2024-05-01
Remove Code Coverage Workflow

khajavi:
As per issue #2812, let's remove this workflow to make the ci green unless we need to maintain the project with such reports.


khajavi:
I am closing this, as it appears that the rate limit has been reset.

PR #2825, opened 2024-05-06
Documentation of Key Features of ZIO HTTP

khajavi:


987Nabil (approved):

nafg:
I assume the rendered HTML is supposed to have line breaks? You need to double-space paragraphs in markdown. https://zio.dev/zio-http/ is a bit overwhelming now.

Also, shouldn't it be a bulleted list?

PR #2826, opened 2024-05-07
Diagrams for ProtocolStack, HandlerAspect and Middleware

khajavi:


987Nabil:
I wonder if this is not too much? I mean all images basically look the same. Is there any significant difference?

khajavi:
> I wonder if this is not too much? I mean all images basically look the same. Is there any significant difference?

The overall structure is the same, but each data type has its unique features e.g. the `ProtocolStack` is not fixed by `Request` and `Response`, the `HandlerAspect` is fixed but also has the `OutCtx` and finally the `Middleware` which has fixed its inputs and outputs, but not have any context and can be applied to route/routes. Looking at each data type diagram, the reader can quickly grasp the concept or may want to compare them with others.

987Nabil (approved):

PR #2828, opened 2024-05-07
Simplify running ZIO effects

kyri-petrou:
Heya folks, I've noticed that there are a few places where ZIO effects are run "synchronously" either unnecessarily or not composed together, so I thought to make a PR for it.

Also, I removed the `NettyRuntime#runtime` method as discussed here: https://github.com/zio/zio-http/pull/2800#discussion_r1579175167. This also allows us to do an ultra-micro-optimization by not having to create the `Runtime#UnsafeAPI` on each invocation of `run`


kyri-petrou:
Flaky test seems to strike again. Probably just needs a retry but I don't have the right permissions to do it

987Nabil:
@kyri-petrou I can't remember I saw that test failing before. I merged main. Let's see what's the outcome

jdegoes (approved):

PR #2829, opened 2024-05-07
Add routes#serve method

GrigoriiBerezin:


zio-http/shared/src/main/scala/zio/http/Routes.scala
```diff
@@ -282,6 +282,10 @@ object Routes extends RoutesCompanionVersionSpecific {
   def singleton[Env, Err](h: Handler[Env, Err, (Path, Request), Response])(implicit trace: Trace): Routes[Env, Err] =
     Routes(Route.route(RoutePattern.any)(h))
 
+  implicit class RouteOps[-Env, +Err <: Response](val routes: Routes[Env, Err]) extends AnyVal {
+    def serve: URIO[Env with Server, Int] = Server.install(routes)
```
kyri-petrou: Since the focus here is on usability, I think this should be calling `Server.serve` instead, so users don't have to call `ZIO.never` afterwards. I believe this was what the issue was aiming for as well

GrigoriiBerezin: Oh, I get it, my bad, thx for notice

GrigoriiBerezin: Fixed

zio-http/shared/src/main/scala/zio/http/Routes.scala
```diff
@@ -282,6 +282,10 @@ object Routes extends RoutesCompanionVersionSpecific {
   def singleton[Env, Err](h: Handler[Env, Err, (Path, Request), Response])(implicit trace: Trace): Routes[Env, Err] =
     Routes(Route.route(RoutePattern.any)(h))
 
+  implicit class RouteOps[-Env, +Err <: Response](val routes: Routes[Env, Err]) extends AnyVal {
```
kyri-petrou: By the way sorry if this a stupid question, but do we need the implicit class? Isn't possible to have this as a method directly implemented on Routes instead or am I missing something?

GrigoriiBerezin: Just wanted to have implementations of server far from route implementation, mb I'm wrong

kyri-petrou: This is a valid point. @khajavi @987Nabil any thoughts on this?

khajavi: This seems to be valid, but I am uncertain if it provides any significant advantages.

GrigoriiBerezin: Should I move it into implementation?

987Nabil: No implicit. Just a method on routes

GrigoriiBerezin: I've done it, but now it's failing with handshake test for some reason

jdegoes (commented):
Good to merge when CI is passing!

jdegoes (commented):
Good to merge when CI passing!

GrigoriiBerezin:
Finally have fixed all issues

987Nabil (approved):

PR #2845, opened 2024-05-14
Attempt fix to `StreamingForm` blocking due to `unsafe.run`

kyri-petrou:
The current `StreamingForm` implementation uses `unsafe.run` in order to avoid wrapping each emitted byte into a ZIO. This means that the underlying implementation is blocking, which is causing CI to hang now that auto-blocking is disabled.

This PR fixes it (as a temporary patch) by running the stream in the blocking threadpool. We need to come up with a better design for it, but I think this should be adequate for RC7

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/ClientHttpsSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ClientHttpsSpec.scala
@@ -73,7 +73,7 @@ object ClientHttpsSpec extends ZIOHttpSpec {
           ),
         ),
       )
-    } @@ nonFlaky(20),
+    } @@ nonFlaky(20) @@ ignore,
   )
     .provideSomeLayer[Client](Scope.default)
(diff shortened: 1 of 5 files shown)


zio-http/jvm/src/test/scala/zio/http/ClientHttpsSpec.scala
```diff
@@ -73,7 +73,7 @@ object ClientHttpsSpec extends ZIOHttpSpec {
           ),
         ),
       )
-    } @@ nonFlaky(20),
+    } @@ nonFlaky(20) @@ ignore,
```
kyri-petrou, line 76: The server the test uses seems to be not working at the moment which is causing CI to fail so I disabled the test. We should open a separate ticket to use a local server instance (perhaps even zio-http now that it supports SSL) instead of relying on a remote server

987Nabil (approved):

PR #2850, opened 2024-05-17
[gen] fix for codegen of sumtypes with reusable fields

### LINKED ISSUE #2847: [gen] faulty code generation for sum types with reusable fields
**Describe the bug**
When defining in open API a component that only reference "subtypes", which all include a shared other component,
the generated code will:
1. not compile 
2. has duplicate classes (subtype are both nested in sealed trait companion, and have standalone redundant copy)
3. shared fields is not referenced by any endpoint, but has its own redundant class generated

**To Reproduce**
generate from the following example:
```yaml
info:
  title: Animals Service
  version: 0.0.1
servers:
  - url: http://127.0.0.1:5000/
tags:
  - name: Animals_API
paths:
  /api/v1/zoo/{animal}:
    get:
      operationId: get_animal
      parameters:
        - in: path
          name: animal
          schema:
            type: string
          required: true
      tags:
        - Animals_API
      description: Get animals by species name
      responses:
        "200":
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Animal'
          description: OK
        "500":
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/HttpError'
          description: Internal Server Error
openapi: 3.0.3
components:
  schemas:
    Animal:
      oneOf:
        - $ref: '#/components/schemas/Alligator'
        - $ref: '#/components/schemas/Zebra'
    AnimalSharedFields:
      type: object
      required:
        - age
        - weight
      properties:
        age:
          type: integer
          format: int32
          minimum: 0
        weight:
          type: number
          format: float
          minimum: 0
    Alligator:
      allOf:
        - $ref: '#/components/schemas/AnimalSharedFields'
        - type: object
          required:
            - num_teeth
          properties:
            num_teeth:
              type: integer
              format: int32
              minimum: 0
    Zebra:
      allOf:
        - $ref: '#/components/schemas/AnimalSharedFields'
        - type: object
          required:
            - num_stripes
          properties:
            num_stripes:
              type: integer
              format: int32
              minimum: 0
    HttpError:
      type: object
      properties:
        messages:
          type: string
```
and examine the generated classes:
```sh
tree target/scala-3.3.3/src_managed
target/scala-3.3.3/src_managed
└── zoo
    └── openapi
        ├── api
        │   └── v1
        │       └── zoo
        │           └── Animal.scala
        └── component
            ├── Alligator.scala
            ├── Animal.scala
            ├── AnimalSharedFields.scala
            ├── HttpError.scala
            └── Zebra.scala

7 directories, 6 files
```

**Expected behaviour**
1. Code needs to compile
2. No duplicate classes
3. No generation of classes not referenced from an endpoint (e.g. shared fields for reuse in other classes)
4. (*) bonus: if all subtypes of a sealed trait include the same shared fields component, add the fields as members of the trait, e.g. in this case:
```scala
@noDiscriminator
sealed trait Animal {
  def age: Int
  def weight: Float
}
```

**Screenshots**
compile_error
all_generated_classes

hochgi:
fixes #2847 

I added a test, initially failing (for the payload in the original issue):
sum-type_codegen_test_failure
Then fixed it, and even extended it to generate abstract fields from a reusable component, e.g:
https://github.com/zio/zio-http/blob/fc51bd75d5bd9870f69c992e30ec60f90219aae5/zio-http-gen/src/test/resources/ComponentAnimal.scala#L7-L34

This PR also avoids generating duplicate copies of the sealed trait subtypes in separate files.


zio-http-gen/src/test/scala/zio/http/gen/scala/CodeGenSpec.scala
```diff
@@ -262,5 +325,5 @@ object CodeGenSpec extends ZIOSpecDefault {
           "/GeneratedUserNameArray.scala",
         )
       },
-    ) @@ java11OrNewer @@ flaky @@ blocking // Downloading scalafmt on CI is flaky
+    ) @@ java11OrNewer /*@@ flaky*/ @@ blocking // Downloading scalafmt on CI is flaky
```
hochgi: Note to self: revert this - flaky was removed just for local runs to not be executed 100 times on the failing test I added…

987Nabil:
@hochgi Thanks for your contribution. And this contains 100% a crucial bugfix.
But, I am not so eager with the change of the shared fields. Why? In OpenAPI you can mark a optional shared field as required for the subtype. This would be a conflict, that is not resolvable on the code generation side. I would like to keep here the more generic solution and keep the super type clean.
I think it is okay for devs to add extension methods if needed or to extend the code manually.
Extension methods will work, even if the code will be generated on each compile/build.

I also remember, that I had a reason for generating the types twice. Once in the sealed trait hierarchy and once separate. But I forgot the reason. I'll think about it and if I can' remember, we might try it with your way.

I like the idea of generating only what is actually used. Even though I think having unused types defined is not good, you can't force anyone (especially not public apis)

hochgi:
> @hochgi Thanks for your contribution. And this contains 100% a crucial bugfix. But, I am not so eager with the change of the shared fields. Why? In OpenAPI you can mark a optional shared field as required for the subtype...

@987Nabil  can you show me how we can define in OpenAPI this example zoo endpoint such that `age` & `weight` are enforced in the `Animal` interface?

I looked at how to do that, and the reusable component was the best I could find.

Of course I'll remove the trait body generation ad requested.
But since I need some mechanism to do this for a real project at work, I'd appreciate a pointer.

If not too hard, perhaps I can chanhe this PR to do it the right way :)

987Nabil:
@hochgi The best way would probably be a construct like having an animal type, that has three fields: `age`, `weight` and `animalType`. Animal type is then `Zebra` and `Alligator` that each have the fields that are unique to them. That should also work with how code is generated and is the kinda FP approach

hochgi:
@987Nabil I agree, and if I were to design the API I'm dealing with, that's how I'd do it.
What I'm attempting is a pragmatic solution on how to deal with existing APIs, were all fields are flattened, and I can't nest like with `AnimalType` containg extra (nested) fields...

EDIT:

Note that in the official docs, a similar approach is described:
https://spec.openapis.org/oas/latest.html#models-with-polymorphism-support
The difference is that in the official docs, `Pet` is an open super type (analogous to a regular `trait` or `abstract class`), whereas here, I'm more intrested with a closed super type (analogous to a `sealed trait`), which is why I made the super type a separate oneOf component.

Perhaps we can:
1. Validate no subtype requires an optional field from any super type (and if so, omit that trait from required mixins)
2. Make (in my example) `AnimalSharedFields` an open trait. (And optionally, if used also as top level class, add a case class with same fields in the trait's companion)
3. Make any type that include such shared fields object extend the open trait
4. Make the closed trait to enforce mixins of all (valid) open traits used by all subtypes.

In this zoo toy example, it could look like:
`AnimalSharedFields.scala`:
```scala
trait AnimalSharedFields {
  def age: Int
  def weight: Float
}
object AnimalSharedFields {
  // "Bare" or any other arbitrary name
  case class Bare (age: Int, weight: Float) extends AnimalSharedFields
}
```
And `Animal.scala`:
```scala
sealed trait Animal { this: AnimalSharedFields =>
}
object Animal {
  case class Alligator (age: Int, weight: Float, num_teeth: Int) extends Animal with AnimalSharedFields
  case class Zebra (age: Int, weight: Float, num_stripes: Int) extends Animal with AnimalSharedFields
}
```

987Nabil:
@hochgi I feel this suggestion is too complex and idk if this will be useful for most users of the code generator.
As discussed in discord, I think a config with a switch would be nice.

I would say the default behaviour is
- no duplicate types are generated
- no common fields on the super type

And opt-in to
- common fields on super type

So basically, your original proposal, but the accessors on the sealed trait are opt in.

I am also not sure, what you are doing actually. Because it seems on the one hand like you can define things in the open api spec, but you said you have to impl. against a given endpoint.

So just some thoughts I had
- does it make more sense to write endpoints and generate open API specs?
- if generating from open API, does it need to be perfect? MIght it be enough to have a starting point and adjust the generated code manually?
- maybe having imperfect data structures generated, even if it is generating repeatedly on build, is good enough and further validation can happen later. For example like `generatedEndpoint.copy(input = generatedEndpoint.input.transformOrFail(/* transform to domain or fail -> results in bad request */))` (we can add `Endpoint#transformInOrFail` if needed)

hochgi:
@987Nabil yes, in hindsight it is a bit complex.
Sure, lets go for the configurable opt in as you suggested.

Might be worth just adding validations such as to not require optional fields (if config flag enabled), or if same field is declared with a different type (perhaps in a separate reusable object), etc'...

But for this PR we can start with just the config :)

My motivation:
An existing service (non scala - python in this particular case), which should be used by several other services.
The API & data types exposed look more like the Pet example from the docs.
The openAPI spec is an afterthought, and goal is to serve this in a way that will: (a) conform existing API, and (b) preserve semantic as best is possible.
I will use it also to generate a (cross versioned) scala "SDK" for the service, but also want to enable other teams to reuse it in their non-scala services, using e.g. https://github.com/OpenAPITools/openapi-generator

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
```diff
@@ -0,0 +1,18 @@
...
+  abstractMembersFromReusableComponents: Boolean,
+)
+object Config {
+
+  val default: Config = Config(
+    abstractMembersFromReusableComponents = false,
+  )
+
+  lazy val config: zio.Config[Config] =
+    zc.boolean("abstract-members-from-reusable-components")
```
hochgi: nit: `"common-fields-on-super-type"` is perhaps better phrasing?
Something else maybe?

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -28,33 +27,164 @@ object EndpointGen {
...
+    def unapply(schema: JsonSchema.AllOfSchema): Some[List[String]] = Some(
+      schema.allOf.foldRight(List.empty[String]) {
+        case (JsonSchema.RefSchema(SchemaRef(simpleName)), simpleNames) =>
+          simpleName :: simpleNames
+        case (_, simpleNames)                                           => simpleNames
+      },
+    )
+  }
+
+  private def extractUsedComponents(openAPI: OpenAPI): List[Code.File] = {
```
hochgi: nit: rename (it's no longer just "used" components. Code still emits separate case class for the reusable component. This will not change in this PR, if at all)

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -527,7 +671,8 @@ final case class EndpointGen() {
               throw new Exception(s"Unexpected subtype $other for oneOf schema $schema")
           }
           .toList
-        val noDiscriminator                      = caseNames.isEmpty
+        val noDiscriminator                         = caseNames.isEmpty
+        val abstractMembersFields: List[Code.Field] = abstractMembers.flatMap(fieldsOfObject(openAPI, annotations))
```
hochgi: validate same field does not appear multiple times (with different type; if same type, we can distinct - also, if same type but optional - i.e. one required, and another optional - we can omit optional and stay with required only)

Same treatment should be done for case classes as well

zio-http-gen/src/test/scala/zio/http/gen/scala/CodeGenSpec.scala
hochgi, line 1: add tests
- fail when union of fields has unresolvable contradictions
- mitigate multiple fields with same type to not fail (both exact type & required-optional for same type)

open issue:
Currently codegen does not support OpenAPI validations (minimum/maximum/pattern/…), which should probably be done in a smart constructor (NOT IN THIS PR!) - so when we "unify" same fields with same (up to required) type, we should also consider how to "unify constraints" in the future…

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
```diff
@@ -0,0 +1,16 @@
...
+
+final case class Config(commonFieldsOnSuperType: Boolean)
+object Config {
+
+  val default: Config = Config(
+    commonFieldsOnSuperType = false,
+  )
+
+  lazy val config: zio.Config[Config] =
+    zc.boolean("common-fields-on-super-type")
```
987Nabil: Please no cryptic renaming. Better use the full path `zio.Config.boolean`

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
```diff
@@ -0,0 +1,16 @@
+package zio.http.gen.openapi
+
+import zio.{Config => zc}
```
987Nabil:
```suggestion
```

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
987Nabil:
```suggestion
    zio.Config.boolean("common-fields-on-super-type")
```

987Nabil (approved):

PR #2852, opened 2024-05-19
Avoid unsafely running effects when handling WS requests

kyri-petrou:
When handling websocket connections, we currently impurely run the ZIO effect within a region that is already running in ZIO's executor. This can be quite bad as it can lead to a deadlock if the effect contains an async computation and there are no `ZScheduler` workers available to execute the effect. ~~This has already introduced some flakiness to Caliban's test suite (see [here](https://app.circleci.com/pipelines/github/ghostdogpr/caliban/6471/workflows/40d2077a-eca2-4d54-8bdf-934c527c0d79/jobs/44829))~~

This PR fixes it by changing the `upgradeToWebSocket` method to return an effect instead, which is then composed with the main effect that is running the request.

EDIT: After some further investigation, it appears that the flakiness issue experienced in Caliban's CI was not related to running the effect unsafely (although I kept that change because I think it's better to do it this way). The issue appears that it was related to the `dispatch` method in `WebSocketAppHandler`. Since `NettyRuntime#runUninterruptible` forks the effect, in some rare cases messages would be offered to the queue in the wrong order.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
@@ -68,9 +68,8 @@ private[zio] final class WebSocketAppHandler(
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
     dispatch(ctx, ChannelEvent.exceptionCaught(cause))
     onComplete match {
-      case Some(promise) =>
-        promise.fail(cause)
-      case None          =>
+      case Some(promise) => promise.unsafe.done(Exit.fail(cause))
+      case None          => ()
     }
   }
 
--- a/zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
@@ -47,13 +47,11 @@ private[zio] final class WebSocketAppHandler(
     event: ChannelEvent[JWebSocketFrame],
     close: Boolean = false,
   ): Unit = {
-    zExec.runUninterruptible(ctx, NettyRuntime.noopEnsuring)(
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
```diff
@@ -68,9 +68,8 @@ private[zio] final class WebSocketAppHandler(
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
     dispatch(ctx, ChannelEvent.exceptionCaught(cause))
     onComplete match {
-      case Some(promise) =>
-        promise.fail(cause)
-      case None          =>
+      case Some(promise) => promise.unsafe.done(Exit.fail(cause))
+      case None          => ()
```
kyri-petrou, line 73: I also noticed this bug where the promise wasn't being fulfilled with the exception as it returned an effect which wasn't then being run

zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
```diff
@@ -47,13 +47,14 @@ private[zio] final class WebSocketAppHandler(
...
-        }),
-    )
+    // IMPORTANT: Offering to the queue must be run synchronously to avoid messages being added in the wrong order
+    // Since the queue is unbounded, this will not block the event loop
+    // TODO: We need to come up with a design that doesn't involve running an effect to offer to the queue
+    zExec.unsafeRunSync(queue.offer(event.map(frameFromNetty)))
+    onComplete match {
+      case Some(promise) if close => promise.unsafe.done(Exit.succeed(ChannelState.Invalid))
+      case _                      => ()
+    }
```
kyri-petrou, line 57: This is the most important change, we need to ensure that messages are offered to the queue synchronously to avoid them being added in the wrong order

987Nabil (approved):

PR #2857, opened 2024-05-20
Add brotli compression (#2646)

### LINKED ISSUE #2646: Adding Brotli as a compression option

987Nabil:
fixes #2646
/claim #2646

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -27,6 +27,7 @@ object Dependencies {
       "io.netty" % "netty-transport-native-kqueue" % NettyVersion,
       "io.netty" % "netty-transport-native-kqueue" % NettyVersion % Runtime classifier "osx-x86_64",
       "io.netty" % "netty-transport-native-kqueue" % NettyVersion % Runtime classifier "osx-aarch_64",
+      "com.aayushatharva.brotli4j" % "brotli4j" % "1.16.0" % "provided",
     )
 
   val `netty-incubator` =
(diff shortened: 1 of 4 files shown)


kyri-petrou (approved):
LGTM 👍

project/Dependencies.scala
```diff
@@ -27,6 +27,7 @@ object Dependencies {
       "io.netty" % "netty-transport-native-kqueue" % NettyVersion,
       "io.netty" % "netty-transport-native-kqueue" % NettyVersion % Runtime classifier "osx-x86_64",
       "io.netty" % "netty-transport-native-kqueue" % NettyVersion % Runtime classifier "osx-aarch_64",
+      "com.aayushatharva.brotli4j" % "brotli4j" % "1.16.0" % "provided",
```
kyri-petrou, line 30: Very small nit: I believe we should be using `optional` here instead of `provided`. A good summary on the semantics from a comment [here](https://stackoverflow.com/questions/40393098/when-to-use-optional-dependencies-and-when-to-use-provided-scope):

> But, as I tried to say, the difference is mainly semantic, i.e. you signal the user that this dependency should come from the container, while for optional dependencies, you signal that you need to explicitly add the necessary dependencies

Unless of course there was some reason for using `provided` in the first place then please ignore me! 😄

987Nabil: Thanks for the hint. I did not know optional

jdegoes (approved):

PR #2862, opened 2024-05-23
fix: Ignore everything after closing boundary when parsing multi form data

### LINKED ISSUE #2411: Parsing multi form data requires CRLF termination
**Describe the bug**
When a client sends a multi form request whose body is terminated by CR instead of CRLF (after the closing boundary), the parsing logic may not parse the body completely. I.e. some form fields are missing. 

**To Reproduce**
Execute the following test suite with the latest ZIO HTTP version (3.0.0-RC2):
```scala
object DemoSuite extends ZIOSpecDefault {
  val CR = '\r'
  val CRLF = "\r\n"
  override def spec: Spec[TestEnvironment with Scope, Any] = suite("Demo")(
    test("encoding") {
      val body = Chunk.fromArray(
        s"""|--(((AaB03x)))${CR}
            |Content-Disposition: form-data; name="hocon-data"${CR}
            |Content-Type: text/plain${CR}
            |${CR}
            |foos: []${CR}
            |--(((AaB03x)))${CR}
            |Content-Disposition: form-data; name="json-data"${CR}
            |Content-Type: text/plain${CR}
            |${CR}
            |{ "bars": [] }${CR}
            |--(((AaB03x)))--${CR}""".stripMargin.getBytes(),
      )

      val form = Form(
        FormField.textField("hocon-data", "foos: []", MediaType.text.`plain`),
        FormField.textField("json-data", """{ "bars": [] }""", MediaType.text.`plain`),
      )

      val boundary = Boundary("(((AaB03x)))")

      val actualByteStream = form.multipartBytes(boundary)

      for {
        form2       <- Form.fromMultipartBytes(body)
        actualBytes <- actualByteStream.runCollect
      } yield assertTrue(
        actualBytes == body,
        form2 == form,
      )
    }
  )
}
```
It will only yield the first entry `hocon-data`.

**Expected behaviour**
It yields both form data entries `hocon-data` as well as `json-data`.

**Additional context**

A client library that will send their multi form data bodies this way is STTP. Example:

```scala
import sttp.client3.httpclient.zio.{HttpClientZioBackend, send}
import sttp.client3.{Identity, RequestT, UriContext, basicRequest, multipart}
import zio.Scope
import zio.test.{Spec, TestEnvironment, ZIOSpecDefault, assertTrue}

object SttpDemo extends ZIOSpecDefault {
  override def spec: Spec[TestEnvironment with Scope, Any] = test("STTP Demo") {
    val request: RequestT[Identity, Either[String, String], Any] = basicRequest
      .post(uri"http://www.example.com")
      .multipartBody(
        multipart("hocon-data", """foos: []"""),
        multipart("json-data", """{ "bars": [] }""")
      )
    send(request).map(response => assertTrue(response.is200))
  }.provide(HttpClientZioBackend.layer())
}
````
Unfortunately, I have not been able to find precise specification on the matter.

### PR DESCRIPTION
**Pull request by seakayone**
Resolves: https://github.com/zio/zio-http/issues/2411

Enables: https://github.com/softwaremill/tapir/pull/3690

Based on an idea by @fliiiix. Co-authored with @fliiiix and @grigala during [a Hackergarten session](https://www.hackergarten.net/).

/claim #2411
/split @fliiiix
/split @grigala

### CODE REVIEW

**Comment by ipetkovic** [comment]
thx @seakayone for pushing this one.
I was just about to make PR myself and then saw yours.

**Comment by seakayone** [comment]
> thx @seakayone for pushing this one. I was just about to make PR myself and then saw yours.

This fix is not yet complete and more of a rough first shot. It currently has a couple of "problems":
- For one we had to replace the `ChunkBuilder` with a `Chunk`. I plan to revert that change. Even though zio's `ChunkBuilder` now [has a `knownSize` method](https://github.com/zio/zio/pull/8892) it would not work on 2.12.
- Although the actual issue is fixed other tests are broken. 

I will not be able to do much the next couple of days and will progress on this in June hopefully.

**Comment by ipetkovic** [comment]
> > thx @seakayone for pushing this one. I was just about to make PR myself and then saw yours.
> 
> This fix is not yet complete and more of a rough first shot. It currently has a couple of "problems":
> 
> * For one we had to replace the `ChunkBuilder` with a `Chunk`. I plan to revert that change. Even though zio's `ChunkBuilder` now [has a `knownSize` method](https://github.com/zio/zio/pull/8892) it would not work on 2.12.
> * Although the actual issue is fixed other tests are broken.
> 
> I will not be able to do much the next couple of days and will progress on this in June hopefully.

can we overcome first issue by having size counter in FormStateBuffer instead?

**Comment by 987Nabil** [comment]
@seakayone if you want to have the bounty, you have to add `/claim` with the issue number to the PR description. You can also use `/split` if you want to split with the co-authors.

File: zio-http/shared/src/main/scala/zio/http/internal/FormState.scala [hunk 1]
```diff
@@ -72,7 +86,12 @@ private[http] object FormState {
           case ClosingBoundary(_)       => BoundaryClosed(tree)
         }
 
-      } else if (crlf && (phase eq Phase.Part2)) {
+      } else if ((crlf || boundaryDetected) && (phase eq Phase.Part2)) {
+        if (boundaryDetected) {
+          val hyphen = "-".getBytes().head
+          buffer += hyphen
+          buffer += hyphen
```
kyri-petrou: Much more performant this way:

```suggestion
          buffer += '-'
          buffer += '-'
```

In fact, I would use `addOne` instead, since `+=` is generic and will cause boxing. The only issue is that I don't know if it cross-compiles to Scala 2.12 (I think we have that method in `ChunkBuilder.Byte` but haven't checked)

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -102,6 +122,8 @@ private[http] object FormState {
     def reset(): Unit = {
       tree0.clear()
       buffer.clear()
+      bufferSize = 0
+      boundaryMatches.map(_ => false)
```
kyri-petrou: What's usually the size of the array? Might be more performant to just make it a `var` and replace the whole thing?

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -29,8 +29,10 @@ private[http] object FormState {
 
   final class FormStateBuffer(boundary: Boundary) extends FormState { self =>
 
-    private val tree0: ChunkBuilder[FormAST] = ChunkBuilder.make[FormAST]()
-    private val buffer: ChunkBuilder.Byte    = new ChunkBuilder.Byte
+    private val tree0: ChunkBuilder[FormAST]    = ChunkBuilder.make[FormAST]()
+    private val buffer: ChunkBuilder[Byte]      = new ChunkBuilder.Byte
+    private var bufferSize: Int                 = 0
```
kyri-petrou: I was going to link [this PR](https://github.com/zio/zio/pull/8892) but just realised that it's only for Scala 2.13+.. And also that it was you that opened it in the first place 😅

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -54,8 +56,20 @@ private[http] object FormState {
 
       val crlf = byte == '\n' && !lastByte.isEmpty && lastByte.get == '\r'
 
+      var boundaryDetected = false;
+      val posInLine        = bufferSize + (if (this.lastByte.isEmpty) 0 else 1)
+      if (posInLine < boundary.closingBoundaryBytes.size) {
```
kyri-petrou: Maybe save `boundary.closingBoundaryBytes.size` in a val during the creation of the buffer since it's used in multiple places

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -54,8 +56,20 @@ private[http] object FormState {
...
+      }
+
+      if (posInLine == boundary.closingBoundaryBytes.size - 1) {
+        boundaryDetected = boundaryMatches.forall(_ == true)
+      }
+
       def flush(ast: FormAST): Unit = {
         buffer.clear()
+        bufferSize = 0
+        boundaryMatches.map(_ => false)
```
kyri-petrou: Same as above

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -87,6 +106,7 @@ private[http] object FormState {
         if (!lastByte.isEmpty) {
           if (isBufferEmpty) isBufferEmpty = false
           buffer += lastByte.get
+          bufferSize = bufferSize + 1
```
kyri-petrou: Sorry for the nitpick:

```suggestion
          bufferSize += 1
```

kyri-petrou (commented):
Just some minor comments from my side, otherwise looks good to me :)

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: see b338c6e347ace6ea3cb0bd5d918e3ce4dcdfa16b

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: see b338c6e347ace6ea3cb0bd5d918e3ce4dcdfa16b

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: Yes, unfortunately I was not able to add Scala 2.12 support for the knownSize of a ChunkBuilder.

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: see b338c6e347ace6ea3cb0bd5d918e3ce4dcdfa16b

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: Thanks, I would like to keep `+=` because in all other places in this file it is used to append to a `ChunkBuilder`.

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: Thanks.

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: Also, I think boxing is not a big concern here as this is only done once per request.

seakayone:
Thank you @kyri-petrou I have addressed all you remarks, tests are passing. Please have another look.

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
```diff
@@ -29,8 +29,11 @@ private[http] object FormState {
 
   final class FormStateBuffer(boundary: Boundary) extends FormState { self =>
 
-    private val tree0: ChunkBuilder[FormAST] = ChunkBuilder.make[FormAST]()
-    private val buffer: ChunkBuilder.Byte    = new ChunkBuilder.Byte
+    private val tree0: ChunkBuilder[FormAST]    = ChunkBuilder.make[FormAST]()
+    private val buffer: ChunkBuilder[Byte]      = new ChunkBuilder.Byte
+    private var bufferSize: Int                 = 0
+    private val closingBoundaryBytesSize        = boundary.closingBoundaryBytes.size
+    private var boundaryMatches: Array[Boolean] = new Array[Boolean](closingBoundaryBytesSize)
```
kyri-petrou: There is something here that I don't understand / failing to see. What is the reason for using an array? Can't this be just a boolean and then L62-68 become this?

```scala
boundaryMatches &&= posInLine < closingBoundaryBytesSize && boundary.closingBoundaryBytes(posInLine) == byte

val boundaryDetected = boundaryMatches && posInLine == closingBoundaryBytesSize - 1
```

kyri-petrou (commented):
Sorry I didn't pick this up in the last review round. I might be missing something but I think we can simplify the logic massively

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
seakayone: Wonderful, you are right we do not need to remember all checked bytes this way.

That works => https://github.com/zio/zio-http/pull/2862/commits/f1f33ba98abc12714d706d0f3a89496c62b082c1

Thanks.

kyri-petrou (approved):
Happy to approve this PR! However, I think you still need an approval from a code owner to get it merged

zio-http/shared/src/main/scala/zio/http/internal/FormState.scala
fliiiix: tbh no idea why i thought we need an array :see_no_evil: so much better without - nice find @kyri-petrou :tada:

jdegoes (approved):

PR #2864, opened 2024-05-25
Update zio schema version and add CurrencyType handling.

### LINKED ISSUE #2766: Endpoints from cases classes where the last field is an Option[Int] or Option[Long] result in BadRequest
**Describe the bug**
Endpoints from cases classes where the last field is an Option[Int] or Option[Long] result in BadRequest when the request body is pretty printed json.

**To Reproduce**
See [this issue and PR](https://github.com/zio/zio-schema/issues/675)

That PR should be acted on and a new version of zio-schema should be released this project should update its dependency to reflect the new version.

**Expected behaviour**
Valid  and correctly encoded JSON should not result in BadRequest.

**Screenshots**

russwyte:
Bumps zio-schema version to latest. 
Fixes #2766 

Note - I needed to handle new `StandardType.CurrencyType` - please check that I did it correctly. 😅


russwyte:
Fixed in another PR

PR #2869, opened 2024-05-27
[gen] fix codegen to reference subtypes nested with their encapsulating object when referenced directly

### LINKED ISSUE #2868: [gen] Concrete subtypes of a sum type referenced directly without nesting or proper import
**Describe the bug**
When we have a sealed trait and subtypes reside inside trait's companion,
we also need to make sure that any other object referencing concrete subtype directly, will nest under companion.

**To Reproduce**
Try to generate from the following yaml:
```yaml
info:
  title: Animals Service
  version: 0.0.1
servers:
  - url: http://127.0.0.1:5000/
tags:
  - name: Animals_API
paths:
  /api/v1/zoo/{animal}:
    get:
      operationId: get_animal
      parameters:
        - in: path
          name: animal
          schema:
            type: string
          required: true
      tags:
        - Animals_API
      description: Get animals by species name
      responses:
        "200":
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Animal'
          description: OK
openapi: 3.0.3
components:
  schemas:
    Animal:
      oneOf:
        - $ref: '#/components/schemas/Alligator'
        - $ref: '#/components/schemas/Zebra'
    AnimalSharedFields:
      type: object
      required:
        - age
        - weight
      properties:
        age:
          type: integer
          format: int32
          minimum: 0
        weight:
          type: number
          format: float
          minimum: 0
    Alligator:
      allOf:
        - $ref: '#/components/schemas/AnimalSharedFields'
        - type: object
          required:
            - num_teeth
          properties:
            num_teeth:
              type: integer
              format: int32
              minimum: 0
    Zebra:
      allOf:
        - $ref: '#/components/schemas/AnimalSharedFields'
        - type: object
          required:
            - num_stripes
          properties:
            num_stripes:
              type: integer
              format: int32
              minimum: 0
    Lion:
      type: object
      required:
        - eats
      properties:
        eats:
          $ref: '#/components/schemas/Zebra'
```

And you'll see the code does not compile:
Screenshot 2024-05-26 at 23 41 11

this is the generated code for `Lion`:
```scala
package test.component

import zio.schema._

case class Lion(
  eats: Zebra
)
object Lion {

  implicit val codec: Schema[Lion] = DeriveSchema.gen[Lion]

}
```
but it should have been:
```scala
package test.component

import zio.schema._

case class Lion(
  eats: Animal.Zebra
)
object Lion {

  implicit val codec: Schema[Lion] = DeriveSchema.gen[Lion]

}
```
Since `Zebra` is nested under `Animal`:
```scala
package test.component

import zio.schema._
import zio.schema.annotation._

@noDiscriminator
sealed trait Animal {
  def age: Int
  def weight: Float
}
object Animal {

  implicit val codec: Schema[Animal] = DeriveSchema.gen[Animal]
  case class Alligator(
    age: Int,
    weight: Float,
    num_teeth: Int
  ) extends Animal
  object Alligator {

    implicit val codec: Schema[Alligator] = DeriveSchema.gen[Alligator]

  }
  case class Zebra(
    age: Int,
    weight: Float,
    num_stripes: Int
  ) extends Animal
  object Zebra {

    implicit val codec: Schema[Zebra] = DeriveSchema.gen[Zebra]

  }
}
```
**Expected behaviour**
when subtypes are referenced directly from outside the shared companion, they should prepend the encapsulating path to the type.
If inside the shared companion, then bare reference should be fine: if `Lion` was part of `Animal`s `oneOf` => it would have extended same sealed trait, and would have resided inside the companion,  thus can (and should) reference `Zebra` directly.

edge case: make sure inner types don't collide with imported types.

hochgi:
fixes #2868

This code will:
- nest when subtypes are referenced directly (e.g. `Animal.Zebra` as in the added test)
- avoid nesting when the encapsulating object reside under same object (i.e. everything defined inside the encapsulating `Animal` object, does not need to prepend `Animal.` to any subtypes referenced directly)
- also deal with type arguments, e.g. `Option[Animal.Alligator]` as in the added tests

### DIFF
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
@@ -235,8 +235,7 @@ final case class EndpointGen(config: Config) {
    * @param f
    *   a function that may alter the type, None means no altering is needed.
    * @return
-   *   [[scala.Option]] of the altered type, or None if no modification was
-   *   needed.
+   *   The altered type, or gives back the input if no modification was needed.
    */
   def mapTypeRef(sType: Code.ScalaType)(f: Code.TypeRef => Code.TypeRef): Code.ScalaType =
     sType match {
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
@@ -235,7 +235,8 @@ final case class EndpointGen(config: Config) {
    * @param f
    *   a function that may alter the type, None means no altering is needed.
    * @return
-   *   [[Option]] of the altered type, or None if no modification was needed.
+   *   [[scala.Option]] of the altered type, or None if no modification was
+   *   needed.
    */
   def mapTypeRef(sType: Code.ScalaType)(f: Code.TypeRef => Code.TypeRef): Code.ScalaType =
     sType match {
@@ -252,7 +253,7 @@ final case class EndpointGen(config: Config) {
    * structure recursively.
    *
    * @param f
-   *   function to transform a [[Code.CaseClass]]
+   *   function to transform a [[zio.http.gen.scala.Code.CaseClass]]
    * @param code
    *   the structure to apply transformation of case classes on
    * @return
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
@@ -39,6 +39,12 @@ object EndpointGen {
         case None    => m - k
       }
   }
+
+  implicit class OptionCompanionCompatOps(o: Option.type) {
+    // scala 2.12 collection does not support Option.when natively, so we're adding this as an extension.
+    def when[A](p: => Boolean)(value: => A): Option[A] =
+      if (p) Some(value) else None
+  }
 }
 
 final case class EndpointGen(config: Config) {
@@ -150,7 +156,7 @@ final case class EndpointGen(config: Config) {
     // we don't need a separate code file, as it will be included in the sealed trait companion.
     // this var stores the bookkeeping of such case classes, and is later used to omit the redundant code files.
     var replacedCasesToOmitAsTopComponents = Set.empty[String]
-    val allComponents                      = componentNameToCodeFile.view.map { case (name, codeFile) =>
+    val allComponents: List[Code.File]     = componentNameToCodeFile.view.map { case (name, codeFile) =>
       traitToSubtypes
         .get(name)
         .fold(codeFile) { subtypes =>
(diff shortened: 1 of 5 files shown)


zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -166,13 +172,96 @@ final case class EndpointGen(config: Config) {
...
+
+    // After filtering out duplicate files, we can make sure  that in all the files left, all fields has proper types.
+    // The types may not be valid in case we reference a concrete subtype of a sealed trait,
+    // as the subtype is defined as an inner class encapsulated inside the trait's companion.
+    // Therefore, we can alter the type to include the enclosing trait/object's name.
+    //
+    // The following function will be used to alter the type of all fields needed.
+    // The `mapCaseClasses` helper takes a function that alters a case class,
+    // and lifts it such that we can apply it to any structure, and it'll take care to recurse when needed.
+    val mapType: String => Code.ScalaType => Code.ScalaType = encapsulatingName =>
```
987Nabil: Too complicated. Can be a `def (st: Code.ScalaType): ScalaType`, since the string is always the scala type name.

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -39,6 +39,12 @@ object EndpointGen {
         case None    => m - k
       }
   }
+
+  implicit class OptionCompanionCompatOps(o: Option.type) {
```
987Nabil: You use this here only once. Please inline. Makes the code to complicated

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -166,13 +172,96 @@ final case class EndpointGen(config: Config) {
...
+   * {{{Chunk[Option[Animal.Zebra]]}}}
+   *
+   * @param sType
+   *   the original type we want to alter
+   * @param f
+   *   a function that may alter the type, None means no altering is needed.
+   * @return
+   *   [[Option]] of the altered type, or None if no modification was needed.
+   */
+  def mapTypeRef(sType: Code.ScalaType)(f: Code.TypeRef => Option[Code.TypeRef]): Option[Code.ScalaType] =
```
987Nabil: I don't have so much time to take a deep look, but can this be simplified to return non option and always use the result, even if it is the previous value. Efficiency is not so important for code gen. We should prefer maintainability.

987Nabil (commented):
I don't have time rn to take a deeper look and think about implications. But please try to simplify the code. I think there is too much generalisation where it might not be needed.

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -166,13 +166,110 @@ final case class EndpointGen(config: Config) {
...
+   * recursively alter the type of a field. Recursion is needed for types that
+   * contain a type parameter. e.g. transforming: {{{Chunk[Option[Zebra]]}}} to
+   * {{{Chunk[Option[Animal.Zebra]]}}}
+   *
+   * @param sType
+   *   the original type we want to alter
+   * @param f
+   *   a function that may alter the type, None means no altering is needed.
+   * @return
+   *   [[Option]] of the altered type, or None if no modification was needed.
```
987Nabil: use full path to fix doc generation

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -166,13 +166,110 @@ final case class EndpointGen(config: Config) {
...
+      case Collection.Opt(inner) => Collection.Opt(mapTypeRef(inner)(f))
+      case _                     => sType
+    }
+
+  /**
+   * Given a function to alter a case class, this function will apply it to any
+   * structure recursively.
+   *
+   * @param f
+   *   function to transform a [[Code.CaseClass]]
```
987Nabil: use full path to fix doc generation

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: done

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: done

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: simplified

hochgi: simplified - no options

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: no need, Option is no longer used

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: done

987Nabil (approved):

PR #2879, opened 2024-06-05
Fix serving of static files over SSL

### LINKED ISSUE #2875: Serving static files with SSL does not work
**Describe the bug**
Serving static files with SSL does not work. Headers are sent but no content is delivered. Note: compression is disabled
This used to work in RC6.

**To Reproduce**

* Setup server SSL certificate
* Add route to a static file, use `Handler.fromFile`
* Browser will not display contents, but headers are available

**Expected behaviour**
The full file contents should be transferred.

kyri-petrou:
/fix #2875
/claim #2875

Unfortunately serving static files in any encoded format (compressed, ssl, etc.) doesn't seem to work. With this PR, we serve files as a stream which we can wrap in a `DefaultHttpResponse` so that Netty can properly encode it

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyBodyWriter.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyBodyWriter.scala
@@ -40,7 +40,6 @@ object NettyBodyWriter {
     body: Body,
     contentLength: Option[Long],
     ctx: ChannelHandlerContext,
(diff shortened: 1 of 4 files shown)


kyri-petrou:
Setting to draft as it seems an unrelated test started failing with this change

kyri-petrou:
Ready for review

jdegoes (approved):

jdegoes:
💪

PR #2886, opened 2024-06-05
Don't share the Scope in `ContentTypeSpec`

### LINKED ISSUE #2885: Fix CI
CI appears to be broken with the following failure:

```
ContentTypeSpec
```

kyri-petrou:
/fix #2885
/claim #2885

@jdegoes I'm fairly certain that this fixes the issue, but it's hard to tell because I can't replicate it locally. I'll run CI 5-10 times (assuming it passes the first time 😅) to make sure it doesn't happen again.

The reason that this started failing now is that since files are served as streams, my assuption is that sharing the same Scope causes issues as the connections are not properly managed across the different test suites. Note that all the other test suites don't share the same Scope

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/ContentTypeSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ContentTypeSpec.scala
@@ -74,6 +74,8 @@ object ContentTypeSpec extends HttpRunnableSpec {
   override def spec = {
     suite("Content-type") {
       serve.as(List(contentSpec))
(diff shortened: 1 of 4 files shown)


kyri-petrou:
There seem to be other tests failing now. I'll mark the PR as draft and mark the PR as ready when they're all stable

kyri-petrou:
@jdegoes PR is ready for review. I rerun CI 4 times and it passed. In case the issue comes back in other test suites I'm happy to fix it there as well in a followup PR

jdegoes (approved):

jdegoes:
Nice work!

PR #2894, opened 2024-06-07
Simpler default `Endpoint#implement` (#2811)

### LINKED ISSUE #2811: Improve the ergonomics of `Endpoint#implement`
Currently, `implement` must be passed a `handler`, which itself requires a function whose input matches the input of the endpoint. It's also necessary to manually ascribe a type to the handler's input parameter.

```scala
  val httpApp = Endpoint(Method.POST / "books").in[BookProps].out[Book]
    .implement(handler { (props: BookProps) => // <-- We must ascribe a type to the input parameter
      BookService.create(props)
    })
``` 

In ZIO land, we don't love manually ascribing types when they can be inferred from the context. Ideally, one could write the following:

```scala
  val httpApp = Endpoint(Method.POST / "books").in[BookProps].out[Book]
    .implement { props => BookService.create(props) }
``` 

As the Endpoint already knows its own input type, it can constrain the provided function to be of the right type. The user's IDE will know exactly what input is required, rather than forcing the user to provide a handler with a logically redundant + manual type ascription. Further, it would be great if `implement` built the `handler` internally, as it's just extra syntactic noise.

987Nabil:
This is actually similar to what we have before we had `ToHandler` and `handler`.

But I think @kitlangton is right, that is is the better user experience because you don't need to make types explicitly.

This is a breaking API change. But I don't see a good migration strategy, unless we don't call it `implement` anymore.

But maybe breaking is okay? @jdegoes 

fixes #2811
/claim #2811

### DIFF
--- a/docs/reference/endpoint.md
+++ b/docs/reference/endpoint.md
@@ -34,7 +34,7 @@ In the above example, we defined an endpoint on the path `/books` that accepts a
 After defining the endpoint, we are ready to implement it. We can implement it using the `Endpoint#implement` method, which takes a proper handler function that will be called when the endpoint is invoked and returns a `Route`:
 
 ```scala
-val booksRoute = endpoint.implement(handler((query: String) => BookRepo.find(query)))
+val booksRoute = endpoint.implement(query => BookRepo.find(query))
 ```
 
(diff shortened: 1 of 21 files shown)


jdegoes:
I wonder how terrible the implications of having handler extend function?

jdegoes (approved):

PR #2897, opened 2024-06-09
feat(#2701): make warning header opt in

### LINKED ISSUE #2701: Why is warning header added by default?
I've noticed that the apps using ZIO HTTP send a warning header on error, containing some information about the backend implementation details (it explicitly mentions ZIO HTTP).

ScalaCLI example:
```scala
//> using scala 3.3.1
//> using dep dev.zio::zio-http:3.0.0-RC4

import zio.*
import zio.http.*

object Application extends ZIOAppDefault:

  private val routes =
    Method.GET / "test" -> Handler.internalServerError

  override def run =
    Server
      .serve(
        // this will remove the "warning" header if the route above is hit
        // in case of "not found" the "warning" header is added anyway
        routes.toHttpApp @@ Middleware.removeHeader("warning")
      )
      .provide(
        Server.defaultWithPort(8080)
      )
```

Some responses:
```bash
$ curl -v localhost:8080/test # no "warning" header - good
…
< content-length: 0
…
$ curl -v localhost:8080/notexists # "warning" header is present, not what I want
…
< HTTP/1.1 404 Not Found
< warning: 404 ZIO HTTP /notexists
< content-length: 0
…
```

Exposing backend implementation details (as I've said, this header mentions ZIO HTTP) can be considered a security issue. Moreover, if I'm not mistaken, the warning header is deprecated or soon to be deprecated (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-cache-06#section-5.5).

After reading the documentation and looking for similar issues on GitHub I still don't  know how to get rid of these headers.

Is there a reason why this is the default behaviour, not opt-in? Is it possible to opt-out? I think such headers shouldn't be added by default.

I understand that I'm expressing a personal opinion, not a proper technical issue; however, I thought that it's worth raising anyway, as the answer may serve as a future reference, in case someone has a similar problem.

wesleymatosdev:
fixes #2701
/claim #2701

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Response.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Response.scala
@@ -140,9 +140,7 @@ object Response {
 
     val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
 
-    val headers = if (includeWarning) Headers(Header.Warning(199, "API", message2)) else Headers.empty
-
-    Response(status = status, headers = headers)
+    Response(status, body = if (includeWarning) Body.fromString(s"warning: $message2") else Body.empty)
   }
 
   def error(status: Status.Error): Response =
@@ -210,7 +208,9 @@ object Response {
    * Creates a new response for the specified throwable. Note that this method
    * relies on the runtime class of the throwable.
    */
-  def fromThrowable(throwable: Throwable, includeWarning: Boolean = false): Response = {
+  def fromThrowable(throwable: Throwable): Response = fromThrowable(throwable, includeWarning = false)
+
+  def fromThrowable(throwable: Throwable, includeWarning: Boolean): Response = {
     throwable match { // TODO: Enhance
       case _: AccessDeniedException           => error(Status.Forbidden, throwable.getMessage, includeWarning)
       case _: IllegalAccessException          => error(Status.Forbidden, throwable.getMessage, includeWarning)
--- a/zio-http/shared/src/main/scala/zio/http/Response.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Response.scala
@@ -133,12 +133,12 @@ object Response {
 
   def badRequest(message: String): Response = error(Status.BadRequest, message)
 
-  def error(status: Status.Error, message: String): Response = {
+  def error(status: Status.Error, message: String, includeWarning: Boolean = false): Response = {
     import zio.http.internal.OutputEncoder
 
     val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
 
-    Response(status = status, body = Body.fromString(message2))
+    Response(status = status, body = Body.fromString(message2), headers = if (includeWarning) Headers(Header.Warning(199, "API", message2)) else Headers.empty)
   }
 
   def error(status: Status.Error): Response =
--- a/zio-http/shared/src/main/scala/zio/http/Response.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Response.scala
@@ -138,7 +138,7 @@ object Response {
 
     val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
 
-    Response(status = status, headers = Headers(Header.Warning(199, "ZIO HTTP", message2)))
+    Response(status = status, body = Body.fromString(message2))
   }
 
   def error(status: Status.Error): Response =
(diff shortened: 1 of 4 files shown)


987Nabil:
@ologbonowiwi this is unrelated to the issue. Take a look at `Response.error`

wesleymatosdev:
Now it's good to go. Thanks for the guidance, @987Nabil. I added some tests also. Please let me know if there's anything else 😄

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -138,7 +138,7 @@ object Response {
 
     val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
 
-    Response(status = status, headers = Headers(Header.Warning(199, "ZIO HTTP", message2)))
+    Response(status = status, body = Body.fromString(message2))
```
987Nabil: This does not offer an opt-in to this behaviour

wesleymatosdev: Fixed. Added a boolean flag to the method; let me know if there's an alternative way.

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -133,12 +133,14 @@ object Response {
...
   def badRequest(message: String): Response = error(Status.BadRequest, message)
 
-  def error(status: Status.Error, message: String): Response = {
+  def error(status: Status.Error, message: String, includeWarning: Boolean = false): Response = {
     import zio.http.internal.OutputEncoder
 
     val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
 
-    Response(status = status, headers = Headers(Header.Warning(199, "ZIO HTTP", message2)))
+    val headers = if (includeWarning) Headers(Header.Warning(199, "API", message2)) else Headers.empty
```
wesleymatosdev: I set the agent as `API` since setting `ZIO HTTP` exposes implementation details, as mentioned in the issue description:
> Exposing backend implementation details (as I've said, this header mentions ZIO HTTP) can be considered a security issue.

wesleymatosdev:
All tests are passing; now, this is ready for review. 

Please let me know if additional changes are required.

987Nabil:
@ologbonowiwi first, thanks for your effort.

But there are two things here that I think don't work out yet.
1. If you take a look at my first comment on the issue, I think we should drop the header (it is deprecated) and have the body opt-in
2. You added a default false to showing the error. This is a problem because we use this methods internally. Which means places the user can't adjust. We would need to have a way for the user to turn this errors on and off. And I think maybe not only globally, but on a per request level would make more sense. This seems to be a great effort though.

Maybe we could try first to make the body error opt-in, with a global config.
@jdegoes do you have any thoughts on this?

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -133,12 +133,14 @@ object Response {
...
-  def error(status: Status.Error, message: String): Response = {
+  def badRequest(message: String, includeWarning: Boolean): Response = error(Status.BadRequest, message, includeWarning)
+
+  def error(status: Status.Error, message: String, includeWarning: Boolean = false): Response = {
     import zio.http.internal.OutputEncoder
 
     val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
 
-    Response(status = status, headers = Headers(Header.Warning(199, "ZIO HTTP", message2)))
+    Response(status, body = if (includeWarning) Body.fromString(s"warning: $message2") else Body.empty)
```
wesleymatosdev, line 143: Added a `"warning: "` prefix because on [`http/HandlerAspect.scala:838`](https://github.com/zio/zio-http/pull/2897/files#diff-319b46aec125e01353188fe3d4b91b66d67c301dee58a03ed0d3510b228bb472R838) and [`http/HandlerAspect.scala:885`](https://github.com/zio/zio-http/pull/2897/files#diff-319b46aec125e01353188fe3d4b91b66d67c301dee58a03ed0d3510b228bb472R885) we used the warning header to build respectively an error body and a console error message, and IDK if it'd be possible to have a non-warning on the body at that step in the code.

jdegoes:
A boolean flag is not the best solution to this problem, as it does not permit opt-in/opt-out. In an ideal world, this might be part of server config, which means it would need to be threaded down from there. Probably there's some subset of config that makes sense to propagate downward for purposes like this.

wesleymatosdev:
> Probably there's some subset of config that makes sense to propagate downward for purposes like this.

I considered this solution but haven't yet found a way to implement it. Is there a place on the codebase where we did something closest to what we need here?

The closest I found was `zio.Config.boolean("error-warning-on-body")`, but I'm still not sure this would work for our use case.

PR #2899, opened 2024-06-10
Implement streaming of multipart field data

### LINKED ISSUE #2734: Support for Streaming of Data Part of Multipart Form Fields
**Describe the bug**
The current implementation allows streaming for form fields on the server side. However, there is a need to stream the body part of each form field individually without waiting for the entire form data to be received from the client.

**To Reproduce**
1. Run the provided server-side code.
2. Use the provided client-side code to send a multipart form with binary fields to the server.
3. Observe that the server waits for the entire form data to be received from the client, before processing them.

```scala
import zio._
import zio.http._
import zio.stream.{ZSink, ZStream}

object MultipartFormDataStreaming extends ZIOAppDefault {
  def logBytes = (b: Byte) => ZIO.log(s"received byte: $b")

  private val app: HttpApp[Any] =
    Routes(
      Method.POST / "upload-stream" / "simple"     -> handler { (req: Request) =>
        for {
          count <- req.body.asStream.tap(logBytes).run(ZSink.count)
        } yield Response.text(count.toString)
      },
      Method.POST / "upload-stream" / "form-field" -> handler { (req: Request) =>
        if (req.header(Header.ContentType).exists(_.mediaType == MediaType.multipart.`form-data`))
          for {
            form  <- req.body.asMultipartFormStream
            count <- form.fields
              .tap(f => ZIO.log(s"started reading new field: ${f.name}"))
              .flatMap {
                case sb: FormField.StreamingBinary => sb.data.tap(logBytes)
                case _                             => ZStream.empty
              }
              .run(ZSink.count)
          } yield Response.text(count.toString)
        else ZIO.succeed(Response(status = Status.NotFound))
      },
    ).sandbox.toHttpApp @@ Middleware.debug

  override def run: ZIO[Any with ZIOAppArgs with Scope, Any, Any] =
    Server.serve(app).provide(ZLayer.succeed(Server.Config.default.enableRequestStreaming), Server.live)
}
```

```scala
for {
  url    <- ZIO.fromEither(URL.decode("http://localhost:8080/upload-stream"))
  client <- ZIO.serviceWith[Client](_.url(url) @@ ZClientAspect.requestLogging())
  form = Form(
    Chunk(
      ("foo", "This is the first part of the foo form field."),
      ("foo", "This is the second part of the foo form field."),
      ("bar", "This is the body of the bar form field."),
    ).map { case (name, data) =>
      FormField.streamingBinaryField(
        name = name,
        data = ZStream.fromChunk(Chunk.fromArray(data.getBytes)).schedule(Schedule.fixed(200.milli)),
        mediaType = MediaType.application.`octet-stream`,
      )
    },
  )
  res <- client.request(
    Request
      .post(
        path = "form-field",
        body = Body.fromMultipartForm(form, Boundary("boundary123")),
      ),
  )
} yield ()
```

**Expected behaviour**
I expect to receive the data portion of each binary field as they are sent to the server, without waiting for the client to finish sending all bytes of the form field data. So when the server starts reading a new form field, the 'logBytes' logs incoming bytes as they are received.

kyri-petrou:
/fixes #2734
/claim #2734

With this PR, we enable streaming of multipart field data by handling each chunk emitted by the input ZStream separately, and optionally flushing the buffer when the last byte(s) don't match the boundary.

I have a small concern with this approach though; in cases that request streaming is enabled, the output stream will be emitting individual bytes which can be quite bad for performance. I'm thinking we should potentially make this behaviour configurable via an extra argument to the `StreamingForm` case class and disable it by default. @jdegoes @khajavi what do you think?

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/FormSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/FormSpec.scala
@@ -338,6 +338,30 @@ object FormSpec extends ZIOHttpSpec {
           )
         }
       } @@ samples(10),
+      test("output stream maintains the same chunk structure as the input stream") {
+        val form           = Form(
+          Chunk(
+            FormField.StreamingBinary(
+              name = "blob",
+              data = ZStream.fromChunk(Chunk(1, 2).map(_.toByte)) ++
+                ZStream.fromChunk(Chunk(3).map(_.toByte)),
+              contentType = MediaType.application.`octet-stream`,
+            ),
+          ),
+        )
+        val boundary       = Boundary("X-INSOMNIA-BOUNDARY")
+        val formByteStream = form.multipartBytes(boundary)
+        val streamingForm  = StreamingForm(formByteStream, boundary)
+        val expected       = Chunk(Chunk[Byte](1, 2), Chunk[Byte](3))
+        streamingForm.fields.flatMap {
+          case sb: FormField.StreamingBinary => sb.data
+          case _                             => ZStream.empty
+        }
+          .mapChunks(Chunk.single)
+          .filter(_.nonEmpty)
+          .runCollect
+          .map { c => assertTrue(c == expected) }
+      },
     ) @@ sequential
 
   def spec =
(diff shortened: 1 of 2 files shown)


kyri-petrou:
Ended up implementing the `collect` method without streaming field data (which was not necessary since we collect the entire stream into memory) and the `fields` method using field data streaming. This way we don't incur any performance penalty wherever `collect` is used (e.g., using `Body#asMultipartForm`)

jdegoes (approved):

PR #2900, opened 2024-06-10
Remove leading/trailing whitespaces from rendered authorization header

### LINKED ISSUE #2742: Header.Authorization.Unparsed with empty authScheme cause code to hang forever
**Describe the bug**

When using client code sending a request with an `Authorization` header with no scheme (`Header.Authorization.Unparsed("", "some-key")`), the client code hangs forever.

**To Reproduce**

Reproduction repository: https://github.com/gaeljw/ziohttp3hanging

Run `sbt test`.

The test hangs forever.

_Note that sometimes the test may fail with a pact server issue like `Cause: io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connexion refused: localhost/[0:0:0:0:0:0:0:1]:33523`, ignore this case, and just retry. It's not the issue I care about for now._

**Expected behaviour**

The test should succeed with output like:
```
[info] MyWsClientPactTest:
[info] MyWsClient
[info] - should retrieve a model
[info] Run completed in 3 seconds, 99 milliseconds.
[info] Total number of tests run: 1
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 1, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
```

It does when changing the `headers` value in `MyWsClient` to any of these 3 other variants:
```diff
- val headers: Headers = Headers(Header.Accept(MediaType.application.json), Header.Authorization.Unparsed("", "plop-key"))
+ val headers: Headers = Headers(Header.Accept(MediaType.application.json), Header.Custom("Authorization", "plop-key"))
+ val headers: Headers = Headers(Header.Accept(MediaType.application.json), Header.Authorization.Unparsed("something", "plop-key"))
+ val headers: Headers = Headers(Header.Accept(MediaType.application.json))
```

**Screenshots**
N/A

 - OS: Linux

**Additional context**

The issue appears when upgrading from zio-http 0.0.5 to 3.0.0-RC4.

kyri-petrou:
/fixes #2742
/claim #2742

Seems that Netty doesn't like the leading whitespace in the Authorization header, which is causing the client to hang

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/ClientSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ClientSpec.scala
@@ -108,7 +108,14 @@ object ClientSpec extends HttpRunnableSpec {
     } @@ timeout(5.seconds) @@ flaky(5),
     test("authorization header without scheme") {
       val app             =
-        Handler.fromFunction[Request](h => Response.text(h.headers.get("authorization").getOrElse(""))).sandbox.toRoutes
+        Handler
+          .fromFunction[Request] { req =>
+            req.headers.get(Header.Authorization) match {
+              case Some(h) => Response.text(h.renderedValue)
+              case None    => Response.unauthorized("missing auth")
+            }
+          }
+          .toRoutes
       val responseContent =
         app.deploy(Request(headers = Headers(Header.Authorization.Unparsed("", "my-token")))).flatMap(_.body.asString)
       assertZIO(responseContent)(equalTo("my-token"))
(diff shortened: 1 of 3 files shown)


gaeljw:
Interesting, could it be an issue in Netty then? Maybe worth letting them know?

987Nabil:
@kyri-petrou is line 1064 not a problem too? We split by space and then scheme is index 0 of that array. But if there is no scheme, index 0 would be params not scheme. For the stringification, I think it does not matter. But the `Unparsed` instance would be wrong or? I think if the splitted result has size one, scheme should be empty.

kyri-petrou:
> @kyri-petrou is line 1064 not a problem too? We split by space and then scheme is index 0 of that array. But if there is no scheme, index 0 would be params not scheme. For the stringification, I think it does not matter. But the `Unparsed` instance would be wrong or? I think if the splitted result has size one, scheme should be empty.

Good points. I did the following:
1. Removed any empty strings from the array after splitting on `" "`
2. Parsing a single non-empty string to `Unparsed("", value)`
3. Otherwise parse the first value as the scheme and then the rest as previously

987Nabil (approved):

PR #2903, opened 2024-06-12
Fix for routes with different parameter names

### LINKED ISSUE #2728: Route parameter names and types can result in routing ambiguities despite parameter arity
**Describe the bug**
Unambiguous routes with different arity of path parameters can result in routes getting ignored depending on the parameter type and name

**To Reproduce**
```scala
test("routes with different path parameter arities should all be handled") {
  val one    = Method.GET / string("first") -> Handler.ok
  val getone = Request.get("/1")

  val two    = Method.GET / string("prefix") / string("second") -> Handler.internalServerError
  val gettwo = Request.get("/2/two")

  val onetwo = Routes(one, two).toHttpApp
  val twoone = Routes(two, one).toHttpApp

  for {
    onetwoone <- onetwo.runZIO(getone)
    onetwotwo <- onetwo.runZIO(gettwo)
    twooneone <- twoone.runZIO(getone)
    twoonetwo <- twoone.runZIO(gettwo)
  } yield {
    assertTrue(extractStatus(onetwoone) == Status.Ok) &&
    assertTrue(extractStatus(onetwotwo) == Status.InternalServerError) &&
    assertTrue(extractStatus(twooneone) == Status.Ok) &&
    assertTrue(extractStatus(twoonetwo) == Status.InternalServerError)
  }
},
```

1. Run this test, see that it fails. For `onetwo`, where route `one` is first, requests with `getone` succeed while those with `gettwo` fail. For `twoone`, where route `two` is first, requests with `gettwo` succeed while those with `getone` fail.
2. In the definition of `two`, change `string("prefix")` to `string("first")`, see that the test passes.
3. In either definition of `one` or `two`, change `string("first")` to `int("first")`, see that the test fails.

All failures are because the status is `NotFound`, indicating it was unable to find a matching route for the request

**Expected behaviour**
The test always succeeds regardless of parameter name and type

kyri-petrou:
/fixes #2728
/claim #2728

The solution to the linked issue is when we have multiple segment codecs that match the predicate, to evaluate them all by passing the partial path to their `get` method, and find the one that returns a result.

In order to avoid adding performance overhead, this evaluation is done only when we determined that we have ambiguities

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
@@ -540,7 +540,6 @@ object PathCodec          {
                   index = 0
                   while (index < n && (subtree eq null)) {
                     val matched = matches(index)
-                    // Checking whether we got a match alone is
                     if (matched > 0) {
                       val (_, subtree0) = flattened(index)
                       val subpath       = path.dropLeadingSlash.drop(i + matched)
--- a/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
@@ -514,24 +514,44 @@ object PathCodec          {
                 result = subtree0.value
                 i = i + matched
               }
-            case _ => // Slowest fallback path. Have to to find the first predicate where the subpath returns a result
-              var index = 0
-              val size  = flattened.length
-              while (index <= size && (subtree eq null)) {
-                val (codec, subtree0) = flattened(index)
-                val matched           = codec.matches(segments, i)
-                // Checking whether we got a match alone is
-                if (
-                  matched > 0 && {
-                    val subpath = path.dropLeadingSlash.drop(i + matched)
-                    subtree0.get(subpath).nonEmpty
-                  }
-                ) {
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
```diff
@@ -501,24 +501,57 @@ object PathCodec          {
...
+                  subtree = flattened(lastPositiveIdx)._2
+                  result = subtree.value
+                  i = i + matches(lastPositiveIdx)
+                case _ =>
+                  index = 0
+                  while (index < n && (subtree eq null)) {
+                    val matched = matches(index)
+                    if (matched > 0) {
+                      val (_, subtree0) = flattened(index)
+                      val subpath       = path.dropLeadingSlash.drop(i + matched)
```
jdegoes, line 545: Any way to eliminate these allocations?

jdegoes (approved):

zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
kyri-petrou: I think we can overload the `get` method to accept an index and use that as the starting `i` in this method. I'll open a separate PR

PR #2904, opened 2024-06-13
Avoid allocation of new path when resolving path ambiguities

kyri-petrou:
See [this comment](https://github.com/zio/zio-http/pull/2903/files/409d4ef89c04fe7d96d3c71f14079df48afc9e34#r1636989779).

Unfortunately I couldn't eliminate the allocation of the Chunk from `get` but that would require us to duplicate a lot of the logic. I think that since this branch is only executed when resolving ambiguities (which is quite rare) I think it's acceptable to leave it like this


jdegoes (approved):
Nicely done! Thanks so much for the followup. 🙏

PR #2843, opened 2024-05-14
[ZClient] Unsafely fulfil promises to avoid race conditions

kyri-petrou:
Opening this as a separate PR since it contains some rather big changes.

Some time ago, [I worked on an issue](https://github.com/zio/zio-http/pull/2610/files) where the client would hang on unexpected channel closures. While it seems the approach worked well till pretty much now, there are some underlying race condition between the channel closing and promises being fulfilled which seem to be present only when trying to update to ZIO 2.1.x (see [CI failure](https://github.com/zio/zio-http/actions/runs/9062096631/job/24895224533?pr=2840#step:9:11345)).

I'm still not 100% sure on why this issue is guaranteed to occur in ZIO 2.1.x, but the fix is to "unsafely" fulfil the promises instead of doing that via effects. A speculation is that it might be related to fork/join being much faster in ZIO 2.1.x, and therefore the "wrong" side of the race condition now almost always wins; although that's very hard to prove.

In either way, I rewrote all the client handlers to work without the need to run ZIO effects. This should end up being much more performant as we no longer need to create fibers just to fulfil promises, but also easier to follow and reason with.

PS: I know the handler API changes were not necessary, but I think it's better not to pass the `NettyRuntime` unnecessarily to the handlers. I hope this discourage the previous pattern of unnecessarily running effects in them

PR #2907, opened 2024-06-14
Fix memory leak in netty connection pool

### LINKED ISSUE #2906: Regression - Netty Connection Exceptions with Large Connection Pool Count
Using snapshot: [3.0.0-RC8+22-5ea81ef6-SNAPSHOT](https://oss.sonatype.org/content/repositories/snapshots/dev/zio/zio-http_2.13/3.0.0-RC8+22-5ea81ef6-SNAPSHOT/)

There is a significant regression in request serving when using both `ConnectionPoolConfig.Dynamic` and `ConnectionPoolConfig.Fixed`. This seems to be a regression as this issue only appeared after migrating from RC6 to RC8. Exceptions include: 

```
[info] Exception in thread "zio-fiber-823178100,1568885665" io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection reset by peer: /0.0.0.0:8080
```
```
[error] Jun 13, 2024 3:25:18 PM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException Jun 13, 2024 3:25:18 PM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException
```
```
Exception in thread "zio-fiber-1756387034" io.netty.handler.codec.PrematureChannelClosureException: Channel closed while executing the request. This is likely caused due to a client connection misconfiguration
```

Reproducer:
```scala
package repro

import zio.{ Config => _, *}
import zio.http.*
import zio.http.ZClient.Config

  private val clientLayer =
    (ZLayer.succeed(Config.default.copy(connectionPool = ConnectionPoolConfig.Dynamic(1, 512, 60.second))) ++ ZLayer
      .succeed(NettyConfig.defaultWithFastShutdown) ++
      DnsResolver.default) >>> Client.live

  private val url = URL.decode("http://0.0.0.0:8080/Test").toOption.get
  private val req = Request.post(url, Body.empty)

  private val routes =
    HttpApp(
      Routes(
        Method.POST / "Test" ->
          handler(ZIO.sleep(1.second).as(Response.text("success")))
      )
    )

  val run: ZIO[Any & ZIOAppArgs & Scope, Throwable, Unit] = {
    for {
      _ <- Server.serve(routes).fork
      client <- ZIO.service[Client]
      _ <- ZIO.foreachParDiscard(0 to 400) { _ =>
        ZIO
          .scoped[Any] {
            client.request(req).flatMap(_.body.asArray)
          }
          .repeatN(100)
      }
    } yield ()
  }.provide(clientLayer, Server.default)
}
```

kyri-petrou:
/fixes #2906

This is totally on me, big apologies on this one. In #2843 I changed the logic to use a listener on Netty's close future instead of forking a fiber to await on the future. However, I was not removing the listener when the channel would be reset. This is causing the `onResult` promise to being referenced and doesn't allow the response to be GC'd, causing a rather severe memory leak until the channel is closed.

In addition, I found that one of the main things that was contributing massively to the memory leak was using a much-bigger-than-needed ZStream buffer in `AsyncBody`. I added dynamic sizing of the buffer when the content length of the response is known. This way we reduce memory allocations significantly for small requests.

Edit:
1. ~~After feedback from @Not-Aru, I reverted the implementation of listening to the closeFuture back to the one used in versions <= 3.0.0-RC6~~
2. I split the logic for handling http / websocket requests `NettyClientDriver#requestOnChannel`. It was making things difficult to reason with plus we were executing effects that were only meant for running websockets when handling http requests and vice-versa. I think the new implementation is much cleaner and easier to follow

Edit 2:
1. After some more extensive testing, I found a small race condition when interrupting the channel and fixed it
2. Using a FutureListener is much more performant than forking a scoped effect to monitor the closeFuture so I changed the implementation back

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
@@ -44,6 +44,7 @@ abstract class AsyncBodyReader extends SimpleChannelInboundHandler[HttpContent](
             case None              => false
             case Some((_, isLast)) => isLast
           }
+          buffer.clear() // GC
 
           if (ctx.channel.isOpen || readingDone) {
             state = State.Direct(callback)
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
@@ -161,8 +161,16 @@ object NettyBody extends BodyEncoding {
             } catch {
               case e: Throwable => emit(ZIO.fail(Option(e)))
             },
-          4096,
+          streamBufferSize,
         )
+
+    // No need to create a large buffer when we know the response is small
+    private[this] def streamBufferSize: Int = {
+      val cl = knownContentLength.getOrElse(4096L)
+      if (cl <= 16L) 16
+      else if (cl >= 4096) 4096
+      else Integer.highestOneBit(cl.toInt - 1) << 1 // Round to next power of 2
+    }
 
     override def isComplete: Boolean = false
 
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyClientDriver.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyClientDriver.scala
@@ -16,8 +16,6 @@
 
 package zio.http.netty.client
 
-import scala.collection.mutable
-
 import zio._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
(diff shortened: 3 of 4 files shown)


hearnadam:
>  Continuous Integration / Publish Artifacts (pull_request) Skipped 

Can this be run? I would like to test the fix

987Nabil:
@hearnadam If you want to test this locally, please check out the branch and do a `publishLocal`.
Or wait for the snapshot after a merge

Not-Aru:
We did a `publishLocal` and saw an improvement when testing performance over RC8, but not on par with RC6. It is unclear if this is because of our implementation or if this commit still does not completely fix the issue. I will deep dive into the logging to investigate further.

kyri-petrou:
@Not-Aru I'll push another commit, let me know if things are improved.

PS: I identified some optimizations for the client, but I want to make them in a separate PR once this one has been merged in. Based on some local testing it improves performance by 15-20% but would really appreciate if you could help me with any benchmarking. I can tag you in the PR once I open it

kyri-petrou:
@Not-Aru you can pull the latest changes now and `publishLocal` again

kyri-petrou:
@hearnadam @Not-Aru if you don't mind, could you test [this branch](https://github.com/kyri-petrou/zio-http/tree/improve-client-non-streaming-performance) against your benchmarks?

It improves the Client performance quite substantially for responses we read as asArray / asChunk / asString etc. According to some local testing it's ~20-25% more performant than the current implementation. I'll open a PR with those changes once this one is merged in but want to check with you if you're not seeing some other regression from those changes

jdegoes (approved):

jdegoes:
Merged right away due to severity. 👍

russwyte:
Hey folks - it seems CI is broken and this snapshot is not publishing. Also can we expect a release since this is an important bug fix?

Issue #2707, opened 2024-02-27
UserAgent regex broken in ScalaJS


swoogles:
**Describe the bug**

This blows up when its referenced in the browser.
 
https://github.com/zio/zio-http/blob/28c9de1d09b9d8da903dd80edda4a5c4277a5b88/zio-http/shared/src/main/scala/zio/http/Header.scala#L4151

**To Reproduce**
On snapshot - [3.0.0-RC4+77-66867e7f-SNAPSHOT/](https://oss.sonatype.org/content/repositories/snapshots/dev/zio/zio-http_2.13/3.0.0-RC4+77-66867e7f-SNAPSHOT/)

```
object Main extends ZIOAppDefault{
  def run =
    ZIO.succeed(Header.UserAgent)
}
```

<img width="1720" alt="Screenshot 2024-02-27 at 4 40 59 PM" src="https://github.com/zio/zio-http/assets/2054940/f6e97e46-bdc4-4c47-a9d1-2779633d4205">

987Nabil:
Is this actually a zio http error? It seems like valid scala code is producing invalid js code. Is this a scalajs bug?

jamesward:
Regex's that are used in both JS and the JVM, must be compatible with both. No way around there being two different Regex engines.

987Nabil:
@jamesward that's my point. I as a lib dev should not make sure that my regex works for both. That should be the scala js compiler/transpiler

jamesward:
I agree. But unfortunately I don't think Scala has any plans to write their own Regex engine. So the reality for now is that users have to write Regex that works on their target platforms.

swoogles:
Related (Almost decade-old) thread-
https://github.com/scala-js/scala-js/issues/997#issuecomment-54197139

987Nabil:
Okay 🙄  At least the issue is clear now.

One way around would be to not use regex at all for the header parsing but `RichTextCodec`

russwyte:
Would zio-parser be a viable option? One possible win would be that it is invertible.

987Nabil:
`RichTextCodec` is similar to zio parser and we would not have another dependency.

pablf:
/attempt #2707

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/pablf) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
[...]

PR #2912, opened 2024-06-15
Fix ScalaJS regex and regression tests

pablf:
/claim #2707

### DIFF
--- a/zio-http/js/src/test/scala/zio/http/RegexSpec.scala
+++ b/zio-http/js/src/test/scala/zio/http/RegexSpec.scala
@@ -1,0 +1,67 @@
+package zio.http
+
+import zio._
+import zio.test._
+
+object RegexSpec extends ZIOSpecDefault {
+  override def spec: Spec[TestEnvironment with Scope, Any] =
+    suite("RegexSpec")(
+      suite("Header")(
+        test("AcceptLanguage") {
+          ZIO.succeed(Header.AcceptLanguage).isSuccess.map(s => assertTrue(s))
+        },
+        test("ContentDisposition") {
+          ZIO.succeed(Header.ContentDisposition).isSuccess.map(s => assertTrue(s))
+        },
+        test("ContentMd5") {
+          ZIO.succeed(Header.ContentMd5).isSuccess.map(s => assertTrue(s))
+        },
+        test("ContentRange") {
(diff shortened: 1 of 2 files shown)


987Nabil:
@pablf This solution changes what is checked by the regex. To avoid this we discussed in the issue to use an alternative to regex, like rich text codec from zio-http. As @jdegoes added $250 bounty I expect that is what he was aiming for as well.

pablf:
@987Nabil Would you mind expanding what is different between those two regex? I really don't mind using the `RichTextCodec` approach, but before the PR I spent some time trying to understand why this approach was too naive to work but wasn't able to see it  :sweat_smile:.

My reasoning is that matching a regex always applies to the full string, unless explicitly unanchoring it using `.unanchored`. In this case, in which there is only a match (find methods could match only part of the expression but is not the case), adding `^` and `$` at the start or the end of the pattern is redundant as it's already checked by the pattern match unless using `unanchored` on the regex. Also, the fact that `s"""^(?i)([a-z0-9]+)(?:/([a-z0-9.]+))(.*)$$"""` can  be changed to `"^(?i)([a-z0-9]+)(?:/([a-z0-9.]+))(.*)$"` without changing the regex and the string interpolation is not really needed, made me think of some possible bug.

Please correct me if I'm wrong, but I don't really see what is failing here  :sweat_smile:. Also, if this modified regex were wrong, I don't see why a really equivalent regex should automatically be discarded. I didn't see this discussed in #2707, so maybe there is also some benefit to `RichTextCodec` that I don't know.

987Nabil:
@pablf actually I think you are right. There is nothing that makes it match on a sub string like surrounding .* . Maybe it is okay like that.
I am not 100% sure, but I could imagine that rich text codec could be more perfomant if done right. But I am not sure as well. Maybe the solution is okay.

pablf:
@987Nabil I've tried to compare both methods: https://scastie.scala-lang.org/KKXGXYeWSTednuDTCXiCfQ. It's a bit messy, but hope it's helpful.

Not too different, although `Regex` seems 2x faster when matching and `RichTextCodec` sometimes fails faster (in JVM). The relevant comparison for this particular case would be Comparison 5. Also, there might be better designed `RichTextCodec` that would perform better than the ones I made...

jdegoes (approved):

PR #2913, opened 2024-06-15
fix: multiple strings to a single Handler

### LINKED ISSUE #2815: Multiple Strings to a single Handler
Request for a more succinct version of

```
val routes: Routes[Any, Response] = Routes(
    Method.GET / "" -> handHome,
    Method.GET / "index.html" -> handHome,
    Method.GET / "index.htm" -> handHome,
    Method.GET / "index" -> handHome
```

So personally my preference would be to for both a general solution to map from a collection of Strings to a Routes collection. But also a specific case as I imagine a good proportion of web sites will want to cover the standard Home page entry points.

I'm just starting to try out ZIO, so apologies, I don't think I can give a more detailed view that would be of value.

varshith257:
fixes #2815 
/claim #2815

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
@@ -413,12 +413,6 @@ object RoutePatternSpec extends ZIOHttpSpec {
         println("Testing routePattern with index.htm path:")
         println(routePattern.matches(Method.GET, Path("index.htm")))
       
-        println("Testing routePattern with index path:")
-        println(routePattern.matches(Method.GET, Path("index")))
-      
-        println("Testing routePattern with other path:")
-        println(routePattern.matches(Method.GET, Path("other")))
-      
         assertTrue(routePattern.matches(Method.GET, Path("")))
         assertTrue(routePattern.matches(Method.GET, Path("index.html")))
         assertTrue(routePattern.matches(Method.GET, Path("index.htm")))
--- a/zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
@@ -408,7 +408,6 @@ object RoutePatternSpec extends ZIOHttpSpec {
         assertTrue(routePattern.matches(Method.GET, Path("index.html")))
         assertTrue(routePattern.matches(Method.GET, Path("index.htm")))
         assertTrue(routePattern.matches(Method.GET, Path("index")))
-        assertFalse(routePattern.matches(Method.GET, Path("other")))
       },
     )
 
--- a/zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
@@ -402,12 +402,11 @@ object RoutePatternSpec extends ZIOHttpSpec {
   def anyOfTests =
     suite("anyOf")(
       test("anyOf with multiple patterns") {
-        val routePattern = Method.GET / anyOf("", "index.html", "index.htm", "index")
+        val routePattern = Method.GET / anyOf("", "index.html", "index.htm")
 
         assertTrue(routePattern.matches(Method.GET, Path("")))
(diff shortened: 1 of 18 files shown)


zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
```diff
@@ -399,12 +399,32 @@ object RoutePatternSpec extends ZIOHttpSpec {
...
+        val routePattern = Method.GET / anyOf("", "index.html", "index.htm")
+
+        println("Testing routePattern with empty path:")
+        println(routePattern.matches(Method.GET, Path("")))
+      
+        println("Testing routePattern with index.html path:")
+        println(routePattern.matches(Method.GET, Path("index.html")))
+      
+        println("Testing routePattern with index.htm path:")
+        println(routePattern.matches(Method.GET, Path("index.htm")))
```
kyri-petrou: Dev leftovers?

zio-http/jvm/src/test/scala/zio/http/RoutePatternSpec.scala
```diff
@@ -399,12 +399,32 @@ object RoutePatternSpec extends ZIOHttpSpec {
...
+      
+        assertTrue(routePattern.matches(Method.GET, Path("")))
+        assertTrue(routePattern.matches(Method.GET, Path("index.html")))
+        assertTrue(routePattern.matches(Method.GET, Path("index.htm")))
```
kyri-petrou: The first 2 are not being evaluated (assertions in `zio-test` need to be composed). Use this instead:

```suggestion
        assertTrue(
          routePattern.matches(Method.GET, Path("")),
          routePattern.matches(Method.GET, Path("index.html")),
          routePattern.matches(Method.GET, Path("index.htm"))
        )
```

Also probably a good idea to add a few assertions on a path that shouldn't be matching, e.g.,

```scala
!routePattern.matches(Method.GET, Path("foo"))
!routePattern.matches(Method.POST, Path("index.html"))
```

varshith257:
@kyri-petrou Any suggestions on failing tests? The anyOf method implemented logic seems to be correct to achieve this behavior:
```
val routes: Routes[Any, Response] = 
        Routes( Method.GET / "" -> handHome, 
                     Method.GET / "index.html" -> handHome, 
                     Method.GET / "index.htm" -> handHome, 
                     Method.GET / "index" -> handHome )
```

// Transformed to: 
```
val routes: Routes[Any, Response] = 
Routes( Method.GET / anyOf("", "index.html", "index.htm", "index") -> handHome )
```

kyri-petrou:
> @kyri-petrou Any suggestions on failing tests? The anyOf method implemented logic seems to be correct to achieve this behavior:

@varshith257 AFAICT the code needs to be formatted. You just need to run `sbt fmt` and commit the changes

varshith257:
@kyri-petrou Done! But test suite seems failing :)

zio-http/shared/src/main/scala/zio/http/RoutePattern.scala
```diff
@@ -164,6 +164,18 @@ object RoutePattern {
    */
   def fromMethod(method: Method): RoutePattern[Unit] = RoutePattern(method, PathCodec.empty)
 
+  /**
+   * Creates a RoutePattern that matches any of the specified paths.
+   */
+  def anyOf(paths: String*): PathCodec[Unit] = {
+    if (paths.isEmpty) {
+      PathCodec.empty
```
hearnadam, line 172: This is unnecessary, no? should be handled by the `getOrElse`

varshith257: Yes, just added to see the test fail causing of the empty path to check beforehand, will remove it 😅.

@hearnadam Any suggestions on why tests failing?

kyri-petrou: @varshith257 The tests that are failing are the ones that you added as part of the PR. It means that the implementation doesn't work

kyri-petrou:
I'm sorry @varshith257 but after looking into this, this implementation is impossible to work. The solution you brought forward is concatenating the paths in `anyOf` which is not what the ticket is meant to be about. Feel free to reopen a PR when you have a working implementation, but please test it locally first to ensure that it works

PR #2915, opened 2024-06-16
Access latest route pattern when mapping route errors (#2823)

### LINKED ISSUE #2823: Nesting routes work strange when handling errors
When I am nesting routes as described at https://zio.dev/zio-http/dsl/routes it works until I need call `handleError`.

This server:
```
val routes = literal("to") / Routes(
    Method.GET / "other" ->  handler(ZIO.fail(MyErrors.IdFormatError)),
    Method.GET / "do" / string("id") -> handler { (id:String, _: Request) => Response.text(s"GET /to/do/${id}") }
).handleError(_ match
    case MyErrors.IdFormatError => Response.badRequest
)

def run = Server.serve(routes.toHttpApp).provide(Server.defaultWithPort(8090))

enum MyErrors:
    case IdFormatError
```

When started and invoked `curl -i http://localhost:8090/to/do/45` outputs:
```
HTTP/1.1 500 Internal Server Error
warning: 199 ZIO HTTP "Exception in thread &quot;zio-fiber-37&quot; java.lang.RuntimeException: Expected path segment &quot;do&quot; but found "
content-length: 0
```

But when the route leading to error (that I have to deal with) is removed:
```
val routes = literal("to") / Routes(
    Method.GET / "do" / string("id") -> handler { (id:String, _: Request) => Response.text(s"GET /to/do/${id}") }
)

def run = Server.serve(routes.toHttpApp).provide(Server.defaultWithPort(8090))
```

Everything works fine and the result of `curl -i http://localhost:8090/to/do/45` is:
```
HTTP/1.1 200 OK
content-type: text/plain
content-length: 18

GET /to/do/45
```

987Nabil:
@jdegoes When handling errors of `Route.Unhandled`, we accessed the current route pattern to decode it based on the incoming `Request`. When nesting, the route pattern of the handler changes, but the decoding was already constructed based on the pre-nesting route pattern.

I fixed this by changed the handler in `Route.Handled` to `handler: Handler[Any, Nothing, RoutePattern[_], Handler[Env, Response, Request, Response]]` and when calling `toHandler`, handing over the latest route pattern.

I considered instead manipulating the Request to remove the prefix when nesting, but decided that this is a bad idea, since the `Request` that the user might access would have a path, that does not reflect the path of the remote request.

fixes #2823
/claim #2823

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/RoutesSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/RoutesSpec.scala
@@ -18,6 +18,8 @@ package zio.http
 
 import zio.test._
 
+import zio.http.codec.PathCodec
+
 object RoutesSpec extends ZIOHttpSpec {
   def extractStatus(response: Response): Status = response.status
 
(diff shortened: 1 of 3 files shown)


jdegoes (approved):

jdegoes:
Good to have this one in! Algebraic laws FTW. 😄

Issue #2117, opened 2023-04-24
Client Performance Not Great OOTB


jamesward:
**Describe the bug**
Setting up a `Client` has high overhead ~2 seconds and without connection pooling, there is ~300ms of overhead on each request.  It'd be great to have better out-of-the-box performance.

**To Reproduce**
Project to test performance: https://github.com/jamesward/zio-http-client-perf

On my connection / machine here are some results compared against the JDK's `java.net.http.HttpClient` hitting `https://google.com`

`java.net.http.HttpClient`
 - 1 req = 1375ms
 - 10 sequential = 708ms / req

`Client.default`
 - 1 req = 3038ms
 - 10 sequential = 896ms / req

`FixedConnectionPool size 10`
 - 1 req = 3105ms
 - 10 sequential = 560ms / req

**Expected behaviour**
Lower `Client` creation overhead and lower client reuse overhead (perhaps `Client.default` should use a connection pool)?

jdegoes:
To be specific, performance goals of this ticket should be:

 - Almost no overhead with connection pooling (e.g. connection right away and spin up connection pool in the background)
 - Achieve performance better than `java.net.http.HttpClient` for both single request, and 10 sequential requests (currently the first is 2x slower, while the latter is 200ms slower)

Essentially, out of the box experience should be measurably better than what one could obtain using `java.net.http.HttpClient`, for both single-use scenarios, and batch / high-volume scenarios.

/bounty $600

jamesward:
Yup! Thanks for the clarified goals.

tachyonicbytes:
Hi! I would like to work on this!

I saw that comments on newer PRs say that I should type /attempt #2217 in order to start, but the older ones don't. Should I do it anyway?

adamgfraser:
@tachyonicbytes I think either way is fine! Most important thing is just to let other people know you are working on it which you already have! 😃

tachyonicbytes:
Yeah, algora-pbc[bot] already gave me the rocket emoji, so I guess I did what it wanted

dkovalenko:
We are comparing such a different implementations. Do we want to have something like `.build()` from java HttpClient  on ZClient? In this case we could do some stuff like preparing thread pools and other heavy initialization eagerly.
Also I see that default `Client.config` pool of 10.

<img width="524" alt="image" src="https://github.com/zio/zio-http/assets/874918/c242b3a3-20ec-49a5-abc0-6d1b5ca715ec">


[...]

PR #2919, opened 2024-06-19
Improve the Client's performance for non-stream bodies

kyri-petrou:
/claim #2117

This PR reimplements the way that we "collect" the response body when the user uses `asArray`, `asChunk`, `asText` etc. (basically any method other than `asStream`).

When those methods are used, regardless of whether we use a ZStream or just a plain `ArrayBuilder` as the buffer, we will be materializing the entire body into memory. Therefore, we can avoid all the overheads of ZStream and simply buffer the entire response into the array builder which performs much better.

~~Since we don't have benchmarks for the Client, I don't have any results to show at the moment, but based on some local testing I'm seeing the following:~~
- ~~20% improvement of using `asChunk` vs `asStream.collectAll`~~
- ~~35-40% improvement over the baseline (series/2.x). This is due to some of the optimizations that benefit both streaming and non-streaming bodies~~

EDIT: Benchmark suite added and results are added below

Note:
The one thing I'm a bit sceptical about this implementation is whether the aggregation should be happening within `UnsafeAsync.Aggregating` instead of within `AsyncBodyReader`. I think semantically it would have been better to have this logic within `UnsafeAsync.Aggregating`, but unfortunately the performance is not _as good_ as doing this within the `AsyncBodyReader` since we can't utilize a single ArrayBuilder to buffer the entire response, and so we need to create an intermediate array when `connect` is called.

@jdegoes Let me know what you think. Do we want to prioritise an implementation that performs better or one that it's cleaner/easier to reason with and follow

## Benchmark results

TLDR:
- ~5% improvement when body is collected `asStream`
- ~45% improvement when body is collected `asChunk` / `asArray` / `asString` 🚀 

Main branch (_NOTE: performance of `asChunk` and `asStream.runCollect` is the same since `asChunk` uses a stream under the hood_)
```
[info] Benchmark                                    (path)   Mode  Cnt     Score      Error  Units
[info] ClientBenchmark.zhttpChunkBenchmark           small  thrpt    3  9978.642 ± 1836.980  ops/s
[info] ClientBenchmark.zhttpChunkBenchmark           large  thrpt    3  9077.854 ± 1626.184  ops/s
[info] ClientBenchmark.zhttpStreamToChunkBenchmark   small  thrpt    3  9959.477 ±  593.264  ops/s
[info] ClientBenchmark.zhttpStreamToChunkBenchmark   large  thrpt    3  9135.447 ± 1355.958  ops/s
```

PR branch:
```
[info] Benchmark                                    (path)   Mode  Cnt      Score      Error  Units
[info] ClientBenchmark.zhttpChunkBenchmark           small  thrpt    3  14554.512 ± 2518.083  ops/s
[info] ClientBenchmark.zhttpChunkBenchmark           large  thrpt    3  13730.751 ± 1534.280  ops/s
[info] ClientBenchmark.zhttpStreamToChunkBenchmark   small  thrpt    3  10506.628 ±  327.098  ops/s
[info] ClientBenchmark.zhttpStreamToChunkBenchmark   large  thrpt    3   9642.722 ±  904.606  ops/s
```


kyri-petrou:
@jdegoes, @987Nabil brought to my attention this issue https://github.com/zio/zio-http/issues/2117 after seeing my PR.

I think the issue is pretty outdated with a part of it already solved in a previous release. However this PR does improve the Client's performance quite a fair bit. Do you think it's applicable to claim it (or part of it) if I were to add benchmarks to showcase the improvement? It's totally okay if you think it's not related, just thought to ask :)

jdegoes (approved):

jdegoes:
@kyri-petrou Yes, adding benchmarks would be great! Complexity can be ignored for 20-40% improvement in performance.

zio-http/shared/src/main/scala-2/zio/http/UrlInterpolator.scala
kyri-petrou, line 1: I realised that these macros did not work while adding the benchmarks because the `fromAbsoluteURI` method is package-private.

I decided to just fix it here instead so that I could use them in the benchmarks

kyri-petrou:
@jdegoes done! See the updated PR description for the results

jdegoes (approved):

jdegoes:
❤️

PR #2920, opened 2024-06-19
Alternative literal path segments for route definitions (#2815)

### LINKED ISSUE #2815: Multiple Strings to a single Handler

987Nabil:
fixes #2815
/claim #2815

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
@@ -16,6 +16,7 @@
 
 package zio.http.codec
 
+import scala.annotation.tailrec
 import scala.collection.immutable.ListMap
 import scala.language.implicitConversions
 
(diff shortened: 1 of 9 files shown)


zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
```diff
@@ -310,6 +352,26 @@ sealed trait PathCodec[A] { self =>
     _optimize
   }
 
+  private def fallbacks(f: Fallback[_]): Set[String] = {
+    var fallbacks = Chunk(f.left, f.right)
+    val result    = ChunkBuilder.make[String]()
+    while (fallbacks.nonEmpty) {
```
jdegoes: `head` / `tail` with `Chunk` is `O(n)`. And assuming `fallbacks` is called in compilation phase (which looks like it is), then no need for `while` loops as you can use more idiomatic, safer, nicer methods of doing the compilation.

zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
```diff
@@ -458,6 +530,7 @@ object PathCodec          {
   private[http] sealed trait Opt
   private[http] object Opt {
     final case class Match(value: String)                     extends Opt
+    final case class MatchAny(values: Set[String])            extends Opt
```
jdegoes, line 558: Good to use a set here. We can do even better than that by producing a finite state machine that produces true for a match, and false otherwise (due to the char-by-char nature of testing for string equality). But, separate issue and PR. 😄

jdegoes (changes requested):
Looks great, just one relatively minor change!

zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
987Nabil: You mean some structure, where we have like `n` states, where `n` is max length of the alternatives and there is a way to progress to the next state with the right char, true if the string ends or false for any other char, right?

zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
987Nabil: Done

jdegoes (approved):

jdegoes (approved):

PR #2921, opened 2024-06-19
Generate maps for additional properties in OpenAPI code gen (#2901)

### LINKED ISSUE #2901: [gen] the codegen from openapi does not respect additionalProperties which is how open products (maps/dictionaries) are defined
**Describe the bug**
Using a type that has an open product member, e.g:
```yaml
type: object
additionalProperties:
  type: string
```
according to openAPI spec:
https://swagger.io/docs/specification/data-models/dictionaries/
should be rendered as `Map[String, String]` but instead an empty case class with capitalized name of the field is generated.

**To Reproduce**
use e.g. the following openAPI yaml:
```yaml
info:
  title: Animals Service
  version: 0.0.1
servers:
  - url: http://127.0.0.1:5000/
tags:
  - name: Animals_API
paths:
  /api/v1/zoo:
    get:
      operationId: get_animals
      tags:
        - Animals_API
      description: Get all animals count
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Animals'
openapi: 3.0.3
components:
  schemas:
    Animals:
      type: object
      required:
        - total
        - counts
      properties:
        total:
          type: integer
          format: int32
        counts:
          type: object
          additionalProperties:
            type: integer
            format: int32
```

generate with zio-http-gen, and  examine the generated component:

```scala
package dev.zoo.openapi.component

import zio.schema._

case class Animals(
  total: Int,
  counts: Counts
)
object Animals {

  implicit val codec: Schema[Animals] = DeriveSchema.gen[Animals]

}
case class Counts(
)
object Counts {

  implicit val codec: Schema[Counts] = DeriveSchema.gen[Counts]

}
```

**Expected behaviour**
The generated case class should look like:

```scala
case class Animals(
  total: Int,
  counts: Map[String, Int]
)
```

987Nabil:
fixes #2901
/claim #2901

### DIFF
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
@@ -721,9 +721,9 @@ final case class EndpointGen(config: Config) {
       case JsonSchema.Integer(_)     => None
       case JsonSchema.String(_, _)   => None // this could maybe be im proved to generate a string type with validation
       case JsonSchema.Boolean        => None
-      case JsonSchema.OneOfSchema(schemas) if schemas.exists(_.isPrimitive) =>
+      case JsonSchema.OneOfSchema(schemas) if schemas.exists(_.isPrimitive)                            =>
         throw new Exception("OneOf schemas with primitive types are not supported")
-      case JsonSchema.OneOfSchema(schemas)                                  =>
+      case JsonSchema.OneOfSchema(schemas)                                                             =>
         val discriminatorInfo                       =
           annotations.collectFirst { case JsonSchema.MetaData.Discriminator(discriminator) => discriminator }
         val discriminator: Option[String]           = discriminatorInfo.map(_.propertyName)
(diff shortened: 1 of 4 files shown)


jdegoes (approved):

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -925,8 +928,10 @@ final case class EndpointGen(config: Config) {
             enums = Nil,
           ),
         )
-
-      case JsonSchema.Enum(enums) =>
+      case JsonSchema.Object(_, _, _)                                                                  =>
+        // properties.isEmpty && additionalProperties.isRight
+        throw new IllegalArgumentException("Top-level maps are not supported")
```
hochgi, line 933: @987Nabil why not actually?
If we want to derive an endpoint of a type like: `Endpoint[Unit, Unit, Nothing, Map[String, String], None]`,
how should we go about to do so?

987Nabil: This type would then need to be generated inline. But this code is for generating a file, not an inline type.

PR #2924, opened 2024-06-21
Fix rare race condition in ZClient causing healthy connections to be discarded

kyri-petrou:
This one has been bugging me for quite a bit of time now but I managed to finally trace it down. In short, there is a race condition which _sometimes_ causes connections to be invalidated from the connection pool without any good reason (request succeeds and server keeps the connection alive)

The issue is that, we need to fulfil the `onComplete` promise before we invoke the final callback when we collect the async body, but _after_ we removed the handler from the pipeline. The race condition happens because when the request Scope is closed, it also interrupts the `onComplete` promise [here](https://github.com/zio/zio-http/blob/554660c6a4a1e4b1a7cbf1f1733aecc985c37409/zio-http/shared/src/main/scala/zio/http/ZClient.scala#L730).

So when we make a request like below, if the callback wins the race, then the scope will be closed _before_ we manage to fulfil the promise, and therefore the Channel will be discarded.

```scala
ZIO.scoped(ZClient.request(???).flatMap(_.body.asChunk))
```

The change in this PR ensures that the onComplete promise is fulfilled right before we call the async body handler on the last message, but _after_ we've removed the handler from the pipeline. This is also important since if we remove the handler after we complete the promise, there is another potential race condition where we try to reuse the channel and end up with an error because the handler hasn't been removed yet.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
@@ -77,6 +77,8 @@ abstract class AsyncBodyReader extends SimpleChannelInboundHandler[HttpContent](
     val _ = ctx.channel().config().setAutoRead(previousAutoRead)
   }
 
+  protected def onLastMessage(): Unit = ()
+
   override def channelRead0(
     ctx: ChannelHandlerContext,
     msg: HttpContent,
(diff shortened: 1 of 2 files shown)


jdegoes (approved):
These kinds of bugs can be incredibly annoying to track down, and it takes a skilled and dedicated programmer to do that. Thank you!

PR #2930, opened 2024-06-22
Tags for Endpoint and OpenAPI generation (#2716)

### LINKED ISSUE #2716: Tags for Endpoint API

987Nabil:
fixes #2716
/claim #2716

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/Doc.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/Doc.scala
@@ -17,7 +17,6 @@
 package zio.http.codec
 
 import zio.Chunk
-import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import zio.schema.Schema
 
@@ -53,6 +52,24 @@ sealed trait Doc { self =>
       case Doc.Sequence(left, right) => left.flattened ++ right.flattened
       case x                         => Chunk(x)
     }
+
+  def tag(tags: Seq[String]): Doc = self match {
+    case Tagged(doc, existingTags) => Tagged(doc, existingTags ++ tags)
+    case _                         => Tagged(self, tags.toList)
+  }
+
+  def tag(tag: String): Doc =
+    self match {
+      case Tagged(doc, tags) => Tagged(doc, tags :+ tag)
+      case _                 => Tagged(self, List(tag))
+    }
+
+  def tag(tag: String, tags: String*): Doc = self.tag(tag +: tags)
+
+  def tags: List[String] = self match {
+    case Tagged(_, tags) => tags
+    case _               => Nil
+  }
 
   def toCommonMark: String = {
     val writer = new StringBuilder
@@ -131,20 +148,27 @@ sealed trait Doc { self =>
         case Doc.Raw(_, docType) =>
           throw new IllegalArgumentException(s"Unsupported raw doc type: $docType")
 
+        case Doc.Tagged(_, _) =>
+
       }
     }
 
     render(this)
+    val tags = self.tags
+    if (tags.nonEmpty) {
+      // Add all tags to the end of the document as an unordered list
+      render(Doc.unorderedListing(tags.map(Doc.p): _*))
+    }
     writer.toString()
   }
 
   def toHtml: template.Html = {
     import template._
 
-    self match {
-      case Doc.Empty                      =>
+    val html: Html = self match {
+      case Doc.Empty                                =>
         Html.Empty
-      case Header(value, level)           =>
+      case Header(value, level)                     =>
         level match {
           case 1 => h1(value)
           case 2 => h2(value)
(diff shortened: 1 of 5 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/Description.scala
```diff
@@ -0,0 +1,11 @@
+package zio.http.endpoint
+
+import zio.http.codec.Doc
+
+final case class Description(doc: Doc, tags: List[String]) {
```
jdegoes: I know this form is more convenient to work with as an implementor, but I feel like just adding another type of `Doc`, e.g. `Doc.Tagged(doc, tag)` will result in fewer breaking changes and a more ergonomic API for the developer, who ordinarily doesn't want tags. Then you can define `Doc#tags` to extract `List[String]` through recursion and otherwise use the same as you use this one.

987Nabil: I get your point, but I am not sure how to handle tags in the middle of a docs ADT tree. Would we just ignore it, when rendering html/md? Do I get also deep nested tags when using recursion on any doc?

These thought lead me to this design

987Nabil: Btw: Your design might be better, if not only endpoints but maybe data structures need to be tagged.

jdegoes: I would collect them and render as `<ul> / <li>`.

So the tags are associated with the endpoint, and not specific items inside the endpoint?

987Nabil: > So the tags are associated with the endpoint, and not specific items inside the endpoint?
I am not sure. For OpenAPI that is the only usecase (afaik). But I think Smithy has tags also on items.

I'll do changes later, I'll be working today on some of the larger issues

987Nabil:
@jdegoes please check again

YassineMEJRI:
@987Nabil Shouldn't a tag have optional description and externalDocs per [OpenAPI 3.0 spec](https://swagger.io/docs/specification/grouping-operations-with-tags/)? I was thinking something like 
`final case class Tag(name: String, description: Option[Doc], externalDocs: Option[ExternalDoc]) extends Doc`
ExternalDoc is already defined [here](https://github.com/zio/zio-http/blob/f5c04110a885d6c79b640665748eadb802a010d8/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPI.scala#L264) but could be moved to Doc too?
I also think the generated JSON should contain the tags section that contains the list of used tags.

zio-http/shared/src/main/scala/zio/http/codec/Doc.scala
```diff
@@ -54,6 +54,17 @@ sealed trait Doc { self =>
       case x                         => Chunk(x)
     }
 
+  def tag(tags: Seq[String]): Doc = tags.foldLeft(self)((doc, tag) => Tagged(doc, tag))
+
+  def tag(tag: String): Doc = Tagged(self, tag)
+
+  def tag(tag: String, tags: String*): Doc = self.tag(tag +: tags)
+
+  def tags: List[String] = self match {
```
jdegoes, line 69: Should there be a method that does recursive tag extraction??

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPIGen.scala
```diff
@@ -542,7 +542,7 @@ object OpenAPIGen {
     def operation(endpoint: Endpoint[_, _, _, _, _]): OpenAPI.Operation = {
       val maybeDoc = Some(endpoint.doc + pathDoc).filter(!_.isEmpty)
       OpenAPI.Operation(
-        tags = Nil,
+        tags = endpoint.tags,
```
jdegoes, line 562: Should this use the recursive method for tag collection?

zio-http/shared/src/main/scala/zio/http/codec/Doc.scala
```diff
@@ -54,6 +54,17 @@ sealed trait Doc { self =>
...
+    case Tagged(doc, tag) => doc.tags :+ tag
```
jdegoes: This is not stack-safe and will blow up on sufficiently nested `Doc`.

jdegoes (changes requested):
One comment on stack safety, otherwise looks great!

PR #2938, opened 2024-06-27
Use `Exit.succeed` when body extraction is side-effect free

kyri-petrou:
When the server evaluates a ZIO effect, if the effect is an `Exit` (i.e., its value is known without evaluating the effect) then we shortcut the effect forking and respond immediately within Netty's event loop.

The issue this PR is trying to address is that this shortcut optimization is currently only possible with `GET` requests. Any kind of request that requires us to extract the body of the request immediately becomes effectful due to `ZIO.succeed`. With this change, we allow for POST requests to also be shortcut by returning an `Exit` when there are no side-effects or copying of arrays that are being captured by `ZIO.succeed`

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Body.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Body.scala
@@ -383,13 +383,13 @@ object Body {
   ) extends Body
       with UnsafeBytes { self =>
 
-    override def asArray(implicit trace: Trace): Task[Array[Byte]] = ZIO.succeed(data.toArray)
+    override def asArray(implicit trace: Trace): Task[Array[Byte]] = Exit.succeed(data.toArray)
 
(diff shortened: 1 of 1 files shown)


zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -531,8 +531,8 @@ object Body {
 
   }
 
-  private val zioEmptyArray = ZIO.succeed(Array.empty[Byte])(Trace.empty)
+  private val zioEmptyArray = Exit.succeed(Array.emptyByteArray)
```
kyri-petrou, line 534: Note on this: Empty arrays are effectively immutable because they have a fixed length of 0, and so they cannot be mutated

jdegoes (approved):
Clever, I like it!

PR #2944, opened 2024-07-06
Expose config option for avoiding context switching when handling requests

kyri-petrou:
In a previous issue (see #2577), I highlighted two issues with using `runOrFork` to run effects within Netty's event loop; one related to using a Loom-based executor and the other regarding CPU-heavy work. This was subsequently fixed by #2782. 

However, after reflecting some more on this, I think a better approach would have been to make this behaviour configurable and give users the option to avoid context switching by running effects within Netty's event loop until the first async/blocking boundary if they wish so. This gives the option to _very_ advanced users to have more fine-grained control over when effect execution is passed on to ZIO's executor (they can force the switch by using `ZIO.yieldNow`)

I tried to document the caveats of enabling this option as well as possible, and disabled it by default. Please let me know if you have any recommendations on the naming or how we propagate the config.

PS: I'm a bit conflicted whether this option belongs in the `Server.Config` or `NettyConfig`. On one hand, I think it's more related to the ZIO-Netty integration, but on the other hand this option is only used for the Server whereas the `NettyConfig` is shared for the server/client. @jdegoes let me know what you think


jdegoes:
@kyri-petrou Looks good to me. I would probably put it in `Server.Config`, on grounds that while it is definitely Netty related, an option like this could make sense in any backend.

kyri-petrou:
@jdegoes sounds good, I moved that param to `Config` and changed the documentation to "IO thread-pool"

987Nabil (approved):

PR #2946, opened 2024-07-07
Allow configuring a default Middleware that applies to all applications

### LINKED ISSUE #2835: Improve OpenTelemetry integration so traces include path
When using the OTEL agent with ZIO-http, we get the benefit of existing Netty instrumentation. This means traces are started/stopped and propagated correctly, but there's a problem.
The server trace names contain only the HTTP method (which is useless, we need the route in the trace name).

From what I can tell, we're hitting the same issue as described here https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/8613

We need to add instrumentation to the OTEL agent that provides the span name.
This involves hooking into specific classes and methods in ZIO-http.

Someone from the core team decides if there's enough stability in the internal API to hook into it with the OTEL java agent.

And if there is, you can provide me the classes/methods where I can see the matched route and I'll try and add the required instrumentation to OTEL.

Someone from the core team writes a plugin and I don't have to do anything? :D

kyri-petrou:
/closes #2835
/claim #2835

In order to allow OTEL and other 3rd party tooling to integrate with ZIO HTTP without requiring changes to user code, we can expose a `FiberRef` that allows setting a Middleware (or more accurately, a HandlerAspect) via a `FiberRef`. This `FiberRef` is applied to an app when we install it. 

This FiberRef can then be used in 2 different ways AFAICT:
1. Set its value via a ZLayer and use it in the application bootstrap
2. Use a javaagent to change the definition of this FiberRef and set its default value to something else.

(2) is extremely similar to how Kamon instruments ZIO 2 applications via the `Runtime.defaultSupervisor` FiberRef (see [here](https://github.com/kamon-io/Kamon/blob/master/instrumentation/kamon-zio-2/src/main/scala/kamon/instrumentation/zio2/ZIO2Instrumentation.scala)).

I think this approach is one of the safest as it doesn't require instrumentation to alter internal classes, and it's likely the most performant as adds 0 overhead under the default operation, and any overhead added by javaagent is incurred only during the app initialization.

@987Nabil let me know what you think of this approach. AFAICT it should get the job done, while also being the least intrusive and easiest to maintain approach I can think of.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/NettyDriver.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/NettyDriver.scala
@@ -54,12 +54,13 @@ private[zio] final case class NettyDriver(
     } yield StartResult(port, serverInboundHandler.inFlightRequests)
 
   def addApp[R](newApp: Routes[R, Response], env: ZEnvironment[R])(implicit trace: Trace): UIO[Unit] =
-    ZIO.fiberId.map { fiberId =>
+    ZIO.fiberId.zipWith(Driver.defaultMiddleware.get) { (fiberId, middleware) =>
       var loop = true
       while (loop) {
         val oldAppAndRt     = appRef.get()
         val (oldApp, oldRt) = oldAppAndRt
-        val updatedApp      = (oldApp ++ newApp).asInstanceOf[Routes[Any, Response]]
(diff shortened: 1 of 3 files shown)


kyri-petrou:
Closing as it seems that it's not really resolving the issue - Issue seems to be related to netty not being instrumented properly

PR #2947, opened 2024-07-08
Rework Endpoint authentication

### LINKED ISSUE #2888: Replace Endpoint middleware concept with Security concept
Right now, we cannot enforce that endpoint middleware is actually implemented (it's a very complicated problem that cannot be solved in time for release). Simultaneously, we cannot document security requirements of endpoints.

To fix this issue, we should instead use a `sealed trait AuthType`, or similar, which is placed into `Endpoint`. Based on this, we can generate OpenAPI documentation, as well as potentially even leverage this information in the automatic HTTP client.

```scala
sealed trait AuthType {
  type ClientRequirements
  type ServerRequirements
}
object AuthType {
  case object Basic extends AuthType {
    type ClientRequirements = ...
    type ServerRequirements = ...
  }
  ... 
}
```

Note that endpoint auth type would still be a type parameter, albeit constrained (Z <: AuthType). The reason is to statically retain the type members inside the `AuthType` value, so they can be used by client and server.

This requires more R&D but should address all of the issues with endpoint middleware that we have, currently. The question is, can we do it in time? 😄

987Nabil:
@jdegoes This is still a draft, because I want some feedback, but have some tidy up work to do and add all auth options.

The idea behind the design is:
- AuthType at Endpoint defines what kind of authorization header is expected
- the server impl adds automatically a check for the existence of the authorization header and validates the format based on the set `AuthType`
- creating client invocations requires to set the authorization header via one of our typed values `Header.Authorization.XXX`. Which of these is required is defined in `AuthType.ClientRequirement`

For documentation purposes, all auth related information should be part of `AuthType`. We could add here also meta information like scopes. 

wdyt?

fixes #2888 
/claim #2888


zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -30,6 +30,18 @@ import zio.http._
...
+sealed trait AuthType {
+  type ClientRequirement
+}
+
+object AuthType {
+  type None = None.type
+
+  case object None   extends AuthType { type ClientRequirement = Unit                        }
+  case object Basic  extends AuthType { type ClientRequirement = Header.Authorization.Basic  }
+  case object Bearer extends AuthType { type ClientRequirement = Header.Authorization.Bearer }
```
guersam: FYI, there could the some more use cases:
1. Custom header like `X-API-Key`. The header name can vary.
2. Combination of multiple auth types: I have some endpoints that support both Bearer and `X-API-Key`.
3. I don't have that kind of use case, but there might be needs for cookie based authentication, when the endpoint need to support plain web browsers.

987Nabil: Thanks for the feedback. I want to get comments about the general design. But we can make auth type a FP DSL with fallback. And I want to add more options for sure. Maybe we can make it work that users can define custom auth types. We actually only need the ClientRequirement and a Codec fo the server. But this all depends on the `authorization` header.
A more generic solution would then also need to define the header type and the client requirement/codec would need to use this header type. This would still work. The question is, if it is worth the effort.

guersam: Sure, take you time. I just wanted to inform my use cases that may affect the design e.g. multiple auth types, not to force you to add all the missing stuff right now :sweat_smile:

In general, I think replacing endpoint middleware with dedicated auth input is way more practical, because it was alway difficult to imagine a general endpoint middleware that takes into account server, client, and documentation with necessary type juggling that may affect all the type members.

swoogles:
@jdegoes Is this still on your radar? If this is truly the last major issue blocking the final RC, it would be great to get @987Nabil some of the feedback he's looking for.

jdegoes:
@987Nabil This is the sort of approach I was envisioning. It looks directionally correct and a big cleanup. Do you need any assistance with anything in particular?

987Nabil:
@jdegoes I just wanted to know if I am walking the right path. I'll continue then and let you know when it is ready for review

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -27,9 +27,66 @@ import zio.schema._
 
 import zio.http.Header.Accept.MediaTypeWithQFactor
 import zio.http._
+import zio.http.codec.HttpCodecType.{RequestType, ResponseType}
 import zio.http.codec._
 import zio.http.endpoint.Endpoint.{OutErrors, defaultMediaTypes}
 
+sealed trait AuthType { self =>
```
guersam: Nitpick: I think it deserves its own file now.

zio-http/jvm/src/test/scala/zio/http/endpoint/AuthSpec.scala
```diff
@@ -44,6 +63,201 @@ object AuthSpec extends ZIOSpecDefault {
...
+              .forkDaemon
+              .catchAllCause(c => ZIO.logInfoCause(c)) <* ZIO.sleep(1.seconds)
+            responseBasic  <- responseBasic
+            responseBearer <- responseBearer
+          } yield assertTrue(responseBasic == "admin" && responseBearer == "bearer-admin")
+        },
+        test("Auth from query parameter with context and endpoint client") {
+          val endpoint = Endpoint(Method.GET / "test" / "query")
+            .out[String](MediaType.text.`plain`)
+            .auth(AuthType.Custom(HttpCodec.query("token")))
```
guersam, line 193: :+1:

zio-http/jvm/src/test/scala/zio/http/endpoint/AuthSpec.scala
```diff
@@ -44,6 +63,201 @@ object AuthSpec extends ZIOSpecDefault {
...
+              .forkDaemon
+              .catchAllCause(c => ZIO.logInfoCause(c)) <* ZIO.sleep(1.seconds)
+            response <- response
+          } yield assertTrue(response == "admin")
+        },
+        test("Auth basic or bearer with context and endpoint client") {
+          val endpoint =
+            Endpoint(Method.GET / "test" / "multiAuth")
+              .out[String](MediaType.text.`plain`)
+              .auth(AuthType.Basic | AuthType.Bearer)
```
guersam, line 159: :clap:

guersam (commented):
Brilliant!

zio-http/jvm/src/test/scala/zio/http/endpoint/AuthSpec.scala
```diff
@@ -44,6 +63,201 @@ object AuthSpec extends ZIOSpecDefault {
...
+          val routes   =
+            Routes(
+              endpoint.implementHandler(handler((_: Unit) => withContext((ctx: AuthContext) => ctx.value))),
+            ) @@ basicOrBearerAuthContext
+
+          val responseBasic = for {
+            client <- ZIO.service[Client]
+            locator    = EndpointLocator.fromURL(url"http://localhost:8080")
+            executor   = EndpointExecutor(client, locator)
+            invocation = endpoint(Left(Header.Authorization.Basic("admin", "admin".reverse)))
```
guersam, line 169: Out of curiosity, would it possible to replace it with some kind of tagged union to avoid nested `Either`?

Scala 3 has union type and `scala.reflect.TypeTest` but I'm not sure of the Scala 2 equivalent.

987Nabil: We can add a Scala 3 specific variant of the or. Like in zio-schema
https://github.com/zio/zio-schema/blob/5a276f7417b3039f46da7c85ca41e00c4f993be0/zio-schema/shared/src/main/scala-3/zio/schema/package.scala#L15

jdegoes (approved):

jdegoes:
@987Nabil Going to merge. However, with an endpoint that requires authentication, it should be mandatory to supply the authentication information to the typesafe Endpoint client. This is not currently implemented, but I think it's a big enough PR to merge anyway.

PR #2951, opened 2024-07-09
Encode SSE based on HttpContentCodec (#2695)

### LINKED ISSUE #2695: ClassCastException when running ServerSentEventEndpoint Example
1. Run the `ServerSentEventEndpoint` from the `example` project.
2. Run `curl -N "http://127.0.0.1:8080/sse" -i`

Expected:
The curl should print the events created by the server to the console

Current Behaviour:
The curl prints the following message:

```bash
> curl -N "http://127.0.0.1:8086/sse" -i
HTTP/1.1 200 OK
content-type: application/json
transfer-encoding: chunked

curl: (18) transfer closed with outstanding read data remaining
```

The app prints these exceptions:

```scala
Feb 23, 2024 11:56:28 AM io.netty.channel.DefaultChannelPipeline onUnhandledInboundException
WARNING: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.ClassCastException: class zio.http.ServerSentEvent cannot be cast to class java.lang.Byte (zio.http.ServerSentEvent is in unnamed module of loader 'app'; java.lang.Byte is in module java.base of loader 'bootstrap')
	at scala.runtime.BoxesRunTime.unboxToByte(BoxesRunTime.java:91)
	at scala.runtime.ScalaRunTime$.array_update(ScalaRunTime.scala:80)
	at zio.Chunk$Singleton.toArray(Chunk.scala:1818)
	at zio.Chunk.toArray(Chunk.scala:1002)
	at zio.Chunk.toArray(Chunk.scala:873)
	at zio.http.netty.NettyBodyWriter$.$anonfun$writeAndFlush$14(NettyBodyWriter.scala:123)
	at zio.http.netty.NettyFutureExecutor$.$anonfun$make$1(NettyFutureExecutor.scala:67)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:904)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:967)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:948)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:967)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:1024)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:967)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:1024)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:1024)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:967)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:890)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:1024)
	at zio.internal.FiberRuntime.runLoop(FiberRuntime.scala:1024)
	at zio.internal.FiberRuntime.evaluateEffect(FiberRuntime.scala:381)
	at zio.internal.FiberRuntime.evaluateMessageWhileSuspended(FiberRuntime.scala:504)
	at zio.internal.FiberRuntime.drainQueueOnCurrentThread(FiberRuntime.scala:220)
	at zio.internal.FiberRuntime.run(FiberRuntime.scala:139)
	at zio.internal.ZScheduler$$anon$4.run(ZScheduler.scala:478)
```

Please note that I have written another server-sent example that works properly:

```scala
object ServerSentExample extends ZIOAppDefault {

  val stream: ZStream[Any, Nothing, ServerSentEvent] =
    ZStream.repeatWithSchedule(ServerSentEvent(ISO_LOCAL_TIME.format(LocalDateTime.now)), Schedule.spaced(1.second))
    
  val app =
    Routes(
      Method.GET / "sse" ->  handler {
          Response.fromServerSentEvents(stream)
      }
    ).toHttpApp
  def run = Server.serve(app).provide(Server.default)
}
```

987Nabil:
The current impl. of SSE uses special casing in the `EncoderDecoder`. Since we now have a generic way of handling en/decoding with `HttpContentCodec`, we don't need this anymore and can use a custom SSE content codec. This has also the benefit, that creating a codec for SSE and trying to change the content type, will lead to an error, since only `text/event-stream` has a codec defined.

I see this also as a fix to #2695. The issues arose because we selected `application/json` has the media type and it was flagged as `binary` in an old zio-http version. Which is no longer the case. But this fix prevents it from ever using any other content type then `text/event-stream`, except it is forced by the user.

Also, the encoding of SSE is currently bugged on main, since it is encoded as json based on the schema. This PR fixes that as well.

fixes #2695
/claim #2695

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
@@ -216,11 +216,6 @@ private[codec] object EncoderDecoder {
       } else {
         false
       }
-    private val isEventStream               = if (flattened.content.length == 1) {
-      isEventStreamBody(flattened.content(0))
-    } else {
-      false
-    }
     private val onlyTheLastFieldIsStreaming =
       if (flattened.content.size > 1) {
         !flattened.content.init.exists(isByteStreamBody) && isByteStreamBody(flattened.content.last)
@@ -533,26 +528,20 @@ private[codec] object EncoderDecoder {
           case SimpleCodec.Specified(method) => Some(method)
         }
       } else None
-    private def encodeBody(inputs: Array[Any], outputTypes: Chunk[MediaTypeWithQFactor]): Body = {
+    private def encodeBody(inputs: Array[Any], outputTypes: Chunk[MediaTypeWithQFactor]): Body =
       if (isByteStream) {
         Body.fromStreamChunked(inputs(0).asInstanceOf[ZStream[Any, Nothing, Byte]])
       } else {
-        if (inputs.length > 1) {
-          Body.fromMultipartForm(encodeMultipartFormData(inputs, outputTypes), formBoundary)
-        } else {
-          if (isEventStream) {
-            Body.fromCharSequenceStreamChunked(
-              inputs(0).asInstanceOf[ZStream[Any, Nothing, ServerSentEvent]].map(_.encode),
-            )
-          } else if (inputs.length < 1) {
+        inputs.length match {
+          case 0 =>
             Body.empty
-          } else {
+          case 1 =>
             val bodyCodec = flattened.content(0)
             bodyCodec.erase.encodeToBody(inputs(0), outputTypes)
-          }
+          case _ =>
+            Body.fromMultipartForm(encodeMultipartFormData(inputs, outputTypes), formBoundary)
         }
       }
-    }
 
     private def encodeMultipartFormData(inputs: Array[Any], outputTypes: Chunk[MediaTypeWithQFactor]): Form = {
       Form(
@@ -581,8 +570,7 @@ private[codec] object EncoderDecoder {
         if (inputs.length > 1) {
           Headers(Header.ContentType(MediaType.multipart.`form-data`))
         } else {
-          if (isEventStream) Headers(Header.ContentType(MediaType.text.`event-stream`))
-          else if (flattened.content.length < 1) Headers.empty
+          if (flattened.content.length < 1) Headers.empty
           else {
             val mediaType = flattened
               .content(0)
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -581,8 +572,7 @@ private[codec] object EncoderDecoder {
         if (inputs.length > 1) {
           Headers(Header.ContentType(MediaType.multipart.`form-data`))
         } else {
-          if (isEventStream) Headers(Header.ContentType(MediaType.text.`event-stream`))
-          else if (flattened.content.length < 1) Headers.empty
+          if (flattened.content.length < 1) Headers.empty
```
guersam, line 573:
```suggestion
          if (flattened.content.isEmpty) Headers.empty
```

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -540,11 +535,7 @@ private[codec] object EncoderDecoder {
         if (inputs.length > 1) {
           Body.fromMultipartForm(encodeMultipartFormData(inputs, outputTypes), formBoundary)
         } else {
-          if (isEventStream) {
-            Body.fromCharSequenceStreamChunked(
-              inputs(0).asInstanceOf[ZStream[Any, Nothing, ServerSentEvent]].map(_.encode),
-            )
-          } else if (inputs.length < 1) {
+          if (inputs.length < 1) {
```
guersam: nitpick: How about replacing the nested `if`s with pattern matching for better readability?

```scala
input.length match {
  case 0 => 
    Body.empty
  
  case 1 =>
    val bodyCodec = flattened.content(0)
    bodyCodec.erase.encodeToBody(inputs(0), outputTypes)
  
  case _ =>  
    Body.fromMultipartForm(encodeMultipartFormData(inputs, outputTypes), formBoundary)
}
```

PR #2954, opened 2024-07-09
#2948: Remove new line character from sandbox warning header

### LINKED ISSUE #2948: Warning header loses original exception when that exception contains invalid-for-netty-headers characters
**Describe the bug**
I am using zio-http and quill to build a REST service. I create a request handler via `Handler.fromFunctionZIO` that calls into my DB via a quill. In one instance, I forgot to `lift` a variable when creating the query (ie I did `id == someId` instead of `id == lift(someId)`. This causes quill to throw an `IllegalStateException` which is then treated as a fatal defect by the quill-created (I think? I don't think that's a relevant detail) ZIO instance. With `sandbox` turned on for my routes, the error handling mechanism attempts to turn the error into a warning response header. However, the error text for the `IllegalStateException` exception has a `\n` character in it, which fails netty's header validation (`verifyValidHeaderValueCharSequence` in `HttpHeaderValidationUtil.java` returns the index of the first `\n` in the string and not `-1`), causing the original error to be swallowed, and the warning response header to then contain the self-fulfilling meta Schrodinger's error `Validation failed for header 'warning'`. 

**To Reproduce**
1. Turn on sandbox (`Server.serve(routes.sandbox)`)
2. Call a route handled by a ZIO via `Handler.fromFunctionZIO`
3. Have this ZIO fail fatally, and include a new line character in the exception error message

**Expected behaviour**
The original error is returned in the warning header.

jgoday:
Removes new line characters (invalid in header values) when using sandbox to expose exceptions as warning headers.

fixes #2948

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Response.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Response.scala
@@ -136,7 +136,7 @@ object Response {
   def error(status: Status.Error, message: String): Response = {
     import zio.http.internal.OutputEncoder
 
-    val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
+    val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message).replace("\n", " ")
 
     Response(status = status, headers = Headers(Header.Warning(199, "ZIO HTTP", message2)))
   }
--- a/zio-http/jvm/src/test/scala/zio/http/HandlerSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/HandlerSpec.scala
@@ -30,6 +30,22 @@ object HandlerSpec extends ZIOHttpSpec with ExitAssertion {
 
   def spec = suite("Handler")(
     suite("sandbox")(
+      test("response failure with exception is passed through") {
+        val message = "long exception\nwith multiple\nlines"
+        val agent   = "ZIO HTTP"
+        val handler =
+          Handler
+            .fromFunctionZIO[Any] { _ =>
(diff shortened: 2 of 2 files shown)


987Nabil:
I actually think we should get rid of the warning header. It is deprecated and does not seem a good fit

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -136,7 +136,7 @@ object Response {
   def error(status: Status.Error, message: String): Response = {
     import zio.http.internal.OutputEncoder
 
-    val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message)
+    val message2 = OutputEncoder.encodeHtml(if (message == null) status.text else message).replace("\n", " ")
```
anqit, line 139: Might be helpful if this was replaced by `"\\n"` instead, I think it would make it easier to recover the original error text if that was ever desired. Also, I believe netty disallows other characters in its headers, would it make sense to handle all of them at once?

jgoday: Seems that escaping with '\\n' should lead to the same problem.

I think netty (or http spec) only accepts visible ascii characters (between 33 and 125),
should we use URL encoding to escape them ?

jgoday:
> I actually think we should get rid of the warning header. It is deprecated and does not seem a good fit

I undesrtand, feel free to remove this PR.

987Nabil:
Closing since we will remove the warning header usage

PR #2959, opened 2024-07-13
Multi-value segments for PathCodecs

### LINKED ISSUE #2679: Improve Routes DSL to support complex URI template patterns
Extend capabilities of ZIO HTTP routes to cover use cases described in [RFC 6570 - URI Template](https://www.rfc-editor.org/rfc/rfc6570)
The issue that I came across is that the current ZIO Route DSL does not support defining complex use cases such as the following:
- multiple path variables in the same path segment, e.g, `http://example.com/people/{firstName}-{lastName}/SSN`
- extractor and literal within the same path segment, e.g `http://example.com/people/{firstName}-some-literal/SSN` or  `http://example.com/people/some-literal-{firstName}/SSN`

Just for reference: https://docs.micronaut.io/latest/guide/#routing, Many other frameworks, including Spring, Quarkus, dotnet, actix-web, implement the URI template specification. But no single scala HTTP library is capable of that.

It would be nice to have the flexibility to define complex routes via Route DSL and Endpoint DSL to have a well-designed and documented API.

image

It seems that it is not possible right now to define a route that will follow [Resource Oriented API with custom methods](https://aip-dev.github.io/aip.dev/136) (more details [here](https://cloud.google.com/apis/design/custom_methods))

I remember that something similar was discussed in the zio-http discord channel - https://discord.com/channels/629491597070827530/819703129267372113/1162396036237623326 but without proper resolution.

I want to have DLS that provide an ability to define
 - multiple path variables in the same path segment:
```scala
POST / "virtual-machines" / string("vmId") / "instances" / PathCodec.combined(string("instanceId"), literal("_"), string("some_other_var"))
```
- extractor and literal within the same path segments
 ```scala
POST / "virtual-machines" / string("vmId") / "instances" / PathCodec.combined(string("instanceId"), literal(":reboot"))
```

@erikvanoosten, probably you have some ideas, coz I remember [discussion](https://discord.com/channels/629491597070827530/819703129267372113/1162443016544456794) in discord about problem related to this issue.

**Additional context**
https://softwaremill.community/t/paths-components-without-separator/317

987Nabil:
fixes #2679
/claim #2679
/split @kyri-petrou

### DIFF
--- a/zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpOptions.scala
+++ b/zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpOptions.scala
@@ -312,7 +312,7 @@ private[cli] object HttpOptions {
   private[cli] def optionsFromSegment(segment: SegmentCodec[_]): Options[String] = {
     def fromSegment[A](segment: SegmentCodec[A]): Options[String] =
       segment match {
-        case SegmentCodec.UUID(name)     =>
+        case SegmentCodec.UUID(name)        =>
           Options
             .text(name)
             .mapOrFail(str =>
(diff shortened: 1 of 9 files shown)


987Nabil:
The full [RFC 6570 - URI Template](https://www.rfc-editor.org/rfc/rfc6570) has no place in zio-http, since it is very focused on query param extraction and zio-http has no query params defined in the path. @kyri-petrou and I went with multi-value segments as the way to implement what the issue creator seemed to care most about.
We added some limitations to what combination of values are possible. For example extracting two `Int`s without a delimiter is not possible, since it is impossible to recognise where one starts and the other ends.

jdegoes (approved):

jdegoes:
Nicely done! Thanks for the attention to performance in this critical section of the code. 🙏

Issue #1522, opened 2022-09-19
Implement a code generator that takes Smithy spec and generates zio.http.api.EndpointSpec values


jdegoes:
**Is your feature request related to a problem? Please describe.**

Some companies begin with Smithy specifications, and then proceed to implement a server that satisfies a given Smithy specification; then they will modify the Smithy specification when a change is necessary, and then re-implement the server.

**Describe the solution you'd like**

In order to support this _spec-first_ driven approach to API development (rather than _code-first_, which is the paradigm already implemented within the `zio.http.api` package), we will need a code generator, which can read in a Smithy specification, and generate the corresponding `zio.http.api.EndpointSpec ` values that declaratively describe the APIs.

Then a user can implement handlers based on the code-generated `EndpointSpec ` values.

jdegoes:
Possibly, this could be done as a plug-in to `smithy4s`.

yisraelU:
I can take on creating a PR in this repo, to have a zio-http interpreter(Smithy4s terminology).
I would wait until Smithy4s 0.18  version is released as it will make it a lot easier.
Happy to discuss further

saadhashmi91:
Hi, I was looking into option of using a recursion scheme(s)/fix-point  parser for Smithy IDL using its ABNF spec. In particular [skeuomorph](https://github.com/higherkindness/skeuomorph)  has an implementation of transforming protobuf schemas. If a similar approach can be used for the parser and also leveraging smithy4s for the spec. implementation.

adamgfraser:
@saadhashmi91 We would like to do this without any third party dependencies.

sideeffffect:
@adamgfraser just to be clear, do you allow other ZIO Organization dependencies, like zio-schema?

adamgfraser:
@sideeffffect If that functionality is necessary for something we are trying to do. For example ZIO Schema is already a dependency since it is necessary to describe arbitrary types as values as part of the endpoints API and we would basically have to implement it ourselves otherwise. But in general would not want to bring in an additional dependency if it was merely a "nice to have".

yisraelU:
> Hi, I was looking into option of using a recursion scheme(s)/fix-point  parser for Smithy IDL using its ABNF spec. In particular [skeuomorph](https://github.com/higherkindness/skeuomorph)  has an implementation of transforming protobuf schemas. If a similar approach can be used for the parser and also leveraging smithy4s for the spec. implementation.

[...]

PR #2964, opened 2024-07-14
Update build.sbt

promisingcoder:
added neccessary dependencies for zio-http/shared/src/main/scala/zio/http/SmithyParser.scala
#1522


kyri-petrou (changes requested):
Let's please not do this. This is adding a lot of dependencies to `zio-http`, with most of them external to the `dev.zio` organization. This will make long-term maintenance of `zio-http` a nightmare

PR #2965, opened 2024-07-14
Create SmithyParser.scala

promisingcoder:
Add SmithyParser to convert Smithy models to OpenAPI and then generate endpoints from OpenAPI
/claim #1522 
I am still in the beginning but due to me not knowing what to do next I made the pull request

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/SmithyParser.scala
+++ b/zio-http/shared/src/main/scala/zio/http/SmithyParser.scala
@@ -1,0 +1,58 @@
+import java.nio.file.{Files, Paths, Path}
+import scala.jdk.CollectionConverters._ // Updated import
+import com.google.gson.{Gson, GsonBuilder}
+import software.amazon.smithy.model.Model
+import software.amazon.smithy.model.node.Node
+import software.amazon.smithy.model.shapes.{ServiceShape, ShapeId}
+import software.amazon.smithy.openapi.OpenApiConfig
+import software.amazon.smithy.openapi.fromsmithy.OpenApiConverter
+import zio.http.endpoint.openapi.OpenAPI
+import zio.http.gen.openapi.EndpointGen
+import zio.http.gen.scala.CodeGen
+
+object SmithyParser {
+  def run(modelPath: Path, outputPath: Path, basePackage: String): Unit = {
+    val converter = OpenApiConverter.create()
+
+    if (!Files.exists(modelPath)) {
+      sys.error(s"Model file not found at path: $modelPath")
+    }
+
+    val model = Model.assembler().addImport(modelPath).discoverModels().assemble().unwrap()
+    val serviceShapes = model.shapes().collect(java.util.stream.Collectors.toList()).asScala.filter(_.isServiceShape)
+
+    if (serviceShapes.isEmpty) {
+      sys.error("No service shapes found in the model.")
+    }
+
+    serviceShapes.foreach { shape =>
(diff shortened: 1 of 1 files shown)


987Nabil:
I don't like this approach. Converting to OpenAPI removes precision in modeling. One reason why smithy exists and is liked is, that it does things different/better than OpenAPI. Reducing it to OpenAPI removes the benefits (for example a better modeling of sum types.) I think we should have dedicated smithy support.

wdyt @jdegoes

promisingcoder:
@jdegoes could you please give me an update on this approach ??

promisingcoder:
@987Nabil 
until now 
when it comes to union types 
there is no problem in the generation from smithy to openapi the problem is in the way that codegen.writefiles generates zio specs from openapi 
I will find a way to pass the right things  to that function so that it generates the required output 
I will detail here in a bit how is this happening

promisingcoder:
and by the way `"EchoUnion": {
                "oneOf": [
                    {
                        "type": "object",
                        "title": "stringMessage",
                        "properties": {
                            "stringMessage": {
                                "type": "string",
                                "pattern": "^[a-zA-Z]+$"
                            }
                        },
                        "required": [
                            "stringMessage"
                        ]
                    },
                    {
                        "type": "object",
                        "title": "intMessage",
                        "properties": {
                            "intMessage": {
                                "type": "number"
                            }
                        },
                        "required": [
                            "intMessage"
                        ]
                    }
                ]
            }`
            ## generated redunant code  from CodeGen.WriteFiles() 
            `package your.base.package.component

import zio.schema._
import zio.schema.annotation._

@noDiscriminator
sealed trait EchoUnion
object EchoUnion {


 implicit val codec: Schema[EchoUnion] = DeriveSchema.gen[EchoUnion]
case class EchoUnion(
 stringMessage: String
)
object EchoUnion {



 implicit val codec: Schema[EchoUnion] = DeriveSchema.gen[EchoUnion]



}
case class EchoUnion(
 intMessage: Double
)
object EchoUnion {



 implicit val codec: Schema[EchoUnion] = DeriveSchema.gen[EchoUnion]



}
         }`

promisingcoder:
> and by the way `"EchoUnion": { "oneOf": [ { "type": "object", "title": "stringMessage", "properties": { "stringMessage": { "type": "string", "pattern": "^[a-zA-Z]+$" } }, "required": [ "stringMessage" ] }, { "type": "object", "title": "intMessage", "properties": { "intMessage": { "type": "number" } }, "required": [ "intMessage" ] } ] }` ## generated redunant code from CodeGen.WriteFiles() `package your.base.package.component
> 
> import zio.schema._ import zio.schema.annotation._
> 
> @noDiscriminator sealed trait EchoUnion object EchoUnion {
> 
> implicit val codec: Schema[EchoUnion] = DeriveSchema.gen[EchoUnion] case class EchoUnion( stringMessage: String ) object EchoUnion {
> 
> implicit val codec: Schema[EchoUnion] = DeriveSchema.gen[EchoUnion]
> 
> } case class EchoUnion( intMessage: Double ) object EchoUnion {
> 
> implicit val codec: Schema[EchoUnion] = DeriveSchema.gen[EchoUnion]
> 
> } }`
2
## I will make sure to fix this If you want @jdegoes @987Nabil

jdegoes:
I don't think we should go through OpenAPI as an intermediate. Too lossy and not what was intented in the original ticket.

PR #2966, opened 2024-07-14
Fix application shutdown issue when `Server#install` is not called

### LINKED ISSUE #2961: Http Server keeps application alive when it should exit
Creating a ZIO http server layer and using it will prevent the application from shutting down once all other effects are completed.

Using RC9 the following code will not exit with code 0, instead remaining running indefinitely.
This behavior does not happen in RC6-8. I have not tested other versions.

```scala
import zio.*
import zio.http.*

object ExampleApp extends ZIOAppDefault {
  def run = (for {
    _ <- ZIO.service[Server]
    exitCode = ExitCode(0)
  } yield (exitCode))
    .provide(
      Server.defaultWithPort(8080)
    )
}
```

Running the application with either `sbt run` or `sbt reStart` produces the same result.
Exiting the application with interrupt works with `sbt run`, reStart forks the process and fails to kill it.

kyri-petrou:
/fixes #2961

Issue is that we're not interrupting the `initialInstall` and `serverStarted` promises when the forked fiber is interrupted

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/ServerStartSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ServerStartSpec.scala
@@ -51,6 +51,15 @@ object ServerStartSpec extends HttpRunnableSpec {
         ZLayer.succeed(NettyConfig.defaultWithFastShutdown),
       )
     },
+    test("application can shutdown if server is not started") {
+      ZIO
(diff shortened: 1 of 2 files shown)


987Nabil (approved):

masonedmison:
Arg, sorry for introducing this issue @kyri-petrou 😢

PR #2969, opened 2024-07-15
Body content type

eyalfa:
refactor body to have `contentType` instead of separate `media-type` and `boundary`


eyalfa:
CC: @987Nabil

eyalfa:
@987Nabil did u get to have a look at this?

987Nabil:
@eyalfa this is not a viable solution. Because you use the a header data type. Bodies have no headers

eyalfa:
> @eyalfa this is not a viable solution. Because you use the a header data type. Bodies have no headers

but I'm not using it as a header, not trying to suggest bodies have headers, they just (usually) have a content type.
information wise this is exactly what's needed in order to describe's a body's content, do u want to introduce a new data type just for that? it'd require converting back and forth with content type header any way

987Nabil:
I would still prefer that I think. Having a data type that is clearly bound to a header used at a place that is not a header feels wrong.

eyalfa:
@987Nabil I've introduced a Body specific ContentType class

eyalfa:
> @987Nabil I've introduced a Body specific ContentType class

@987Nabil  ☝️

987Nabil (approved):

PR #2970, opened 2024-07-17
Fix sbt module compilation issue when using IntelliJ IDE

kyri-petrou:
See https://discord.com/channels/629491597070827530/629491597070827534/1262347147517628486

When compiling via IntelliJ, compilation fails for the `sbt-zio-http-grpc-tests` module as the files are only generated for Scala 2.12. To improve DX when using IntelliJ IDE, we exclude this module when we're working via the IDE.

Note that this change will have no effect via sbt, so we won't see any changes in CI or when using the sbt shell directly


987Nabil:
Can't we make it work for all Scala versions?

jdegoes (approved):

Issue #2942, opened 2024-07-03
StackOverflowError when too many "optional" query parameters defined for Endpoint


Andycharalambous:
**Describe the bug**
If an endpoint has 15 or more optional query parameters any attempt to call it will result in a StackOverflowError unless at least one query param value is provided.

**To Reproduce**
Steps to reproduce the behaviour:

```
type SOIn = ((Option[String],Option[String],Option[String],Option[String],Option[String],Option[String],Option[String],Option[String],Option[String],Option[String]),Option[String],Option[String],Option[String],Option[String],Option[String])
  private val soEndpoint =
    Endpoint(Method.GET / "so")
      .query[Option[String]](query("a").optional)
      .query[Option[String]](query("b").optional)
      .query[Option[String]](query("c").optional)
      .query[Option[String]](query("d").optional)
      .query[Option[String]](query("e").optional)
      .query[Option[String]](query("f").optional)
      .query[Option[String]](query("g").optional)
      .query[Option[String]](query("h").optional)
      .query[Option[String]](query("i").optional)
      .query[Option[String]](query("j").optional)
      .query[Option[String]](query("k").optional)
      .query[Option[String]](query("l").optional)
      .query[Option[String]](query("m").optional)
      .query[Option[String]](query("n").optional)
      .query[Option[String]](query("o").optional)
      .out[String]

private val soHandler: Handler[Any, zio.ZNothing, SOIn, String] = Handler.fromZIO(ZIO.succeed(""))
private val soRoute: Route[Any, Nothing] = soEndpoint.implementHandler(soHandler)

soRoute.run(Request.get("/so"))
Results in:
 java.lang.StackOverflowError
	at zio.http.Handler$$anon$4.apply(Handler.scala:314)

soRoute.run(Request.get("/so?a=test"))
Results in successfull call
```

**Expected behaviour**
the request should not result in a StackOverflowError

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
3.0.0-RC9

gregor-rayman:
I have played with this a bit. The problem seems to be in the class `Handler.scala` method `foldCauseHandler` (line 308 currently). 

Although the method is not really recursive, it delegates to another handler, and there is plenty of them. In my tests, I was getting stack overflow errors when the query had 13 optional parameters. With 12 it run successfully, but there were over 4000 generated handlers. 

Here a primitive change to highlight the issue:

```
[...]

PR #2980, opened 2024-07-28
Fix for many optional query parameters

gregor-rayman:
Solves #2942


gregor-rayman:
Added alternative encoding of query parameters. Instead of `query[String](query("name")).optional` which doubles the number of handlers, `query[Option[String]](queryOptional("name")` can read the optional query parameter directly without the need of a Fallback handler.

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -287,9 +287,10 @@ private[codec] object EncoderDecoder {
 
         val params = queryParams.queryParamsOrElse(query.name, Nil)
 
-        if (params.isEmpty)
-          throw HttpCodecError.MissingQueryParam(query.name)
-        else {
+//        if (params.isEmpty)
```
jdegoes: This code is commented out. Why?

jdegoes:
Closed in favor of @987Nabil 's solution, which allow's deserializing any number of query parameters into a case class.

gregor-rayman:
Great! His solution is much better

Issue #2953, opened 2024-07-09
Support receiving `ServerSentEvents` in the (Endpoint) client


987Nabil:
Currently users can sent SSE with zio-http but not receive them. We should make this possible. For the Endpoint API as well as for the low level API. This means we need a way to decode an incoming stream of data to an SSE and integrate this in `Response` as well as in the EndpointAPI

kyri-petrou:
Excuse my ignorance, but is client-sent SSE a thing? I thought SSE was meant to be the server sending back events in response to a `text/event-stream` request 🤔 I also can't find any reference to it anywhere

987Nabil:
But where is the server sending the events too? The client. And we need to be able to read them, no? Why should we not have parser logic that generates instances of `ServerSentEvent` in the Endpoint API client?

kyri-petrou:
🤦‍♂️ I read the issue as "the server being able to receive SSE from a client"

jdegoes:
/bounty $400 but has to come quickly due to final release.

PR #2985, opened 2024-07-30
Generic ServerSentEvents (#2952)

### LINKED ISSUE #2952: Generic `ServerSentEvents`
Currently the data of `ServerSentEvents` is a string. But I think we can make it a generic type that needs a `Schema` and use `HttpContentCodec` to encode the data. This should be more user friendly, since users don't need to transform data to strings by themselves.

987Nabil:
fixes #2952
/claim #2952

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
+++ b/zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
@@ -20,7 +20,7 @@ import zio._
 
 import zio.stream.ZPipeline
 
-import zio.schema.codec.{BinaryCodec, DecodeError}
+import zio.schema.codec._
 import zio.schema.{DeriveSchema, Schema}
 
 import zio.http.codec.{BinaryCodecWithSchema, HttpContentCodec}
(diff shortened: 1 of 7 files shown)


zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -64,23 +62,60 @@ final case class ServerSentEvent(
...
+        decoded.map(value => ServerSentEvent(value, eventType, id, retry))
+      }
 
-      override def streamDecoder: ZPipeline[Any, DecodeError, Byte, ServerSentEvent] =
-        throw new UnsupportedOperationException("ServerSentEvent decoding is not yet supported.")
+      override def streamDecoder: ZPipeline[Any, DecodeError, Byte, ServerSentEvent[T]] =
+        ZPipeline.chunks[Byte].map(_.asString) >>> ZPipeline
+          .splitOn("\n\n")
+          .mapZIO(s => {
+            val lines     = s.linesIterator
```
jdegoes, line 91: This logic for encoding and decoding should have some tests associated with it. 👍

987Nabil: @jdegoes This is actually more something for #2953 since receiving SSE is not yet impl. I have a draft PR open for that #2987 that I am working on right now. Are you okay if I add the tests there?

jdegoes (approved):

PR #2987, opened 2024-07-30
ServerSentEvents client support (#2953)

### LINKED ISSUE #2953: Support receiving `ServerSentEvents` in the (Endpoint) client

987Nabil:
fixes #2953
/claim #2953


987Nabil:
Depends on #2985

jdegoes (approved):

PR #2988, opened 2024-07-31
Add a test using `TestServer` and the `Endpoint` API

guizmaii:


jdegoes:
Looks fine to me. The body is a JSON encoded string, which would be `"Hi"`, including the quotes. But if you embed that into a Scala string, it becomes `"\"Hi\""` due to the double quote character being a string delimiter.

guizmaii:
Not a bug. I was missing this:
```scala
.addHeaders(Headers(Header.Accept(MediaType.text.`plain`)))
```
in my request definition

Issue #2945, opened 2024-07-06
Unseal `TextCodec` to allow custom query parameter enc/decoding


anqit:
**Is your feature request related to a problem? Please describe.**
I would like to decode query parameters to custom objects. Currently, we can use functions like `URL#queryParamTo[T]` to decode query params to any type that has a `TextCodec` available, however, unlike other codec related classes (`BinaryCodec`, `Schema`, etc), `TextCodec` is sealed. This means we can only decode query params to a small, pre-determined set of types.

**Describe the solution you'd like**
Maybe there's some limitation I'm unaware of, but I don't see why `TextCodec` has to be sealed. It would be useful if it weren't, so that we could make custom ones to decode query parameters to any custom type. Additionally, it would then be helpful to have methods like `transform` and ilk to be able to generate new `TextCodec` instances from existing ones.

**Describe alternatives you've considered**
Currently, I have to do something like 

```scala
(req: Request) =>
                req.url.queryParamTo[String]("myParam").map(stringToMyCustomTypeConverter)...
```
whereas, with the ability to create custom `TextCodec`s, we could just do:

```scala
(req: Request) =>
                req.url.queryParamTo[MyCustomType]("myParam")...
```

987Nabil:
Unsealing `TextCodec` is not possible. It would break garanties that we rely on in other places.
If you want to use easy custom mapping of query params, please use the Endpoint API.

There was an idea before though, to extract query params to case classes (not by opening `TextCodec`)
@jdegoes do you know what happened to it, I remember someone tried to implement some schema based solution.

anqit:
I see, I've been meaning to check out the Endpoint API anyways, maybe this will be a good impetus to do that.
That said, seems like having variants of the query param extraction functions that use `Schema` instead of `TextCodec` could work

jdegoes:
I don't think the correct solution is to unseal `TextCodec`, but rather, to allow use of any type that has a ZIO Schema.

If the query parameter does not have the structure necessary to encode / decode from a text-based query string parameter, then that would become a runtime failure.

/bounty $250 for a solution and tests

guersam:
@987Nabil I'm not sure it's fully resolved by #2937, because the extracter in `QueryGetters` still require `TextCodec` rather than `QueryCodec`. Is there any way to make them work with `QueryCodec` keeping binary compatible?

anqit:
[...]

PR #2991, opened 2024-07-31
Schema based query codecs including validation (#2937)

### LINKED ISSUE #2937: Endpoint: add query codec validation for max/min length and regex

987Nabil:
fixes #2937
/claim #2937

### DIFF
--- a/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliEndpoint.scala
+++ b/zio-http-cli/src/main/scala/zio/http/endpoint/cli/CliEndpoint.scala
@@ -127,11 +127,8 @@ private[cli] object CliEndpoint {
       case HttpCodec.Path(pathCodec, _) =>
         CliEndpoint(url = HttpOptions.Path(pathCodec) :: List())
 
-      case HttpCodec.Query(name, textCodec, _, _) =>
-        textCodec.asInstanceOf[TextCodec[_]] match {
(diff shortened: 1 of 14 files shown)


987Nabil:
@jdegoes PR was already big enough. But there is some follow up work to do.
1. Integrate the validation information in OpenAPI. I started with this for the bodies already. The PR is blocked by a zio-schema bug. But now queries needs it too. I will open a new issue.
2. I modified the `TextBinaryCodec`, so it works for Records with exactly one field. This is used in query codec, but only works if the user gives the query param name. To improve this, we should add that one can define a case class and the query codec looks for query params that fit all field names `HttpCodec.query[Params]` where `case class Params(perPage: Int, page: Int)` to read two query params

jdegoes:
> To improve this, we should add that one can define a case class and the query codec looks for query params that fit all field names HttpCodec.query[Params] where case class Params(perPage: Int, page: Int) to read two query params

It would be nice to get this one done as part of this PR, if possible.

jdegoes (approved):

987Nabil:
I just realized, this fixes #2945 as well
/claim #2945

anqit:
It'd be great if user-facing changes like this included doc updates as well

987Nabil:
@anqit we will check and update all docs before the final 3.0 release.
Also, there are no API changes.
Take a look at this for API changes including adjusted docs. https://github.com/zio/zio-http/pull/3000/files#diff-af8fd425043c6947f3e6b86e9b2fb5fdefa737968e9b63458fe6ceb8938ce5ce

PR #2994, opened 2024-08-01
Remove warning header

### LINKED ISSUE #2955: Remove internal usage of warning header
The warning header is currently used by zio-http. But is deprecated. I think errors should not add any info about stack traces or error messages in headers at all.
We should have error messages as opt-in in the body.
This should be a setting that can be set globally by a config, as well as by the user for a specific call to `Response.error` and similar methods.

pablf:
fixes #2955
/claim #2955

Removed use of warning header for errors. If an user specifies an error messsage, it's contained in the body. Empty body otherwise.


987Nabil:
@pablf  This does not fulfill the requirements of the issue. We want opt-in errors in the body independent of the source. Like a setting.

pablf:
@987Nabil I thought that this could be controlled when creating a `Handler` to implement a `Server` as this calls `Response.error`. Do you mean putting it additionally in `Server.Config` for example and overriding possible user-defined handler's behaviour?

987Nabil:
No, we call the error methods internally. But users want to control it. On for Dev mode off for prod for example.

pablf:
Understood, thanks!

pablf:
Is this the correct direction?

987Nabil:
Clothing in favor of #3050

Issue #2916, opened 2024-06-17
OpenAPI generation fails with a Map[String, List[String]]


countfloyd:
For a field of type Map[String, List[String]] (or List of anything for that matter), the OpenAPI generator fails

Run this program to reproduce:

```
import zio.http.endpoint.openapi.JsonSchema
import zio.http.endpoint.openapi.JsonSchema.SchemaStyle
import zio.schema.*

import scala.collection.immutable.List

@main def schema(): Unit =
  val schema = Schema.map[String, List[String]]
  val s = JsonSchema.fromZSchemaMulti(schema, SchemaStyle.Reference)
  println(s.toString)
```

The OpenAPI generation should succeed. Instead, get this exception:

```
Exception in thread "main" java.util.NoSuchElementException: None.get
	at scala.None$.get(Option.scala:627)
	at scala.None$.get(Option.scala:626)
	at zio.http.endpoint.openapi.JsonSchema$.fromZSchemaMulti(JsonSchema.scala:378)
	at com.g1.esg1.v10n.web.schema$package$.schema(schema.scala:25)
	at com.g1.esg1.v10n.web.schema.main(schema.scala:23)
```

 - OS: macOS 14.5
 - Version Java 22

Using zio-http 3.0.0-RC8 and Scala 3.4.2

jdegoes:
/bounty $75 for test case and fix.

pawelsadlo:
/attempt #2916

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clz8f6eyx0002l009hn796gsl/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

PR #2996, opened 2024-08-01
Fix fromZSchemaMulti

pawelsadlo:
/claim #2916


jdegoes:
There needs to exist, at minimum, a new failing test case that, when this patch is applied, makes the test case succeed.

pawelsadlo:
@jdegoes added test

987Nabil (approved):

PR #2997, opened 2024-08-02
fix: OpenAPI generation fails with a Map[String, List[String]]

asr2003:


asr2003:
Added failing testcase to check its behaviour https://github.com/zio/zio-http/pull/2997/commits/9cedde059f43a0768bf6468dae42eeb7247b3a7b as intended it passes , it fails after patch:

image

asr2003:
@987Nabil Can you take a look? Added test case ensures the both cases of evaluating patch and the exception case too for future purpose if it indeed fails with any changes

987Nabil:
Already merged a fix

PR #2998, opened 2024-08-02
Simplify Encoderdecoder

### LINKED ISSUE #2958: Simplify `EncoderDecoder`
As shown in #2951 with `HttpContentCodec` we can generalize cases in `EncoderDecoder` that had special handling before.
We should investigate if more such simplifications can be made, since `EncoderDecoder` is an important part of the Endpoint API. A simple version would increase maintainability and readability. Also it would be nice, if we can reduce most use cases to some core functionality. For example, would it be possible to use `HttpContentCodec` for form data?

This is an issue that needs research and careful adjustments.

pablf:
fixes #2958
/claim #2958
Grouped logic in `genericDecode`, `decodeBody`, `genericEncode`, `simpleEncode` and `encodeBody`. Decoding and encoding of bodies are done through `BodyCodec` with different cases based on length. Overall, `EncoderDecoder` simplified from 592 to 420 lines. I don't think it's possible to simplify it further.


987Nabil:
@pablf I fear this convicts with something I am working on. I converted it to draft, since my work is more complex and probably necessary. I'll let you know when this is unblocked.

pablf:
@987Nabil The conflicting part was already merged if I'm not mistaken. Mark it as a draft otherwise.

jdegoes:
@987Nabil I think we should get this in after conflicts fixed.

pablf:
@jdegoes conflicts merged! Unrelated test about server shutdown seemed to fail

987Nabil:
@pablf I restarted the pipeline. If it still fails, please check why. I did not see this test failing before.

pablf:
@987Nabil It passes now, it seems like it might be flaky

987Nabil (approved):

PR #2999, opened 2024-08-02
Efficient query optional encoding (#2942)

### LINKED ISSUE #2937: Endpoint: add query codec validation for max/min length and regex

### LINKED ISSUE #2942: StackOverflowError when too many "optional" query parameters defined for Endpoint

987Nabil:
This PR is based on the changes in #2991.
It encodes the optionality of the query params by adjusting the type hint and the binary codec.
This has multiple benefits.
1. the codecs are not blowing up the http codec alternatives when calculated
2. the more accurate modeling of the types for the query codec (Unit vs single value vs chunk) safes unnecessary chunk allocation when en/decoding
3. the change to BinaryCodecWithSchema allows us, to offer an API that manipulates the schema and generating a new codec from the changes schema. We could use this to add option or validation calls to the codec, instead of forcing the user to use zio-schema annotations on the data type. This would also work with primitive types. There would be no need to wrap them in a case class to annotate the field with validations.

fixes #2942
/claim #2942

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
@@ -18,34 +18,37 @@ package zio.http.codec
 import zio.Chunk
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
+import zio.schema.Schema
+
 import zio.http.codec.HttpCodec.Query.QueryParamHint
+import zio.http.codec.internal.TextBinaryCodec
 
 private[codec] trait QueryCodecs {
 
-  def query(name: String): QueryCodec[String] = singleValueCodec(name, TextCodec.string)
+  def query(name: String): QueryCodec[String] = singleValueCodec(name, Schema[String])
 
-  def queryBool(name: String): QueryCodec[Boolean] = singleValueCodec(name, TextCodec.boolean)
+  def queryBool(name: String): QueryCodec[Boolean] = singleValueCodec(name, Schema[Boolean])
 
-  def queryInt(name: String): QueryCodec[Int] = singleValueCodec(name, TextCodec.int)
+  def queryInt(name: String): QueryCodec[Int] = singleValueCodec(name, Schema[Int])
 
-  def queryTo[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] = singleValueCodec(name, codec)
+  def queryTo[A](name: String)(implicit codec: Schema[A]): QueryCodec[A] = singleValueCodec(name, codec)
 
-  def queryAll(name: String): QueryCodec[Chunk[String]] = multiValueCodec(name, TextCodec.string)
+  def queryAll(name: String): QueryCodec[Chunk[String]] = multiValueCodec(name, Schema[String])
 
-  def queryAllBool(name: String): QueryCodec[Chunk[Boolean]] = multiValueCodec(name, TextCodec.boolean)
+  def queryAllBool(name: String): QueryCodec[Chunk[Boolean]] = multiValueCodec(name, Schema[Boolean])
 
-  def queryAllInt(name: String): QueryCodec[Chunk[Int]] = multiValueCodec(name, TextCodec.int)
+  def queryAllInt(name: String): QueryCodec[Chunk[Int]] = multiValueCodec(name, Schema[Int])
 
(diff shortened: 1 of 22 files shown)


987Nabil:
This needs some migration with changes from other PRs

zio-http/shared/src/main/scala/zio/http/codec/BinaryCodecWithSchema.scala
```diff
@@ -3,9 +3,21 @@ package zio.http.codec
 import zio.schema.Schema
 import zio.schema.codec.BinaryCodec
 
-final case class BinaryCodecWithSchema[A](codec: BinaryCodec[A], schema: Schema[A])
+trait CodecBuilder {
+  def build[A](schema: Schema[A]): BinaryCodec[A]
+}
+
+final case class BinaryCodecWithSchema[A](codecBuilder: CodecBuilder, schema: Schema[A]) {
```
jdegoes: I don't like the structure of this:

1. `CodecBuilder` says, "I can give you a binary codec for any schema"
2. `BinaryCodecWithSchema` says "Here is a schema and a way to get a binary codec for ANY schema"

Why was this change necessary?

zio-http/shared/src/main/scala/zio/http/codec/QueryCodecs.scala
```diff
@@ -17,35 +17,31 @@
...
+  def query(name: String): QueryCodec[String] = singleValueCodec(name, Schema[String])
 
-  def queryBool(name: String): QueryCodec[Boolean] = singleValueCodec(name, TextCodec.boolean)
+  def queryBool(name: String): QueryCodec[Boolean] = singleValueCodec(name, Schema[Boolean])
 
-  def queryInt(name: String): QueryCodec[Int] = singleValueCodec(name, TextCodec.int)
+  def queryInt(name: String): QueryCodec[Int] = singleValueCodec(name, Schema[Int])
 
-  def queryTo[A](name: String)(implicit codec: TextCodec[A]): QueryCodec[A] = singleValueCodec(name, codec)
+  def queryTo[A](name: String)(implicit codec: Schema[A]): QueryCodec[A] = singleValueCodec(name, codec)
```
jdegoes, line 34: Thinking about this more, I believe that if we have a schema, we should get the name from the field(s) inside the Schema. Rather than accepting it explicitly. So no `query*` method that takes a `Schema[A]` should also take a `String` name. We can get the arity from the schema too (if it's a sequence rather than a single one; NonEmptyChunk can take care of non-empty sequences). This also provides a path to deserializing multiple fields at once, which will solve the stack overflow problems.

jdegoes (changes requested):
Definitely the right direction, but a few enhancements and modifications will help push this over the finish line.

987Nabil:
closing in favor of #3000

PR #3000, opened 2024-08-03
Efficient query optional encoding (#2942)

### LINKED ISSUE #2937: Endpoint: add query codec validation for max/min length and regex

### LINKED ISSUE #2942: StackOverflowError when too many "optional" query parameters defined for Endpoint

### LINKED ISSUE #2992: Multi field case classes as input for query codec
In https://github.com/zio/zio-http/pull/2991 I modified the TextBinaryCodec, so it works for Records with exactly one field. This is used in query codec, but only works if the user gives the query param name. To improve this, we should add that one can define a case class and the query codec looks for query params that fit all field names. 

The goal is to be able to define `HttpCodec.query[Params]` where `case class Params(perPage: Int, page: Int)` to read two query params named `perPage` and `page`, for example.

987Nabil:
More efficient optional query param encoding and query codes based on case class support

fixes #2942
fixes #2992
/claim #2942
/claim #2992

### DIFF
--- a/docs/reference/endpoint.md
+++ b/docs/reference/endpoint.md
@@ -25,7 +25,7 @@ object Book {
 
 val endpoint =
   Endpoint(RoutePattern.GET / "books")
-    .query(QueryCodec.queryTo[String]("q") examples (("example1", "scala"), ("example2", "zio")))
+    .query(HttpCodec.query[String]("q") examples (("example1", "scala"), ("example2", "zio")))
     .out[List[Book]]
 ```
 
@@ -109,7 +109,7 @@ import zio.http.codec._
 ```scala mdoc:compile-only
 val endpoint: Endpoint[Unit, String, ZNothing, ZNothing, AuthType.None] =
   Endpoint(RoutePattern.GET / "books")
-    .query(QueryCodec.queryTo[String]("q"))
+    .query(HttpCodec.query[String]("q"))
 ```
 
 QueryCodecs are composable, so we can combine multiple query parameters:
@@ -117,7 +117,7 @@ QueryCodecs are composable, so we can combine multiple query parameters:
 ```scala mdoc:compile-only
 val endpoint: Endpoint[Unit, (String, Int), ZNothing, ZNothing, AuthType.None] =
   Endpoint(RoutePattern.GET / "books")
-    .query(QueryCodec.queryTo[String]("q") ++ QueryCodec.queryTo[Int]("limit"))
+    .query(HttpCodec.query[String]("q") ++ HttpCodec.query[Int]("limit"))
 ```
 
 Or we can use the `query` method multiple times:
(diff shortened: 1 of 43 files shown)


jdegoes:
@987Nabil 

As you're working on this, I'd suggest the following API:

```scala
// Parses a single named field with the specified schema:
val limit: QueryCodec[Int] = HttpCodec.query[Int]("limit")

// Parses a single named optional field with the specified schema:
val optionalLimit: QueryCodec[Option[Int]] = HttpCodec.query[Option[Int]]("limit")

// Parses a sequence of named optional fields with the specified schema:
val seqLimit: QueryCodec[List[Int]] = HttpCodec.query[List[Int]]("limit")

// Parses all named fields with whatever types they have:
case class QueryParams(limit: Option[Int], skip: Option[Int], order: Option[SortOrder], language: String = "EN")
val params: QueryCodec[QueryParams] = HttpCodec.queryAll[QueryParams]
```

Note the distinction between `query`, which takes the name of a field, and `queryAll`, which does not take a name, but rather, derives names from the schema.

Note further that in a future improvement, even `query[QueryParams]("params")` could be supported for composite schema through namespacing, e.g. `?params.limit=12&params.skip=123&params.order=ASC` (however, only primitive types need be supported for now).

jdegoes:
@987Nabil Can you please fix conflicts? 🙏

jdegoes (approved):

PR #3002, opened 2024-08-04
[gen] newtype aliases

### LINKED ISSUE #2962: [gen] add support for type aliases (newtype / opaque type) when generating code from spec

In business logic code we want to use Newtype to make sure we don't mix up different fields of the same type.
Generated code is great, but all primitives can still be mixed up.
This is especially true when we don't have a "stable source",
e.g.  when an `openapi.yaml` spec is retrieved during build from a live service, which can change.

consider a simple example like:
```yaml
components:
  schemas:
    Person:
      type: object
      properties:
        firstName:
          type: string
        lastName:
          type: string
```
generated class would be:
```scala
case class Person(
  firstName: String,
  lastName: String
)
```
and example code usage might be:
```scala
val johnDoe = Person("John", "Doe")
```
but a simple change in the order of fields in yaml, or a field rename may result in the generated class to look different, e.g:
```scala
case class Person(
  surname: String,
  firstName: String
)
```
which will cause our code to still compile, but be wrong:
```scala
val johnDoe = Person("John", "Doe")
```
code didn't change, but now "John" is the surname, and "Doe" is the first name.

Ideally, we control the OpenAPI spec, and can define it "safer", in a way that enables generated code to use safe type aliases.
e.g. by defining aliases directly, and using `ref`s , e.g:
```yaml
components:
  schemas:
    FirstName:
      type: string
    LastName:
      type: string
    Person:
      type: object
      properties:
        firstName:
          $ref: '#/components/schemas/FirstName'
        lastName:
          $ref: '#/components/schemas/LastName'
```
such definition can be picked up by zio-http-gen to render `FirstName` & `LastName` as e.g. prelude Newtypes:
```scala
// file FirstName.scala
object FirstName extends Newtype[String] {
  given schema: ZSchema[FirstName] = derive[ZSchema]
}
type FirstName = FirstName.Type
```
```scala
// file LastName.scala
object LastName extends Newtype[String] {
  given schema: ZSchema[LastName] = derive[ZSchema]
}
type LastName = LastName.Type
```
```scala
// file Person.scala
case class Person(
  firstName: FirstName,
  lastName: LastName
)
```

Alternatively, we can consider some `x-codegen` annotations as described [here](https://openapispec.com/docs/documentation-and-annotation-in-openapi/#annotations-for-code-generation-and-integration).
e.g:
```yaml
components:
  schemas:
    Person:
      type: object
      properties:
        firstName:
          type: string
          x-codegen-type-alias: FirstName
        lastName:
          type: string
          x-codegen-type-alias: LastName
```
to achieve a similar goal.

What's currently possible, if we're being very meticulous, we can still avoid the issue of silent breakage if we keep a "manual" safe copy of the classes with safe Newtype aliases. We then can use `Schema.migrate` to help us transform all safe instances to the generated ones, and vice versa.
While this works (if we are careful to always use named fields - thus code won't compile when changes like the one mentioned above are introduced), it is still inferior to an ideal solution when we generate safe code to begin with.
- for every class we'll have 2 almost identical copies
- we pay the transform overhead from DTO to/from internal domain class on every call.

**Additional context**

Using a made up annotation like `x-codegen-type-alias` would work, but is "local" to the component using it.
If we have several components that "need" the same semantics over a primitive, it makes sense to make it "global".
Relying on "same named type" might be ambiguous.
Consider a type alias like `ID`. It can mean a person ID number, or order ID, etc'…
using a `ref` makes it clear different components use the SAME semantics of the field.
Perhaps an annotation like `x-codegen-type-alias` can be considered in the future for non-opaque type aliases?

Perhaps the feature would be configurable.
I think that by default primitives should generate Newtype aliases.
The configuration can control whether to generate prelude Newtypes, regular scala 3 opaque types, opt-out aliasing altogether to render the primitive type directly, or anything else we might want to enable (e.g. 3rd party support for libs like neotype)

hochgi:
basic implementation.
Aliased types is only supported for primitives in this PR.

fixes #2962 
/claim #2962

TODO:
- [x] make it configurable
- [x] when opt-out, replace alias with original primitive name
- [x] add more tests to cover most common scenarios:
  * [x] collections of aliased newtypes
  * [x] collections of un-aliased primitives
  * [x] path aliased newtype params
  * [x] path un-aliased primitive params 
  * [x] query aliased newtype params
  * [x] query un-aliased primitive params 
- [x] documentation

### DIFF
--- a/zio-http-cli/src/test/scala/zio/http/endpoint/cli/EndpointGen.scala
+++ b/zio-http-cli/src/test/scala/zio/http/endpoint/cli/EndpointGen.scala
@@ -3,7 +3,7 @@ package zio.http.endpoint.cli
 import zio.ZNothing
 import zio.test._
 
-import zio.schema.{Schema, StandardType}
+import zio.schema.Schema
 
 import zio.http._
 import zio.http.codec.HttpCodec.Query.QueryParamHint
--- a/zio-http-gen/src/main/scala/zio/http/gen/grpc/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/grpc/EndpointGen.scala
@@ -43,6 +43,7 @@ object EndpointGen {
     }
     val obj       = Code.Object(
       name = name.capitalize,
+      extensions = Nil,
       schema = false,
       endpoints = endpoints.toMap,
       objects = Nil,
--- a/zio-http-gen/src/main/scala/zio/http/gen/grpc/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/grpc/EndpointGen.scala
@@ -2,7 +2,6 @@ package zio.http.gen.grpc
 
 import zio.http.Method
 import zio.http.gen.scala.Code
-import zio.http.gen.scala.Code.ScalaType
 
 object EndpointGen {
 
@@ -44,7 +43,7 @@ object EndpointGen {
     val obj       = Code.Object(
       name = name.capitalize,
       extensions = Nil,
-      schema = false,
+      schema = None,
       endpoints = endpoints.toMap,
       objects = Nil,
       caseClasses = Nil,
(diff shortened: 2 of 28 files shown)


jdegoes (approved):

jdegoes:
Looks good to me, may want to get @987Nabil to take a quick look too.

987Nabil:
Will do tomorrow. Already too sleepy 😴

987Nabil (approved):

Issue #2908, opened 2024-06-14
Add a hybrid `RequestStreaming` option that streams requests only if they're larger than X bytes


kyri-petrou:
**Is your feature request related to a problem? Please describe.**
Currently, `zio-http` provides 2 config options for `RequestStreaming`:
1. `Disabled` (with a maximum number of bytes that the request can be before rejecting it)
2. `Enabled` (requests are streamed regardless of size)

For servers with low latency / high performance requirements, disabling streaming is a favourable option as it's more performant than streaming the requests. However, in some cases we might want to be able to be able to handle very large requests, so it would be good if `zio-http` provided an option to aggregate the response if it's less than X bytes, and stream it if it's larger than that

**Describe the solution you'd like**
Another case class extending `RequestStreaming` similar to `Disabled` but instead of failing the request it streams the bytes instead

**Describe alternatives you've considered**
Enabling streaming and using a middleware to batch the `ZStream`, but this doesn't perform as well as I would have hoped

jdegoes:
/bounty $250 but has to ship soon because of final release coming up.

abhi267266:
/attempt #2908

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clz8f2qzr0002jt097l3hiqsk/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

Ahmadkhan02:
/attempt #2908

<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/Ahmadkhan02) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @Ahmadkhan02 | 1 bounty from 1 project | <div align="center">TypeScript, Jupyter Notebook</div> | <div align="center"></div> | [Cancel attempt](https://console.algora.io/api/bounties/clz8f2qzr0002jt097l3hiqsk/cancel-attempt) |

PR #3003, opened 2024-08-04
Added HybridContentLengthHandler 

Ahmadkhan02:
/claim #2908


zio-http/jvm/src/main/scala/zio/http/netty/server/ServerChannelInitializer.scala
```diff
@@ -86,6 +87,16 @@ private[zio] final case class ServerChannelInitializer(
       case RequestStreaming.Enabled                        =>
       case RequestStreaming.Disabled(maximumContentLength) =>
         pipeline.addLast(Names.HttpObjectAggregator, new HttpObjectAggregator(maximumContentLength))
+      case RequestStreaming.Hybrid(maxAggregatedLength)    =>
+        pipeline.addLast(
+          Names.HttpObjectAggregator,
+          new HttpObjectAggregator(maxAggregatedLength) {
+            override def handleOversizedMessage(ctx: ChannelHandlerContext, oversized: HttpMessage): Unit = {
+              ctx.pipeline().replace(this, "chunkedWriter", new ChunkedWriteHandler())
```
kyri-petrou: Won't this change the pipeline for all persisted connections to the server that are served through this channel?

Ahmadkhan02: Ohh right, updating this

zio-http/jvm/src/test/scala/zio/http/HybridStreamingSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+import zio._
+import zio.test.Assertion.equalTo
+import zio.test.TestAspect.{diagnose, sequential, shrinks, withLiveClock}
+import zio.test.{assertTrue, assertZIO}
+
+import zio.http.ServerSpec.requestBodySpec
+import zio.http.internal.{DynamicServer, HttpRunnableSpec}
+import zio.http.netty.NettyConfig
+
+object HybridRequestStreamingServerSpec extends HttpRunnableSpec {
```
kyri-petrou, line 27: I think we need a few extra tests / tweeks in here:

1. We need to ensure that the pipeline is properly reset (and functional) after we changed the handler. We need to make multiple requests within the same test that reuses the same underlying connection, with bodies smaller and larger than the limit.
2. You need to use `body.isComplete` to check if the request body is streaming or not instead of `body.asStream`. For cases that the `size < maxSize` it should be `true`, or `false` when its larger than `maxSize`.

Ahmadkhan02:
hi @kyri-petrou fixed the code by adding a new HybridContentLengthhander to switch between ChunkedWriter and HttpObjectAggregator as per whats needed. Also added the tests as you suggested. Removed the previous one as those were redundant now

narma:
I think we need make this new mode default for default Config

kyri-petrou:
> I think we need make this new mode default for default Config

I think that's a great idea. @987Nabil @jdegoes WDUT?

EDIT: Actually, there is a downside to it; It can potentially be exploited if the user is not explicitly handling (oversized) streaming requests; which I'm guessing no one that uses the default will be doing.

zio-http/jvm/src/test/scala/zio/http/HybridStreamingSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+      suite("app with hybrid request streaming") {
+        appWithHybridReqStreaming.as(List(requestBodySpec, hybridStreamingServerSpec))
+      }
+    }.provideSome[DynamicServer & Server & Client](Scope.default)
+      .provideShared(
+        DynamicServer.live,
+        ZLayer.succeed(configAppWithHybridRequestStreaming),
+        Server.customized,
+        ZLayer.succeed(NettyConfig.defaultWithFastShutdown),
+        Client.default,
```
kyri-petrou: Use a connection pool with a size of 1 here to ensure that the tests are run using a single connection to the server:
```suggestion
        Client.live,
        ZLayer.succeed(Client.Config.default.fixedConnectionPool(1)),
        DnsResolver.default,
```

kyri-petrou (commented):
Sorry it's taken me so long to review this, been quite busy the past few weeks. Just a minor comment below

Ahmadkhan02:
HybridStreamingSpec Test should be fine now. Sorry didnt keep this issue in mind earlier. Build test is failing though. Not due to this PR I believe. Its ConnectionPoolIssuesSpec thats causing error. Looking into this

kyri-petrou:
> Build test is failing though. Not due to this PR I believe. Its ConnectionPoolIssuesSpec thats causing error

Yeah that's a separate issue. Test appears to be flaky for some reason; I'll look into that separately.

@Ahmadkhan02 can you please sign the CLA? It seems that it hasn't been signed yet 
image

Ahmadkhan02:
> > Build test is failing though. Not due to this PR I believe. Its ConnectionPoolIssuesSpec thats causing error
> 
> Yeah that's a separate issue. Test appears to be flaky for some reason; I'll look into that separately.
> 
> @Ahmadkhan02 can you please sign the CLA? It seems that it hasn't been signed yet image

Ohh. Signed the CLA

PR #2838, opened 2024-05-12
Added Comprehensive documentation for ZIO HTTP

varshith257:
- Fixes #2198 
/claim #2198
-  Introduction 
-  Concept
- [x] Routing 
- [x] Request Handling 
- [x] Server 
- [x] Client 
- [x] Middleware 
- [x] Endpoint 
-  Tutorials
- [x] Your First ZIO HTTP App 
- [x] Deploying a ZIO HTTP App
- [x] Testing Your ZIO HTTP App

- [x] How-to-guides
(This should be based off the list of topics at http4k, but should also include ZIO HTTP use cases around Endpoint, middleware, etc.) 

> [!NOTE]  
~~sidebar.js will be updated soon with added new files~~

PR #3006, opened 2024-08-06
fix: uuid PathCodec is broken in snapshot versions

### LINKED ISSUE #3005: uuid PathCodec is broken in snapshot versions
My handler with definition
`Method.GET / "api" / uuid("entityId")` 
is stop matches and requests against `GET /api/$UUID` returns 404 instead of response

Reproducer, works with 3.0.0-RC9, but broken in latest snaphots.

```scala
//> using repositories sonatype:snapshots
//> using dep "dev.zio::zio-http:3.0.0-RC9+25-5aa5d9ec-SNAPSHOT"
/// 3.0.0-RC9 works
//>> using dep "dev.zio::zio-http:3.0.0-RC9"

import java.util.UUID
import zio.Chunk
import zio.http.codec.SegmentCodec

val c = SegmentCodec.uuid("p")
val v = UUID.randomUUID().toString()
val result = c.matches(Chunk("api", v), 1)
println(s"result = $result")
if (result == 1) then println("good") else println("not matches")
```

varshith257:
/claim #3005
fixes #3005 

The issue stemmed from the recent changes to the UUID `SegmentCodec` implementation. Specifically, the matches method was modified to use the inSegmentUntil method which introduced an error in the way UUID segments were validated causing valid UUID segments to be mismatched.

I have simplified the `matches` method by using `java.util.UUID.fromString` directly to validate UUID strings. This approach ensures that the entire segment is validated as a proper UUID and handles segment boundaries correctly. This change ensures that only valid UUID strings will match, resolving the issue and restoring the correct behavior and the endpoint now correctly matches UUID segments.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/SegmentCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/SegmentCodec.scala
@@ -439,9 +439,12 @@ object SegmentCodec          {
     def matches(segments: Chunk[String], index: Int): Int = {
       if (index < 0 || index >= segments.length) -1
       else {
-        val lastIndex = inSegmentUntil(segments(index), 0)
-        if (lastIndex == -1 || lastIndex + 1 != segments(index).length) -1
-        else 1
+        try {
+          java.util.UUID.fromString(segments(index))
+          1 // Valid UUID
+        } catch {
+          case _: IllegalArgumentException => -1 // Not a valid UUID
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/codec/SegmentCodec.scala
```diff
@@ -439,9 +439,12 @@ object SegmentCodec          {
     def matches(segments: Chunk[String], index: Int): Int = {
       if (index < 0 || index >= segments.length) -1
       else {
-        val lastIndex = inSegmentUntil(segments(index), 0)
-        if (lastIndex == -1 || lastIndex + 1 != segments(index).length) -1
-        else 1
+        try {
+          java.util.UUID.fromString(segments(index))
```
kyri-petrou: This is extremely inefficient. We should avoid throwing an exception that fills the stack trace.

The format of UUIDs is well known, you should instead manually validate that the string is a UUID or not

987Nabil:
Fix already merged

varshith257:
Okay 😄

@987Nabil Can you review this PR? It's been a while it's opened https://github.com/zio/zio-http/pull/2838

PR #3008, opened 2024-08-07
Add a `batched` API to `ZClient`

### LINKED ISSUE #2956: ZClient memory leaks finalizers on each request
**Describe the bug**
We are using ZIO HTTP in production and have encountered an issue with a constantly increasing amount of memory consumed by the service. At some point, the service is unable to free up memory and subsequently restarts.
Screenshot 2024-07-12 at 13 45 44

We investigated this issue by taking a memory dump and inspecting it using Eclipse Memory Analyzer. Here is the result:
Screenshot 2024-07-12 at 13 51 04

We checked the `zio.Scope` class which has a `Running` state with `finalizers: LongMap[Finalizer]`. We set a breakpoint at `zio.Scope:156` within the `addFinalizerExit` method and discovered that each `ZClient` request adds a new finalizer but does not remove it. The longer the application runs, the more finalizers are added.
Screenshot 2024-07-12 at 13 41 27

You can download [this memory dump](https://drive.google.com/file/d/17QLbG5--zrjnc7f0Tui_HU_wlYELHOTe) for investigation. 

**To Reproduce**
1. Clone this demo project to reproduce the problem: https://github.com/ntngel1/zio-http-zclient-leak
2. Set a breakpoint at `zio.Scope:156` in the `addFinalizerExit` method.
3. Run the program.
4. Let it run for 10 seconds (it makes 10 requests). Attach a debugger and you will see that there are 10 finalizers added and none removed. The more requests made, the more finalizers are added, leading to increased memory consumption.

**Expected behaviour**
Finalizers should be removed after they are no longer needed to prevent memory leaks.

### LINKED ISSUE #2957: Improve ZClient UX with regards to Scope
Currently the Client requires a scope to close some kinds of bodies. The user has to wrap his code in `ZIO.scoped` to avoid memory leaks (which is also not documented well).
From a users perspective, this is not needed most of the time. He probably just wants to handle the body directly and clean up the request.
We should have a simpler API that does not require an explicit scope. We can let the user handover a function, that handles the response and after its execution we just clean up for the user.

This should not replace the current more powerful API, but should be promoted as the default in documentation.
For the current docs, we need to move away from using the global scope and should change most examples to this new API.

cc @kyri-petrou

kyri-petrou:
/fixes #2956
/fixes #2957
/claim #2956
/claim #2957

In a nutshell, this PR adds the following methods and updates docs/examples/tests to change _most_  usages of `Client.request` to these new methods. The first 3 will materialize the body in memory, whereas the latter will embed the Scope into the ZStream, which will be closed when the stream ends.

```scala
object Client {
  def quick(request: Request): RIO[Client, Response]

  def quickWith[A](request: Request)(f: Response => A): RIO[Client, A]

  def quickWithZIO[R, A](request: Request)(f: Response => RIO[R, A]): RIO[Client & R, A]

  def quickWithStream[R, A](request: Request)(f: Response => ZStream[R, Throwable, A]): ZStream[Client & R, Throwable, A]
}
```

@987Nabil @jdegoes, I'm _REALLY_ tempted to also change the `get`, `post`, `put`, `patch`, `delete` and `head` methods on `Client` to also materialize the body in memory. The issue with this though is that if there are users using these methods to stream responses, then we'll be messing things up for them.

Wdut? Should we change those methods as well and document it very clearly in the release notes, or leave them as is and hope for the best that users are handling the `Scope` properly?


987Nabil:
@kyri-petrou then you would add explicit streaming methods?

kyri-petrou:
> @kyri-petrou then you would add explicit streaming methods?

@987Nabil You mean adding add methods for `getStreaming`, `putStreaming` etc?

987Nabil:
Yes. Make the `quick` the default and add streaming variants. Do you think it makes sense? But if we keep the `quick` I'd try to find another name.

kyri-petrou:
@987Nabil I ended up going down a little different path in the end. Instead of re-implementing all of the methods to have variants that don't require a `Scope`, instead I added the `simple` method that exposes the HTTP request methods of the Client but without the Scope. Example usages:

```scala
client.simple.get("/foo")
client.simple(Request.get(???))
client.simple.request(Request.get(???))
```

987Nabil:
@kyri-petrou I thought this is not working for streaming bodies, when we don't handover a callback or?

kyri-petrou:
@987Nabil it is working for streaming bodies. What happens with streaming bodies is that the request body will be materialized in memory. Basically these two are equivalent when we have a streaming body:

```scala
client.simple.get("/foo").flatMap(_.body.asString)

ZIO.scoped(client.get("/foo").flatMap(_.body.asString))
```

The only usage where something is different is this:

```scala
ZStream.unwrap(client.simple.get("/foo").map(_.body.asStream))

ZStream.unwrapScoped(client.get("/foo").map(_.body.asStream))
```

In this case, the resulting stream from the `simple` API will not actually stream data as it arrives, it is just a `Chunk[Byte]` already materialized into memory lifted into a `ZStream`

987Nabil:
@kyri-petrou idk if this is not a problem. You don't necessarily know, if the server returns a chunked stream. Or if it will fit in memory. I see the bug report already coming 🙈

kyri-petrou:
> @kyri-petrou idk if this is not a problem. You don't necessarily know, if the server returns a chunked stream. Or if it will fit in memory. I see the bug report already coming 🙈

But this would still be a problem if they don't explicitly stream the response, even with the current API. And it's only a problem if a user decides to use the `simple` API, the "standard" one behaves the same way as before. There is also a `stream` method on Client now, so users can use that one if they want to stream the response and not worry about the `Scope`.

The only way to make this can be made more safe is for the `simple` API to return a supertype of Response where the body cannot be streamed.

I documented the caveat of the `simple` API both in the documentation and in the scaladoc. I think that if after documenting it very clearly users still end up using it to stream responses, experience said bug, again not read the scaladoc, and end up filing an issue then at this point it's a bit up to them.

987Nabil:
@kyri-petrou I am not sure about simple as a name I'll give it some thought

jdegoes:
@kyri-petrou Here's an idea I just want to float for purposes of discussion -- I am not 100% convinced, but it has some promising elements.

Let's say `Client` has a type parameter that represents "environmental dependencies added during request execution". Then `Client[Any]` becomes the type of a client that adds no environmental dependencies during request execution. However, `Client[Scope]` becomes the type of a client that adds a dependency on `Scope` during request execution.

Then a streaming client must clearly be `Client[Scope]`, but a batched client can be `Client[Any]`, and, moreover, there can exist a single unary operator on `Client` that can batch up streaming responses, allowing taking a `Client[Scope]` into a `Client[Any]` (`.batched`?). Finally, choice of default constructors should be re-evaluated, and possibly, batched is the default, with streaming being opt-in for those use cases where it makes sense.

What have we achieved in this design?

 - No more methods nor duplication of any methods
 - Safe (loss-less) default behavior, with streaming as an opt-in
 - Identical API to what we have today, with no need to learn anything new
 - Ability to abstract over alternate client implementations that might impose other requirements on a per request basis, e.g. `Auth` (not that such would be part of this PR)

jdegoes:
One more comment, only partially related to this PR, and mostly just for purposes of future PRs: essentially, for purposes of best-practice architecture, I think all functionality should be directly exposed on trait/class. This means no functionality can be implemented only in terms of `R` (such as deprecated accessor pattern). Because as convenient as that is, sometimes, it encourages people to go directly to such "accessors", then makes it difficult to do performant and clean layer composition (in a pure layer architector, `R` is used only for scope, or other locally eliminable context; never for dependency management).

kyri-petrou:
@jdegoes I reimplemented the client so that now there is an `S >: Scope` bound on it, and when it's `Any` the `batched` API erases the `Scope` dependency. The `stream` and `socket` methods are only available when `S =:= Scope` to make it more type-safe.

In line with your last comment, I'm considering is to remove the `-Env` bound on the Client altogether (also from `BodyEncoder`, `BodyDecoder` and `Driver`. However I didn't do this as I wanted to first see if you agree with this approach and hear your thoughts

987Nabil:
@kyri-petrou I am not sure, if we should remove env. Right now, it is buried deep in the lib and average joe would never touch it. But I could imagine, that it could enable some future features. Since the user facing API is only the returned env that by default is always Any, I'd tend to leaving it in.

987Nabil (approved):

zio-http/shared/src/main/scala/zio/http/ZClient.scala
```diff
@@ -19,25 +19,27 @@ package zio.http
...
+import zio.stacktracer.TracingImplicits.disableAutoTrace
+
+import zio.stream.ZStream
 
 import zio.http.URL.Location
 import zio.http.internal._
-//import zio.http.netty.client._
 
-final case class ZClient[-Env, -In, +Err, +Out](
+final case class ZClient[-Env, S >: Scope, -In, +Err, +Out](
```
jdegoes: 1. In general, I am not a fan of constraining type parameters at the declaration site, unless absolutely strictly required. The constraints add a lot of work for no real benefit and also interfere with evolution.
2. Because this `S` is added to each request response (covariant), and there it appears in contravariant position, we should be able to make the parameter contravariant, unless I'm missing something.

This would change the signature like so:

```suggestion
final case class ZClient[-Env, -S, -In, +Err, +Out](
```

In which case I might rename:

```suggestion
final case class ZClient[-Env, -ReqEnv, -In, +Err, +Out](
```

"The env added to requests."

Or something similar (`AddEnv`).

docs/reference/client.md
```diff
@@ -34,9 +34,28 @@ object Client {
...
 
+### "Batched" API
+
+The handling of `Scope` can quickly become cumbersome in cases where we simply want to execute an HTTP request and not handle the lifetime of the HTTP request.
+For this reason, the Client can be converted to a batching client where the HTTP request methods don't have Scope in the environment via the `batched` method.
+It is also possible to execute batched requests
+
+```scala
+object Client {
+  def batched(request: Request): ZIO[Client, Throwable, Response] = ???
```
jdegoes: Thinking more about this (good) naming convention: could it make sense to rename `Client.request` to `Client.streaming`?

jdegoes (changes requested):
Just a couple comments, but looks great! Beautiful API.

zio-http/shared/src/main/scala/zio/http/ZClient.scala
kyri-petrou: > I am not a fan of constraining type parameters at the declaration site

Main reason that I went with the constraining type is that removing it doesn't affect binary compatibility, so I thought perhaps it's better to be more strict and remove it if needed. But I agree I don't think that it adds much benefit so I'll remove it 

> we should be able to make the parameter contravariant, unless I'm missing something

@jdegoes the problem with making it contravariant is that then this becomes possible:

```scala
  val batchedClient: ZClient[Any, Any, Body, Throwable, Response] = ???
  val streamingClient: ZClient[Any, Scope, Body, Throwable, Response] = batchedClient
```

This is a problem because once we've eliminated the `Scope` the client cannot stream thereafter. The only way to enforce it AFAICT is via an invariant type param

docs/reference/client.md
kyri-petrou: Since this is a widely used method, I ended up keeping `Client.request` and annotated it deprecated so that we can point users to the new methods without breaking their code.

kyri-petrou:
@jdegoes I applied all your recommendations minus making the `ReqEnv` contravariant (see https://github.com/zio/zio-http/pull/3008#discussion_r1730254534).

I also updated the docs to first introduce the "streaming" client (i.e., default) followed up by the "batched" client/methods. I hope this will be clearer for users to follow

docs/reference/client.md
987Nabil: I am not sure about the deprecated. Then we have to keep this method in until 4.x
I think we should remove it and write migration docs

jdegoes (approved):

PR #3011, opened 2024-08-11
Cache parsing of content-type header

kyri-petrou:
#2969 unknowingly introduced a ~10% regression in performance for simple requests when the content-type header is present. This regression was introduced because parsing of the content-type header is a fairly expensive (~1us to parse `application/json` based on some basic benchmarking), and to make things worse, it is executed within Netty's event loop which should not be performing CPU-intensive work.

On one hand, we could simply revert the PR or parse the content-type header only when needed, but I doubt this is the first time someone will accidentally do this. In order to guard against such usages in the future, this PR adds a cache infront of the content-type header parser. Since this can be potentially, we guard against this by clearing the cache if it grows too big. This size (8192 entries) should never be reached for standard usages

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Header.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Header.scala
@@ -2557,7 +2557,7 @@ object Header {
     def parse(s: String): Either[String, ContentType] = {
       // Guard against malicious registering of invalid content types
       if (cache.size >= CacheMaxSize) cache.clear()
-      cache.computeIfAbsent(s.toLowerCase, parseFn)
+      cache.computeIfAbsent(s, parseFn)
     }
 
     def render(contentType: ContentType): String = codec.encode(contentType).toOption.get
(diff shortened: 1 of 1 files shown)


zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -2543,12 +2544,21 @@ object Header {
   }
 
   object ContentType extends HeaderType {
+    private final val CacheInitialSize = 256
```
987Nabil: Is this not already big? Most servers will probably see 10 content types max I would assume.

kyri-petrou: We can lower it, although I don't think there is much benefit in making it smaller memory-wise since this is initialized just once. Plus HashMaps with a larger capacity have a better read performance due to a better hash key distribution

kyri-petrou: I'll lower it to 64 though cause 256 does seem a bit extensive

987Nabil (approved):

PR #3017, opened 2024-08-14
Simplify the interface of `URL.decode` by only returning one possible `Exception`, instead of all possible `Exception`

guizmaii:


zio-http/shared/src/main/scala/zio/http/Request.scala
```diff
@@ -222,7 +222,7 @@ object Request {
    */
   private def pathOrUrl(path: String): URL =
     if (path.startsWith("http://") || path.startsWith("https://")) {
-      URL.decode(path).toOption.getOrElse(URL(Path(path)))
```
guizmaii, line 226: No need for `.toOption` here

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -281,24 +282,31 @@ final case class URL(
...
         case Some(value) => Right(value)
       }
-
     } catch {
-      case e: Exception => Left(e)
+      case NonFatal(e) => invalidURL(Some(e))
     }
   }
 
-  def config: Config[URL] = Config.string.mapAttempt(decode(_).toTry.get)
```
guizmaii, line 301: No need to instantiate a `Try` here

zio-http/shared/src/main/scala-2/zio/http/UrlInterpolator.scala
```diff
@@ -32,7 +32,7 @@ private[http] object UrlInterpolatorMacro {
     c.prefix.tree match {
       case Apply(_, List(Apply(_, Literal(Constant(p: String)) :: Nil))) =>
         val result = URL.decode(p) match {
-          case Left(error) => c.abort(c.enclosingPosition, s"Invalid URL: ${error.getMessage}")
```
guizmaii, line 35: The error message already contains `Invalid URL:`

987Nabil:
@guizmaii just FYI: using `url"http://my-awsome-service"` whenever possible is my recommendation ;)

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -281,24 +282,31 @@ final case class URL(
...
   def empty: URL = URL(Path.empty)
 
-  def decode(string: String): Either[Exception, URL] = {
-    def invalidURL(string: String) = Left(new MalformedURLException(s"""Invalid URL: "$string""""))
+  /**
+   * Proper implementation of [[java.net.MalformedURLException]] which, unlike
+   * the Java version, can propagate the cause.
+   */
+  final case class MalformedURLException(message: String, cause: Option[Throwable])
+      extends IOException(message, cause.orNull)
```
kyri-petrou: I think we can change this so that we avoid building the string for cases that we discard the error:

```scala
  final case class MalformedURLException(invalidUrl: String, cause: Option[Throwable]) extends IOException(cause.orNull) {
    override def getMessage: String = s"""Invalid URL: "$invalidUrl""""
  }
```

kyri-petrou: Also, should we change `URL.empty` above to a `val` instead of a `def`?

zio-http/shared/src/main/scala-2/zio/http/UrlInterpolator.scala
```diff
@@ -68,9 +68,8 @@ private[http] object UrlInterpolatorMacro {
         val exampleParts = staticParts.zipAll(injectedPartExamples, "", "").flatMap { case (a, b) => List(a, b) }
         val example      = exampleParts.mkString
         URL.decode(example) match {
-          case Left(error) =>
-            c.abort(c.enclosingPosition, s"Invalid URL: ${error.getMessage}")
-          case Right(url)  =>
+          case Left(error) => c.abort(c.enclosingPosition, error.getMessage)
+          case Right(_)    =>
```
kyri-petrou, line 72: Thank you, the compiler warning from this was annoying 😄

guizmaii: It was! 😄

guizmaii:
@987Nabil 
> just FYI: using url"http://my-awsome-service" whenever possible is my recommendation ;)

What's the difference?

The thing is that I receive an `url:String` from the user of my HTTP API, that's why I use `URL.decode(url)`.
Should I prefer `url"$url"`? That's a bit of weird code to me.

987Nabil:
No, for that case probably not. But if you read configs or have predefined urls or for testing. As I said, where poasible

zio-http/shared/src/main/scala/zio/http/URL.scala
guizmaii: Changed both. Thanks

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -279,26 +280,35 @@ final case class URL(
...
 
 object URL {
-  def empty: URL = URL(Path.empty)
+  val empty: URL = URL(path = Path.empty)
+
+  /**
+   * Proper implementation of [[java.net.MalformedURLException]] which, unlike
+   * the Java version, can propagate the cause.
+   */
+  final case class MalformedURLException(rawUrl: String, cause: Option[Throwable])
```
guizmaii: I tend to write `cause: Throwable = null` instead of using an `Option` here but Scala people tend to get shocked when `null` is used anywhere.

@987Nabil @kyri-petrou What do you prefer?

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -279,26 +280,35 @@ final case class URL(
...
 
 object URL {
-  def empty: URL = URL(Path.empty)
+  val empty: URL = URL(path = Path.empty)
+
+  /**
+   * Proper implementation of [[java.net.MalformedURLException]] which, unlike
+   * the Java version, can propagate the cause.
+   */
+  final case class MalformedURLException private[http] (rawUrl: String, cause: Option[Throwable])
```
guizmaii: I make the constructor private so people can't build a `MalformedURLException` themselves as it's IMO internals.

WDYT?

kyri-petrou: I don't think we need this to be public at all. Wdut about something like this:

```scala
  final private class Err(rawUrl: String, cause: Throwable) extends MalformedURLException {
    override def getMessage: String  = s"""Invalid URL: "$rawUrl""""
    override def getCause: Throwable = cause
  }
```

Note that I changed it to a `class` as well instead of a `case class`

guizmaii: It needs to be public I think as it's part of the result of the public function `URL.decode`, no? 🤔

kyri-petrou: The return of `decode` can be `Either[java.net.MalformedURLException, URL]` (since `Err extends java.net.MalformedURLException`)

987Nabil:
@guizmaii please run `fmt` 🙂

zio-http/shared/src/main/scala/zio/http/URL.scala
guizmaii: @kyri-petrou I'm sorry I'm not sure to understand the benefit of your suggestion 🤔
What if the user wants to have access to the `cause`? 
The fact that the `Err` class is private will prevent him/her to get access to the `getCause` method, no?

guizmaii:
@987Nabil Done :)

zio-http/shared/src/main/scala/zio/http/URL.scala
kyri-petrou: No the user can still access it because the `getCause` method is defined on `Throwable`. My suggestion is so that we don't have to return a custom error (which might cause issues with binary compatibility in the future if we want to extend it)

kyri-petrou: https://scastie.scala-lang.org/3TFy9PhyTPqomnoOcfXl4Q

guizmaii: Ok cool thanks. I'll make the change :)

guizmaii: Done

guizmaii:
All comments took into account :)

Thanks for your review mates :)

987Nabil (approved):

PR #3018, opened 2024-08-15
Add streaming body constructors with env access (#2817)

### LINKED ISSUE #2817: Stream body with environmentment access or scope
Currently `Body.fromStream*` and `Response.fromServerSentEvents` fix the `R` type of the output stream to be `Any`. It blocks or makes difficult to implement some use cases such as: 

- A Server Sent Event response stream might want access internal modules via environment.
- Use a resourceful ZIO effect with a stream body, i.e. A temporary file `RIO[Scope, File]` that should be deleted after the stream terminates.

There are some workarounds like `ZIO.environmentWith(stream.provideEnvironment)`, but it does not go well with `Scope` because the stream outlives the handler in the current design.

987Nabil:
fixes #2817
/claim #2817

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Body.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Body.scala
@@ -324,6 +324,13 @@ object Body {
     StreamBody(stream, knownContentLength = Some(contentLength))
 
   /**
+   * Constructs a [[zio.http.Body]] from a stream of bytes with a known length
+   * that depends on `R`.
+   */
+  def fromStreamEnv[R](stream: ZStream[R, Throwable, Byte], contentLength: Long)(implicit trace: Trace): RIO[R, Body] =
+    ZIO.environmentWith[R][Body](r => fromStream(stream.provideEnvironment(r), contentLength))
+
+  /**
    * Constructs a [[zio.http.Body]] from stream of values based on a zio-schema
    * [[zio.schema.codec.BinaryCodec]].<br>
    *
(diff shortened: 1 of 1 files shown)


zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -339,13 +346,36 @@ object Body {
...
+   * Constructs a [[zio.http.Body]] from stream of values based on a zio-schema
+   * [[zio.schema.codec.BinaryCodec]] and depends on `R`<br>
+   *
+   * Example for json:
+   * {{{
+   * import zio.schema.codec.JsonCodec._
+   * case class Person(name: String, age: Int)
+   * implicit val schema: Schema[Person] = DeriveSchema.gen[Person]
+   * val persons = ZStream(Person("John", 42))
+   * val body = Body.fromStream(persons)
```
guersam: I think this example should be updated using `fromStreamR`, reflecting the return tyep is `RIO[R, Body]` instead of `Body`.

zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -339,13 +346,36 @@ object Body {
...
+   * }}}
+   */
+  def fromStreamR[A, R](stream: ZStream[R, Throwable, A])(implicit codec: BinaryCodec[A], trace: Trace): RIO[R, Body] =
```
jdegoes: This should be `fromStreamEnv` or `fromStreamEnvironment`, as we don't use the name of the type parameters in method names anymore (since ZIO 2.x).

zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -339,13 +346,36 @@ object Body {
...
    * using chunked transfer encoding.
    */
   def fromStreamChunked(stream: ZStream[Any, Throwable, Byte]): Body =
     StreamBody(stream, knownContentLength = None)
 
+  /**
+   * Constructs a [[zio.http.Body]] from a stream of bytes of unknown length
+   * that depends on `R`, using chunked transfer encoding.
+   */
+  def fromStreamChunkedR[R](stream: ZStream[R, Throwable, Byte])(implicit trace: Trace): RIO[R, Body] =
```
jdegoes: Ditto.

jdegoes (changes requested):
Just a quick change to naming convention, then will be good to merge.

PR #3020, opened 2024-08-15
Add convenient `AuthType.Basic` and `AuthType.Bearer` type aliases

guizmaii:


jdegoes (approved):
Another way to do this:

```scala
sealed trait None extends AuthType { 
...
}
val None = new None {}
```

The advantage is that there is no de-aliasing of `None` / `Basic` / `Bearer`, so they will always appear in that form. The disadvantage is you have to implement your own consistent equals/hashCode/toString.

PR #3024, opened 2024-08-16
Optimise `HttpContentCodec#lookup`

guizmaii:
See also https://github.com/ghostdogpr/caliban/pull/2366


jdegoes (approved):
Damn you're getting good. 😆

jdegoes (approved):

guizmaii:
@jdegoes Thanks! I'm copying from some of the optimisations I see made by Kyri in ZIO! 🙂 

We have a discussion with @987Nabil in Discord tho.
The default implementation of `Map#getOrElse` is instantiating an `Option`, which is annoying as it's in the hot path.
So I replaced the `scala.Map` by a `scala.HashMap` because its `#getOrElse` implementation doesn't instantiate any `Option`:
Screenshot 2024-08-16 at 9 04 43 PM

What's your opinion about that?

987Nabil:
@jdegoes I was also wondering, if small HashMaps (4 or 5 elements max) are slower then the default map.

kyri-petrou:
I might be missing something, but this looks like an ideal situation to use a mutable map instead of an immutable one with a `var`. What's the reasoning for using an immutable one?

kyri-petrou:
I mentioned this on Discord but just putting this here as well for visibility; Most (if not all) immutable Map implementations override the `getOrElse` implementation on the base Map trait. E.g., Map1:

image

987Nabil:
@kyri-petrou I don't think in this case it makes any significant difference between var or mutable. Because the writes are probably just a very few. This is read heavy

kyri-petrou:
@987Nabil In that case I'd leave the implementation as is then (using `Map` instead of `HashMap`)

kyri-petrou:
@guizmaii by the way I recently came across a small trick that might be good to apply here. To avoid the Function0 allocation in `getOrElse`, you can store the `() => null` in a val and do this:

```scala
val map: Map[String, String] = ???
private val orElseNull: () => String = () => null

map.getOrElse("foo", orElseNull())
```
This avoids the allocation because `() => String` and `=> String` are the same thing.

_Scala 3 only_:
Store the `val` in the companion object and add the `@static` annotation to it

kyri-petrou:
@guizmaii I just found out something interesting (or maybe a bug in dotty?). In Scala 2 this optimization works out of the box, but in Scala 3 it only works if you use the `@static` annotation 🤔

zio-http/shared/src/main/scala-3/zio/http/syntax.scala
```diff
@@ -0,0 +1,17 @@
...
+private[http] object syntax {
+  @static val NullFn: () => AnyRef = () => null
+
+  extension [K, V <: AnyRef](inline map: Map[K, V]) {
+    transparent inline def getOrElseNull(key: K): V = map.getOrElse(key, NullFn()).asInstanceOf[V]
+  }
+}
+
+// Required for @static fields
+private final class syntax private
```
guizmaii, line 17: @kyri-petrou What is this for? Do I need it? 🤔

guizmaii:
@987Nabil @kyri-petrou I made the changes you asked for :)

987Nabil:
@guizmaii I think something of the optimizing is not working with scala js

987Nabil:
@guizmaii I close PRs to have a better overview what is needed for 3.0
Feel free to reopen when it is building

PR #3026, opened 2024-08-16
Provide Authentication information via EndpointExecutor (#3019)

### LINKED ISSUE #3019: Handle auth for endpoint without explicit handover
Currently auth info has to be handed over manually. For many endpoints that might be cumbersome.

We want a solution, where we can handover auth information for a call in a way that it can be computed, since it might need to refresh a token.
Probably this can be bound to an `EndpointExecutor`. An executor that has a way of authenticating, should also be able to execute requests that need no auth, since apis often have a public subset of endpoints.

987Nabil:
fixes #3019
/claim #3019


jdegoes (approved):
Such a huge cleanup and big improvement to usability! 🎉

PR #3032, opened 2024-08-18
Configurable EndpointAPI codecs (#2911)

### LINKED ISSUE #2911: Allow users to be able to toggle ignoreEmptyCollections when returning JSON from their Endpoints
I need to be able to return JSON from APIs where empty collections are serialized as `"myList": []` instead of excluded completely from the output. I noticed that in the HttpContentCodec there is a line that has the ignoreEmptyCollections config hard coded to be true:

https://github.com/zio/zio-http/blob/36df30fd5d02d33a4fe624ca57fde0076162f084/zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala#L264

There  is no way to change this that I can see. I can understand the reasoning, however I live in the real world where other downstream consumers of my APIs are expecting a non-missing field and they can't/won't change their systems to accommodate this.

Some way to configure how the JsonCodec is setup in zio-http rather than have it hard coded. 

Creating something like the following for each case class model works, but is clunky, just to get around the hard coded config:
```
case class Test(list: List[String])
object Test {
    implicit val schema : Schema[Test] = DeriveSchema.gen
    val codec: BinaryCodecWithSchema[Test] = BinaryCodecWithSchema(
      zio.schema.codec.JsonCodec.schemaBasedBinaryCodec,
      schema,
    )
    implicit val testHttpContentCodec: HttpContentCodec[Test] = zio.http.codec.HttpContentCodec.from(
      (MediaType.application.json, codec )
    )
 }

val endpoint = Endpoint(Method.GET / "test")
    .out[Test]

val route = endpoint.implement(handler(ZIO.succeed(Test(Nil))))

route.apply(Request.get("/test"))
// will output: {"list":[]}
```

**Additional context**

987Nabil:
This PR also contains the changes of #3026 to avoid conflicts later on.
fixes #2911
/claim #2911

### DIFF
--- a/zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala-2.13/zio/http/benchmarks/EndpointBenchmark.scala
@@ -11,7 +11,7 @@ import zio.{Scope => _, _}
 import zio.schema.{DeriveSchema, Schema}
 
 import zio.http._
-import zio.http.codec.HttpCodec
+import zio.http.codec.{CodecConfig, HttpCodec}
 import zio.http.endpoint.Endpoint
 
 import cats.effect.unsafe.implicits.global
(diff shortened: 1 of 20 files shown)


jdegoes:
@987Nabil Can you merge conflicts and I can take another look? Hard to tell exactly what's going on here because the PR is so big.

jdegoes (approved):

PR #3038, opened 2024-08-19
clean some tests repetitive boilerplate

hochgi:
extracted some repetitive code to shared method, and used it for all tests in the suite

### DIFF
--- a/zio-http-gen/src/test/scala/zio/http/gen/scala/CodeGenSpec.scala
+++ b/zio-http-gen/src/test/scala/zio/http/gen/scala/CodeGenSpec.scala
@@ -1,6 +1,6 @@
 package zio.http.gen.scala
 
-import java.nio.file.{Path => JPath, _}
+import java.nio.file._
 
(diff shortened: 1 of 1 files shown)


zio-http-gen/src/test/scala/zio/http/gen/scala/CodeGenSpec.scala
```diff
@@ -37,12 +37,12 @@ object CodeGenSpec extends ZIOSpecDefault {
   )
   implicit val validatedDataSchema: Schema[ValidatedData] = DeriveSchema.gen[ValidatedData]
 
-  private def fileShouldBe(dir: java.nio.file.Path, subPath: String, expectedFile: String): TestResult = {
+  private def fileShouldBe(dir: JPath, subPath: String, expectedFile: String): TestResult = {
```
987Nabil: I'd prefer the full name, since it had only two usages now anyway

987Nabil (approved):

PR #3039, opened 2024-08-19
Security audit

### LINKED ISSUE #1535: Security audit of ZIO HTTP

pablf:
/claim #1535
fixes #1535

### Conclusions
1. Static file serving and resource serving is safe through `serveDirectory` and `serveResources`. Not tested for scalajs as it's not implemented.
2. Exceptions can throw parts of stacktrace through `Response.fromCause`. Possible vulnerabilities.
3. There are some timing vulnerabilities through `basicAuth`, and `bearerAuth`. Some of this are due to `zio.Config.Secret` vulnerability as it leaks the length of the secret (see zio/zio#9129). `basicAuth(u: String, p: String)` is unsafe due to not comparing passwords if the user is wrong. `bearerAuth` using a function directly on the secret can be unsafe. Also `customAuth` might potentially create vulnerabilities.
4. Limits on length of parts of the requests are exposed and working. Their defaults are reasonable for all given parameters: 4096 for initial line, 8192 for headers and 1024 * 100 for body. Other frameworks have higher parameters (like twice zio's) that would also be reasonable. I think that for most basic uses this is enough and they coincide with netty's ones. 
5. Metrics middleware doesn't register (and can't as middlewares are implemented) responses generated when limits are crossed. This means that they can't reach user's code which would be a vulnerability but `Middleware.metrics` can't be used to measure these types of bad requests.
6. XSS exploits could be posible depending on the user employing or not `Dom.raw` (or using a custom DOM library). Appropriate escaping otherwise. Other type of injections attacks are possible as these are not prevented in any manner.

### Changes done
- Modify `Handler.basicAuth` to make it time constant regardless of whether username is correct or not.
- Modify` Handler.bearerAuth` and `Handler.bearerAuthZIO` to use zio.Config.Secret. `customAuth` can be used to replicate current use, so I would opt to make the default `bearerAuth` safer.
-  The implementation of `equals` of `zio.Config.Secret` leaks the length of the secret. I will make a pull request for that in zio repository.

### Possible changes to discuss:
- Regarding 2., #2994 could make it safe unless `Response.fromCause` is used by the user or  `includeErrorDetails` is used. It can be useful to have access to the stacktrace so maybe the best option is to leave it like this and always use excludeBodyError (which would be the default) for production environments.
- Regarding 3., it would be safer to implement a safe comparison for `zio.http.Credentials` and make both user name and password invisible. This would defend against all attacks based on the server manipulating `Header.Authorization` through `customAuth` at the cost of making inaccesible the user name. Maybe a `def toUnsafe(): (String, Secret)` or `def unsafeUser: String` could avoid users making unsafe decisions and at the same time allowing access to the user name.
- Regarding 6., `Dom.raw` should be exposed but maybe it should be clearer that it is unsafe. Maybe changing method name or description. Also, DOM-based XSS vulnerabilities are client-side and thus difficult to prevent. Any client-side vulnerable code could be sent by the server. This is demonstrated through some tests marked with `TestAspect.failing` and the only way to avoid this is through good server implementation.

### Suggerences:
- A full-fledged library for input sanitation in other cases like SQL, XML, JSON,... would be a big plus. This is the user's responsability, but having a sanitation module integrated within `zio-http` would be a great upgrade. This can be done using schemas for the input.
- Some clients (like browsers) might implement escaping before sending requests and `zio-http` could have some tools to make safer the use of a client given a compromised server.

### Other
I've seen several errors in the client about `PrematureClosedChannelException` when sending a lot of requests, which I don't think is expected. @kyri-petrou has made some work related to this, so he might find this useful. In particular when testing multipart forms in `SizeLimitsSpec`, it sometimes breaks the client in this way. Maybe I'm just using something bad, but when testing configurable limits usually arise in some of the tests versions, so I've skipped testing bad requests in that particular test.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -5,7 +5,7 @@ object Dependencies {
   val NettyVersion                  = "4.1.112.Final"
   val NettyIncubatorVersion         = "0.0.25.Final"
   val ScalaCompactCollectionVersion = "2.12.0"
-  val ZioVersion                    = "2.1.7"
+  val ZioVersion                    = "2.1.8"
   val ZioCliVersion                 = "0.5.0"
   val ZioJsonVersion                = "0.7.1"
   val ZioSchemaVersion              = "1.4.1"
--- a/zio-http/jvm/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/internal/middlewares/AuthSpec.scala
@@ -29,9 +29,10 @@ object AuthSpec extends ZIOHttpSpec with HttpAppTestExtensions {
 
   private val successBasicHeader: Headers  = Headers(Header.Authorization.Basic("user", "resu"))
   private val failureBasicHeader: Headers  = Headers(Header.Authorization.Basic("user", "user"))
-  private val bearerToken: String          = "dummyBearerToken"
+  private val bearerContent: String        = "dummyBearerToken"
+  private val bearerToken: Secret          = Secret(bearerContent)
   private val successBearerHeader: Headers = Headers(Header.Authorization.Bearer(bearerToken))
-  private val failureBearerHeader: Headers = Headers(Header.Authorization.Bearer(bearerToken + "SomethingElse"))
+  private val failureBearerHeader: Headers = Headers(Header.Authorization.Bearer(bearerContent + "SomethingElse"))
 
   private val basicAuthM     = HandlerAspect.basicAuth { c =>
     Secret(c.uname.reverse) == c.upassword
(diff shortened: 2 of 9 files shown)


kyri-petrou:
> I've seen several errors in the client about PrematureClosedChannelException when sending a lot of requests, which I don't think is expected

Would you be able to open an issue with a repro for this? It doesn't need to be deterministic, just something that can be used as a starting point

jdegoes (approved):

Issue #2960, opened 2024-07-13
Improve `Literal` path matching via state machine


987Nabil:
Matching on path literals is currently done by set operations on a set of strings.
The idea would be to improve this, by creating a state machine, that transitions to the next state based on the chars of all the possible alternative path segments.
For example, if we have two segments on the same level `aa` and `ab`, we can only proceed if the first char is an `a`, the second `a` or `b`.
Also we might be able to make a check based on the input length and fast exist if non of the possible segments has the same length.

This requires benchmarking first. So we can compare the current to the state machine version.

jdegoes:
/bounty $1250 for benchmarks and compilation to super fast state machines. Benchmarks should prove superior performance and happy path should be zero allocation, or as close as is theoretically possible.

987Nabil:
/attempt #2960


<div id="algora-attempt" />

| [Algora profile](https://console.algora.io/@/987Nabil) | Completed bounties | Tech | Active attempts | Options |
| --- | --- | --- | --- | --- |
| @987Nabil | &nbsp;&nbsp;&nbsp;56 ZIO bounties <br/> + 1 bounty from 1 project | <div align="center">Scala</div> | <div align="center">﹟1522</div> | [Cancel attempt](https://console.algora.io/api/bounties/clz8gldv70004kz09oi0l1bz0/cancel-attempt) |

RohanMishra315:
hey @987Nabil you working on these, i would like to do this ??

987Nabil:
Yes, i am

Mbazooka:
@987Nabil May I work on this?

hoklims:
/attempt #2960

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clz8gldv70004kz09oi0l1bz0/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

promisingcoder:
/attempt #2960

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/clz8gldv70004kz09oi0l1bz0/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

promisingcoder:
@jdegoes @987Nabil 
please review my pull request

promisingcoder:
@vigoo

[...]

PR #3040, opened 2024-08-19
Fix Route with literal path segment makes route with variable path segment unavailable

### LINKED ISSUE #3036: Route with literal path segment makes route with variable path segment unavailable
**Describe the bug**

Consider the following zio-http app:

```scala
//> using dep dev.zio::zio-http::3.0.0-RC9

import zio.*
import zio.http.*


object TestServer extends ZIOAppDefault:

  private val app = Routes(
    Method.GET / "base" / "type1" / "fixed" -> handler(Response.text("Fixed endpoint")),
    Method.GET / "base" / string("type") / "dynamic" -> Handler.fromFunctionHandler[(String, Request)] { case (tpe: String, _: Request) =>
      Handler.text(s"Dynamic endpoint: $tpe")
    }
  )


  override def run =
    Server
      .serve(app)
        .provideSomeLayer[Scope](
          ZLayer.succeed(Server.Config.default.port(8060)) >>> Server.live
        )

```

When I do the following requests, this happens:

Correct:
```
GET http://localhost:8060/base/type1/fixed

HTTP/1.1 200 OK
content-type: text/plain
date: Mon, 19 Aug 2024 13:17:46 GMT
content-length: 14

Fixed endpoint

Response code: 200 (OK); Time: 582ms (582 ms); Content length: 14 bytes (14 B)
```

Wrong:
```
GET http://localhost:8060/base/type1/dynamic

HTTP/1.1 404 Not Found
warning: 199 ZIO HTTP "/base/type1/dynamic"
date: Mon, 19 Aug 2024 13:18:29 GMT
content-length: 0

<Response body is empty>

Response code: 404 (Not Found); Time: 11ms (11 ms); Content length: 0 bytes (0 B)


```

---------------------------

If I comment out the fixed route:
```scala
//    Method.GET / "base" / "type1" / "fixed" -> handler(Response.text("Fixed endpoint")),
```

the dynamic path does work as expected:
```
GET http://localhost:8060/base/type1/dynamic

HTTP/1.1 200 OK
content-type: text/plain
date: Mon, 19 Aug 2024 13:20:09 GMT
content-length: 23

Dynamic endpoint: type1

Response code: 200 (OK); Time: 609ms (609 ms); Content length: 23 bytes (23 B)


```

It seems that the routing algorithm falls into the fixed "base" / "type1" path, and no longer considers the path with the path variable, even though it is still valid.

--------------

"Why would you even use routes like this?", you might wonder.

Say you have the following API:
GET /pets/{animalType}/is-four-legged
GET /pets/dog/list
GET /pets/snake/list

The first endpoint you would like to define with animalType as a variable, because the response type is the same for any animalType.
The 2nd and 3rd endpoint you would like to define statically per animal type, because the response type is different (snake and dog have different properties/fields).

narma:
Resolves #3036

So case happens when literal and other (codec) in subtree are both valid for path segment we try only literal one but tail is valid only for not-literal one
To fix this, I've added backtracking to other codec way in unhappy case when literal one was give us empty results. Also we don't do backtracking if others segments doesn't match literals.

This solution is optimized for cases without these conflicts, but when this case happened - backtracking is always slow, perhaps futher we need to investigate different approach to solve this issue.


narma:
Implementation of https://github.com/zio/zio-http/issues/2960 will rewrite all of these code, but I think we should merge this to get new tests for this issue.

987Nabil:
I saw the PR, but did not find the time to take a good look

narma:
@987Nabil No worries! Take your time

jdegoes (approved):

PR #3041, opened 2024-08-19
fix User-Agent header parsing

### LINKED ISSUE #3035: UserAgent parsing is broken
The parsing of the UserAgent is broken.
This example can not be parsed. But is a valid user agent.
```scala
UserAgent.parse("some-thing/v1.2.0")
```
This SO post seems to be a good summary of the requirements.
https://stackoverflow.com/a/2601492

Also, user agents can have multiple products. We currently only parse one.

geeeezmo:
Resolves #3035


987Nabil (changes requested):
This is still not implementing the spec. Please check the so post again

geeeezmo:
> This is still not implementing the spec. Please check the so post again

I've re-read the SO post and the relevant RFC portion, and I'm still not entirely sure what should be considered `comment` in, for example, this user agent string:
```
Mozilla/v5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36
```

Current regular expression captures everything after the first opening parenthesis as `comment`
```
(Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36
```
but the RFC (and the SO post) says comment is text that is enclosed in parentheses, so it should be
```
(Macintosh; Intel Mac OS X 10_15_7)
```

In that case, everything after the closing parenthesis should be parsed as another product.

Are we happy with the current behaviour (capture everything after the opening parenthesis), or do would it be better to slightly change the structure of `UserAgent` variants? Or perhaps add a 4th variant with multiple products like it's done for `Via`
```
object UserAgent extends HeaderType {
  ...
  final case class Multiple(values: NonEmptyChunk[Complete]) extends UserAgent
  ...
}
```
?

987Nabil:
I am not sure, if an extra multiple case is not too much. Also the structure is not correct. The multiples might contain also comments only parts if I read the spec correctly

geeeezmo:
> I am not sure, if an extra multiple case is not too much. Also the structure is not correct. The multiples might contain also comments only parts if I read the spec correctly

Indeed it can be comments only after the first product, but according to [rfc9110#name-comments](https://www.rfc-editor.org/rfc/rfc9110#name-comments)
> Comments can be included in some HTTP fields by surrounding the comment text with parentheses.

So this user agent
```
Mozilla/v5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36
```
would have the structure
```
product comment product comment product product
```
while if the parts after the first product were only comments, the user agent would look like this
```
Mozilla/v5.0 (Macintosh; Intel Mac OS X 10_15_7) (KHTML, like Gecko) (comment 2) (comment 3)
```

I've understood it this way. And regardless, still have to change how the string is parsed.

987Nabil:
Yes, that sounds about right.

geeeezmo:
In that case, if we don't want/need multiple variant of the header, the current implementation covers all other cases:
```
Foobar
Foobar-baz
Foobar/abc
Foobar/1.0.0
Foobar/2021.44.30.15-b917dc
MyProduct Foobar/2021.44.30.15-b917dc
Tsom/OfraHaza (Life is short and love is always over in the morning) AnotherProduct
Mozilla/v5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.114 Safari/537.36
CERN-LineMode/2.15 libwww/2.17b3
Mozilla/v5.0 (Macintosh; Intel Mac OS X 10_15_7) (KHTML, like Gecko) (comment 2) (comment 3)
```
capturing only the first product and comment.
I just don't think there's a third way of doing it.
I can add tests for different kinds if necessary.

987Nabil:
I am not sure, I got you right before. I'd expect, that the structure should look like `UserAgent(product: Product, parts: List[ProductOrComment])`

geeeezmo:
> I am not sure, I got you right before. I'd expect, that the structure should look like `UserAgent(product: Product, parts: List[ProductOrComment])`

Ah, got it. Makes sense.
Will try to implement it that way.

geeeezmo:
> I am not sure, I got you right before. I'd expect, that the structure should look like `UserAgent(product: Product, parts: List[ProductOrComment])`

To be honest, I do not like to model it exactly like that:
```scala
final case class UserAgent(
    product: UserAgent.ProductOrComment.Product,
    parts: List[UserAgent.ProductOrComment] = Nil
  ) extends Header

object UserAgent extends HeaderType {
  sealed trait ProductOrComment
  object ProductOrComment {
    final case class Product(name: String, version: Option[String]) extends ProductOrComment

    final case class Comment(comment: String) extends ProductOrComment
  }
}
```
and even if I move `Product` and `Comment` back outside of `ProductOrComment`
```
object UserAgent extends HeaderType {
  sealed trait ProductOrComment
  
  final case class Product(name: String, version: Option[String]) extends ProductOrComment

  final case class Comment(comment: String) extends ProductOrComment
}
```
it still looks quite clumsy.
But I'd probably be OK with using it like that.
It will work either way, just a matter of library user experience I guess.

987Nabil:
We always place the impl of the sealed trait in the sealed trait companion. How it looks like in user code depends on the imports. Of you have a better design in mind to model the spec. Feel free to make a suggestion

geeeezmo:
> We always place the impl of the sealed trait in the sealed trait companion. How it looks like in user code depends on the imports. Of you have a better design in mind to model the spec. Feel free to make a suggestion

Nah, I think I'm alright with the one you've suggested.

987Nabil:
Looks good for me so far. @jdegoes can you take a look too.

jdegoes (approved):

987Nabil (approved):

PR #3043, opened 2024-08-21
[gen] dictionaries with referenced keys extension

### LINKED ISSUE #3030: [gen] support a schema for dictionary keys
Open dictionaries always have unconstrained string keys.
Typically one would define it like:
```yaml
components:
  schemas:
    Order:
      …
    UserOrderHistory:
      type: object
      properties:
        user_id:
          type: string
          format: uuid
        order_history:
          type: object
          additionalProperties:
            $ref: '#/components/schemas/Order'
```
which when used with codegen would yield code like:
```scala
case class UserOrderHistory(
  user_id: UUID,
  order_history: Map[String, Order]
)
```
But a common scenario is when keys are constrained.
e.g. in this case, we may want `order_history` to have only UUID keys instead of unconstrained strings.

Anything other than unconstrained strings for dictionary keys is not supported in OpenAPI by default.
But we can consider leveraging aliasing as introduced in #2962 with a new custom extension property to enable such functionality.
e.g:
```yaml
components:
  schemas:
    OrderId:
      type: string
      format: uuid
    Order:
      …
    UserOrderHistory:
      type: object
      properties:
        user_id:
          type: string
          format: uuid
        order_history:
          type: object
          additionalProperties:
            $ref: '#/components/schemas/Order'
            x-string-key-ref:
              $ref: '#/components/schemas/OrderId'
```
codegen would then verify that the ref actually points to a string (and string only) type.
Any other type should yield a failure.

with this example we should now be able to generate code like:
```scala
case class UserOrderHistory(
  user_id: UUID,
  order_history: Map[UUID, Order]
)
```
or with aliased flag turned on:
```scala
case class UserOrderHistory(
  user_id: UUID,
  order_history: Map[OrderId.Type, Order]
)
```

There isn't really an alternative, as this is not a supported feature in OpenAPI.
Best we can do, is maybe add textual description with runtime validations in code.

**Additional context**
By using this mechanism, we can add any sort of `format` or validation supported by OpenAPI (e.g. `pattern`).
This would be helpful with the aliased field as Newtypes flag turned either on or off.
We could still use better types like `UUID` or use a validated `String` in generated code.

hochgi:
* works with or without alias flag properly.
* validates non-string references.
* tests for all UUID/aliased/non-string

fixes #3030 
/claim #3030

### DIFF
--- a/zio-http-gen/src/test/resources/ComponentAliasOrderId.scala
+++ b/zio-http-gen/src/test/resources/ComponentAliasOrderId.scala
@@ -1,0 +1,9 @@
+package test.component
+
+import zio.prelude.Newtype
+import zio.schema.Schema
+import java.util.UUID
+
+object OrderId extends Newtype[UUID] {
+  implicit val schema: Schema[OrderId.Type] = Schema.primitive[UUID].transform(wrap, unwrap)
+}
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -1,18 +1,21 @@
 package zio.http.endpoint.openapi
 
 import scala.annotation.{nowarn, tailrec}
+import scala.util.chaining.scalaUtilChainingOps
 
 import zio._
 import zio.json.ast.Json
 
 import zio.schema.Schema.CaseClass0
+import zio.schema.StandardType.Tags
 import zio.schema._
 import zio.schema.annotation._
 import zio.schema.codec._
 import zio.schema.codec.json._
 import zio.schema.validation._
 
-import zio.http.codec.{SegmentCodec, TextCodec}
+import zio.http.codec.{PathCodec, SegmentCodec, TextCodec}
+import zio.http.endpoint.openapi.BoolOrSchema.SchemaWrapper
 import zio.http.endpoint.openapi.JsonSchema.MetaData
 
 @nowarn("msg=possible missing interpolator")
@@ -26,6 +29,7 @@ private[openapi] case class SerializableJsonSchema(
   @fieldName("enum") enumValues: Option[Chunk[Json]] = None,
   properties: Option[Map[String, SerializableJsonSchema]] = None,
   additionalProperties: Option[BoolOrSchema] = None,
+  @fieldName("x-string-key-schema") optionalKeySchema: Option[SerializableJsonSchema] = None,
   required: Option[Chunk[String]] = None,
   items: Option[SerializableJsonSchema] = None,
   nullable: Option[Boolean] = None,
@@ -101,7 +105,7 @@ private[openapi] object BoolOrSchema {
 
   object SchemaWrapper {
     implicit val schema: Schema[SchemaWrapper] =
-      Schema[SerializableJsonSchema].transform(SchemaWrapper(_), _.schema)
+      Schema[SerializableJsonSchema].transform(SchemaWrapper.apply, _.schema)
   }
 
   final case class BooleanWrapper(value: Boolean) extends BoolOrSchema
@@ -254,9 +258,18 @@ object JsonSchema {
 
   private def fromSerializableSchema(schema: SerializableJsonSchema): JsonSchema = {
     val additionalProperties = schema.additionalProperties match {
-      case Some(BoolOrSchema.BooleanWrapper(false)) => Left(false)
-      case Some(BoolOrSchema.BooleanWrapper(true))  => Left(true)
-      case Some(BoolOrSchema.SchemaWrapper(schema)) => Right(fromSerializableSchema(schema))
+      case Some(BoolOrSchema.BooleanWrapper(bool))  => Left(bool)
+      case Some(BoolOrSchema.SchemaWrapper(schema)) =>
+        val valuesSchema = fromSerializableSchema(schema)
+        Right(
+          schema.optionalKeySchema.fold(valuesSchema)(keySchema =>
+            valuesSchema.annotate(
+              MetaData.KeySchema(
+                fromSerializableSchema(keySchema),
+              ),
+            ),
+          ),
+        )
       case None                                     => Left(true)
     }
 
@@ -432,10 +445,10 @@ object JsonSchema {
                 minItems = Some(1),
                 uniqueItems = identity == "NonEmptySet",
               )
-            case Schema.Map(_, valueSchema, _)                             =>
-              mapSchema(refType, ref, seenWithCurrent, valueSchema)
-            case Schema.NonEmptyMap(_, valueSchema, _)                     =>
-              mapSchema(refType, ref, seenWithCurrent, valueSchema)
+            case Schema.Map(keySchema, valueSchema, _)                     =>
+              mapSchema(refType, ref, seenWithCurrent, keySchema, valueSchema)
+            case Schema.NonEmptyMap(keySchema, valueSchema, _)             =>
+              mapSchema(refType, ref, seenWithCurrent, keySchema, valueSchema)
             case Schema.Set(elementSchema, _)                              =>
               arraySchemaMulti(refType, ref, elementSchema, seenWithCurrent)
           }
@@ -492,30 +505,25 @@ object JsonSchema {
     }
   }
 
-  private def mapSchema[V](
+  private def mapSchema[K, V](
     refType: SchemaStyle,
     ref: Option[java.lang.String],
     seenWithCurrent: Set[java.lang.String],
+    keySchema: Schema[K],
     valueSchema: Schema[V],
   ) = {
-    val nested = fromZSchemaMulti(valueSchema, refType, seenWithCurrent)
+    val nested          = fromZSchemaMulti(valueSchema, refType, seenWithCurrent)
+    val mapObjectSchema = annotateMapSchemaWithKeysSchema(nested.root, keySchema)
+
     if (valueSchema.isInstanceOf[Schema.Primitive[_]]) {
       JsonSchemas(
-        JsonSchema.Object(
-          Map.empty,
-          Right(nested.root),
-          Chunk.empty,
-        ),
+        mapObjectSchema,
         ref,
         nested.children,
       )
     } else {
       JsonSchemas(
-        JsonSchema.Object(
-          Map.empty,
-          Right(nested.root),
-          Chunk.empty,
-        ),
+        mapObjectSchema,
         ref,
         nested.children ++ nested.rootRef.map(_ -> nested.root),
       )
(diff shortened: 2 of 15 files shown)


zio-http-gen/src/test/resources/ComponentAliasOrderId.scala
```diff
@@ -0,0 +1,9 @@
+package test.component
+
+import zio.prelude.Newtype
+import zio.schema.Schema
+import java.util.UUID
+
+object OrderId extends Newtype[UUID] {
```
jivanic-demystdata, line 7: You need to add this when you declare a newtype
```suggestion
type OrderId = OrderId.Type
object OrderId extends Newtype[UUID] {
```

zio-http-gen/src/test/resources/ComponentAliasOrderId.scala
```diff
@@ -0,0 +1,9 @@
...
+  implicit val schema: Schema[OrderId.Type] = Schema.primitive[UUID].transform(wrap, unwrap)
```
jivanic-demystdata, line 8: You don't need to write them. They can be automatically derived:
```suggestion
  implicit val schema: Schema[Type] = derive
```

hochgi: This is also a part of #3002, not really changed in this PR.
I did tried it at first though:
https://github.com/zio/zio-http/pull/3002/commits/f11b05d5d71c03fb0675ae8447690d2d0aff0aa3#diff-05e59888caf2ce39f045befc5e4ffc68051f5a68cb3d001410545aa7ae2b96cbR7

Couldn't get it to work on all cross versions, so I went for the more verbose code.

I figured its generated code anyway, and not using the macro gives better compilation times, so there's a silver lining...

Perhaps in a future PR it can be improved, but as said - this is not a part of current PR.

zio-http-gen/src/test/resources/ComponentAliasOrderId.scala
hochgi: @jivanic-demystdata thanks for taking the time to look at this :)

scala 2 does not support top level type declarations.
Generated code needs to be valid scala 2 as well.

Regardless, the Newtype aliases has already been implemented in #3002 

This PR isn't a direct extension, but it needs to work with the alias flag as well.

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -1209,7 +1220,7 @@ object JsonSchema {
 
   final case class Object(
     properties: Map[java.lang.String, JsonSchema],
-    additionalProperties: Either[Boolean, JsonSchema],
+    additionalProperties: Either[Boolean, Object.Dictionary],
```
987Nabil: This is a similar discussion as we had today in the other issues. Your focus here is only the code gen. But this is a general impl. of json schema. And it does not know Dictionary.
I think Schema is right and you need to fix your requirements in the gen

987Nabil (changes requested):

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
hochgi: Indeed.
In fact, after giving it more thought, the extension can take any schema, not just a reference. as long as we validate it is a string schema.
So I'll generalize to also allow inline a regular schema instead of restricting to ref.
Will update :+1:

hochgi: @987Nabil both on write & read side, I'll need some sort of "tuple" to hold the optional key schema with the required value schema.

So I have several options here:
- make JsonSchema.Object take another optional key schema field
- make additionalProperties a tuple `(Option[JsonSchema], Either[Boolean, JsonSchema])` - for optional key schema with current either
- don't touch structure, but use a JsonSchema "product" for the two optional schemas - currently perhaps only `AllOfSchema` makes sense

I don't like the first, since it's not "bound" to where it's eligible, which is the `additionalProperties` field.
The last  would be ambiguous whether top level allOf refer to key & value schema, or just an allOf for the values.
The second option is too verbose.
IIUC, `Left(false)` means no `additionalProperties`. `Left(true)` means unconstrained `additionalProperties` values. And `Right(schema)` means strict schema for values.
If that's the case, it could have been `Option[Option[JsonSchema]]`:
- None == Left(false)
- Some(None) == Left(true)
- Some(Some(schema)) == Right(schema)
Such that with optional key schema it can be `Option[(Option[JsonSchema], Option[JsonSchema])]` - the tuple is optional key schema + optional value schema.
But if that's the case, why not make it a named tuple (case class) like `Dictionary`?

I would make sure to do both read/write sides to `SerializableJsonSchema` work with it, but it does seem like the best option. no?

987Nabil: Leave additional properties as is. When you read the key extension, add a new annotation with the schema to the schema of additional properties. Less complicated. And fits the idea of it being an extension better.

hochgi: @987Nabil annotate how?
The current annotations are: Examples,  Default,  Discriminator,  Nullable,  Description,  ContentEncoding,  ContentMediaType,  Deprecated
Should I just add a new one?

~~Also, in case we have `Left(true)` together with my extension, i.e. only restricting keys with schema, not values.
I have nothing to annotate…~~

edit:
in case of `additionalProperties: true` there's no place to put the extension key either 🤦‍♂️

987Nabil: Yes, just add a new one. Is true even a valid map? I think true is just a way to allow extra fields (instead of failing validation) or for dynamic payload.

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -527,6 +535,38 @@ object JsonSchema {
     }
   }
 
+  private def annotationForKeySchema[K](keySchema: Schema[K]) = {
+    val stringType = StandardType.fromString("string")
+    keySchema match {
+      case Schema.Primitive(`stringType`, annotations) if annotations.isEmpty => None
```
987Nabil: Just use `StandartType.String`

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -527,6 +535,38 @@ object JsonSchema {
...
+          case JsonSchema.String(None, None, None, None) => None // no need for extension
+          case s: JsonSchema.String                      => Some(MetaData.KeySchema(s))
+          case _                                         => None // only string keys are allowed
+        }
+    }
+  }
+
+  private def annotateMapSchemaWithKeysSchema[K](valueSchema: JsonSchema, keySchema: Schema[K]) = {
+    val keySchemaOpt: Option[MetaData.KeySchema] = annotationForKeySchema(keySchema)
+    val resultSchema                             = keySchemaOpt.fold[JsonSchema](valueSchema)(valueSchema.annotate)
```
987Nabil: Please use pattern match instead of fold. It is easier to read.

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -1248,23 +1285,79 @@ object JsonSchema {
     def required(required: Chunk[java.lang.String]): Object =
       this.copy(required = required)
 
+    private def reconcileTwoOptions[T](left: Option[T], right: Option[T], noneIfOne: Boolean)(
+      combine: (T, T) => Option[T],
+    ): Option[T] = {
+      if (noneIfOne) left.fold(Option.empty[T])(l => right.fold(Option.empty[T])(r => combine(l, r)))
```
987Nabil: If we keep this, make it a pattern match please

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -1248,23 +1285,79 @@ object JsonSchema {
...
-        case (Left(false), _)            => Left(false)
-        case (_, Left(_))                => left
-        case (Left(true), _)             => right
-        case (Right(left), Right(right)) =>
-          Right(AllOfSchema(Chunk(left, right)))
+        case (Left(false), _)                 => Left(false)
+        case (_, Left(_))                     => left
+        case (Left(true), _)                  => right
+        case (Right(lSchema), Right(rSchema)) =>
+          val (leftKey, lAnnotations)  = Object.extractKeySchemaFromAnnotations(lSchema)
```
987Nabil, line 1398: I am not sure we should make it this complicated. Maybe it is okay we say we only support a certain sub set. Like only one key annotated schema?

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -1248,23 +1285,79 @@ object JsonSchema {
...
+          val (rightKey, rAnnotations) = Object.extractKeySchemaFromAnnotations(rSchema)
+
+          val keySchema = reconcileTwoOptions(leftKey, rightKey, noneIfOne = true) {
+            // TODO: what happens in case of annotated key schemas?
+            //       Is it possible?
+            //       How should we combine if so?
+            case (JsonSchema.String(lfmt, lptn, lmxl, lmnl), JsonSchema.String(rfmt, rptn, rmxl, rmnl)) =>
+              Some(
+                JsonSchema.String(
+                  format = reconcileTwoOptions(lfmt, rfmt, noneIfOne = true)(someWhenEq),
```
987Nabil: Not sure, if specialized methods with some duplicated code are not easier to maintain.

987Nabil (changes requested):

987Nabil:
Please also resolve the conflicts

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
hochgi: only one key means that of the 2 dictionaries: 1 is unconstrained, and 1 has a schema. in this case, code resorts to `None`, since we cannot have any schema.
The only applicable case is when we have 2 schemas.
I chose to implement "intersection" semantics when possible, over "union".

This is an important decision, which I'm not sure about.
What's your take? should we relax the key schema to be a union of both, or restrict to be the intersection of both?

Also, when does combine is called anyway?

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
hochgi: you mean instead of having the flag?
OK, makes sense.

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
987Nabil: I think, id we have something that we can't resolve successfully, we should throw. At least something like not impl exception. Dropping it silently is hell for debugging

hochgi: In general I agree with the principle, not sure it applies here.
- We're not dropping silently
- it's an extension that is ignored by OpenAPI (anything that starts with `x-` is passed on transparently), thus I think we should avoid throwing as much as possible.

lets start with the obvious: if both are the same (either None, or identical Some(x)) - preserve.

The question is when they differ.
If only 1 is defined, i.e. combining 2 dictionaries, one restricted (e.g. UUID keys), the other is not.
So I assume the combined should not be constrained?

If both are defined, then some properties can be reconciled (e.g. min/max length or pattern), while others less so (format).
So I tried to do the best here.

What I lack, is the context of when combining actually happens.

hochgi: @987Nabil IIUC, combine is called only when generating JsonSchema from zio Schema, i.e. when we define the endpoint in scala, and generate the schema from it.
If that's the case, I think that my decision to do an "intersection" was wrong, and I should do a "union".
I.e. max of maxLength (instead min of maxLength), `|` instead of lookahead assertions in pattern, etc'…

I'll change accordingly.

987Nabil (approved):

PR #3044, opened 2024-08-22
Fix some `Endpoint::doc` field usages && Rename it to `documentation` to avoid this kind of mistake in the future

guizmaii:


zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -575,7 +575,7 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
       output = ((HttpCodec.content[Output2] ++ StatusCodec.status(status)) ?? doc) | self.output,
       error,
       codecError,
-      Doc.empty,
```
guizmaii, line 578: I don't understand why using `myEndpoint.out[...](status = ..., doc = ...)` would reset the doc to an empty one. Probably a typo

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -593,7 +593,7 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
       output = (HttpCodec.content[Output2](mediaType) ++ StatusCodec.Ok ?? doc) | self.output,
       error,
       codecError,
-      doc,
```
guizmaii, line 596: This was not using the correct `doc`. It was using the one passed in parameter and not the one from `this`

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -612,7 +612,7 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
       output = ((HttpCodec.content[Output2](mediaType) ++ StatusCodec.status(status)) ?? doc) | self.output,
       error,
       codecError,
-      doc,
```
guizmaii, line 615: This was not using the correct `doc`. It was using the one passed in parameter and not the one from `this`

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPI.scala
```diff
@@ -451,7 +451,7 @@ object OpenAPI {
 
     // todo maybe not the best regex, but the old one was not working at all
     // safe chars from RFC1738 "$" | "-" | "_" | "." | "+"
-    val validPath: Regex = """\/[\/a-zA-Z0-9\-_{}$.+]*""".r
+    private val validPath: Regex = """\/[\/a-zA-Z0-9\-_{}$.+]*""".r
```
guizmaii, line 454: Not related to the main subject of this PR but I've seen that this implementation detail was public. I think we can make it private

987Nabil (approved):

PR #3046, opened 2024-08-22
fix OpenAPI doc for Endpoint not rendered

### LINKED ISSUE #3045: [OpenAPI] The doc passed to  `Endpoint::out[A](doc = ...)` are not rendered
Reproducer:
```scala
val testEndpoint = 
  (Endpoint(RoutePattern.POST / "test") ?? Doc.p("This is my 'POST /test' endpoint doc"))
    .in[String](doc = Doc.p("this is the input doc"))
    .out[String](mediaType = MediaType.application.json, doc = Doc.p("this is the output doc"))

val spec: String =
  OpenAPIGen.fromEndpoints(
    title = "This is my OpenAPI doc title",
    version = "0.0.0",
    endpoints = List(testEndpoint)
  ).toJson
  
spec.contains("this is the output doc") // == false
```

geeeezmo:
As per [OpenAPI 3.0.3 spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md), `description` field for responses should be nested under status code (e.g. `responses.200`).
Components may have a `description` field, but it is not described for `schema` part by the official spec.

Also, eliminated merging of output documentation into `Endpoint` documentation when calling `.outX` family of endoint methods.

Resolves #3045

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
@@ -165,7 +165,7 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
       (GET / "static" / int("id") / uuid("uuid") ?? Doc.p("user id") / string("name")) ?? Doc.p("get path"),
     )
       .in[SimpleInputBody](Doc.p("input body"))
-      .out[SimpleOutputBody](Doc.p("output body"))
+      .out[SimpleOutputBody](mediaType = MediaType.application.json, Doc.p("output body"))
       .outError[NotFoundError](Status.NotFound, Doc.p("not found"))
 
   private val queryParamEndpoint =
@@ -269,21 +269,21 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
                              |        },
                              |        "responses" : {
                              |          "200" : {
+                             |            "description" : "output body\n\n",
                              |            "content" : {
                              |              "application/json" : {
                              |                "schema" : {
-                             |                  "$ref" : "#/components/schemas/SimpleOutputBody",
-                             |                  "description" : "output body\n\n"
+                             |                  "$ref" : "#/components/schemas/SimpleOutputBody"
                              |                }
                              |              }
                              |            }
                              |          },
                              |          "404" : {
+                             |            "description" : "not found\n\n",
                              |            "content" : {
                              |              "application/json" : {
                              |                "schema" : {
-                             |                  "$ref" : "#/components/schemas/NotFoundError",
-                             |                  "description" : "not found\n\n"
+                             |                  "$ref" : "#/components/schemas/NotFoundError"
                              |                }
                              |              }
                              |            }
@@ -1539,24 +1539,24 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
             |        "responses" : {
             |          "200" :
             |            {
+            |            "description" : "output body\n\n",
             |            "content" : {
             |              "application/json" : {
             |                "schema" :
             |                  {
-            |                  "$ref" : "#/components/schemas/SimpleOutputBody",
-            |                  "description" : "output body\n\n"
+            |                  "$ref" : "#/components/schemas/SimpleOutputBody"
             |                }
             |              }
             |            }
             |          },
             |          "404" :
             |            {
+            |            "description" : "not found\n\n",
             |            "content" : {
             |              "application/json" : {
             |                "schema" :
             |                  {
-            |                  "$ref" : "#/components/schemas/NotFoundError",
-            |                  "description" : "not found\n\n"
+            |                  "$ref" : "#/components/schemas/NotFoundError"
             |                }
             |              }
             |            }
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPIGen.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPIGen.scala
@@ -275,7 +275,9 @@ object OpenAPIGen {
     metadata: Chunk[HttpCodec.Metadata[_]] = Chunk.empty,
     referenceType: SchemaStyle = SchemaStyle.Inline,
     wrapInObject: Boolean = false,
+    omitDescription: Boolean = false,
   )(mediaType: MediaType): JsonSchema = {
+    val descriptionFromMeta = if (omitDescription) None else description(metadata)
     codec match {
       case atom: HttpCodec.Atom[_, _]                              =>
         atom match {
@@ -285,7 +287,7 @@ object OpenAPIGen {
             JsonSchema.obj(
               name -> JsonSchema
                 .fromZSchema(codec.lookup(mediaType).map(_.schema).getOrElse(codec.defaultSchema), referenceType)
-                .description(description(metadata))
+                .description(descriptionFromMeta)
                 .deprecated(deprecated(metadata))
                 .nullable(optional(metadata)),
             )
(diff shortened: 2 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -609,10 +609,10 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
     Endpoint(
       route,
       input,
-      output = ((HttpCodec.content[Output2](mediaType) ++ StatusCodec.status(status)) ?? doc) | self.output,
+      output = (HttpCodec.content[Output2](mediaType) ++ StatusCodec.status(status) ?? doc) | self.output,
```
987Nabil: As far as I can tell, this has either no effect on the codec or applies the doc only to status. In either way, I'd prefer to keep this. Unless you can tell me why that would turn tests read.

987Nabil (changes requested):

zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
```diff
@@ -165,7 +165,7 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
       (GET / "static" / int("id") / uuid("uuid") ?? Doc.p("user id") / string("name")) ?? Doc.p("get path"),
     )
       .in[SimpleInputBody](Doc.p("input body"))
-      .out[SimpleOutputBody](Doc.p("output body"))
+      .out[SimpleOutputBody](mediaType = MediaType.application.json, Doc.p("output body"))
```
987Nabil, line 168: Why?

jivanic-demystdata:
@geeeezmo Thanks for fixing this issue. I tried yesterday but couldn't find where the source of the issue was.
You'll have to rebase your branch on the `main` branch as we merged this branch: https://github.com/zio/zio-http/pull/3044, which renamed the `OpenAPI::doc` field to `OpenAPI::documentation`

zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
guizmaii: Probably to test this: https://github.com/zio/zio-http/pull/3044#discussion_r1726522056, which he also fixes in his PR here: https://github.com/zio/zio-http/pull/3046/files#diff-70f00f4d860129984cdbd655e9ea96e1c0b74ba77457b8a95a546a39739c1d96R578

guizmaii:
FYI, I tested it on my project and it does fix the issue I reported :)

zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
geeeezmo: Yes, this is to test the exact issue described in #3044

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
geeeezmo: You're right, this and the other change in `Endpoint` seem to have no effect on the codec - the tests are still green with the old code.
I think I was just playing around, trying to find the culprit, and thought this was it 😅

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -575,7 +575,7 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
       output = ((HttpCodec.content[Output2] ++ StatusCodec.status(status)) ?? doc) | self.output,
       error,
       codecError,
-      documentation,
+      self.documentation,
```
987Nabil: the self is now not needed anymore

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -662,7 +662,7 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
     alt: Alternator[Err2, Err],
   ): Endpoint[PathInput, Input, alt.Out, Output, Auth] =
     copy[PathInput, Input, alt.Out, Output, Auth](
-      error = ((ContentCodec.content[Err2]("error-response") ++ StatusCodec.status(status)) ?? doc) | self.error,
+      error = (ContentCodec.content[Err2]("error-response") ++ StatusCodec.status(status) ?? doc) | self.error,
```
987Nabil: you did not revert here

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -794,7 +794,7 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
     Endpoint(
       route,
       input,
-      output = ((contentCodec ++ StatusCodec.status(status)) ?? doc) | self.output,
+      output = (contentCodec ++ StatusCodec.status(status) ?? doc) | self.output,
```
987Nabil: same

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
geeeezmo: Right, because the method parameter is still called `doc`.
Good catch, fixed

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
geeeezmo: I did not revert, because that is exactly what causes the incorrect doc to be rendered - the `description` is placed under `schema` as well as under `responses.<CODE>`, which does not comply with the OpenAPI 3.0.3 spec.

987Nabil: Then this is a bug in the generator. Not in the endpoint. The call is clearly: Set the content with that status and document the **whole**. The OpenAPI gen is then incorrectly adding the doc on the wrong place.
So even if this fixes the generated code, it is not the right fix.

geeeezmo: Right, OK, I'll take another look at it over the weekend.
Thanks for the review :)

geeeezmo: Found where the issue was.
Please, give this another review pass 😃

guizmaii:
Oh fuckkkk! I pushed on your PR @geeeezmo. I'm really sorry. I'll fix that.

guizmaii:
Oh. I'm really sorry @geeeezmo. I completely messed up. I overwrote your work. I'm so sorry. I was trying to push on another branch but didn't realise I was writing on yours.

Don't pull from the remote branch locally, that could erase your work.
Could you please overwrite what I did, please?
Maybe the simplest would be that we close this PR and you open a new one?

I'm so sorry again. I apologise for this. 🫣😟

987Nabil:
I guess [this](https://github.com/zio/zio-http/commit/8dbe505ee964b66aff31bddb5838284454960061) is the version that should be restored

geeeezmo:
> Oh. I'm really sorry @geeeezmo. I completely messed up. I overwrote your work. I'm so sorry. I was trying to push on another branch but didn't realise I was writing on yours.
> 
> Don't pull from the remote branch locally, that could erase your work. Could you please overwrite what I did, please? Maybe the simplest would be that we close this PR and you open a new one?
> 
> I'm so sorry again. I apologise for this. 🫣😟

No worries 😆 
I've just force pushed from my local branch again

guizmaii (approved):
LGTM

987Nabil (approved):

PR #3050, opened 2024-08-23
Remove warning header; Configurable generated error responses (#2955)

### LINKED ISSUE #2701: Why is warning header added by default?

### LINKED ISSUE #2948: Warning header loses original exception when that exception contains invalid-for-netty-headers characters

### LINKED ISSUE #2955: Remove internal usage of warning header

987Nabil:
* I removed the typed warning header code. The intention is, to not encourage usage of a deprecated header.
* I redesigned how error messages are created if generated from `Throwable` through internal code and certain user calls, like `sandbox`. The basic idea is, that the user can auto generate error responses and based on config the details are shown or not. The config is saved in a fiber ref. That makes it possible to not only set a global default, but override config based on the route or even the request. I added for that some middleware to support this in an easy way. For example, the following code will change the config based on a set header, to get maximum details for the error.
```scala
myRoute @@ ErrorResponseConfig.debug.whenHeader(_.get("zio-debug").contains("true"))
```
* By default, the generated error responses only have a status code.
* The error response will be in HTML, JSON or plaintext, based on the accept header^
 
fixes #2955
fixes #2701
fixes #2948

/claim #2955
/claim #2701
/claim #2948


zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -80,7 +80,10 @@ private[zio] final case class ServerInboundHandler(
         try {
           if (jReq.decoderResult().isFailure) {
             val throwable = jReq.decoderResult().cause()
-            attemptFastWrite(ctx, Response.fromThrowable(throwable))
+            attemptFastWrite(
+              ctx,
+              Response.fromThrowable(throwable, runtime.unsafeRunSync(ErrorResponseConfig.configRef.get)),
```
kyri-petrou: It seems a bit wasteful to start a fiber only to extract a `FiberRef` from it.

How about adding this method to `NettyRuntime`:

```scala
def getRef[A](ref: FiberRef[A]): A = zioRuntime.fiberRefs.getOrDefault(ref)
```

And then this becomes:
```scala
Response.fromThrowable(throwable, runtime.getRef(ErrorResponseConfig.configRef))
```

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -58,6 +58,9 @@ final case class Response(
         self.copy(body = Body.fromChunk(bytes))
       }
 
+  def contentType(json: MediaType): Response =
+    self.addHeader(Header.ContentType(json))
```
kyri-petrou: Rendering of the `ContentType` header is quite expensive. For this reason, at the bottom of the the `Response` companion object the main 3 content-types are cached after calling `untyped` on them (which invokes the `render` method).

Might be better to pattern match here and check if the media type here matches one of those 3 and use the cached value or fallback to `Header.ContentType.apply` if not?

kyri-petrou: Also the param name should be changed to `mediaType`

987Nabil: Maybe it is just the best to use the untyped API. Probably the cheapest.

zio-http/shared/src/main/scala/zio/http/ErrorResponseConfig.scala
```diff
@@ -0,0 +1,43 @@
...
+ *   if true, includes the stack trace in the response body
+ * @param maxStackTraceDepth
+ *   maximum number of stack trace lines to include in the response body. Set to
+ *   0 to include all lines.
+ */
+case class ErrorResponseConfig(
+  withErrorBody: Boolean = false,
+  withStackTrace: Boolean = false,
+  maxStackTraceDepth: Int = 10,
+  errorFormat: ErrorResponseConfig.ErrorFormat = ErrorResponseConfig.ErrorFormat.Html,
```
kyri-petrou, line 20: I'm thinking plaintext might make more sense as a default as it accounts for a wider range of usecases (APIs & serving user content)

zio-http/shared/src/main/scala/zio/http/ErrorResponseConfig.scala
```diff
@@ -0,0 +1,43 @@
...
+    ErrorResponseConfig(withErrorBody = true, withStackTrace = true, maxStackTraceDepth = 0)
+
+  lazy val debug: HandlerAspect[Any, Unit] =
+    Middleware.runBefore(configRef.set(debugConfig))
+
+  def withConfig(config: ErrorResponseConfig): HandlerAspect[Any, Unit] =
+    Middleware.runBefore(configRef.updateSome { case oldConfig if oldConfig != config => config })
+
+  private[http] lazy val configRef: FiberRef[ErrorResponseConfig] =
+    FiberRef.unsafe.make(ErrorResponseConfig())(Unsafe)
```
kyri-petrou: For performance reasons, `FiberRefs` use reference equality to determine whether a `FiberRef` was changed. It's probably not an issue but perhaps use `default` here _just incase_

zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -746,10 +746,49 @@ object Handler extends HandlerPlatformSpecific with HandlerVersionSpecific {
...
   def error(status: => Status.Error, message: => String): Handler[Any, Nothing, Any, Response] =
-    fromResponse(Response.error(status, message))
+    (fromResponse(Response.status(status)) @@ Middleware.interceptHandlerStateful(
+      handler((req: Request) => (req.header(Header.Accept), (req, ()))),
+    ) {
+      handler { (accept: Option[Header.Accept], res: Response) =>
+        ErrorResponseConfig.configRef.get.map { cfg =>
+          if (cfg.withErrorBody) {
+            val mediaType: MediaType = accept
+              .flatMap(_.mimeTypes.sortBy(mt => -mt.qFactor.getOrElse(1.0)).map(_.mediaType).collectFirst {
```
kyri-petrou: Should we add an implicit `Ordering` in the companion object of `Accept`? 🤔

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -58,6 +58,9 @@ final case class Response(
         self.copy(body = Body.fromChunk(bytes))
       }
 
+  def contentType(mediaType: MediaType): Response =
+    self.addHeader("content-type", mediaType.fullType)
```
kyri-petrou, line 64: Outside the scope of this PR, but I'm wondering whether we should change the `fullType` to a `val` instead of a `lazy val` to avoid the penalty of lazy vals.

zio-http/shared/src/main/scala/zio/http/Response.scala
```diff
@@ -206,21 +221,83 @@ object Response {
...
+
+  private def throwableToMessage(throwable: Throwable, status: Status, config: ErrorResponseConfig): Body =
+    if (!config.withErrorBody) Body.empty
+    else {
+      val stackTrace =
+        if (config.withStackTrace && throwable.getStackTrace.nonEmpty)
+          (if (config.maxStackTraceDepth == 0) throwable.getStackTrace
+           else throwable.getStackTrace.take(config.maxStackTraceDepth))
+            .mkString("\n", "\n", "")
+        else ""
```
kyri-petrou: Is `getStackTrace` cheap? Or should we rewrite this to invoke it only once?

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -51,13 +51,13 @@ sealed trait Route[-Env, +Err] { self =>
...
   final def handleError(f: Err => Response)(implicit trace: Trace): Route[Env, Nothing] =
-    self.handleErrorCause(Response.fromCauseWith(_)(f))
+    self.handleErrorCauseZIO(c => ErrorResponseConfig.configRef.get.map(Response.fromCauseWith(c, _)(f)))
 
   final def handleErrorZIO(f: Err => ZIO[Any, Nothing, Response])(implicit trace: Trace): Route[Env, Nothing] =
     self.handleErrorCauseZIO { cause =>
       cause.failureOrCause match {
         case Left(err)    => f(err)
-        case Right(cause) => ZIO.succeed(Response.fromCause(cause))
+        case Right(cause) => ErrorResponseConfig.configRef.get.map(Response.fromCause(cause, _))
```
kyri-petrou: Not as nice, but it requires 1 FlatMap less:
```suggestion
        case Right(cause) => ErrorResponseConfig.configRef.getWith(c => ZIO.succeed(Response.fromCause(cause, c)))
```

There are a few other places this can be applied to

kyri-petrou (commented):
Apologies, I missed some minor things in my last review. Otherwise LGTM

zio-http/shared/src/main/scala/zio/http/Response.scala
987Nabil: I'll double check

zio-http/shared/src/main/scala/zio/http/Response.scala
987Nabil: Since both are static strings, I'll expect the val to be cheap anyway

987Nabil: If this would be scala 3 only, inline def is probably best?

zio-http/shared/src/main/scala/zio/http/Handler.scala
987Nabil: We could also ensure it is always sorted.

zio-http/shared/src/main/scala/zio/http/Route.scala
987Nabil: Still nice enough

zio-http/shared/src/main/scala/zio/http/ErrorResponseConfig.scala
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param withErrorBody
+ *   if true, includes the error message in the response body
+ * @param withStackTrace
+ *   if true, includes the stack trace in the response body
+ * @param maxStackTraceDepth
+ *   maximum number of stack trace lines to include in the response body. Set to
+ *   0 to include all lines.
+ */
+case class ErrorResponseConfig(
```
jdegoes:
```suggestion
final case class ErrorResponseConfig(
```

zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -549,7 +547,7 @@ sealed trait Handler[-R, +Err, -In, +Out] { self =>
     self(in)
 
   final def sandbox(implicit trace: Trace): Handler[R, Response, In, Out] =
-    self.mapErrorCause(Response.fromCause(_))
+    self.mapErrorCauseZIO(c => ErrorResponseConfig.configRef.get.map(Response.fromCause(c, _)).flip)
```
jdegoes, line 550: There's a fair amount of overhead here--of course, there was before, but now there is even more.

Could be useful to unsafely get the fiber ref value in as many places as you can "safely" bundle it up in lazy evaluation (like here).

zio-http/shared/src/main/scala/zio/http/Handler.scala
jdegoes: Best to ensure it's always sorted, we don't want to be doing this on each request.

987Nabil: It has to be done on each request, since this the the requests accept header. But since the amount of accepted media types will be small, I thought a sorted list might be easier to understand. It is just sorted by q factor when ever you deal with the typed interface.

zio-http/shared/src/main/scala/zio/http/Response.scala
987Nabil: I created a local val. I think we could optimize printing limited stack traces, since `getStackTrace` captures the full stack. We could probably walk the trace only until n. But I think this is okay for now.

zio-http/shared/src/main/scala/zio/http/Response.scala
kyri-petrou: Let's tackle this in another issue/PR where we can revise the usages of `lazy val` in ZIO HTTP. I think there are quite a few unnecessary ones

kyri-petrou: https://github.com/zio/zio-http/issues/3051

zio-http/shared/src/main/scala/zio/http/Handler.scala
987Nabil: I tried to figure out, what you mean. But I don't get it.
Currently, this config is saved within a fiber ref. Therefore idk how to get it unsafe. Also idk if the call to sandbox needs to be that optimized? Don't we advertise this as a non-prod feature?
But what you try to say might be true for other places?

One other design option I thought of was, to use a ConcurrentHashMap with the route pattern as a key. That way we could avoid fiber refs and zio based access to our config storage.

zio-http/shared/src/main/scala/zio/http/Handler.scala
jdegoes: Maybe you could check and avoid sorting it if it's already sorted, which would avoid the overhead entirely for the happy path.

zio-http/shared/src/main/scala/zio/http/Handler.scala
jdegoes: Ah right, FiberRef instead of Ref.

jdegoes: Wondering what the thought in using a fiber ref is? To make it compatible with scoped changes?

987Nabil: Yes. That you can override the behavior per route/request.

jdegoes (approved):

PR #3053, opened 2024-08-26
[OpenAPI] case class field being an Option of an ADT are not rendered as nullable

### LINKED ISSUE #3052: [OpenAPI] case class field being an `Option` of an ADT are not rendered as nullable
```scala
sealed trait Example
final case class A(a: Int) extends Example
final case class B(b: String) extends Example

/**
 * This is the Input documentation
 */
final case class Input(example: Option[Example])

val testEndpoint = 
  (Endpoint(RoutePattern.POST / "test") ?? Doc.p("This is my 'POST /test' endpoint doc"))
    .in[Input]
    .out[String](mediaType = MediaType.application.json, doc = Doc.p("this is the output doc"))

val spec: String =
  OpenAPIGen.fromEndpoints(
    title = "This is my OpenAPI doc title",
    version = "0.0.0",
    endpoints = List(testEndpoint)
  ).toJson
```

The generated OpenAPI spec will be:
```json
"Input": {
  "type": "object",
  "properties": {
    "example": {
      "$ref": "#/components/schemas/Example"
    },
  }
}
```
According to https://stackoverflow.com/a/48114924/2431728, it should be:
```json
"Input": {
  "type": "object",
  "properties": {
    "example": [
      "anyOf": [
        { "type": "null" },
        { "$ref": "#/components/schemas/Example" }
      ]
    }
  }
}
```

guizmaii:
Fixes https://github.com/zio/zio-http/issues/3052

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
@@ -2839,7 +2839,6 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
             |          }
             |        },
             |        "required" : [
-            |          "nestedOption",
             |          "nestedList",
             |          "nestedMap",
             |          "nestedSet",
@@ -3067,10 +3066,7 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
             |              { "$ref": "#/components/schemas/SealedTraitCustomDiscriminator" }
             |            ]
             |          }
-            |        },
-            |        "required" : [
-            |          "optionalAdtField"
-            |        ]
+            |        }
             |      },
             |      "Two" :
             |        {
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
@@ -160,6 +160,11 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
     implicit def schema[T: Schema]: Schema[WithGenericPayload[T]] = DeriveSchema.gen
   }
 
+  final case class WithOptionalAdtPayload(optionalAdtField: Option[SealedTraitCustomDiscriminator])
+  object WithOptionalAdtPayload {
+    implicit val schema: Schema[WithOptionalAdtPayload] = DeriveSchema.gen
+  }
+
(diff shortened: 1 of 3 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPIGen.scala
```diff
@@ -988,7 +988,7 @@ object OpenAPIGen {
   }
   private def schemaReferencePath(nominal: TypeId.Nominal, referenceType: SchemaStyle): String = {
     referenceType match {
-      case SchemaStyle.Compact => s"#/components/schemas/${nominal.typeName}}"
```
guizmaii, line 991: Nasty typo, which was sadly validated by the tests here https://github.com/zio/zio-http/pull/3053/files#diff-0b5e5bc79ca85db4957d0ab782e9509a15f69b70d84ffd76a15a712367a860f3L2234-L2236

zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
```diff
@@ -2989,6 +2997,114 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
...
+            |        {
+            |        "type" :
+            |          "object",
+            |        "properties" : {
+            |          "optionalAdtField" : {
+            |            "anyOf": [
+            |              { "type": "null" },
+            |              { "$ref": "#/components/schemas/SealedTraitCustomDiscriminator" }
+            |            ]
+            |          }
```
guizmaii, line 3068: We now correctly have the `anyOf`

987Nabil: I wonder, if this should be anyOf, or if the field should just be missing from the required fields list. I'd tend to the second. Most Scala json serialization does not write the field if the option is None

987Nabil: Maybe both. To be accept all possible encodings

guizmaii: I made the change. Now, we have both 🙂

guizmaii: Done. They are now both nullable and non-required

jdegoes (approved):

PR #3059, opened 2024-08-28
Fixed #3058 - added securityRequirementSchema to ignore securitySchem…

### LINKED ISSUE #3058: [OpenAPI] SecurityRequirement case class produces invalid open api json
**Describe the bug**
When generating OpenApi json. SecurityRequirement adds securitySchemes object that is not a part of OpenApi 3.1.0 specification. For this reason, Swagger Ui is ignoring the security object and it is not forwarding the header/apiKey that is defined to the backend.

**To Reproduce**
Go to OpenAPISpec test. In the expected result you see security defined as
```
  "security" : [
    {
      "securitySchemes" : {
        "apiKeyAuth" : []
      }
    }
  ]
```
**Expected behaviour**
The test is passing because this is not a valid OpenApi 3.1.0 json. If you read the [speficifation](https://spec.openapis.org/oas/v3.1.0.html#security-requirement-object) you will see that  securitySchemes is redundant and the test should look like this:
```
  "security" : [
    {
      "apiKeyAuth" : []
    }
  ]
  ```

kdelija:
…es field when generating open api json

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPI.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPI.scala
@@ -252,6 +252,15 @@ object OpenAPI {
         t => Right(Map(t._1 -> t._2)),
       )
 
+  implicit def securityRequirementSchema: Schema[SecurityRequirement] = {
+    zio.schema
+      .Schema[Map[String, List[String]]]
+      .transform[SecurityRequirement](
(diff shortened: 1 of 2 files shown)


jdegoes (approved):

jdegoes (approved):
I would love it if the implicit were in the companion object of `SecurityRequirement`, rather than just randomly here, but it seems that's the existing pattern in this part of the code base, so 👍 .

PR #3060, opened 2024-08-28
Use an unbounded queue in `AsyncBody#asStream`

### LINKED ISSUE #2297: Server streams request bodies forever when not completed normally

kyri-petrou:
/fixes #2297
/claim #2297

As I mentioned in https://github.com/zio/zio-http/pull/2402#issuecomment-2315114560, the problem with this issue (as well as what #2402 was trying to resolve) is that in the case of an ultra-fast producer, the ZStream buffer queue becomes full, which ends up blocking Netty's event loop. Since Netty's event loop is... event-based, once a connection is is closed, reading from the channel will keep returning messages until all of them have been consumed.

The solution here is to use an unbounded queue to construct the `ZStream`. By using an unbounded queue, we no longer block Netty's event loop in cases that the buffer becomes full. This means that when the connection is interrupted we can shutdown the queue / ZStream via `callback.fail(cause)`.

PS: While it might sound dangerous not to have a backpressure mechanism on the ZStream, in reality the backpressure needs to be applied on Netty's side via Netty's config. Otherwise, even if the ZStream has backpressure, Netty's queues will still be filled and memory issues will arise. See https://github.com/zio/zio-http/pull/2402#issuecomment-2315114560 for more info

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
@@ -19,14 +19,13 @@ package zio.http.netty
 import java.nio.charset.Charset
 
 import zio._
-import zio.internal.OneShot
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
-import zio.stream.ZStream
+import zio.stream.{Take, ZChannel, ZStream}
 
 import zio.http.Body.UnsafeBytes
 import zio.http.internal.BodyEncoding
-import zio.http.{Body, Boundary, Header, MediaType}
+import zio.http.{Body, Header}
 
 import io.netty.buffer.{ByteBuf, ByteBufUtil}
 import io.netty.util.AsciiString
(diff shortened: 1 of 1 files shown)


jdegoes (approved):

jdegoes:
@kyri-petrou I love it! ❤️ 🚀  Good to merge when CI is passing.

PR #3061, opened 2024-08-28
Avoid secondary super type-checking on `UnsafeBytes`

kyri-petrou:
There are a few places in zio-http where we pattern-match / use `isInstanceOf[UnsafeBytes]` on a `Body`.

There is a [bug in the JVM](https://bugs.openjdk.org/browse/JDK-8331341) (fixed in JDK23+) that affects type-checking on secondary supers causing them to be slow / not scaling well under contention. To prevent this issue, we make `UnsafeBytes` an abstract class and make any concrete implementations of it extend it directly (since it also extends `Body`)


jdegoes (approved):

kyri-petrou:
Linked the wrong JDK issue, this is the OG one: https://bugs.openjdk.org/browse/JDK-8180450

PR #3063, opened 2024-08-29
Fixed Server.Config enforcing kebab case

### LINKED ISSUE #2803: ZIO HTTP Doesn't Respect Bash Naming Style (Environment Variables)
Assume I have the following ZIO HTTP App:

```scala
object MainApp extends ZIOAppDefault {
  def run =
    Server.serve(
      Routes(
        Method.GET / "hello" -> handler(Response.ok),
      ),
    ).provide(
      Server.live,
      ZLayer.fromZIO(ZIO.config(Server.Config.config)).debug)
    )
}
```

If of set:

```
export BINDING_PORT=3000
```

But it doesn't respect it and after debugging find out that it requires `BINDING-PORT`

The workaround is to map key names using `Server.Config.config.mapKey(_.replace('-', '_'))` which is not idiomatic.

42njaber:
/fixes #2803
/claim #2803

Hi, this is my first pull request on this project so please let me know if I'm doing anything incorrectly.

The issue was caused by using "-" in the config names instead of letting the ConfigProvider class handle it using its own path separator.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Server.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Server.scala
@@ -200,22 +200,25 @@ object Server extends ServerPlatformSpecific {
   object Config {
     lazy val config: zio.Config[Config] = {
       SSLConfig.config.optional ++
-        zio.Config.string("binding-host").optional ++
-        zio.Config.int("binding-port").withDefault(Config.default.address.getPort) ++
-        zio.Config.boolean("accept-continue").withDefault(Config.default.acceptContinue) ++
-        zio.Config.boolean("keep-alive").withDefault(Config.default.keepAlive) ++
-        Decompression.config.nested("request-decompression").withDefault(Config.default.requestDecompression) ++
-        ResponseCompressionConfig.config.nested("response-compression").optional ++
-        RequestStreaming.config.nested("request-streaming").withDefault(Config.default.requestStreaming) ++
-        zio.Config.int("max-initial-line-length").withDefault(Config.default.maxInitialLineLength) ++
-        zio.Config.int("max-header-size").withDefault(Config.default.maxHeaderSize) ++
-        zio.Config.boolean("log-warning-on-fatal-error").withDefault(Config.default.logWarningOnFatalError) ++
-        zio.Config.duration("graceful-shutdown-timeout").withDefault(Config.default.gracefulShutdownTimeout) ++
-        zio.Config.duration("idle-timeout").optional.withDefault(Config.default.idleTimeout) ++
-        zio.Config.boolean("avoid-context-switching").withDefault(Config.default.avoidContextSwitching) ++
-        zio.Config.int("so-backlog").withDefault(Config.default.soBacklog) ++
-        zio.Config.boolean("tcp-nodelay").withDefault(Config.default.tcpNoDelay)
-
+        zio.Config.string.nested("binding", "host").optional ++
+        zio.Config.int.nested("binding", "port").withDefault(Config.default.address.getPort) ++
+        zio.Config.boolean.nested("accept", "continue").withDefault(Config.default.acceptContinue) ++
+        zio.Config.boolean.nested("keep", "alive").withDefault(Config.default.keepAlive) ++
+        Decompression.config.nested("request", "decompression").withDefault(Config.default.requestDecompression) ++
+        ResponseCompressionConfig.config.nested("response", "compression").optional ++
+        RequestStreaming.config.nested("request", "streaming").withDefault(Config.default.requestStreaming) ++
+        zio.Config.int.nested("max", "initial", "line", "length").withDefault(Config.default.maxInitialLineLength) ++
+        zio.Config.int.nested("max", "header", "size").withDefault(Config.default.maxHeaderSize) ++
+        zio.Config.boolean
+          .nested("log", "warning", "on", "fatal", "error")
+          .withDefault(Config.default.logWarningOnFatalError) ++
+        zio.Config.duration
+          .nested("graceful", "shutdown", "timeout")
+          .withDefault(Config.default.gracefulShutdownTimeout) ++
+        zio.Config.duration.nested("idle", "timeout").optional.withDefault(Config.default.idleTimeout) ++
+        zio.Config.boolean.nested("avoid", "context", "switching").withDefault(Config.default.avoidContextSwitching) ++
+        zio.Config.int.nested("so", "backlog").withDefault(Config.default.soBacklog) ++
+        zio.Config.boolean.nested("tcp", "nodelay").withDefault(Config.default.tcpNoDelay)
     }.map {
       case (
             sslConfig,
@@ -281,7 +284,9 @@ object Server extends ServerPlatformSpecific {
     object ResponseCompressionConfig {
       lazy val config: zio.Config[ResponseCompressionConfig] =
         (
-          zio.Config.int("content-threshold").withDefault(ResponseCompressionConfig.default.contentThreshold) ++
+          zio.Config.int
+            .nested("content", "threshold")
+            .withDefault(ResponseCompressionConfig.default.contentThreshold) ++
             CompressionOptions.config.repeat.nested("options")
         ).map { case (contentThreshold, options) =>
           ResponseCompressionConfig(contentThreshold, options)
@@ -428,7 +433,7 @@ object Server extends ServerPlatformSpecific {
 
     lazy val config: zio.Config[RequestStreaming] =
       (zio.Config.boolean("enabled").withDefault(true) ++
-        zio.Config.int("maximum-content-length").withDefault(1024 * 100)).map {
+        zio.Config.int.nested("maximum", "content", "length").withDefault(1024 * 100)).map {
         case (true, _)          => Enabled
         case (false, maxLength) => Disabled(maxLength)
       }


42njaber:
Would it be preferable that I create a test case for this issue and add it to the PR?

jdegoes (approved):

jdegoes:
@42njaber A test case would be great, if you can! 🙏

guersam:
AFAIK, the purpose of `.nested` is for namespacing rather than loosening naming convention. For example, `keep-alive` is a whole concept by itself and having a separate `keep` namespace is nonsence.

To adapt different naming conventions, you can configure the config providers using the helpers like `ConfigProvider#kebabCase`.

guersam:
BTW, there are some inconsistency how ZIO HTTP describes configs.

For example, there are some cases where the config names are in camelCase rather than kebab-case. ([code](https://github.com/zio/zio-http/blob/b4dde6afae757d136885e08a263475c9abae329d/zio-http/shared/src/main/scala/zio/http/SSLConfig.scala#L145-L152))

987Nabil:
The real question is, if I have a config name in kebab case, how should it be translated for the ENV var.
I don't think this is a zio-http but zio issue. The fix here does not solve the actual problem, but changes the config in a way to fit the expected env var naming.
But if I would have had a hocon `binding-port` before, it would fail now.

987Nabil:
There is some info on how spring is doing it [here](https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.files.property-placeholders)

Anyhow, I think this should be fixed in the config provider not here (Probably translating `-` to `_` or just removing it)

guersam:
`zio.ConfigProvider.env` is already doing it for us. 

https://github.com/zio/zio/blob/dd98613cd887ef538f378b3b26cb161d9cd4ee53/core/shared/src/main/scala/zio/ConfigProvider.scala#L450

987Nabil:
@guersam why is it then failing currently? Something seems odd

42njaber:
Hi everyone, thanks for commenting, I think I should have explained my thought process more in the original message.

As I investigated the issue, I noticed that the main issue here is that namespacing and word spacing cannot be consistently represented with the way the `ConfigProvider` trait currently works with different casings. To be clear, word spacing shows hardly any means for cross-compatibility between different config providers while namespacing does have a consitent interface through `.nested()`. Furthering the issue is that it's hard to conceptualize an implementation that meaningfully distinguishes between word spacing and namespacing across all config providers, and that's exactly the issue with bash variables.

The `path_seq` variables in the various `ConfigProvider` implementations provide the means for cross-compatible namespacing, but they are not parsed in `Config` definitions, that do not have access to them and are assumed to be well-formed (which, in my opinion, means casing-independant). Instead, their functionality is accessed through `.nested()` so I chose to go with that solution.

Other solutions would include :
 - Re-defining the `ConfigProvider` trait to provide namespacing in a different way that might however be more awkward to use across different configuration methods (like nesting variable definitions in the variable values), thus freeing up casing to be standardized as the word spacing mechamism across providers.
 - Providing a method for `Config` to specify the difference between word spacing and namespacing, while dissallowing any case-specific characters in variable name definitions, which might break backwards compatibility, and requires a good rework of how configuration "paths" are read and stored, since `path_seq` will need to be interpreted as either word spacing or namespacing depending  on the provided `Config` *and* there needs to be some sort of conflict resolution between the 2.

While I believe that the later method is the ideal one, I think that this amount of change is not warranted simply by this issue and I can't tell if this would generalize into more frequent user issues in the future.

I'm hoping for your feedback on this and I'll think on what tests needs to be implemented - and whether it should be here or in zio - based on what people more experienced with this project think about my reasoning.

zio-http/shared/src/main/scala/zio/http/Server.scala
```diff
@@ -200,22 +200,25 @@ object Server extends ServerPlatformSpecific {
...
-        zio.Config.boolean("tcp-nodelay").withDefault(Config.default.tcpNoDelay)
-
+        zio.Config.string.nested("binding", "host").optional ++
+        zio.Config.int.nested("binding", "port").withDefault(Config.default.address.getPort) ++
+        zio.Config.boolean.nested("accept", "continue").withDefault(Config.default.acceptContinue) ++
+        zio.Config.boolean.nested("keep", "alive").withDefault(Config.default.keepAlive) ++
+        Decompression.config.nested("request", "decompression").withDefault(Config.default.requestDecompression) ++
+        ResponseCompressionConfig.config.nested("response", "compression").optional ++
+        RequestStreaming.config.nested("request", "streaming").withDefault(Config.default.requestStreaming) ++
+        zio.Config.int.nested("max", "initial", "line", "length").withDefault(Config.default.maxInitialLineLength) ++
```
987Nabil, line 210: in relation to your latest comment @42njaber, I think this kind of nesting is not what we want. Why? Because we create a lot of name spaces. In my opinion, namespaces should be few and preferable not contain only one member. Also they should logically group things.
While I can see that for `binding`, I can't see that for `max`.
My suggestion would be, that one config object in the code should relate to one namespace

42njaber: I understand and semantically that makes a lot of sense but this does not reflect the implementation of `Config` and `ConfigProvider`. I am not knowledgeable of the project to understand the intent behind the implementation, and thus what can be changed, but I can explain my view of the issue.

The current implementation is very simple :
In zio's `Config.scala`, line 76
```scala
  def nested(name: => String): Config[A] =
    Config.defer(Config.Nested(name, self))
```
And line 317
```scala
  final case class Nested[A](name: String, config: Config[A])                                    extends Composite[A]
```

In `ConfigProvide.scala`, line 268
```scala
            case Nested(name) :: tail =>
              loop(name +: path, tail)
```
And line 622
```scala
  /**
   * Constructs a ConfigProvider using a map and the specified delimiter string,
   * which determines how to split the keys in the map into path segments.
   */
  def fromMap(map: Map[String, String], pathDelim: String = ".", seqDelim: String = ","): ConfigProvider =
    fromFlat(new Flat {
      val escapedSeqDelim   = java.util.regex.Pattern.quote(seqDelim)
      val escapedPathDelim  = java.util.regex.Pattern.quote(pathDelim)
      val mapWithIndexSplit = splitIndexInKeys(map, unmakePathString, makePathString)
```

So, `.nested()` does not create any namespace, and simply defers the resolution of the config name to a config provider that has split the string along a "path" which might or might not refer to namespaces.

There is not much mechanism for case conversions, although these could be extended, but I'm getting the feeling that `.nested()` does not actually do what it is semantically meant to do, since it only matches "path segments", so it is either miss-named or is planned to be developed further in the future?

That's why I proposed different solutions, but I'm not quite sure what the intended design is, what would be useful for users, and how these features are being used in general. There seems to simply be no mechanism for distinguishing between casing and nesting in `zio`, and I'd want to have the opinion of somebody more experienced on how it *should* be, since you pointed out `.nested()` should not be used for casing conversion, only namespacing.

This bug report implies that it is intended for config providers to convert between different casings, creating compatibility between different config providers, but there seems to be no apparent generinc method for that other then using `.nested()`.

kyri-petrou:
One of the main issues with this PR is that it'll potentially alter user's applications behaviour, perhaps in a way that's not easy to detect during tests.

I think a better approach that's also backwards compatible is to do something along these lines:

```scala
zio.Config.string("binding-host") || zio.Config.string.nested("binding", "host")
```

You can probably keep things DRY by creating a method to do this automatically as well.

As a side-note: This might be something that should be done/fixed in ZIO directly, to have a fallback that splits on `-` and uses a nested structure instead

zio-http/shared/src/main/scala/zio/http/Server.scala
guersam: There are other providers like Typesafe Config or AWS Parameter Store that have similar concepts to namespace. Here's an example using Typesafe Config (HOCON):

```hocon
server {
  port = 8080
  keep-alive = true
}

logging {
  root-level = INFO
  mappings {
    "some.zio.module" = WARN
  }
}
```

In this case, the server section can be described in ZIO config like `(int("port") ++ boolean("keep-alive")).nested("server")`, which has different semantic from `(int("port") ++ boolean("alive").nested("keep")).nested("server")`.

guersam:
Here's an attempt to reproduce in Scala 3.5.0 with ZIO HTTP 3.0.0-RC10.

`Main.scala`
```scala
//> using scala "3.5.0"
//> using dep "dev.zio::zio-http:3.0.0-RC10"

import zio.* 
import zio.http.*

  val run =
    ZIO.config(Server.Config.config)
      .map(_.address.getPort())
      .debug("binding port")
}
```

```bash
$ scala Main.scala
binding port: 8080
$ BINDING_PORT=9008 scala Main.scala
binding port: 9008
```

At first I suspected the missing config provider in `ZIOApp#bootstrap`, but it works as expected without additional setup.


p.s. out of the scope, but I'd want the configs to be implicit to align with the new `ZIO.config` API ([code](https://github.com/zio/zio/blob/v2.1.9/core/shared/src/main/scala/zio/ZIO.scala#L3107-L3112)) so that `ZIO.config(Server.Config.config)` can be `ZIO.config[Server.Config]`.

987Nabil:
The code of @guersam would suggest that no code changes are need or? Is this bug reproduceable on main with any scala version?

kyri-petrou:
Aha, seems this was fixed in ZIO v2.1.0 https://github.com/zio/zio/pull/8680

987Nabil:
Since this was a bug in zio that is already fixed, I'll close this PR and the issue.

PR #3065, opened 2024-08-29
Fix OpenApi gen components from case class with arity > 22

### LINKED ISSUE #3064: OpenAPI generation doesn't work for case classes with >22 fields
**Describe the bug**
I'm using `3.0.0-RC9` version.
When generating an OpenAPI from schema which describes a case class with >22 fields, invalid specification is produced which generates an error in SwaggerUI.

**To Reproduce**
Here's the code snippet
```scala
import zio._
import zio.http._
import zio.http.endpoint.Endpoint
import zio.http.endpoint.openapi.OpenAPIGen
import zio.schema.{ DeriveSchema, Schema }

object HttpApp extends ZIOAppDefault {

  override def run: ZIO[Any, Any, Any] =
    ZIO.debug(OpenAPIGen.fromEndpoints(Endpoint(Method.GET / "foo").out[BigModel]).toJsonPretty)

  private implicit lazy val bigModelSchema: Schema[BigModel] = DeriveSchema.gen

  final case class BigModel(
    f1: Boolean,
    f2: Boolean,
    f3: Boolean,
    f4: Boolean,
    f5: Boolean,
    f6: Boolean,
    f7: Boolean,
    f8: Boolean,
    f9: Boolean,
    f10: Boolean,
    f11: Boolean,
    f12: Boolean,
    f13: Boolean,
    f14: Boolean,
    f15: Boolean,
    f16: Boolean,
    f17: Boolean,
    f18: Boolean,
    f19: Boolean,
    f20: Boolean,
    f21: Boolean,
    f22: Boolean,
    f23: Boolean
  )

}
```
**Actual behaviour**
The code prints specification without components schemas
```json
{
  "openapi" : "3.1.0",
  "info" : {
    "title" : "",
    "version" : ""
  },
  "paths" : {
    "/foo" : {
      "get" : {
        "responses" : {
          "200" : 
            {
            "content" : {
              "application/json" : {
                "schema" : 
                  {
                  "$ref" : "#/components/schemas/BigModel"
                }
              }
            }
          }
        }
      }
    }
  },
  "components" : {
    
  }
}
```

**Expected behaviour**
It should print the referenced schema. For example, here's what it prints if we leave only 2 fields (I don't want to include result for 22 fields since it's too large)
```json
{
  "openapi" : "3.1.0",
  "info" : {
    "title" : "",
    "version" : ""
  },
  "paths" : {
    "/foo" : {
      "get" : {
        "responses" : {
          "200" : 
            {
            "content" : {
              "application/json" : {
                "schema" : 
                  {
                  "$ref" : "#/components/schemas/BigModel"
                }
              }
            }
          }
        }
      }
    }
  },
  "components" : {
    "schemas" : {
      "BigModel" : 
        {
        "type" : 
          "object",
        "properties" : {
          "f1" : {
            "type" : 
              "boolean"
          },
          "f2" : {
            "type" : 
              "boolean"
          }
        },
        "required" : [
          "f1",
          "f2"
        ]
      }
    }
  }
}
```

**Screenshots**
image

**Additional context**
The issue seems to be in `zio.http.endpoint.openapi.OpenAPIGen#nominal` method. When dealing with case classes with > 22 fields, zio-schema wraps `GenericRecord` into `Transform` class (idk why), which prevents it from matching and just skips it. Both Scala 2 and Scala 3 versions are affected

narma:
Resolves #3064 
/claim  #3064 

Currently zio-schema derives schema for case classes with arity > 22 fields via Transform from/to map. Both For [scala 2](https://github.com/zio/zio-schema/blob/v1.4.0/zio-schema-derivation/shared/src/main/scala-2/zio/schema/DeriveSchema.scala#L380) and [scala 3](https://github.com/zio/zio-schema/blob/v1.4.0/zio-schema-derivation/shared/src/main/scala-3/zio/schema/DeriveSchema.scala#L305).

Current PR add support for this case as it done in other places, for example:
https://github.com/zio/zio-http/blob/b4dde6afae757d136885e08a263475c9abae329d/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala#L442
https://github.com/zio/zio-http/blob/b4dde6afae757d136885e08a263475c9abae329d/zio-http-cli/src/main/scala/zio/http/endpoint/cli/HttpOptions.scala#L142

### DIFF
--- a/zio-http-gen/src/test/scala/zio/http/gen/openapi/EndpointGenSpec.scala
+++ b/zio-http-gen/src/test/scala/zio/http/gen/openapi/EndpointGenSpec.scala
@@ -9,7 +9,7 @@ import zio.http._
 import zio.http.codec._
 import zio.http.endpoint._
 import zio.http.endpoint.openapi.JsonSchema.SchemaStyle.{Compact, Inline}
-import zio.http.endpoint.openapi.{OpenAPI, OpenAPIGen}
+import zio.http.endpoint.openapi.{JsonSchema, OpenAPI, OpenAPIGen}
 import zio.http.gen.model._
 import zio.http.gen.scala.Code
 import zio.http.gen.scala.Code.Collection.Opt
@@ -593,6 +593,24 @@ object EndpointGenSpec extends ZIOSpecDefault {
             enums = Nil,
           )
           assertTrue(scala.files.head == expected)
+        },
+        test("response are case class with 22+ fields") {
+          val endpoint = Endpoint(Method.GET / "api" / "v1" / "users").out[BigModel]
+          val openAPI  = OpenAPIGen.fromEndpoints(endpoint)
+          assertTrue(
+            openAPI.components.is(_.some).schemas.nonEmpty,
+            openAPI.components.is(_.some).schemas.headOption.is(_.some)._1.name == "BigModel",
+            openAPI.components
+              .is(_.some)
+              .schemas
+              .headOption
+              .is(_.some)
+              ._2
+              .asJsonSchema
+              .is(_.subtype[JsonSchema.Object])
+              .properties
+              .size == 23,
+          )
         },
         test("seq request") {
           val endpoint = Endpoint(Method.GET / "api" / "v1" / "users").in[Chunk[User]]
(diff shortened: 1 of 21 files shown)


narma:
Looks like current failure in test not related to this PR. Probably it because of https://github.com/zio/zio/issues/9152

zio-http-gen/src/test/scala/zio/http/gen/openapi/EndpointGenSpec.scala
```diff
@@ -594,6 +594,24 @@ object EndpointGenSpec extends ZIOSpecDefault {
           )
           assertTrue(scala.files.head == expected)
         },
+        test("response are case class with 22+ fields") {
```
987Nabil: You need to add a test to the OpenAPIGenSepc. You can keep this spec, but please let it look like the other specs

987Nabil (changes requested):

987Nabil (approved):

PR #3069, opened 2024-08-30
[gen] normalized fields with annotations holding original field name

### LINKED ISSUE #3031: [gen] support for normalized fields
APIs are messy, fields come in all flavors: snake_case, kebab-case, or even whitespaced tokens.
While illegal scala variable names are backtick escaped since #2839, we still end up with unconventional (ugly) generated code.
Even legal names, like snake_case, or UPPER_SNAKE which are not escaped does not look "pretty" in generated code.

Add a flag in configuration, to allow change the field name, and add an annotation like:
```scala
import zio.schema.annotation.fieldName

case class User(
  @fieldName("user_id") userId: String,
  …
)
```
There would have to be some parser with heuristics involved. Which tokens casing to preserve, capitalize, lowercase, etc'…

putting up with ugly generated code? 😅 

**Additional context**
There is no "right" answer on how to parse & transform field names.
e.g: perhaps `arn_API-Gateway` should preserve some casing and be converted to `arnAPIGateway`?
Or perhaps we prefer to always use capitalized tokens, and even break `arn_APIGatway` to `arnApiGateway`?
Anything would be better than backtick escaped
```
`arn_API-Gateway`
```

Moreover, maybe the configuration can also allow customization for some "unparseable" replacements (like when starting with a non-letter char), e.g:
```
"1st" -> "first
"2nd" -> "second"
"3rd" -> "third"
…
```

This might be only "cosmetics", but it can really improve generated code quality.

hochgi:
PR ready for review

fixes #3031 
/claim #3031

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -8,6 +8,7 @@ object Dependencies {
   val ZioVersion                    = "2.1.9"
   val ZioCliVersion                 = "0.5.0"
   val ZioJsonVersion                = "0.7.1"
+  val ZioParserVersion              = "0.1.10"
   val ZioSchemaVersion              = "1.4.1"
   val SttpVersion                   = "3.3.18"
   val ZioConfigVersion              = "4.0.2"
@@ -36,7 +37,11 @@ object Dependencies {
 
   val zio                   = "dev.zio" %% "zio"                 % ZioVersion
   val `zio-cli`             = "dev.zio" %% "zio-cli"             % ZioCliVersion
+  val `zio-config`          = "dev.zio" %% "zio-config"          % ZioConfigVersion
+  val `zio-config-magnolia` = "dev.zio" %% "zio-config-magnolia" % ZioConfigVersion
+  val `zio-config-typesafe` = "dev.zio" %% "zio-config-typesafe" % ZioConfigVersion
   val `zio-json-yaml`       = "dev.zio" %% "zio-json-yaml"       % ZioJsonVersion
+  val `zio-parser`          = "dev.zio" %% "zio-parser"          % ZioParserVersion
   val `zio-streams`         = "dev.zio" %% "zio-streams"         % ZioVersion
   val `zio-schema`          = "dev.zio" %% "zio-schema"          % ZioSchemaVersion
   val `zio-schema-json`     = "dev.zio" %% "zio-schema-json"     % ZioSchemaVersion
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -41,7 +41,6 @@ object Dependencies {
   val `zio-config-magnolia` = "dev.zio" %% "zio-config-magnolia" % ZioConfigVersion
   val `zio-config-typesafe` = "dev.zio" %% "zio-config-typesafe" % ZioConfigVersion
   val `zio-json-yaml`       = "dev.zio" %% "zio-json-yaml"       % ZioJsonVersion
-  val `zio-parser`          = "dev.zio" %% "zio-parser"          % ZioParserVersion
   val `zio-streams`         = "dev.zio" %% "zio-streams"         % ZioVersion
   val `zio-schema`          = "dev.zio" %% "zio-schema"          % ZioSchemaVersion
   val `zio-schema-json`     = "dev.zio" %% "zio-schema-json"     % ZioSchemaVersion
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
@@ -1,5 +1,9 @@
 package zio.http.gen.openapi
 
+import zio.config.ConfigOps
+
+import zio.http.gen.openapi.Config.NormalizeFields
+
 final case class Config(
   commonFieldsOnSuperType: Boolean /*
    * oneOf expressions in openapi result in sealed traits in generated scala code.
@@ -8,7 +12,7 @@ final case class Config(
    * then the common fields from that shared object(s) will result in abstract fields
    * defined on the sealed trait.
    */,
-  generateSafeTypeAliases: Boolean, /*
+  generateSafeTypeAliases: Boolean /*
    * Referencing primitives, and giving them a name makes the openapi spec more readable.
    * By default, the generated scala code will resolve the referenced name,
    * and replace it with the primitive type.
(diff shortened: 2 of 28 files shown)


zio-http-gen/src/test/scala/zio/http/gen/scala/CodeGenSpec.scala
```diff
@@ -771,5 +771,5 @@ object CodeGenSpec extends ZIOSpecDefault {
           }
         }
       },
-    ) @@ java11OrNewer @@ flaky @@ blocking // Downloading scalafmt on CI is flaky
+    ) @@ java11OrNewer /*@@ flaky*/ @@ blocking // Downloading scalafmt on CI is flaky
```
hochgi: TODO: don't forget to re-enable

zio-http-gen/src/test/resources/inline_schema_weird_field_names.yaml
hochgi, line 1: this file is C&Ped from a different test, and I only test the order class - I should ditch large parts of its less relevant content.

zio-http-gen/src/test/resources/ComponentOrderWithAliases.scala
```diff
@@ -1,12 +1,14 @@
...
+import zio.schema.annotation.validate
+import zio.schema.validation.Validation
 
 case class Order(
   id: OrderId.Type,
   product: String,
-  @zio.schema.annotation.validate[Int](zio.schema.validation.Validation.greaterThan(0)) quantity: Int,
-  @zio.schema.annotation.validate[Double](zio.schema.validation.Validation.greaterThan(-1.0)) price: Double,
+  @validate[Int](Validation.greaterThan(0)) quantity: Int,
+  @validate[Double](Validation.greaterThan(-1.0)) price: Double,
```
hochgi, line 11: Not related to current PR, but the float/double validation logic is faulty.
`minimum 0` does not mean `> -1.0`
Since e.g: `-0.5 > -1.0`, but does not adhere to original minimum requirement.

CC @987Nabil

project/Dependencies.scala
```diff
@@ -36,7 +37,11 @@ object Dependencies {
 
   val zio                   = "dev.zio" %% "zio"                 % ZioVersion
   val `zio-cli`             = "dev.zio" %% "zio-cli"             % ZioCliVersion
+  val `zio-config`          = "dev.zio" %% "zio-config"          % ZioConfigVersion
+  val `zio-config-magnolia` = "dev.zio" %% "zio-config-magnolia" % ZioConfigVersion
+  val `zio-config-typesafe` = "dev.zio" %% "zio-config-typesafe" % ZioConfigVersion
   val `zio-json-yaml`       = "dev.zio" %% "zio-json-yaml"       % ZioJsonVersion
+  val `zio-parser`          = "dev.zio" %% "zio-parser"          % ZioParserVersion
```
hochgi: zio-parser implementation wasn't good enough (parser was very slow - about a minute runtime per call. equivalent scala-parser-combinators implementation was fast).
I ended up replacing with a good enough regex (with the help of co-pilot).
This dependency is currently not used, and can be removed.

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
```diff
@@ -19,20 +23,55 @@ final case class Config(
...
 )
 object Config {
 
+  final case class NormalizeFields(
+    enabled: Boolean /*
+     * If enabled, the generator will attempt to normalize field names to camelCase,
+     * unless original field is defined in the specialReplacements map.
+     */,
+    specialReplacements: Map[String, String], /*
+     * When normalization is enabled, a heuristic parser will attempt to normalize field names.
```
hochgi: Map can be used also when auto normalization is not enabled. I should amend the documentation to be clearer

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -729,14 +730,14 @@ final case class EndpointGen(config: Config) {
...
-    obj.properties.map { case (name, schema) =>
+    obj.properties.toList.sortBy(_._1).map { case (name, schema) =>
       val field = schemaToField(schema, openAPI, name, annotations)
         .getOrElse(
           throw new Exception(s"Could not generate code for field $name of object $name"),
         )
         .asInstanceOf[Code.Field]
       if (obj.required.contains(name)) field else field.copy(fieldType = field.fieldType.opt)
-    }.toList
+    }
```
hochgi: the `obj.properties: Map[String, JsonSchema]` when converted to a list has inconsistent ordering in different Scala versions. Ideally, we should preserve the order of fields as appears originally in OpenAPI .
Not sure how to pull that off, or if its even worth the hassle.
Any suggestions @987Nabil ?

hochgi: I opted for sorting by (original) name, which is at least consistent.

project/Dependencies.scala
987Nabil: Please remove it then. Btw, we have RichTextCodec, which might help.

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
```diff
@@ -19,20 +23,55 @@ final case class Config(
...
+   * OpenAPI can declare fields that have unconventional "casing" in scala,
+   * like snake_case, or kebab-case.
+   * This configuration allows to normalize these fields.
+   * The original casing will be preserved via a @fieldName("<original-name>") annotation.
    */
 )
 object Config {
 
+  final case class NormalizeFields(
+    enabled: Boolean /*
```
987Nabil: Block comments belong before not after what is commented. Anyhow, I think this should be comments for the config class
```
  /**
     * 
     * @param enabled
     */
```

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
987Nabil: If `specialReplacements` can be used out side of normalization, does it belong in the normalization config?

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
```diff
@@ -19,20 +23,55 @@ final case class Config(
...
+        zio.Config
+          .table("special-replacements", zio.Config.string)
+          .withDefault(Config.default.fieldsNormalizationConf.specialReplacements)
+    ).to[NormalizeFields]
+  }
+
   val default: Config = Config(
     commonFieldsOnSuperType = false,
     generateSafeTypeAliases = false,
+    fieldsNormalizationConf = NormalizeFields(
```
987Nabil: if you keep this variable, remove the conf from the name. We are in conf context anyway

zio-http-gen/src/main/scala/zio/http/gen/openapi/Config.scala
hochgi: I meant of course it's part of normalization. but it overrides **auto** normalization (i.e: using the parser to construct a different term).
Normalization has 2 modes:
- manual replacements (user control exactly what to replace with what)
- automatic (a heuristic parser will try to do the best it can)

both fall under normalization.
Perhaps I should rename as: `automaticEnabled` & `manualReplacements`?

987Nabil: Would turning the automated off lead to invalid code? Should it be always on? And the manual just be tried first, if the map is not empty?

987Nabil: Would turning the automated off lead to invalid code? Should it be always on? And the manual just be tried first, if the map is not empty?

hochgi: no, not at all.
These are decoupled.

having e.g. auto off, and replacement on can yield from e.g:
```yaml
Thing:
  type: object
  properties:
    S/N:
      type: string
      format: uuid
    N_A_M_E:
      type: String
```
with only replacing `"N_A_M_E" -> "name"` something like:
```scala
case class Thing (
  `S/N`: UUID,
  name: String
)
```

which is perfectly fine.

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: As discussed over discord (and to persist in a visible place), fields inconsistent ordering stem from zio-schema internally using a `Map` to represent a Dictionary:
https://github.com/zio/zio-schema/blob/c741f6b3b771d7e08c6936f9f54539735a71aa78/zio-schema/shared/src/main/scala/zio/schema/DynamicValue.scala#L95-L104

We've managed to avoid it so far, since almost all test cases are small (4 fields or less), and Map1, …, Map4 implementations across different scala versions remain consistent in term of element ordering.

When I used a test with 7 fields, it exposed the fact that real maps (> 4 elements) are inconsistent in different scala versions.

I tried replacing Map with ListMap, but the derived deserializer is already inconsistent.

@987Nabil suggests it may even be an issue in zio-json (is internal Json object relies on Map or Chunk[(String, Json)]?)

I ended up reducing the test to only 4 fields, and now it's consistent.
But the issue remains.

What's the risk?
Someone relies on generated code, and have same type fields, e.g: `case class Thing(a: String, b: String, …, e: String)`
by simply upgrading scala version, fields order may change, but code will still compile.
So having code like `Thing("A", "B", …, "E")` is now wrong and hard to track.

zio-http-gen/src/test/resources/ComponentOrderWithAliases.scala
987Nabil: True. But I think this is unsolvable then. There is no way to translate it to zio-schema correctly right now. We probably need to add new validations to zio schema

hochgi: @987Nabil in second thought, perhaps you need a change for greaterOrEqualTo, but we can still use `java.lang.Math.nextDown` (or `nextAfter`) to get the closest representable floating point number that is smaller (or larger) than the input.

It might be ugly, e.g:
```sacla
@ java.lang.Math.nextDown(0f)
res0: Float = -1.4E-45F
```
but at least it'll be (effectively) correct.

WDYT?

987Nabil: I think that works work. But I would not write out the number, but generate the expression

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: issue: https://github.com/zio/zio-schema/issues/736

zio-http-gen/src/test/resources/ComponentOrderWithAliases.scala
hochgi: Yup...
```scala
@validate(Validation.greaterThan(Math.nextDown(0f)))
```
Would be much better (and understandable) than generating weird epsilons

987Nabil (approved):

PR #3073, opened 2024-08-31
refactor:  `HttpContentCodec` for Improved Caching

### LINKED ISSUE #3029: Improvements for HttpContentCodec
I think we can do the following, to avoid allocating to many codecs that are more or less all the same.

1. The different schema based constructors of `HttpContentCodec` should cache the crated instances. This will reduce the amount of codecs in the running server and also improve the efficiency of the `lookupCache`
2. Because of 1. we might want to use `ConcurrentHashMap`. I am not sure if it is needed, since a concurrent update would only lead to a cache miss. Might be okay to just keep it as is.
3. Make `HttpContentCodec` a sealed trait and have a `Filtered` subtype, that keeps a view on an actual `HttpContentCodec` instance. This way, when we call `only` to limit the codec to a certain media type, we can still use the `lookupCache` of the original codec.

varshith257:
### Changes Done:
1. Made `HttpContentCodec` a sealed trait with `Default` and `Filtered` subtypes.
2. The `lookupCache` has been optimized to avoid unnecessary recomputation
3. Refactored the `only` method to utilize the `Filtered` subtype to ensure that the original codecs cache is utilized effectively to reduce overhead.

fixes #3029 
/claim #3029

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
@@ -15,9 +15,9 @@ import zio.http.codec.internal.TextBinaryCodec
 import zio.http.internal.HeaderOps
 import zio.http.template._
 
-final case class HttpContentCodec[A](
-  choices: ListMap[MediaType, BinaryCodecWithSchema[A]],
-) { self =>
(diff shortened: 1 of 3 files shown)


jdegoes (approved):

987Nabil:
Already solved

PR #3075, opened 2024-09-01
Various optimizations

kyri-petrou:
This PR contains a few optimizations and fixes that I spotted over time while working on other issues.

Server:
1. Disable auto-read and manually read from the channel. This is mostly important for requests where the response is an effect, since Netty will be poll-reading the channel while we're processing the request in ZIO's threadpool
2. Increment/decrement the request count whenever the ServerInboundHandler is added/removed. The addition is done when a connection to a client is established, whereas the removal is done when the connection is closed. This reduces contention on the LongAdder for persisted connections
3. Place the `HttpServerExpectContinueHandler` and `HttpServerKeepAliveHandler` handlers _before_ the `HttpObjectAggregator` (this is documented in the javadoc of these handlers)
4. Use `fiber.tellInterrupt` in `NettyRuntime#closeFuture` to avoid blocking Netty's event-loop awaiting for the interruption of the effect

Client:
1. Use the Promise's unsafe API to avoid spinning up a fiber in the CloseFutureListener
2. Fail the promise with a new `PrematureChannelClosureException` each time so that the stack traces are properly filled

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/NettyRuntime.scala
@@ -75,8 +75,11 @@ private[zio] final class NettyRuntime(zioRuntime: Runtime[Any]) {
   private def closeListener(fiber: Fiber.Runtime[_, _])(implicit
     unsafe: Unsafe,
     trace: Trace,
-  ): GenericFutureListener[Future[_ >: Void]] =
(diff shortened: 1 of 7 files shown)


kyri-petrou:
I'm going to set this to Draft till I figure out why one of the tests is being flaky

kyri-petrou:
Closing this for the time being in favour of #3114, might revisit the auto-read optimization in the future

PR #3078, opened 2024-09-02
Wrap responses of `SimpleEffectBenchmarkServer` in `ZIO.succeed`

kyri-petrou:
One of the benchmarks that runs in CI is meant to record the performance of effectful endpoints, however the responses are not wrapped in an effect


kyri-petrou:
Note that I noticed that the benchmarks were using NIO (instead of epoll) as the channel type so I fixed that as well

jdegoes (approved):

jdegoes:
@kyri-petrou You going to return to this one? Feel free to re-open when ready!

PR #3084, opened 2024-09-03
Use user-defined content-length for HEAD responses

### LINKED ISSUE #3080: Content-Length always reset to 0 for HEAD requests
For a HEAD requests, where the content-length is (dynamically) set to represent whatever would be returned for a GET request, it's always reset to 0. Consider:

```scala
//> using scala 3
//> using lib dev.zio::zio-http:3.0.0-RC10

import zio.ZIOAppDefault
import zio.http.{Body, Header, Headers, Method, Request, Response, Routes, Server, Status, handler}

object ZioDemo extends ZIOAppDefault {
  val routes =
    Routes(
      Method.HEAD / "test" -> handler { (req: Request) =>
        Response(
          status = Status.Ok,
          headers = Headers(Header.ContentLength(123))
        )
      }
    )

  def run = Server.serve(routes).provide(Server.default)
}
```

Test:

```
[~]% curl -X HEAD -v localhost:8080/test
> HEAD /test HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
< date: Mon, 02 Sep 2024 13:57:58 GMT
< content-length: 0
<
```

This used to work with RC8. For more details on content-length and HEAD requests, see [here](https://stackoverflow.com/questions/3854842/content-length-header-with-head-requests).

Blocks https://github.com/softwaremill/tapir/pull/4007.

kyri-petrou:
/fixes #3080

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/NettyResponseEncoder.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/NettyResponseEncoder.scala
@@ -14,51 +14,55 @@
  * limitations under the License.
  */
 
-package zio.http.netty
+package zio.http.netty.server
 
 import zio._
 import zio.stacktracer.TracingImplicits.disableAutoTrace
 
 import zio.http._
+import zio.http.netty.CachedDateHeader
 import zio.http.netty.model.Conversions
 
 import io.netty.buffer.Unpooled
-import io.netty.channel.ChannelHandlerContext
 import io.netty.handler.codec.http._
 
-private[zio] object NettyResponseEncoder {
+private object NettyResponseEncoder {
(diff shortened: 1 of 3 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/server/NettyResponseEncoder.scala
```diff
@@ -14,51 +14,55 @@
  * limitations under the License.
  */
 
-package zio.http.netty
+package zio.http.netty.server
```
kyri-petrou, line 17: Moved this under the `server` package since these methods are _very_ server specific

987Nabil (approved):

PR #3097, opened 2024-09-06
gen: failing test: code is only generated for last response

nafg:
If an OpenAPI spec operation defines multiple responses (e.g. one success and one or more errors, or just two or more error responses), the generated code only acknowledges the last response in the list. It won't contain multiple `.outError` or a `.out` and `.outError`.


987Nabil:
#3103

PR #3098, opened 2024-09-06
gen: failing test: zio.schema import missing

nafg:
For some reason even if the request body is just referencing a component schema, the operation file has its own `case class RequestBody` which defines a `Schema`. However it doesn't import `zio.schema.*`.

1. I would expect it to just use a type alias. The code generated for the component schema ends up being useless and redundant.
2. In any case since it's using ZIO Schema, it needs to import it.


987Nabil:
#3101

PR #3099, opened 2024-09-06
gen: failing test: header name isn't quoted

nafg:
It generates this:

```scala
  .header(HeaderCodec.name[String](token))
```

And of course `token` is undefined. It should be quoted as a string.

Also, I suppose it doesn't matter but I don't know why it lower-cased it from `"Token"`.


987Nabil:
#3102

PR #3106, opened 2024-09-06
Remove http app wording

987Nabil:


zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -347,14 +347,14 @@ private[zio] final case class ServerInboundHandler(
 object ServerInboundHandler {
 
   val live: ZLayer[
-    AppRef & Server.Config,
+    RoutesRef & Server.Config,
```
kyri-petrou, line 350: Out of the scope of this PR, but should this object (`ServerInboundHandler`) be package private instead?

kyri-petrou (approved):
LGTM

PR #3114, opened 2024-09-09
Optimize netty integration and default config

kyri-petrou:
## Change to default Netty configuration

The main improvement in this PR is a change in the sizing of Netty's EvenLoopGroup. By default, Netty will size the EventLoopGroup at `nCPU * 2`, and will use this pool for 2 purposes:
1. Accepting new connections (boss)
2. Running the pipeline and IO (worker)

The sizing of this pool is extremely conservative and assumes that _some_ blocking IO might occur in Netty's worker threads. It also guards against cases where the underlying transport is blocking. However, since we're delegating all user-defined logic to ZIO's executor, and since the transports we're using are all non-blocking (Epoll, KQueue, NIO), a better configuration is to:
1. Use a separate EventLoopGroup with `nThreads=1` for the boss group. This way, we avoid delays in accepting new connections when the worker threads are busy serving existing connections
2. Size the worker EventLoopGroup with `nThreads=nCPU` since there's no blocking code running in the worker threads

With this change, we see a ~10% increase in throughput vs the default configuration

## Use `ctx.channel.write` instead of `ctx.write` in ServerInboundHandler

This part is not really well documented in Netty, but their main differences are
- `ctx.write` / `ctx.writeAndFlush` will walk through the pipeline from the current handler until the head of the pipeline when writing a response
- `ctx.channel.write` / `ctx.channel.writeAndFlush` will start at the end of the pipeline and walk through all the handlers in the pipeline when writing a response

In our case, short-cutting writes to the pipeline by starting at the current handler doesn't bring any benefit because the ServerInboundHandler is the last one in the pipeline. In addition, as [this video on Netty's best practises](https://youtu.be/WsMOJqAYW5M?si=piFsYwMvpKrfNY3v&t=1725) seems to suggest, when we're writing to the channel from a different thread it's better to use `ctx.channel.write`.

## Other changes

- removed the `runtime` scope from epoll / kqueue native transports. These are generally tiny (~20kb each) and I think it's better for users to have them available by default rather than having to install them manually and having to align netty versions
- updated the `./.devcontainer` files to install `wrk` and updated the Java / SBT versions in order to make it quicker to spin up a devcontainer for benchmarking purposes
- updated `build.sbt` to fork processes when using `sbt zioHttpExample/runMain example.xx` to avoid having issues with restarting servers
- added sensible java opts when running examples

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/NettyDriver.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/NettyDriver.scala
@@ -18,16 +18,18 @@ package zio.http.netty.server
 
 import java.lang.{Boolean => JBoolean}
 import java.net.InetSocketAddress
-
 import zio._
-
 import zio.http.Driver.StartResult
 import zio.http.netty._
 import zio.http.netty.client.NettyClientDriver
 import zio.http.{ClientDriver, Driver, Response, Routes, Server}
-
 import io.netty.bootstrap.ServerBootstrap
 import io.netty.channel._
+import io.netty.channel.epoll.{EpollChannelOption, EpollEventLoopGroup, EpollMode}
+import io.netty.channel.kqueue.KQueueEventLoopGroup
+import io.netty.channel.nio.NioEventLoopGroup
+import io.netty.channel.unix.UnixChannelOption
+import io.netty.incubator.channel.uring.IOUringEventLoopGroup
 import io.netty.util.ResourceLeakDetector
 
 private[zio] final case class NettyDriver(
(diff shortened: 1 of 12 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/server/NettyDriver.scala
```diff
@@ -42,9 +45,21 @@ private[zio] final case class NettyDriver(
...
+        // The boss group is responsible for accepting new connections.
+        // It's recommended to use a separate, single-threaded group for this purpose to avoid contention between
+        // accepting new connections and processing I/O.
+        val cfg = nettyConfig.maxThreads(1).copy(shutdownQuietPeriodDuration = 0.seconds)
+        eventLoopGroup match {
+          case _: EpollEventLoopGroup   => EventLoopGroups.epoll(cfg)
+          case _: KQueueEventLoopGroup  => EventLoopGroups.kqueue(cfg)
+          case _: IOUringEventLoopGroup => EventLoopGroups.uring(cfg)
+          case _                        => EventLoopGroups.nio(cfg)
+        }
```
kyri-petrou: @987Nabil another way here would be to create an `EventLoopGroups` case class that contains the boss and worker group and add it as a requirement to the ZIO environment, that way users could technically configure it. What do you think, do we have time before the 3.0.0 release to do that (since it'll be a breaking change)?

kyri-petrou: After discussing it on Discord, we decided to go ahead and push this into a layer

jdegoes (approved):

PR #3118, opened 2024-09-10
Test latest zio snapshot

kyri-petrou:
Testing whether the ZPool related flakiness is gone with the latest ZIO snapshot

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -167,6 +167,12 @@ lazy val zioHttp = crossProject(JSPlatform, JVMPlatform)
       `zio-test`,
       `zio-test-sbt`,
     ),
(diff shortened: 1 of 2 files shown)


jdegoes:
@kyri-petrou Please re-open if you need to do further testing! I take it from the failure, ZPool has more work to do??

kyri-petrou:
@jdegoes the issue is with a badly written test which provides the `Client` layer _and then_ closes the request `Scope`. Previously `ZPool` would leak the connection pool resources so the test would succeed, but with the new implementation the connections are correctly closed when the `Client.default` finalizers are run so the test is failing.

I'll open a PR to fix the test

PR #3122, opened 2024-09-10
Changes the default Endpoint.outStream[X] encoding to produce a JSON array

### LINKED ISSUE #3113: Default encoding of `Enpoint.outStream[X]` should be JSON array, or the Content Type should be `application/json-seq`
Conceptually when an endpoint returns a list of X, it is the same as when it returns a ZStream[_, _, X]. Currently the default output encoding for `.out[X]` is an JSON array. For `.outStream[X]` it is however a not separated collection of JSONs. The `Content-Type` in the response is `application/json`, correct one would be `application/json-seq`.

Either the data should be returned as a JSON array, so prefixed with `[`, separated by `,` and terminated by `]`, or the `Content-Type` should be `application/json-seq`. I'd prefer the array, because that makes the change change from `.out[List[X]]` to `.outStream[X]` an internal change of the server, without the need to change the clients.

A viable solution currently (RC10) is to provide custom codes for endpoints with `.outStream[X]` stream outputs.

gregor-rayman:
Changes the default response encoding of `Endpoint.outStream[X]` for the MIME type `application/json` so that is produces a valid JSON array instead of just concatenated JSON entities.

Fixes #3113, /claim #3113

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -12,6 +12,8 @@ val _ = sys.props += ("ZIOHttpLogLevel" -> Debug.ZIOHttpLogLevel)
 ThisBuild / githubWorkflowEnv += ("JDK_JAVA_OPTIONS" -> "-Xms4G -Xmx8G -XX:+UseG1GC -Xss10M -XX:ReservedCodeCacheSize=1G -XX:NonProfiledCodeHeapSize=512m -Dfile.encoding=UTF-8")
 ThisBuild / githubWorkflowEnv += ("SBT_OPTS" -> "-Xms4G -Xmx8G -XX:+UseG1GC -Xss10M -XX:ReservedCodeCacheSize=1G -XX:NonProfiledCodeHeapSize=512m -Dfile.encoding=UTF-8")
 
+ThisBuild / resolvers ++= Resolver.sonatypeOssRepos("snapshots")
+
 ThisBuild / githubWorkflowJavaVersions := Seq(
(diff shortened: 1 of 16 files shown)


gregor-rayman:
Now the OpenAPI generation is done.

I hope I have implemented all the needed changes and tests. 

There are two small things that could improve this in my opinion:

* In the encoder I use `ZPipeline.intersperse` with `ZPipeline.prepend` and `ZPipeline.append`. With https://github.com/zio/zio/pull/9182 only the `intersperse` would be needed.

* In `HttpContentCodec` I have added the array producing and parsing `BinaryCodec` that is very similar to `zio.schema.codec.JsonCodec.schemaBasedBinaryCodec` from https://github.com/zio/zio-schema. https://github.com/zio/zio-schema/pull/739 already provides the needed functionality, so when the new version of zio-schema is available, this can be simplified here.

It will also mitigate the codeconv-commenter complaint, as in `zio-schema` the corresponding parts are covered.

987Nabil:
@gregor-rayman please use the snapshot build of zip schema and make the simplifications you mentioned. If it works, I'll release zio-schema and we can use the released version before merging your pr.

gregor-rayman:
@987Nabil , done.

gregor-rayman:
@987Nabil , I have upgraded the `zio-schema` dependency to 1.5.0

987Nabil:
1.5.0 has a bin incompatibility I think I'll fix that first

987Nabil (approved):

PR #3126, opened 2024-09-11
Add Support for Intersection Types in Scala 2 

### LINKED ISSUE #3112: Attaching HandlerAspect to a Routes object with mulitple dependencies does not work (using Scala2)
The operator '@@' is used to attach a HandlerAspect to a Routes object.

It is working fine if applying results in a Routes object with no dependencies.

If the result is a Routes object with one dependency, it still works, but that dependency needs to be declared explicitly, like:
`.@@[Int]`

However if the result is a Routes object with multiple dependencies, it won't work even if we declare those dependencies explicitly.

Using scala 3 everything works as expected, and the type does not need to be declared explicitly either.

Example code to reproduce the issue:

```scala

//> using dep dev.zio::zio-http::3.0.0-RC10
//> using scala 2.13.14

import zio.http.codec.{HttpCodec, StatusCodec}
import zio.http.endpoint.{AuthType, Endpoint}
import zio.http.{HandlerAspect, Header, Route, RoutePattern, Routes, withContext}
import zio._

object Issue extends App {
  val effectWithNoDependency: IO[Nothing, String] = for {
    _ <- ZIO.log("executing effect with no dependency")
  } yield "hardcodedResult - effectWithNoDependency"

  val effectWithOneDependency: ZIO[Int, Nothing, String] = for {
    magicNumber <- ZIO.service[Int]
    _ <- ZIO.log(s"executing effect with one dependency; magic number is $magicNumber")
  } yield "hardcodedResult - effectWithOneDependency"

  val effectWithTwoDependency: ZIO[Int & Long, Nothing, String] = for {
    magicNumber <- ZIO.service[Int]
    magicNumber2 <- ZIO.service[Long]
    _ <- ZIO.log(s"executing effect with two dependency; magic numbers are: $magicNumber and $magicNumber2")
  } yield "hardcodedResult - effectWithTwoDependency"

  val authContext: HandlerAspect[Any, String] = HandlerAspect.customAuthProviding[String] { request =>
    {
      request.headers.get(Header.Authorization).flatMap {
        case Header.Authorization.Basic(uname, secret) if uname.reverse == secret.value.mkString =>
          Some(uname)
        case _ =>
          None
      }
    }
  }

  val endpoint = Endpoint(RoutePattern.GET / "route1")
    .outCodec[String](StatusCodec.Ok ++ HttpCodec.content[String])
    .auth(AuthType.Basic)

  val route1: Route[String, Nothing] = endpoint.implement((_: Unit) =>
    withContext((username: String) => ZIO.log(s"username is $username")) *>
      effectWithNoDependency
  )

  // The authContext eliminates the String dependency
  val routes1: Routes[Any, Nothing] = Routes(route1) @@ authContext

  val route2: Route[Int & String, Nothing] = endpoint.implement((_: Unit) =>
    withContext((username: String) => ZIO.log(s"username is $username")) *>
      effectWithOneDependency
  )

  // The authContext eliminates the String dependency, so only the Int dependency remains
  // But what remains - it has to be declared explicitly
  val routes2: Routes[Int, Nothing] = Routes(route2).@@[Int](authContext)

  val route3: Route[Int & Long & String, Nothing] = endpoint.implement((_: Unit) =>
    withContext((username: String) => ZIO.log(s"username is $username")) *>
      effectWithTwoDependency
  )

  // Like in routes2 we would like to eliminate the String dependency so only Int and Long would remain.
  // However, the below gives this compilation error:
  // "A Tag may not contain an intersection type, yet have provided: Int with Long"
  val routes3: Routes[Int & Long, Nothing] = Routes(route3).@@[Int & Long](authContext)
}
```

asr2003:
**Changes Made:**
- Introduced special handling for Scala 2, where intersection types (with) require casting via asInstanceOf to ensure proper type alignment when applying handler aspects.
- In Scala 3, intersection types (&) are natively supported and I think do not require casting, so the handler aspect is applied directly.

/claim #3112
fixes #3112

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Routes.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Routes.scala
@@ -39,10 +39,13 @@ final case class Routes[-Env, +Err](routes: Chunk[zio.http.Route[Env, Err]]) { s
   def @@[Env0](aspect: HandlerAspect[Env0, Unit]): Routes[Env with Env0, Err] =
     aspect(self)
 
-  def @@[Env0, Ctx <: Env](
-    aspect: HandlerAspect[Env0, Ctx],
-  )(implicit tag: Tag[Ctx]): Routes[Env0, Err] =
-    self.transform(_ @@ aspect)
+  def @@[Env0, Ctx <: Env](aspect: HandlerAspect[Env0, Ctx])(implicit tag: Tag[Ctx]): Routes[Env0, Err] = {
+    if (isScala2 && isIntersectionType[Ctx]) {
+      self.transform(_ @@ aspect).asInstanceOf[Routes[Env0, Err]]
+    } else {
+      self.transform(_ @@ aspect)
+    }
+  }
 
   def @@[Env0]: ApplyContextAspect[Env, Err, Env0] =
     new ApplyContextAspect[Env, Err, Env0](self)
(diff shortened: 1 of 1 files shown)


asr2003:
@987Nabil How can we make a test to run in Scala2 Only to test the patch. I have test ready but stuck in how to make it run for Scala2 only.

I didn't see any @@Scala2Only something similiar like in ZIO

987Nabil (changes requested):
There is no test

asr2003:
Just to confirm to test patch for scala2, I have waited. I will add it now

987Nabil:
Closing in favor of #3128

PR #3143, opened 2024-09-17
Optimizations for request execution happy path

kyri-petrou:
Main changes:
1. Convert `Routes` to `Handler` in `ServerInboundHandler` to avoid allocations & method invocations each time we call `Routes#apply`.
2. Store the wildcards tree in `RoutePattern` in a val to avoid doing that on each invocation to `RoutePattern#get`
3. Use `Map#getOrElse(key, null)` in a couple of places
4. Extract `tree` outside of the `Function1` in `toHandler` to avoid checking the class `var` on each request
5. Use `null` as the default for `Set[Int]` in `PathCodec#get`. This avoids the boxing of the Int on each iteration.

Note on (5): We could also use `BitSet` here, but since almost always `skipLiteralsFor` will be empty / null, it's better to avoid the method invocation and just do a null check

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -48,8 +48,8 @@ private[zio] final case class ServerInboundHandler(
 
   implicit private val unsafe: Unsafe = Unsafe.unsafe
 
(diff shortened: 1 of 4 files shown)


jdegoes (approved):

jdegoes:
Nice work! Can't believe (1) wasn't already done. 🤦‍♂️

PR #3149, opened 2024-09-18
Only call handle error lambda on fatal cause for handled routes (#3134)

### LINKED ISSUE #3134: Using a handled route with error handling functions causes a ClassCastException
**Describe the bug**

Affected versions: 3.0.0-RC5+ **including 3.0.0**

This [pr](https://github.com/zio/zio-http/commit/7458b6f0ee29a2e6f9c03e61789ca8c4241c51d1#diff-9647adc8988f3499f424096635a41c9c07904f9cb3e7ed3ead01fb0c227307a1R66) introduced changes to the error handling behaviour, allowing error handling functions to be used together with routes that don't fail / use the failure channel for responses.

Problem is that the type signature lies for handled handlers. A Handled route will not fail with a `Cause[Nothing]`, but rather with a `Cause[Response]`. This can cause ClassCastExceptions at runtime.

For example trying to use `.handleErrorCauseZIO(c => ZIO.logInfo(Try(c.squash).toString()).as(Response.ok))` with a handled route will log a `class zio.http.Response cannot be cast to class scala.runtime.Nothing$`.

Root cause are this [line](https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/Route.scala#L76) which casts the cause to a `Cause[Nothing]`.

And [this](https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/RoutePattern.scala#L73) constructor of a Route using a RoutePattern and a handled handler. This function has an unused type parameter [Err] with ends up being `Nothing` unless explicitly specified.

**To Reproduce**

1. Create a simple zio-http application:
```scala
package example

import zio._
import zio.http._
import scala.util.Try

  val routes = Routes(
    Method.GET / "foo" -> handler { _: Request =>
      ZIO.fail(Response.ok)
    }
  ).handleErrorCauseZIO(c => ZIO.logInfo(Try(c.squash).toString()).as(Response.ok))

  def run: ZIO[Environment with ZIOAppArgs with Scope,Any,Any] =
    Server.serve(routes).provideLayer(Server.default)

}
```
2. Run it
3. Send a http get request to http://localhost:8080/foo
4. See error in the logs

**Expected behaviour**
No Classcast exceptions should happen at runtime. I suspect one of two solutions would fit here:
1. Error type of handled routes should be Cause[Response] not Cause[Nothing].
2. When using error handling operators on Handled routes we should strip all failures and only leave interrupts and defects to be logged. IMO this has better ergonomics than options (1) and is what I would prefer here.

**Additional Notes**

Looks like this issue was already reported here https://github.com/zio/zio-http/issues/2580#issuecomment-2139543267

987Nabil:
fixes #3134
/claim #3134

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/RouteSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/RouteSpec.scala
@@ -23,7 +23,6 @@ import zio.http.codec.{HttpCodec, StatusCodec}
 import zio.http.endpoint.{AuthType, Endpoint}
 
 object RouteSpec extends ZIOHttpSpec {
-  def extractStatus(response: Response): Status = response.status
 
   def spec = suite("RouteSpec")(
     suite("Route#prefix")(
@@ -45,7 +44,7 @@ object RouteSpec extends ZIOHttpSpec {
 
         for {
           result <- ignored.toHandler.run().merge
-        } yield assertTrue(extractStatus(result) == Status.Ok)
+        } yield assertTrue(result.status == Status.Ok)
       },
       test("route dying with throwable ends in internal server error") {
         val route =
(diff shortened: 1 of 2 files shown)


mschuwalow:
seems to be duplicate of https://github.com/zio/zio-http/pull/3146

PR #3150, opened 2024-09-18
Allow applying aspects only to handlers with `Request` input (#3141)

### LINKED ISSUE #3141: ClassCassException when using HandlerAspect and Path parameters.
**Describe the bug**

**To Reproduce**
Run this snipped

```scala
import zio.http._
import zio._

        ZIO.succeed((req, None))
      }
    )

      })
    }) @@ maybeWebSession



    (for {
      _ <- ZIO.never

  }
}
```

**Expected behaviour**
Log Hello and return 200 Ok

**Additional context**
Same error in scala 2 and 3.

987Nabil:
The actual bug was, that the handler aspect could be applied to a handler that has a input different then request. There was a construct that should have prevented it, but it did not work.

To make it easier to add aspects to a single route, I added `@@` methods to `Route` too. And a test that shows, that it works with path parameters

fixes #3141
/claim #3141

### DIFF
--- a/docs/reference/aop/handler_aspect.md
+++ b/docs/reference/aop/handler_aspect.md
@@ -426,7 +426,7 @@ We notice in the response that first `basicAuth` handler aspect responded `HTTP/
 We can abort the requests by specific response using `HandlerAspect.fail` and `HandlerAspect.failWith` aspects, so the downstream handlers will not be executed:
 
 ```scala mdoc:invisible
-val myHandler = Handler.identity
+val myHandler = Handler.ok
 ```
 
 ```scala mdoc:compile-only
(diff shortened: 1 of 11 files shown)


987Nabil:
@jdegoes This is not binary compatible. In fact it would make code not compile that compiled before. But that code would be impossible to run anyway.

I tried to make it possible to apply context aspects to arbitrary handlers, but from what I see, that is not possible.
If you have an idea how it could work let me know. But I fear this breaking change is our best option 😞 

The only other good way I see is to deprecate the method and find a new operator.

Petter-K:
Any update on this?

PR #3154, opened 2024-09-19
Deprecatre ZClient.request  method

### LINKED ISSUE #3133: Flag ZClient.request as deprecated
I've seen that with 3.0.0-RC10 (and 3.0.0), the `ZClient` was refactored regarding the need for `Scope` which is great.

However I noticed that the `ZClient.request` method is deprecated in favour of `batched` and `streaming` only in the companion object: https://github.com/zio/zio-http/blob/8c70090aa01442c3e09193e468917adfa0aa2286/zio-http/shared/src/main/scala/zio/http/ZClient.scala#L296-L297

If you already have an instance of `ZClient`, the method `request` is still accessible and it's not obvious that it's oriented for streaming usage: https://github.com/zio/zio-http/blob/8c70090aa01442c3e09193e468917adfa0aa2286/zio-http/shared/src/main/scala/zio/http/ZClient.scala#L178

As a user, I would've expected to see that the `request` method (of the class) should not be used directly.

I guess there's not that much we can do now without breaking binary compatibility though. :shrug:

varshith257:
/claim #3133
fixes #3133

### DIFF
--- a/zio-http-cli/src/main/scala/zio/http/endpoint/cli/Retriever.scala
+++ b/zio-http-cli/src/main/scala/zio/http/endpoint/cli/Retriever.scala
@@ -2,6 +2,8 @@ package zio.http.endpoint.cli
 
 import java.nio.file.Path
 
+import scala.annotation.nowarn
+
 import zio._
 
 import zio.http._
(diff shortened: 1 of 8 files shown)


zio-http-cli/src/main/scala/zio/http/endpoint/cli/Retriever.scala
```diff
@@ -27,10 +29,13 @@ private[cli] object Retriever {
   final case class URL(name: String, url: String, mediaType: MediaType) extends Retriever {
 
     val request                                                = Request.get(http.URL(http.Path.decode(url)))
-    override def retrieve(): ZIO[Client, Throwable, FormField] = for {
-      client <- ZIO.serviceWith[Client](_.batched)
-      chunk  <- client.request(request).flatMap(_.body.asChunk)
-    } yield FormField.binaryField(name, chunk, mediaType)
+    @nowarn("cat=deprecation")
```
varshith257, line 32: @987Nabil Any suggestions here?
I have supplied Scope explicitly with `Client with Scope' and also tried wrapping up req in `ZIO.scoped` by using streaming/batched appropriately wherever needed but it throwing compile error of not providing Scope.

zio-http/jvm/src/test/scala/zio/http/ZClientAspectSpec.scala
```diff
@@ -86,20 +86,24 @@ object ZClientAspectSpec extends ZIOHttpSpec {
...
+        def followRedirectsTest = {
+          for {
+            port       <- Server.install(redir ++ routes)
+            baseClient <- ZIO.service[Client]
+            client = baseClient
+              .url(
+                URL(Path.empty, Location.Absolute(Scheme.HTTP, "localhost", Some(port))),
+              )
+              .batched @@ ZClientAspect.followRedirects(2)((resp, message) => ZIO.logInfo(message).as(resp))
+            response <- client.request(Request.get(URL.empty / "redirect"))
```
varshith257, line 100: same here too

varshith257 (commented):
.

varshith257:
@987Nabil Any suggestions?

varshith257:
@987Nabil Any review on this?

PR #3156, opened 2024-09-19
Endpoint Scala 3 API for union types output (#3125)

### LINKED ISSUE #3125: Update `Alternator` and perhaps others to include implicits for scala 3 type updates
I am trying to create an `Endpoint` with multiple output error possibilities, and was hoping the combined error type would end up being a union type, but it looks like the implicit `Alternator` being used results in `Either` being used instead. `Alternator` is sealed, so adding my own, while a simple implementation, is not possible.

I would like an `Alternator` instance that combines types using union types instead of nesting `Either`s.

None, just living with nested `Either`s

**Additional context**
should be as simple as:
```scala
    given unionAlternator[A, B]: Alternator.WithOut[A, B, A | B] =
        new Alternator[A, B]:
            type Out = A | B

            def left(l: A): Out = l

            def right(r: B): Out = r

            def unleft(out: Out): Option[A] = out match
                case a: A => Some(a)
                case _ => scala.None

            def unright(out: Out): Option[B] = out match
                case b: B => Some(b)
                case _ => scala.None
```
I'm wondering if there are other places where converting to union types over either's, or other scala 3 type updates in general, could be useful

987Nabil:
@jdegoes we can't get rid of alternator or combiner under the hood, since it is part of the codec. This PR is just a transformation on top of `HttpCodec.Fallback`. This also means, there is no need for the to implement the Endpoint methods that use `Combiner`, since the type is always an n-tuple. Scala 3 or 2.

Combiner could have only one generic tuple as hlist impl. in Scala 3 I guess, but there is not need to split 2 and 3 in my opinion. I don't think there is a significant performance benefit. And the Scala 2 version needs to be there anyway.

fixes #3125 
/claim #3125

### DIFF
--- a/docs/examples/endpoint-scala3.md
+++ b/docs/examples/endpoint-scala3.md
@@ -1,0 +1,36 @@
+---
+id: endpoint-scala3
+title: "Endpoint Scala 3 Syntax"
+sidebar_label: "Endpoint Scala 3 Syntax"
+---
+
+```scala
+import zio.http.*
+import zio.http.codec.*
+import zio.http.endpoint.*
+
+import java.util.UUID
+
+type NotFound[EntityId] = EntityId
+type EntityId           = UUID
+
+val union: ContentCodec[String | UUID | Boolean] =
+  HttpCodec.content[String] || HttpCodec.content[UUID] || HttpCodec.content[Boolean]
+
+val unionEndpoint =
+  Endpoint(Method.GET / "api" / "complex-union")
+    .outCodec(union)
+
+val unionWithErrorEndpoint
+    : Endpoint[Unit, Unit, NotFound[EntityId] | String, UUID | Unit, AuthType.None] =
+  Endpoint(Method.GET / "api" / "union-with-error")
+    .out[UUID]
+    .orOut[Unit](Status.NoContent)
+    .outError[NotFound[EntityId]](Status.NotFound)
+    .orOutError[String](Status.BadRequest)
+
+val impl = unionWithErrorEndpoint.implementEither { _ =>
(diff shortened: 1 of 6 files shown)


jdegoes:
@987Nabil It's not clear to me how this looks in user-land code, can you provide a more complicated example?

987Nabil:
Does this help?
```scala
import zio.http.*
import zio.http.codec.*
import zio.http.endpoint.*

import java.util.UUID

type NotFound[EntityId] = EntityId
type EntityId           = UUID

val union: ContentCodec[String | UUID | Boolean] =
  HttpCodec.content[String] || HttpCodec.content[UUID] || HttpCodec.content[Boolean]

val unionEndpoint =
  Endpoint(Method.GET / "api" / "complex-union")
    .outCodec(union)

val unionWithErrorEndpoint
    : Endpoint[Unit, Unit, NotFound[EntityId] | String, UUID | Unit, AuthType.None] =
  Endpoint(Method.GET / "api" / "union-with-error")
    .out[UUID]
    .orOut[Unit](Status.NoContent)
    .outError[NotFound[EntityId]](Status.NotFound)
    .orOutError[String](Status.BadRequest)

val impl = unionWithErrorEndpoint.implementEither { _ =>
  val result: Either[NotFound[EntityId] | String, UUID | Unit] = Left("error")
  result
}
```

jdegoes:
@987Nabil I think this looks good but should be accompanied by documentation and at least one example. 🙏

anqit:
curious when the `orOut...` API will be released?

987Nabil:
I am in preparation for a release

Issue #3066, opened 2024-08-30
[Bug] The passed `Env` is not the good one


guizmaii:
Reproducer:

```scala
import zio.http.*
import zio.http.endpoint.{AuthType, Endpoint}
import zio.http.netty.NettyConfig
import zio.http.netty.server.NettyDriver
import zio.test.TestAspect.shrinks
import zio.test.{Gen, Spec, TestEnvironment, ZIOSpecDefault, assertTrue, check}
import zio.{Scope, UIO, ULayer, ZIO, ZLayer, ZNothing}

object MyApiSpec extends ZIOSpecDefault {

  trait MyService  {
    def code: UIO[Int]
  }
  object MyService {
    def live(code: Int): ULayer[MyService] = ZLayer.succeed(new MyServiceLive(code))
  }
  final class MyServiceLive(_code: Int) extends MyService {
    def code: UIO[Int] = ZIO.succeed(_code)
  }

  val endpoint: Endpoint[Unit, String, ZNothing, Int, AuthType.None] =
    Endpoint(RoutePattern.POST / "api").in[String].out[Int]

  val api = endpoint.implement(_ => ZIO.serviceWithZIO[MyService](_.code))

  override def spec: Spec[TestEnvironment & Scope, Any] =
    test("test") {
      check(Gen.fromIterable(List(1, 2, 3, 4, 5))) { code =>
        (
          for {
            client <- ZIO.service[Client]
            port   <- ZIO.serviceWithZIO[Server](_.port)
            url     = URL.root.port(port) / "api"
            request = Request
                        .post(url = url, body = Body.json("this is some input"))
                        .addHeader(Header.Accept(MediaType.application.json))
            _      <- TestServer.addRoutes(api.toRoutes)
            result <- client.batched(request)
            output <- result.body.asString
          } yield assertTrue(output == s"$code")
        ).provideSome[TestServer & Client](
          ZLayer.succeed(new MyServiceLive(code))
        )
      }.provide(
        ZLayer.succeed(Server.Config.default.onAnyOpenPort),
        TestServer.layer,
        Client.default,
        NettyDriver.customized,
        ZLayer.succeed(NettyConfig.defaultWithFastShutdown),
      )
    } @@ shrinks(0)
}

```

The test will fail on the second iteration of the `check` with the following assertion error:
```scala
Assertion failed:
Expected :"2"
Actual   :"1"
```

The `Env` of the `api` handler is never updated. It always contains the instance of `MyService` returning `1`

[...]

PR #3160, opened 2024-09-23
Fix Route merge order to fix Issue #3066

exi:
In case of duplicate routes, the old logic gives precedence to older routes, which breaks cases where the routes are updated with new environments and re-added.

This gives precedence to the newer ones.

/claim #3066

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Routes.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Routes.scala
@@ -251,19 +251,13 @@ final case class Routes[-Env, +Err](routes: Chunk[zio.http.Route[Env, Err]]) { s
         val chunk = tree.get(req.method, req.path)
         chunk.length match {
           case 0 => Handler.notFound
-          case 1 => chunk(0)
-          case n => // TODO: Support precomputed fallback among all chunk elements
-            var acc = chunk(0)
-            var i   = 1
-            while (i < n) {
-              val h = chunk(i)
-              acc = acc.catchAll { response =>
+          case _ => // TODO: Support precomputed fallback among all chunk elements
+            chunk.reverseIterator.reduceLeft { (acc, h) =>
+              acc.catchAll { response =>
                 if (response.status == Status.NotFound) h
                 else Handler.fail(response)
               }
-              i += 1
             }
-            acc
         }
       }
       .merge


kyri-petrou (commented):
I'm a little bit confused. `++` should be left-associative, so why / how does this work exactly? AFAICT this should be breaking things 😕

Also, can you please add the reproducer as a test that was added in the original issue?

EDIT / PS: If this is indeed the solution to the issue, I think we should be fixing `++` instead and adding tests to ensure that it's left-associative

exi:
@kyri-petrou 

> EDIT / PS: If this is indeed the solution to the issue, I think we should be fixing `++` instead and adding tests to ensure that it's left-associative

As far as understand, doing `++` on routes just concatenates them in a list. That list is then checked left-to-right during calls to search for appropriate request handlers.
Since in the linked issue, we add the same route twice, with different environments, before this PR the old route takes precedence, after this PR, the new one does.

I agree that I would have expected routes to be merged route-by-route, but figured that there probably was a good reason to just put then in a list.

exi:
Adding the test case is complicated because it depends on TestServer, which is afaik not usable in zio-http because it would lead to circular dependencies.
For my local testing I added a new test top level sbt module named `zio-http-regressiontests`. I'm happy to include it but it seems a bit misplaced since it only contains one test.

987Nabil:
Even if this works, this is not how `++` should work. Like on `Map` the right should win. If there is a bug, it is way deeper in the code

exi:
@987Nabil Do you mean that the request handlers should be constructed from the routes iterating back to front over them or do you mean that the request handler tree building code should read front to back but override older values or at least prefer newer ones?
I could implement that.

exi:
@987Nabil @kyri-petrou 

I updated the PR to change the handler merging logic in cases of multiple handlers for the same path.
The old code iterated over the handler from left to right, the new one flips that and also gets rid of explicit loops in favor of a reduction.

exi:
Since the new code is more general, i also got rid of the special casing between one handler and more than one handler.

exi:
Lot's of tests failing, I'll check what's up with that.

987Nabil:
This change does not look right either. Neither the place nor the fact that we replace a while loop with allocation of an interator

exi:
@987Nabil Then i'm not sure what the intended behaviour is.

Routes uses the normal `++` behaviour, which when doing `oldApp ++ newApp`, will create Chunks of routes and put the new one at the end/right.
This later gets dumped into then handler tree building where routes are aggregated by path, while still keeping the chunk order of the handlers.
The handler tree building at the lowest level creates the real handler where in the case of the code I changed here, the old code only ever looks at the second handler in the list when the first handler returned `Status.NotFound`.

And since the chunks and therefore the handlers are put into "oldest left" order and the handler building iterates left-to-right, the oldest one gets preference every time.

# Alternative
I could change the `++` implementation for Routes and do an actual merge instead of a concatenation, and throw away the old handler in case of duplicate paths. Some other `++` implementations seem to be doing that.

exi:
@guizmaii @jdegoes I went through a couple of iterations in this PR already but I'm now a bit lost on what the desired solution is.
I think the key issue is that `++` for `Routes` just appends the new routes to the list of old routes and the handler building walks this list front-to-back and uses whatever handler works first.
I could either change the handler building logic to work backwards, or I could implement a different `++` for `Routes` where that discards old handlers during the merge. Which one is the preferred solution here?

exi:
@kyri-petrou @987Nabil Turns out, our notion of how `++` is supposed to work is wrong for `Routes`. I just found this comment: 

https://github.com/zio/zio-http/blob/502331c7c8ec548b326120a9596864d30fabf4a7/zio-http/shared/src/main/scala/zio/http/Routes.scala#L51-L57

It seems like "old (left) has precedence over new (right)" is the intended behaviour.
Which makes `addRoutes` for the same route unusable during tests.
I also think that many of the tests zio-http has that use `addRoutes` do not expect this behaviour.

987Nabil:
@jdegoes I find this behaviour confusing. Everywhere else we give "the new thing" precedence. Should we change it?

zio-http/shared/src/main/scala/zio/http/Routes.scala
```diff
@@ -251,19 +251,13 @@ final case class Routes[-Env, +Err](routes: Chunk[zio.http.Route[Env, Err]]) { s
...
         chunk.length match {
           case 0 => Handler.notFound
-          case 1 => chunk(0)
-          case n => // TODO: Support precomputed fallback among all chunk elements
-            var acc = chunk(0)
-            var i   = 1
-            while (i < n) {
-              val h = chunk(i)
-              acc = acc.catchAll { response =>
+          case _ => // TODO: Support precomputed fallback among all chunk elements
```
jdegoes, line 254: I like the idea here but this is too slow. It's in a hot path and needs to be optimized. You should definitely not delete the special cases (`1`, `n`), and if you need to do reverse iterator and reduction, it needs to be done somewhere else (not here, in the hot path), as a pre-computation.

987Nabil:
closing in favor of #3337

PR #3166, opened 2024-09-24
Issue 3165: Fix Forwarded Header rendering

### LINKED ISSUE #3165: "Frowarded" Header renders incorrectly which causes it to loose information
**Describe the bug**
The `Forwarded` header incorrectly renders optional values when those values are `None` 

For example:
```
Header.Forwarded(forValues = List("""127.0.0.1"""))
```

Renders as:
```
None; for=127.0.0.1; None; None
```

This causes the header to loose data when pulled out of a `Request`.

**To Reproduce**
```
 val header      = Header.Forwarded(forValues = List("""127.0.0.1"""))
// This renders as `None; for=127.0.0.1; None; None` which is not valid

val r = Request(headers = Headers(header))
println(r.header(Header.Forwarded)) // Output: Some(Forwarded(None,List(),None,None))
```

As can be seen above, the `Forwarded` is loosing information when pulled back out of the request because it fails to correctly parse the incorrectly rendered header string.

**Expected behaviour**
- `Forwarded` should correctly render unset values as an empty string, "" and not as `None; `.
- A round trip of rendering and parsing the `Forwarded` should NOT loose data.

**Additional context**
I should have a fix for this forth coming, it will look something like this:

```
   def render(forwarded: Forwarded): String = {
      def formatDirective(directive: Option[ String ]) = directive.map(_+"; ").getOrElse("")

      s"${formatDirective(forwarded.by)}${forwarded.forValues.map(v => s"for=$v").mkString(",")};${formatDirective(forwarded.host)}${formatDirective( forwarded.proto )}"
    }
```

scottweaver:
This set of change address the issue described in https://github.com/zio/zio-http/issues/3165.

For instance the header, 'Header.Forwarded(forValues = List("127.0.0.1"))', previously rendered incorrectly as 'None; for=127.0.0.1; None; None'.

It now properly renders as 'for=127.0.0.1;'.

It includes an additional test spec that uses a property check to verify all permutations optional directives.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Header.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Header.scala
@@ -2856,8 +2856,13 @@ object Header {
       Right(Forwarded(by, forValue, host, proto))
     }
 
-    def render(forwarded: Forwarded): String =
-      s"${forwarded.by}; ${forwarded.forValues.map(v => s"for=$v").mkString(",")}; ${forwarded.host}; ${forwarded.proto}"
+    def render(forwarded: Forwarded): String = {
+      def formatDirective(directive: Option[String]) = directive.map(_ + ";").getOrElse("")
+
+      val forValues = if (forwarded.forValues.nonEmpty) forwarded.forValues.map(v => s"for=$v").mkString(",") + ";" else ""
+
+      s"${formatDirective(forwarded.by.map("by=" + _))}${forValues}${formatDirective(forwarded.host.map("host=" + _))}${formatDirective(forwarded.proto.map("proto=" + _))}"
+    }
   }
 
   /** From header value. */
--- a/zio-http/jvm/src/test/scala/zio/http/headers/ForwardedSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/headers/ForwardedSpec.scala
@@ -19,7 +19,7 @@ package zio.http.headers
 import zio.Scope
 import zio.test._
 
-import zio.http.{Header, ZIOHttpSpec}
+import zio.http.{Header, Headers, Request, ZIOHttpSpec}
 
 object ForwardedSpec extends ZIOHttpSpec {
   override def spec: Spec[TestEnvironment with Scope, Any] = suite("Forwarded suite")(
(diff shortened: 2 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -2856,8 +2856,13 @@ object Header {
       Right(Forwarded(by, forValue, host, proto))
     }
 
-    def render(forwarded: Forwarded): String =
-      s"${forwarded.by}; ${forwarded.forValues.map(v => s"for=$v").mkString(",")}; ${forwarded.host}; ${forwarded.proto}"
+    def render(forwarded: Forwarded): String = {
+      def formatDirective(directive: Option[String]) = directive.map(_ + ";").getOrElse("")
```
987Nabil, line 2860: is this not just `mkString(";")`?

zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -2856,8 +2856,13 @@ object Header {
...
+
+      val forValues = if (forwarded.forValues.nonEmpty) forwarded.forValues.map(v => s"for=$v").mkString(",") + ";" else ""
```
987Nabil, line 2862:
```suggestion
      val forValues = if (forwarded.forValues.nonEmpty) forwarded.forValues.mkString("for=",",for=",";") else ""
```
I think this is the same, but might be more efficient

zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -2856,8 +2856,13 @@ object Header {
...
+
+      s"${formatDirective(forwarded.by.map("by=" + _))}${forValues}${formatDirective(forwarded.host.map("host=" + _))}${formatDirective(forwarded.proto.map("proto=" + _))}"
```
987Nabil, line 2864: if you already have a method, maybe handover a prefix as well?

987Nabil (commented):
some small stuff

jdegoes (approved):

PR #3167, opened 2024-09-24
Special case encode optional for HttpCodec.Content (#3144)

### LINKED ISSUE #3144: Optional content codec failing on missing request body
**Describe the bug**
I'd like to define an endpoint where the request body is optional.  I've passed an optional content codec to the endpoint, but the implemented route is still returning an error indicating not being able to read the body there is no body in the request:
```html
<html><body><h1>Codec Error</h1><p>There was an error en-/decoding the request/response</p><p id="name">MalformedBody</p><p id="message">Malformed request body failed to decode: Unexpected end of input</p></body></html>%
```

**To Reproduce**

I've defined an extension method as 

```scala
    extension [P, I, E, O, M <: EndpointMiddleware](e: Endpoint[P, I, E, O, M])
        def optionalIn[I1: HttpContentCodec](using combiner: Combiner[I, Option[I1]]): Endpoint[P, combiner.Out, E, O, M] =
            e.copy(input = e.input ++ HttpCodec.content[I1].optional)
```
created and endpoint using that:
```scala
val myEndpoint = Endpoint(Method.POST / path / "some-path")
        .optionalIn[OptionalBody]
        // ... other endpoint definition calls
```
and implemented the route:
```scala
        Handler.fromFunctionZIO: (maybeBody: Option[OptionalBody]) =>
                // ... implement business logic
```
Make a request with no body to that endpoint results in the error. Passing a correctly formed body does not result in the error.

**Expected behaviour**
Specifying an optional `in` declaration should allow requests with no body and result in an `Option` being added to the expected `In` type and empty option being passed to the handler function if the request contains no body.

**Additional context**

987Nabil:
While this is a real bug fix, the preferred way should be `HttpCodec.content[Option[MyType]]`

fixes #3144 
/claim #3144

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
@@ -148,6 +148,36 @@ sealed trait HttpContentCodec[A] { self =>
     choices.headOption.map(_._2).getOrElse {
       throw new IllegalArgumentException(s"No codec defined")
     }
+
+  def optional: HttpContentCodec[Option[A]] =
+    self match {
+      case HttpContentCodec.Choices(choices)           =>
+        HttpContentCodec.Choices(
+          choices.map { case (mediaType, BinaryCodecWithSchema(fromConfig, schema)) =>
+            mediaType -> BinaryCodecWithSchema(fromConfig.andThen(optBinaryCodec), schema.optional)
+          },
+        )
+      case HttpContentCodec.Filtered(codec, mediaType) =>
+        HttpContentCodec.Filtered(codec.optional, mediaType)
+    }
+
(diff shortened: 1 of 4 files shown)


987Nabil:
Is this MIMA issue a real issue? 🤔

zio-http/shared/src/main/scala/zio/http/codec/HttpContentCodec.scala
```diff
@@ -144,6 +144,8 @@ sealed trait HttpContentCodec[A] { self =>
     choices.headOption.map(_._2).getOrElse {
       throw new IllegalArgumentException(s"No codec defined")
     }
+
+  def optional: HttpContentCodec[Option[A]]
```
jdegoes: This is not backward compatible because it has no default implementation.

jdegoes (approved):

jdegoes (approved):

PR #3169, opened 2024-09-26
HTTP spec conformance during CI

### LINKED ISSUE #3083: Test for HTTP spec conformance during CI
I think this is a bit of an ambitious request, but hopefully it's something that we can achieve. 

The HTTP spec contains a lot of "gotchas" and unless a contributor knows the HTTP spec inside-out (which is almost impossible), we're bound to make a change at some point that violates the spec. A good example of this is #3080, where a previous fix to an issue that allowed users to provide an invalid content-length (which is against the spec) seems to have broken HEAD requests, where the content-length should match the length of the content as if the request is a GET request.

In order to guard against this, we need an HTTP Spec conformance test suite that runs during CI. A potentially good starting point is [this repo](https://github.com/cispa/http-conformance), but there are probably other / better tools for it (haven't looked too extensively). I'm sure that adding such a suite will very likely unearth violations, so we should probably do it in 2-3 steps:

1. Add a CI workflow that tests a server for HTTP spec conformance.
2. Fix any failing checks that are required by the spec.
3. Fix as many recommendations as possible.

@jdegoes @987Nabil what are your thoughts on having this? Is this something that's too ambitious or do you think it might be doable?

Saturn225:
/claim #3083
fixes #3083

This PR integrates new HTTP conformance tests derived from the research paper _"Who's Breaking the Rules? Studying Conformance to the HTTP Specifications and its Security Impact"_ by Jannis Rautenstrauch and Ben Stock. These tests now acts as a guardrails  to ZIO -HTTP implementations adhere to the specifications and help identify potential security issues.

**Conclusions**

1. The tests taken reference from [http-conformance](https://github.com/cispa/http-conformance) are categorised into 3 levels, **Requirement, Recommendations and ABNF**. The initial process is to add the conformance suite and I have added the Requirement and Recommendation level conformance tests which are critical to be tested to safeguard.

2.  I have ran http-conformance tool with simple zio-http server setup and observed analysis of tool with different categories Dangerous broken, Dangerous not broken, Not dangerous broken and Not Dangerous not broken. I have shifted towards first add tests for Dangerous ones and added them broken/not-broken then added not-dangerous ones.

**Changes done:**

**Status Codes:**

This  specs verifies behaviour of the different Status Codes in Violations

- **`204 No Content`**  which verifies no body is sent.
- **`205 Reset Content`** checks no body is sent.
- **`206 Partial Content`** checks the presence of Content-Range.
- **`206 Multipart Content`** checks Content-Range is excluded in multipart responses.
- **`206 Headers`** checks headers like ETag and Cache-Control are present.
- **`401 Unauthorized`** 
     - checks the presence of WWW-Authenticate header.
     - checks  401 status when Authorization header is missing.
- **`405 Method Not Allowed`** checks the Allow header is present.
- **`407 Proxy Authentication Required`** verifies the Proxy-Authenticate header is present.
- **`304 Not Modified`** checks no body is returned for 304 Not Modified and verifies consistency with 200 OK and more....

**Redirection (Location Header):**

This tests added validates the presence of Location header in 300 Multiple Choices, 301 Moved Permanently, 302 Found, 303 See Other, 307 Temporary Redirect and 308 Permanent Redirect responses.

**Headers and Metadata:**

- **`Range Header (206)`** checks Content-Range is present in 206 responses.
- **`Content-Range (416)`** validates Content-Range in 416 Range Not Satisfiable.
- **`Content-Length in CONNECT`** checks no Content-Length for 2XX CONNECT.
- **`Transfer-Encoding in CONNECT`** checks no Transfer-Encoding for 2XX CONNECT.
- **`CSP Header`** validates that only one Content-Security-Policy header is sent.

**HTTP Methods:**

- **`HEAD Request (no body)`** verifies no body is sent in HEAD requests.
- **`GET and HEAD Headers Consistency`** checks for matching headers in GET and HEAD.
- **`POST Invalid Response Codes`** validates POST does not return 206, 304, or 416.
- **`501 Unknown HTTP Methods`** validates unknown methods return 501 Not Implemented.
- **`405 Blocked Methods`** validates non-allowed methods return 405 Method Not Allowed.

**HTTP Versions:**

- **`Transfer-Encoding with HTTP/1.0`** validates no Transfer-Encoding for HTTP/1.0.
- **`Transfer-Encoding with HTTP/1.1`** validates Transfer-Encoding for HTTP/1.1 or higher.
- **`Whitespace in Start-Line`** validates rejection of requests with improper start-line formatting.
- **`Bare CR in Headers`** checks compliance with CRLF formatting in headers.

**Connection Management and Headers:**

- **`Close in Final Response`** validates Connection close is included when requested.
- **`Bad Headers with CR, LF, or NUL`** validates headers with illegal characters are rejected.
- **`Upgrade and Expect Headers`** validates 100 Continue before 101 Switching Protocols when both Upgrade and Expect headers are present.

Some of other added are **`426 Upgrade Required`** and **`101 Switching Protocols`**

**Cache-Control Directives:**

This spec added checks for unquoted values for max-age and s-maxage and quoted-string form for no-cache and private.

**Content-Length and Transfer-Encoding:**

This spec validates no Content-Length or Transfer-Encoding in 1xx and 204 responses and also to validate consistency of Content-Length between HEAD and GET as well as between 304 Not Modified and 200 OK.

**Cookie Handling:**

This spec guardrails no duplicate cookie attributes in Set-Cookie headers and no duplicate cookies with the same name and also the use of IMF-fixdate format for cookie expiration dates.

**Miscellaneous:**

Some of miscellaneous compliance tests added checks with headers such as Server, Content-Type, Accept-Patch and Date.

**Possible Changes to Discuss:**

Some tests related to Content-Security-Policy-Report-Only and Strict-Transport-Security (STS) are not yet supported in ZIO-HTTP, but I have raised tickets for their future inclusion #3171 and #3172 . Once these headers are supported, we can add more tests to validate their behavior. HTTP/2 is not yet supported in ZIO-HTTP. Once it is implemented, additional tests for HTTP/2 conformance will be necessary. I have mainly focused on to add the Requirement and Recommended tests which sets initial process to add conformance guardrail for zio-http. The ABNF checks could be added gradually into the suite

**Other**

I have fixed the all violating checks added as per the conformance suite added.
Thanks @JannisBush for the detailed paper and the excellent tool!


987Nabil:
@Saturn225 HTTP/2 is out of scope. The failing test should be fixed first. I set this PR to draft for now. Once you fixed the issues open again. If you are stuck, you can contact me on discord

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -108,6 +115,32 @@ private[zio] final case class ServerInboundHandler(
 
   }
 
+  private def validateHostHeader(req: Request): Boolean = {
+    req.headers.get("Host") match {
```
987Nabil: Use gerOrElse(null) to avoid allocation

zio-http/shared/src/main/scala/zio/http/Routes.scala
```diff
@@ -248,22 +248,34 @@ final case class Routes[-Env, +Err](routes: Chunk[zio.http.Route[Env, Err]]) { s
...
-            var i   = 1
-            while (i < n) {
-              val h = chunk(i)
-              acc = acc.catchAll { response =>
-                if (response.status == Status.NotFound) h
-                else Handler.fail(response)
-              }
-              i += 1
+        val chunk          = tree.get(req.method, req.path)
+        val allowedMethods = tree.getAllMethods(req.path)
```
987Nabil: Should be a def, since it is used only once and only if chunk is empty

Saturn225:
The failing test is that cors header is adding OPTIONS method automatically in that spec. I will need to take care of this case in toHandler. I'll fix it

Saturn225:
The validation added in ServerInboundHandler.scala gone too aggressive for the zioHttpTestKit. Tackling it

Saturn225:
@987Nabil Any suggestions on failing tests on ServerSpec?

Saturn225:
@987Nabil Any Any guidance here?

987Nabil:
Sorry, I am busy these days. I come back to you when I have time.

Saturn225:
Thanks @987Nabil

Saturn225:
@987Nabil @jdegoes Seems tricky why the entire workflows failing across the repo with following:

```
sbt: command not found
```

Any recent updates to the workflows?

Saturn225:
@jdegoes @987Nabil Can this get reviewed? It's been open from a long time

Saturn225:
@jdegoes @987Nabil Please take a look into this PR at your free time?

Saturn225:
@jdegoes I wanted to follow up on this PR which has been open for 4 months. Could you please take a moment to review it? Your approval would help it move forward 🙏

jdegoes:
@987Nabil Can you review this, please? 🙏

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -108,6 +114,27 @@ private[zio] final case class ServerInboundHandler(
 
   }
 
+  private def validateHostHeader(req: Request): Boolean = {
+    val host = req.headers.get("Host").getOrElse(null)
```
987Nabil:
```suggestion
    val host = req.headers.getUnsafe("Host")
```

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -108,6 +114,27 @@ private[zio] final case class ServerInboundHandler(
...
+      if (!isValid) {
+        ZIO
+          .logWarning(
+            s"Invalid Host header for request ${req.method} ${req.url}. " +
+              s"Host: $host, isValidHost: $isValidHost, isValidPort: $isValidPort",
+          )
+      }
+      isValid
+    } else {
+      ZIO.logWarning(s"Missing Host header for request ${req.method} ${req.url}")
```
987Nabil: This is dead code

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -108,6 +114,27 @@ private[zio] final case class ServerInboundHandler(
...
 
+  private def validateHostHeader(req: Request): Boolean = {
+    val host = req.headers.get("Host").getOrElse(null)
+    if (host != null) {
+      val parts       = host.split(":")
+      val isValidHost = parts(0).forall(c => c.isLetterOrDigit || c == '.' || c == '-')
+      val isValidPort = parts.length == 1 || (parts.length == 2 && parts(1).forall(_.isDigit))
+      val isValid     = isValidHost && isValidPort
+      if (!isValid) {
+        ZIO
```
987Nabil: Dead code. Please remove

zio-http/shared/src/main/scala/zio/http/RoutePattern.scala
```diff
@@ -185,6 +185,12 @@ object RoutePattern                                                       {
       else forMethod ++ wildcardsTree.get(path)
     }
 
+    def getAllMethods(path: Path): Set[Method] = {
```
987Nabil:
```suggestion
    private[http] def getAllMethods(path: Path): Set[Method] = {
```

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -384,7 +384,11 @@ private[codec] object EncoderDecoder {
                 .getOrElse(throw HttpCodecError.MalformedHeader(codec.name, codec.textCodec))
 
             case None =>
-              throw HttpCodecError.MissingHeader(codec.name)
+              if (codec.name.equalsIgnoreCase("Authorization")) {
```
987Nabil: why is this needed? That does not seem to change a thing

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -340,7 +340,11 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
               case Some(HttpCodecError.CustomError("SchemaTransformationFailure", message))
                   if maybeUnauthedResponse.isDefined && message.endsWith(" auth required") =>
                 maybeUnauthedResponse.get
-              case Some(_) =>
+              case Some(HttpCodecError.MissingHeader(header)) if header.equalsIgnoreCase("Authorization") =>
+                Handler.succeed(
+                  Response.unauthorized.addHeaders(Headers(Header.WWWAuthenticate.Bearer(realm = "Restricted Area"))),
```
987Nabil: While this needs spezial handling, it is not always the right header to add

zio-http/shared/src/main/scala/zio/http/internal/HeaderModifier.scala
```diff
@@ -32,7 +32,11 @@ import zio.http._
  */
 trait HeaderModifier[+A] { self =>
   final def addHeader(header: Header): A =
-    addHeaders(Headers(header))
+    if (header.headerName == Header.XFrameOptions.name) {
```
987Nabil: special case handling for a deprecated header. Please revert and remove related tests if there are any

zio-http/shared/src/main/scala/zio/http/Routes.scala
```diff
@@ -248,23 +248,40 @@ final case class Routes[-Env, +Err](routes: Chunk[zio.http.Route[Env, Err]]) { s
...
-          case n => // TODO: Support precomputed fallback among all chunk elements
-            var acc = chunk(0)
-            var i   = 1
-            while (i < n) {
-              val h = chunk(i)
-              acc = acc.catchAll { response =>
-                if (response.status == Status.NotFound) h
-                else Handler.fail(response)
+        val chunk          = tree.get(req.method, req.path)
+        def allowedMethods = tree.getAllMethods(req.path)
```
987Nabil: This can be done more efficient. I think I might open a separate PR for that.

987Nabil: @jdegoes @kyri-petrou after giving that a thought, I think we should keep 404 as a default.
It makes it more performant and a default 405 (MethodNotAllowed) could make it easier to find out what routes exist on a server.
If someone needs 405, it can be implemented in user land

987Nabil: @Saturn225 please revert this. We decided to keep 404

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
Saturn225: @987Nabil It's returning a default 400 Bad Request instead 401 Unauthorized error when header is missing. So we need this to conform RFC to return Unauthorized error with header

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
Saturn225: But to conform this https://datatracker.ietf.org/doc/html/rfc7235#section-3.1 we need to add header along 401. If we need flexibility i remove adding header and instead return only 401

zio-http/shared/src/main/scala/zio/http/internal/HeaderModifier.scala
Saturn225: Okay

zio-http/shared/src/main/scala/zio/http/Routes.scala
Saturn225: Done!

Saturn225:
Done @987Nabil

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
987Nabil: But `codec.name` is `authorization` you call the same missing header constructor with the same arguments

zio-http/shared/src/main/scala/zio/http/codec/HttpCodecError.scala
```diff
@@ -31,7 +31,8 @@ sealed trait HttpCodecError extends Exception with NoStackTrace with Product wit
 }
 object HttpCodecError {
   final case class MissingHeader(headerName: String)                                           extends HttpCodecError {
-    def message = s"Missing header $headerName"
+    def message = if (headerName.equalsIgnoreCase("Authorization")) "Missing Authorization header"
```
987Nabil: Please revert. It only changes the order of `header` and `Authorization` for no reason

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
987Nabil: Just keep it unchanged for now and set the test to ignore. This needs a smarter solution

Saturn225: Okay! Done

zio-http/shared/src/main/scala/zio/http/codec/HttpCodecError.scala
Saturn225: Done!

Saturn225:
@987Nabil I have cleaned up the PR and marked the test as ignore for now

987Nabil:
@kyri-petrou wants to take a look too

Saturn225:
@jdegoes @987Nabil @kyri-petrou Can you review this, please? 🙏

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
987Nabil: This is still dead code

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -87,12 +87,17 @@ private[zio] final case class ServerInboundHandler(
             )
             releaseRequest()
           } else {
-            val req  = makeZioRequest(ctx, jReq)
-            val exit = handler(req)
-            if (attemptImmediateWrite(ctx, req.method, exit)) {
+            val req = makeZioRequest(ctx, jReq)
+            if (!validateHostHeader(req)) {
```
987Nabil: @kyri-petrou I am not sure we should check this here. I think it is overhead and only if we parse it to the typed version we should check it, I think.

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
Saturn225: We are using it at validation

Saturn225:
@kyri-petrou Can you review this, please? 🙏

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
kyri-petrou: How are you using it in validation? This makes no sense to me

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -108,6 +113,24 @@ private[zio] final case class ServerInboundHandler(
 
   }
 
+  private def validateHostHeader(req: Request): Boolean = {
+    val host = req.headers.getUnsafe("Host")
+    if (host != null) {
+      val parts       = host.split(":")
+      val isValidHost = parts(0).forall(c => c.isLetterOrDigit || c == '.' || c == '-')
```
kyri-petrou: This can be rewritten so that you iterate the string until the first `:` instead of splitting it and then calling `forall`. Also, _do not_ call `forall` on Strings as that will cause each individual char to be boxed

zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -1034,6 +1034,18 @@ object Handler extends HandlerPlatformSpecific with HandlerVersionSpecific {
   def notFound(message: => String): Handler[Any, Nothing, Any, Response] =
     error(Status.NotFound, message)
 
+  /**
+   * Creates a handler which always responds with a 501 status code.
+   */
+  def notImplemented: Handler[Any, Nothing, Any, Response] =
```
kyri-petrou:
```suggestion
  val notImplemented: Handler[Any, Nothing, Any, Response] =
```

zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
kyri-petrou: I think the overhead should be relatively very small, although according to [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Host#:~:text=A%20400%20Bad%20Request%20status%20code%20may%20be%20sent%20to%20any%20HTTP/1.1%20request%20message%20that%20lacks%20or%20contains%20more%20than%20one%20Host%20header%20field.) the server _MAY_ respond with 400 if the host header is missing or is invalid.

To be honest I think a better approach would be to add a flag to the config `validateHeaders` (with a default to `false`) which will:
1. set [this](https://github.com/zio/zio-http/blob/22caa5a7920603955a500848edb4fc0ce892c575/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerChannelInitializer.scala#L68) to true
2. enable any custom header validations

kyri-petrou: Also, ideally we want to keep all header validation happening in one place. I think the only way we can do this is by extending the default encoder and adding any additional validation within it.

https://github.com/zio/zio-http/blob/22caa5a7920603955a500848edb4fc0ce892c575/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerChannelInitializer.scala#L63

kyri-petrou: Also, ideally we want to keep all header validation happening in one place. I think the only way we can do this is by extending the default encoder and adding any additional validation within it.

https://github.com/zio/zio-http/blob/22caa5a7920603955a500848edb4fc0ce892c575/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerChannelInitializer.scala#L63

Saturn225:
@kyri-petrou @987Nabil Can you review this? Made mima safe approach to introduce validateHeaders flag and all checks are passed now except netifly deployments checks

Saturn225:
@kyri-petrou  @987Nabil could you please review/merge PR?

zio-http/shared/src/main/scala/zio/http/Server.scala
```diff
@@ -578,3 +585,18 @@ object Server extends ServerPlatformSpecific {
 
   }
 }
+
+final case class ServerRuntimeConfig(
```
987Nabil, line 589: We can't have a new class ever time we add a config. And while this is bin compatible, it is not source compatible.
It might be more work, but please make the other config a class instead of a case class and write all necessary methods by hand

987Nabil (changes requested):

Excellencedev:
This is clearly AI slop. TRASH!!!!!!!. Please don't waste your time and close this rubbish

PR #3174, opened 2024-09-26
Prevent OOM when receiving large request streams

### LINKED ISSUE #3173: OOM when receiving large streaming requests
**Describe the bug**
We have a service that proxies large streams to different backends. It's compiled with GraalVM native-image and is running in docker with a quite small heap ~80Mb. It has worked fine for a long time but when we updated to zio-http 3.0.0 from 3.0.0-RC9 it started to get OOM instantly. 
I did some investigation and the problem is the unbounded queue in AsyncBody.asStream that was introduced. It completely disables back pressure for incoming streams. So if the producer is faster than the consumer, the data will be buffered in the unbounded queue. This could lead to OOM. 

**To Reproduce**
1. Create a service with request streaming enabled and limit the heap. Make processing of incoming data a bit slow
2. Create a test case that sends a request larger than the heap of the service. 
3. Watch OOM in the service

**Expected behaviour**
It should work like it did before (but without blocking netty, of course)

davidlar:
/claim #3173

This PR aims to fix [3173](https://github.com/zio/zio-http/issues/3173) by making sure we don't read more data than we can handle.
Now ctx.read() is called after the queue is drained by the consuming end of the stream. The call to ctx.read() will fill the queue with more data asynchronously. 
Nettys AUTO_READ flag is already turned off in zio-http and ctx.read should be called when we are ready for more data, not before.

I'm not sure if we should keep the unbounded queue or not. I guess it will only contain one chunk at most.


zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
```diff
@@ -166,7 +174,7 @@ object NettyBody extends BodyEncoding {
               maybeError =>
                 ZChannel.fromZIO(queue.shutdown) *>
                   maybeError.fold[ZChannel[Any, Any, Any, Any, E, Chunk[A], Unit]](ZChannel.unit)(ZChannel.fail(_)),
-              a => ZChannel.write(a) *> loop,
+              a => ZChannel.write(a) *> ZChannel.fromZIO(nettyRead.whenZIO(queue.isEmpty)) *> loop,
```
kyri-petrou: I'm wondering whether we should buffer reads here, e.g., `nettyRead.whenZIO(queue.size.map(_ < XXX))`

Previously this was set to 4096 (using a bounded queue which we can't use anymore since it was blocking Netty's threads) and there weren't any issues right?

kyri-petrou: Also, 2 notes on performance. 
1. Use `whenZIODiscard` instead of `whenZIO`
2. I think it might be better to store this ZChannel in a val to avoid creating it on each loop:

```scala
val maybeRead = ZChannel.fromZIO(nettyRead.whenZIODiscard(???))

// And then
a => ZChannel.write(a) *> maybeRead *> loop,
```

kyri-petrou (commented):
Thanks for the fix! AFAICT the fix works, although the part that worries me a bit is that with these changes we're delegating `ctx.read()` to a different threadpool which is will add a fair bit of overhead due to context switching.

I'm wondering whether we should add a `continueReading: () => Boolean` argument to `UnsafeAsync.Streaming`. This way we can decide in `AsyncBodyReader` whether we'll invoke a `ctx.read()` or defer it. The part that worries me with this is that we might be adding too much complexity. What do you think?

zio-http/jvm/src/main/scala/zio/http/netty/NettyBody.scala
davidlar: I'll fix the suggested changes in NettyBody!

davidlar:
Not sure I understand how ctx.read() can be called from AsyncBodyReader as it has be called from the consuming side of the queue. So it's probably to complex for me at least :-)

davidlar:
I tried locally with queue.size.map(_ < 4096) and it can be a lot of data since the chunks in the queue can be big in some cases. 
Maybe we should count the queued bytes instead in a ref or something. Or lower that limit to 128 or so. 
What do you think @kyri-petrou ?

davidlar:
It feels like the buffering before #3060 was meant to be maximum 4096 bytes, but was really max 4096 chunks which is not what we want, I guess...
On the other hand I haven't seen smaller chunks than 8kb when testing this, so maybe queue.isEmpty is a good choice after all.
I leave it like that if you want to approve! I will go away for the weekend...

davidlar:
Is there anything else I need to do to get this reviewed? 

Regards David

kyri-petrou (approved):

davidlar:
Great, thanks! Could you please merge this before the next release?

SamuelBoerlin:
Hi, thanks for the fix.
I think we're currently running into this issue. Large requests lead to OOM despite directly streaming the data to a file, but only when the network connection is fast enough.
When is the next release with this fix planned?

guizmaii:
@SamuelBoerlin You can already use snapshot releases

See https://github.com/zio/zio-http/actions/runs/16792940892/job/47558545318#step:16:1422

PR #3178, opened 2024-09-27
[gen] verify issue #3077 is fixed

### LINKED ISSUE #3077: [bug][gen] aliased type only in key schema - Newtype file is not generated
**Describe the bug**
When the OpenAPI schema contains a referenced primitive, when the `generateSafeTypeAliases` flag is enabled,
but is used only as key schema with `x-string-key-schema` extension, the generated code is OK, but the Newtype file is not created.

**To Reproduce**
```yaml
components:
  schemas:
    Key:
      type: string
      format: uuid
    ObjectWithDictionary:
      type: object
      required:
        - dict
      properties:
        dict:
          type: object
          additionalProperties:
            type: string
            x-string-key-schema:
              $ref: '#/components/schemas/Key'
```

and generate with `generateSafeTypeAliases = true`.
The `Key.scala` file is not generated

**Expected behaviour**
files for Newtypes that are referenced only as key schemas are generated

hochgi:
/fixes #3077

I wanted to fix this issue I had with pre-3.0.0 version.
Added a test that has previously failed.
and it passed…

Issue has already been fixed in another PR.

@jdegoes / @987Nabil I will not claim the bounty, as I did not do any fixing here.
But the issue can be closed now :)

### DIFF
--- a/zio-http-gen/src/test/resources/ComponentAliasKey.scala
+++ b/zio-http-gen/src/test/resources/ComponentAliasKey.scala
@@ -1,0 +1,9 @@
+package test.component
+
+import zio.prelude.Newtype
+import zio.schema.Schema
+import java.util.UUID
+
+object Key extends Newtype[UUID] {
+  implicit val schema: Schema[Key.Type] = Schema.primitive[UUID].transform(wrap, unwrap)
+}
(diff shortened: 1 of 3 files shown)


jdegoes (approved):

987Nabil:
So you actually fixed it before? Or someone else?

jdegoes (approved):

hochgi:
> So you actually fixed it before? Or someone else?

I'm not sure where exactly was the problem. Figured it would be easiest to just write the should-be-failing test, and debug it ¯\\\_(ツ)\_/¯

PR #3190, opened 2024-10-11
Update ZIO to 2.1.11

kyri-petrou:
For some _really_ weird reason, one of the tests added by @pablf fails when updating ZIO to 2.1.11. The weirdest part is that it fails for all `-SNAPSHOT` versions after `2.1.9` (even the first one which was just an NPM dependency update).

Based on the comment, I see that the margin was added to account for differences between Scala 2/3. @pablf do you have any idea why do we need to increase this margin now that we're updating the ZIO version even if it doesn't contain any changes that would explain this?


kyri-petrou:
@987Nabil I think CI is broken. I'm not sure why but I have a feeling that it might be due to [this release](https://github.com/actions/setup-java/releases/tag/v4.4.0)

987Nabil:
@kyri-petrou I think I know why. I read from Eugene the sbt guy, that the new Ubuntu images are missing some sbt script and you need a new action he created. Or I guess the cousier setup action can do it as well

pablf:
@kyri-petrou I don't see why something like this would happen. It means that something has changed in the length of the actual request that is being sent by zio-http, which is not per se a problem (I'm assuming that  changing the `ERROR_MARGIN` was enough to solve this). The only thing that comes to mind at the moment is something like a transitive dependency impacting encoding of requests.

kyri-petrou:
@987Nabil could you take a look at the PR when you have 2 mins 🙏

987Nabil (approved):

PR #3203, opened 2024-11-09
Fix: Mimetype application/x-zip-compressed interpreted as a text field 

### LINKED ISSUE #3200: Mimetype application/x-zip-compressed interpreted as a text field
I am running into an issue where parsed form fields (using `.body.asMultipartFormStream`) are recognized/decoded as `FormField.Text`, when Windows browsers (seen with Chrome and Edge) attach files with mimetype `application/x-zip-compressed` instead of `application/zip`. This means the binary file data is coerced into a string and mangled and unusable.

The Mediatypes are generated from `jshttp` which does not include this mimetype, which would be one thing that could address this issue, but there seems to be the intention to fall back to `application/octet-stream` which however is not happening, instead falling back to a non-binary Text type thereby mangling the data irreversibly.

Specifically in `FormField.scala`:

```
      contentType      = extract._2
        .flatMap(x => MediaType.forContentType(x.value))
        .getOrElse(MediaType.application.`octet-stream`)
```

however in `MediaType.forContentType(...)`:

```
contentTypeMap.get(contentTypeLC).orElse(parseCustomMediaType(contentTypeLC))
```

which does:

```
def parseCustomMediaType(customMediaType: String): Option[MediaType] = {
    val contentTypeParts = customMediaType.split('/')
    if (contentTypeParts.length == 2) {
      val subtypeParts = contentTypeParts(1).split(';')
      if (subtypeParts.length >= 1) {
        Some(
          MediaType(
            mainType = contentTypeParts.head,
            subType = subtypeParts.head,
            parameters = if (subtypeParts.length >= 2) parseOptionalParameters(subtypeParts.tail) else Map.empty,
          ),
        )
      } else None
    } else None
  }

```

which is going to return a `MediaType` with `binary` set to false and not fall back to `application/octet-stream`.

I do prefer preserving the mimetype the browser set as opposed to setting it to `application/octet-stream`, since I can deal with that in my app, but the data should IMHO be treated as binary unless certain it really is text.

That would mean that `parseCustomMediaType`  would always explicitly set `binary=true` if the mimetype does not start with `text/` (and possibly other mimetype families known to be text)

varshith257:
/claim #3200 
Fixes #3200

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/MediaTypeSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/MediaTypeSpec.scala
@@ -37,7 +37,7 @@ object MediaTypeSpec extends ZIOHttpSpec {
       }
     },
     test("custom mime type parsing") {
-      assertTrue(MediaType.parseCustomMediaType("custom/mime").contains(MediaType("custom", "mime")))
+      assertTrue(MediaType.parseCustomMediaType("custom/mime").contains(MediaType("custom", "mime", binary = true)))
     },
     test("optional parameter parsing") {
       assertTrue(
--- a/zio-http/jvm/src/test/scala/zio/http/MediaTypeSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/MediaTypeSpec.scala
@@ -48,5 +48,13 @@ object MediaTypeSpec extends ZIOHttpSpec {
           ),
       )
     },
+    test("application/x-zip-compressed should be binary") {
+      val mediaType = MediaType.forContentType("application/x-zip-compressed")
+      assertTrue(mediaType.exists(_.binary))
+    },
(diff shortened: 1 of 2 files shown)


varshith257:
It seems CI is flaky
https://github.com/zio/zio-http/actions/runs/11757111013/job/32753986518?pr=3203#step:10:8446

Dennis4b:
Not sure if it is my place to comment, but `application/json` and `application/xml` should never make it to `parseCustomMediaType` so that check would be superfluous.

Second I am not sure what `unsafeParseCustomMediaType` is intended for, but should it not have the same fix?

varshith257:
Thanks @Dennis4b. Yes, I think for both methods we need the same and updated

987Nabil (approved):

PR #3205, opened 2024-11-12
[gen] adding support for "Any" and "AnyObject" as json in codegen

### LINKED ISSUE #3199: [gen] accept OpenAPI "open" types like "any" & "anyObject"
defining "Any" in OpenAPI according to the docs: https://swagger.io/docs/specification/v3_0/data-models/data-types/
Screenshot 2024-11-06 at 8 50 01

These won't work in zio-http-gen.

Expectation:
```yaml
components:
  schemas:
    Foo:
      type: object
      properties:
        name:
          type: string
        anything: {}
```
should be rendered as:
```scala
case class Foo(name: String, anything: Any)
```

Also, support "free form objects:
Screenshot 2024-11-06 at 17 33 55
e.g:
```yaml
components:
  schemas:
    Foo:
      type: object
      properties:
        name:
          type: string
        anything:
          type: object
          additionalProperties: {} # or true, or nothing at all
```
should be rendered as:
```scala
case class Foo(name: String, anything: Map[String, Any])
```

Also, since `Any` is not really nice to use, and a more common use case would be to generate with Json:
```scala
case class Foo(name: String, anything: Json)
```
or:
```scala
case class Foo(name: String, anything: Json.Obj)
```
instead of `Any`/`Map[String, Any]`…

we should add an ability to configure this.
Obviously we need to provide a way to generate a zio-json compatible code,
but perhaps other Json ASTs could be supported (like circe/playjson/etc'…)?
This controls both the code, and the required imports.

default should be `Any`/`Map[String, Any]`, since it would be surprising to generate code that would fail to compile unless you add a 3rd party dependency.

hochgi:
/fixes #3199 
/claim #3199

### DIFF
--- a/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
+++ b/zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
@@ -673,85 +673,44 @@ final case class EndpointGen(config: Config) {
     } else None
 
   @tailrec
-  private def schemaToPathCodec(schema: JsonSchema, openAPI: OpenAPI, name: String): Code.PathSegmentCode = {
+  private def schemaToPathCodec(schema: JsonSchema, openAPI: OpenAPI, name: String): Code.PathSegmentCode =
     schema match {
       case JsonSchema.AnnotatedSchema(s, _) => schemaToPathCodec(s, openAPI, name)
       case JsonSchema.RefSchema(ref)        => schemaToPathCodec(resolveSchemaRef(openAPI, ref), openAPI, name)
-      case JsonSchema.Integer(JsonSchema.IntegerFormat.Int32, _, _, _, _, _)     =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.Int)
-      case JsonSchema.Integer(JsonSchema.IntegerFormat.Int64, _, _, _, _, _)     =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.Long)
-      case JsonSchema.Integer(JsonSchema.IntegerFormat.Timestamp, _, _, _, _, _) =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.Long)
-      case JsonSchema.String(Some(JsonSchema.StringFormat.UUID), _, _, _)        =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.UUID)
-      case JsonSchema.String(Some(JsonSchema.StringFormat.Date), _, _, _)        =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.LocalDate)
-      case JsonSchema.String(Some(JsonSchema.StringFormat.DateTime), _, _, _)    =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.Instant)
-      case JsonSchema.String(Some(JsonSchema.StringFormat.Time), _, _, _)        =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.LocalTime)
-      case JsonSchema.String(Some(JsonSchema.StringFormat.Duration), _, _, _)    =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.Duration)
-      case JsonSchema.String(_, _, _, _)                                         =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.String)
-      case JsonSchema.Boolean                                                    =>
-        Code.PathSegmentCode(name = name, segmentType = Code.CodecType.Boolean)
-      case JsonSchema.OneOfSchema(_)           => throw new Exception("Alternative path variables are not supported")
-      case JsonSchema.AllOfSchema(_)           => throw new Exception("Path variables must have exactly one schema")
-      case JsonSchema.AnyOfSchema(_)           => throw new Exception("Path variables must have exactly one schema")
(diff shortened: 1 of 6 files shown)


zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
```diff
@@ -1077,7 +1078,10 @@ final case class EndpointGen(config: Config) {
           ),
         )
       case JsonSchema.Null    => throw new Exception("Null query parameters are not supported")
-      case JsonSchema.AnyJson => throw new Exception("AnyJson query parameters are not supported")
+      case JsonSchema.AnyJson => {
+        // throw new Exception("AnyJson query parameters are not supported")
```
hochgi: why did we previously throw exception?
Am I breaking some other validation now?

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -254,111 +254,123 @@ object JsonSchema {
...
-          schema.optionalKeySchema.fold(valuesSchema)(keySchema =>
-            valuesSchema.annotate(
-              MetaData.KeySchema(
-                fromSerializableSchema(keySchema),
+
+    // if type: object with additionalProperties defined,
+    // but nothing else, we should assume a free form object
+    def anyObject: Boolean = schema.additionalProperties.collect { case BoolOrSchema.BooleanWrapper(true) =>
+      schema.schemaType.contains(TypeOrTypes.Type("object")) &&
+      schema.productIterator.count(_.asInstanceOf[Option[_]].isDefined) == 2
```
hochgi: should probably also allow no type declaration at all, and only `additionalProperties`?

zio-http-gen/src/test/resources/inline_schema_any_and_any_object.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - name
+        - eats
+        - extra_attributes
+      properties:
+        name:
+          type: string
+        eats: {}
+        extra_attributes:
+          type: object
+          additionalProperties: true
```
hochgi, line 49: should also test for all the 3 out of 4 possible cases we handle?
```yaml
  extra_attributes:
    type: object
    additionalProperties: true

  extra_attributes:
    type: object
    additionalProperties: {}

  extra_attributes:
    additionalProperties: true
```

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
hochgi: this was handled in this commit:
https://github.com/zio/zio-http/pull/3205/commits/d4ef6d4498729ab9e2f209fb014f257a825f1ba0

out of 4 possible cases, 3 work, and one seems to require changes to zio-schema

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: @987Nabil any idea regarding this one?

987Nabil: 1. Please no `{}` block for a case handling :)
2. The exception was that there is no handling implemented. Not that there can't be one
3. `None` seems odd. Can you explain?

zio-http-gen/src/test/resources/inline_schema_any_and_any_object.yaml
987Nabil: Sorry, help me out. This dynamic fields open api always confuses me

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -254,111 +254,138 @@ object JsonSchema {
...
-      case Some(BoolOrSchema.BooleanWrapper(bool))  => Left(bool)
-      case Some(BoolOrSchema.SchemaWrapper(schema)) =>
-        val valuesSchema = fromSerializableSchema(schema)
-        Right(
-          schema.optionalKeySchema.fold(valuesSchema)(keySchema =>
-            valuesSchema.annotate(
-              MetaData.KeySchema(
-                fromSerializableSchema(keySchema),
+
+    val definedAttributesCount = schema.productIterator.count(_.asInstanceOf[Option[_]].isDefined)
```
987Nabil, line 258: One thing I learned from the zio http open api integration: I hate it and open API is a crap format haha 😂

zio-http-gen/src/main/scala/zio/http/gen/openapi/EndpointGen.scala
hochgi: 1. sure, will do
2. the exception refers to query parameters, so that was weiird.
3. the method is `def schemaToCode(…): Option[Code.File]`, there is no need to generate a file for `Json`. It's just an import. So `None` was obvious to me.

zio-http-gen/src/test/resources/inline_schema_any_and_any_object.yaml
hochgi: this is straight out of the [docs](https://swagger.io/docs/specification/v3_0/data-models/data-types/):
Screenshot 2024-12-02 at 17 05 06

987Nabil (approved):

PR #3209, opened 2024-11-19
Add missing `final` in `Condition` trait

guizmaii:


987Nabil:
Mima is complaining. What besides aesthetics we try to fix here?

PR #3211, opened 2024-11-19
Optimize `zio.http.codec.internal.EncoderDecoder.Single.decodeBody` code

guizmaii:
Removes an `Option` allocation and some method calls


987Nabil:
fmt

987Nabil (approved):

PR #3214, opened 2024-11-19
Reproducer for #3144

guizmaii:
See #3144 

The issue seems to be that for a `Schema[Option[A]]` we need to create a `HttpContentCondec.fromSchema(Schema[A]).optional`

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/OptionalBodySpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/OptionalBodySpec.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2021 - 2023 Sporta Technologies PVT LTD & the ZIO HTTP contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package zio.http.endpoint
+
+import zio._
+import zio.http._
+import zio.http.codec.Doc
+import zio.http.endpoint.AuthType.None
+import zio.json.ast.Json
+import zio.test._
+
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/test/scala/zio/http/endpoint/OptionalBodySpec.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  override def spec: Spec[TestEnvironment with Scope, Throwable] =
+    suite("OptionalBodySpec")(
+      test("accepts empty body") {
+        val body = Option.empty[Json]
+
+        for {
+          response <- api.runZIO(makeRequest(body))
+          body <- response.body.asString
+        } yield assertTrue(body == """{"no":"body"}""")
```
guizmaii, line 57: Fails with:
```scala
  - OptionalBodySpec / accepts empty body - 78 ms
    ✗ "{\"name\":\"MalformedBody\",\"message\":\"Malformed request body failed to decode: Unexpected end of input\"}" was not equal to "{\"no\":\"body\"}"
    body == """{"no":"body"}"""
    body = "{\"name\":\"MalformedBody\",\"message\":\"Malformed request body failed to decode: Unexpected end of input\"}"
    at /Users/jules/jules_ivanic/workspace/zio-http/zio-http/jvm/src/test/scala/zio/http/endpoint/OptionalBodySpec.scala:57 
```

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -367,66 +363,69 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
...
-  def inCodec[Input2](codec: HttpCodec[HttpCodecType.RequestType, Input2])(implicit
+  def inCodec[Input2: EnvironmentTag](codec: HttpCodec[HttpCodecType.RequestType, Input2])(implicit
     combiner: Combiner[Input, Input2],
-  ): Endpoint[PathInput, combiner.Out, Err, Output, Auth] =
-    copy(input = input ++ codec)
+  ): Endpoint[PathInput, combiner.Out, Err, Output, Auth] = {
+    val isOptional = EnvironmentTag[Input2] <:< EnvironmentTag[Option[?]]
+    println(s"==================== Is optional: $isOptional ==================== ")
+    copy(input = input ++ (if (isOptional) codec.optional.asInstanceOf[zio.http.codec.HttpCodec[zio.http.codec.HttpCodecType.RequestType,Input2]] else codec))
+  }
```
guizmaii, line 428: I attempted a fix here but the result is that it makes fail the only test which was passing and the test which was failing still fails

PR #3225, opened 2024-11-30
[sbt-gen] sbt plugin for openapi to endpoints codegen

### LINKED ISSUE #3180: Create a OpenAPI to Endpoint Code Gen sbt plugin
- Needs to have a config to switch between automatic (on file change) and manual generation
- config for file/folder path(s) of OpenAPI files
- config for package prefix
- setting to handover a global openapi config
- setting for openapi config based on a sbt file filter

hochgi:
/fixes #3180
/claim #3180

* based on a plugin I collaborated with @runtologist on, in another closed repo
* automatic mode integrated in regular `~compile` continuous compilation mode, manual trigger via direct task call
* static openapi files read from `src/main/oapi/**`
* dynamic download/create/… of openapi files via regular use of `sourceGenerators` scoped to the custom `ZIOpenApi` configuration. Files should reside under `ZIOpenApi / sourceManaged` (which is by default evaluated as `target/scala-{cross version}/src_managed/oapi/` in same dir as package convention (see example in tests).
* package prefix determined by relative path under `/src/main/oapi/` => allows for multiple different spec files in different packages in a conventional and expected manner
* sbt config to control codegen
* using sbt caching mechanism to avoid re-generating sources from unchanged spec files on every compile

known gaps:
* for some (currently unknown) reason, using `.dependsOn(…)` does not work, and I was forced to use `zio-http-gen` as an external dependency
* cross compilation fails, currently commented out, so we only support sbt 1.x (sbt 2 is not yet released anyway)
* ~~only "static" openapi spec files are supported as of this first commit. Ideally, we should allow for dynamic generation of spec files like any other generated sources.~~ (done!)
* collisions of generated sources from different spec files is not checked (but users should really use different packages for different services they're generating endpoints for).
* only supporting `Compile` (main) configuration - generating endpoints  for e.g. `Test` (by placing openapi files under `src/test/oapi/`) will not work out of the box (but is it really needed anyway?)
* no README/docs yet - should probably add to the site, and document after we converge on actual functionality (I assume review might change enough to render any docs I'll write now obsolete)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,12 @@ jobs:
         with:
           apps: sbt
 
+      - name: Test sbt plugin
+        if: ${{ github.event_name == 'pull_request' }} && matrix.scala == '2.12.19'
+        run: sbt ++2.12.19 zioHttpGenSbt/scripted
+
+      - uses: coursier/setup-action@v1
+
       - name: Check doc generation
         if: ${{ github.event_name == 'pull_request' }}
         run: sbt ++2.13.14 doc
@@ -106,7 +112,7 @@ jobs:
         run: sbt '++ ${{ matrix.scala }}' zioHttpShadedTests/test
 
       - name: Compress target directories
-        run: tar cf targets.tar sbt-zio-http-grpc/target zio-http-cli/target target zio-http/jvm/target zio-http-docs/target sbt-zio-http-grpc-tests/target zio-http-gen/target zio-http-benchmarks/target zio-http-tools/target zio-http-example/target zio-http-testkit/target zio-http/js/target zio-http-htmx/target project/target
+        run: tar cf targets.tar sbt-zio-http-grpc/target zio-http-gen-sbt-plugin/target zio-http-cli/target target zio-http/jvm/target zio-http-docs/target sbt-zio-http-grpc-tests/target zio-http-gen/target zio-http-benchmarks/target zio-http-tools/target zio-http-example/target zio-http-testkit/target zio-http/js/target zio-http-htmx/target project/target
 
       - name: Upload target directories
         uses: actions/upload-artifact@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,14 +92,14 @@ jobs:
         run: sbt '++ ${{ matrix.scala }}' test
 
       - uses: coursier/setup-action@v1
-        with:
-          apps: sbt
 
       - name: Test sbt plugin
         if: ${{ github.event_name == 'pull_request' }} && matrix.scala == '2.12.19'
         run: sbt ++2.12.19 zioHttpGenSbt/scripted
 
       - uses: coursier/setup-action@v1
+        with:
+          apps: sbt
 
       - name: Check doc generation
         if: ${{ github.event_name == 'pull_request' }}
--- a/build.sbt
+++ b/build.sbt
@@ -93,6 +93,22 @@ ThisBuild / githubWorkflowBuildPreamble := Seq(
   ),
 )
 
+ThisBuild / githubWorkflowBuild := {
+  (ThisBuild / githubWorkflowBuild).value ++ WorkflowJob(
+    "testSbtPlugin",
+    "Test sbt plugin",
+    List(
+      WorkflowStep.Use(UseRef.Public("coursier", "setup-action", "v1")),
+      WorkflowStep.Run(
+        name = Some(s"Test sbt plugin"),
+        commands = List(s"sbt ++${Scala212} zioHttpGenSbt/scripted"),
+        cond = Some(s"$${{ github.event_name == 'pull_request' }} && matrix.scala == '$Scala212'"),
+      ),
+    ),
+    scalas = List(Scala212),
+  ).steps
+}
+
 ThisBuild / githubWorkflowBuildPostamble :=
   WorkflowJob(
     "checkDocGeneration",
@@ -348,6 +364,26 @@ lazy val zioHttpGen = (project in file("zio-http-gen"))
     },
   )
   .dependsOn(zioHttpJVM)
+
+lazy val zioHttpGenSbt = (project in file("zio-http-gen-sbt-plugin"))
+  .enablePlugins(SbtPlugin)
+  .settings(publishSetting(true))
+  .settings(
+    name := "zio-http-sbt-codegen",
+    sbtPlugin    := true,
+    scalaVersion := Scala212,
+    semanticdbEnabled := true,
+    semanticdbVersion := scalafixSemanticdb.revision,
+    scalacOptions ++= stdOptions ++ extraOptions(scalaVersion.value),
+    sbtTestDirectory := sourceDirectory.value / "sbt-test",
+    scriptedLaunchOpts += ("-Dplugin.version=" + version.value),
+    scriptedBufferLog := false,
+    libraryDependencies ++= Seq(
+      `zio-json-yaml`,
+      `zio-test`,
+      `zio-test-sbt`,
+    )
+  ).dependsOn(LocalProject("zioHttpGen"))
 
 lazy val sbtZioHttpGrpc = (project in file("sbt-zio-http-grpc"))
   .settings(stdSettings("sbt-zio-http-grpc"))
--- a/zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/Format.scala
+++ b/zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/Format.scala
@@ -1,0 +1,11 @@
+package zio.http.gen.sbt
+
+sealed trait Format
+object Format {
+  case object YAML extends Format
+  case object JSON extends Format
+  def fromFileName(fileName: String): Format =
+    if (fileName.endsWith("yml") || fileName.endsWith("yaml")) YAML
+    else if (fileName.endsWith("json")) JSON
+    else throw new Exception(s"Unknown format for file $fileName")
+}
--- a/zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/ZioHttpCodegen.scala
+++ b/zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/ZioHttpCodegen.scala
@@ -1,0 +1,146 @@
+package zio.http.gen.sbt
+
+import java.io.File
+import java.nio.charset.StandardCharsets
+import java.nio.file.StandardOpenOption.{CREATE, TRUNCATE_EXISTING}
+import java.nio.file.{Files, Path}
+
+import scala.io.Source
+
+import zio.json.yaml._
+
+import zio.schema.codec.JsonCodec
+
+import zio.http.endpoint.openapi.OpenAPI
+import zio.http.gen.openapi.{Config, EndpointGen}
+import zio.http.gen.scala.CodeGen
+
+import sbt.Defaults.configSrcSub
+import sbt.Keys._
+import sbt._
+import sbt.util.FileInfo
+
+object ZioHttpCodegen extends AutoPlugin {
+
+  object autoImport {
+
+    val ZIOpenApi = config("oapi") extend Compile
+
+    val zioHttpCodegenMake    = taskKey[Seq[File]]("Generate ADTs & endpoints from OpenAPI spec files")
+    val zioHttpCodegenConf    = settingKey[Config]("Configuration for codegen")
+    val zioHttpCodegenSources = settingKey[File]("Source dir. analoguous to e.g: scalaSource or javaSource")
+  }
+
+  import autoImport._
+
+  override lazy val projectSettings: Seq[Setting[_]] = inConfig(ZIOpenApi)(
+    Seq[Setting[_]](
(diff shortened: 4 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -89,6 +89,9 @@ jobs:
       - name: Build project
         run: sbt '++ ${{ matrix.scala }}' test
 
+      - name: Test sbt plugin
+        run: sbt 'zioHttpGenSbt/scripted'
```
hochgi, line 93: @987Nabil this was naive, not sure how to add sbt plugin testing to CI.
this does not work:
```
[error] java.lang.RuntimeException: ci.yml does not contain contents that would have been generated by sbt-github-actions; try running githubWorkflowGenerate
[error] 	at scala.sys.package$.error(package.scala:30)
[error] 	at sbtghactions.GenerativePlugin$.reportMismatch$1(GenerativePlugin.scala:919)
[error] 	at sbtghactions.GenerativePlugin$.compare$1(GenerativePlugin.scala:925)
[error] 	at sbtghactions.GenerativePlugin$.$anonfun$projectSettings$5(GenerativePlugin.scala:929)
[error] 	at sbtghactions.GenerativePlugin$.$anonfun$projectSettings$5$adapted(GenerativePlugin.scala:906)
[error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:63)
[error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:69)
[error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:283)
[error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:24)
[error] 	at sbt.Execute.work(Execute.scala:292)
[error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:283)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
[error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:65)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
[error] 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
[error] 	at java.base/java.lang.Thread.run(Thread.java:1583)
[error] (githubWorkflowCheck) ci.yml does not contain contents that would have been generated by sbt-github-actions; try running githubWorkflowGenerate
```

How can I add a CI step to run `sbt zioHttpGenSbt/scripted`?

zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/Format.scala
```diff
@@ -0,0 +1,11 @@
+package zio.http.gen.sbt
+
+sealed trait Format
+object Format {
+  case object YAML extends Format
+  case object JSON extends Format
+  def fromFileName(fileName: String): Format =
+    if (fileName.endsWith("yml") || fileName.endsWith("yaml")) YAML
+    else if (fileName.endsWith("json")) JSON
+    else throw new Exception(s"Unknown format for file $fileName")
```
hochgi, line 10: pretty naive - if needed, this can be enhanced by simply trying to parse as json, and then yaml. or even use apache-tika to detect file's mimetype.
For now, this seems good enough to me.

build.sbt
```diff
@@ -86,6 +86,22 @@ ThisBuild / githubWorkflowBuildPreamble := Seq(
...
+      WorkflowStep.Use(UseRef.Public("coursier", "setup-action", "v1")),
+      WorkflowStep.Run(
+        commands = List(s"sbt zioHttpGenSbt/scripted"),
+        name = Some("Test sbt plugin"),
+        cond = Some("${{ github.event_name == 'pull_request' }}"),
+      ),
+    ),
+    scalas = List(Scala212),
+  ).steps
+}
```
hochgi: wasteful.
I wanted to add sbt testing to CI, and this works. but…
it has scalaVersion set to 2.12 (only supporting sbt 1.x for now),
and it runs for every matrix value. which means we run the same tests multiple times.
Not sure how to opt out.

build.sbt
```diff
@@ -342,6 +359,43 @@ lazy val zioHttpGen = (project in file("zio-http-gen"))
...
+    // crossScalaVersions := Seq(Scala212, Scala213),
+    // pluginCrossBuild / sbtVersion := {
+    //   // Not sure why version 1.5.8 specifically is set to minimum version,
+    //   // but when I searched the plugins under sbt org, this seems like the most common:
+    //   // https://github.com/search?q=org%3Asbt%20pluginCrossBuild&type=code
+    //   scalaBinaryVersion.value match {
+    //     case "2.12" => "1.5.8"
+    //     case _      => "2.0.0-M2"
+    //   }
+    // },
```
hochgi: this does not work in CI.
Since sbt 2.x is not published yet (only milestone versions currently available),
I commented it out.
When sbt 2.0 comes out, this should be commented in with stable version, and it should just work.

build.sbt
```diff
@@ -342,6 +359,43 @@ lazy val zioHttpGen = (project in file("zio-http-gen"))
...
+      `zio-json-yaml`,
+      `zio-test`,
+      `zio-test-sbt`,
+      // when I tried to use the regular .dependsOn(zioHttpGen) syntax,
+      // sbt tried to resolve as an external dependency with current SNAPSHOT version,
+      // and fails (since it is not published).
+      // For now, and until this issue is resolved, we'll just use latest stable published version.
+      //
+      // FIXME: replace the dependency with .dependsOn(zioHttpGen) syntax
+      "dev.zio" %% "zio-http-gen" % "3.0.1",
```
hochgi: I tried, and wasted a few hours trying to figure out why it won't work.
In a clean multi module project, it works.
Only in zio-http repo it acts weird.

similar plugin modules as part of a larger build work in other repos (playframework, tapir, …)

Any idea what's special here?

zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/ZioHttpCodegen.scala
```diff
@@ -0,0 +1,142 @@
...
+
+      }
+    }
+  }
+
+  def listFilesRec(dir: File): List[File] =
+    sbt.io.IO.listFiles(dir).foldRight(List.empty[File]) { case (f, acc) =>
+      if (f.isDirectory) listFilesRec(f) ::: acc
+      else f :: acc
+    }
```
hochgi: nit: can be tail recursive
```scala
@tailrec def listFilesRec(dir: File, acc: List[File] = Nil): List[File] =
  sbt.io.IO.listFiles(dir).foldRight(acc) { case (f, tail) =>
    if (f.isDirectory) listFilesRec(f, tail)
    else f :: tail
  }
```

zio-http-gen-sbt-plugin/src/sbt-test/zio-http-sbt-codegen/dynamic/build.sbt
```diff
@@ -0,0 +1,44 @@
...
+  .settings(
+    name := "uspto-sdk",
+    organization := "com.example",
+    scalaVersion := "2.13.15",
+    libraryDependencies +="dev.zio" %% "zio-http" % "3.0.1",
+    ZIOpenApi / sourceGenerators += Def.task[Seq[File]] {
+
+      val outFile = (ZIOpenApi / sourceManaged).value / "gov" / "uspto" / "ibd" / "api.json"
+      val http = Gigahorse.http(Gigahorse.config)
+      val request = Gigahorse.url("https://developer.uspto.gov/ibd-api/v3/api-docs")
```
hochgi, line 19: arbitrary openapi spec. I tried so many, and every single one I checked had usage of features we don't support in zio-http-gen yet.
not the official swagger example: https://petstore3.swagger.io
nothing in this collection (didn't do an exhaustive check though): https://petstore.dev
So I tried looking for the most minimal example I could find, and even the one I found didn't work out of the box.
So I pruned the JSON, to only be left with the acceptable APIs definitions. 

In retrospect, it kind of nice to have this as an example.
Most services we deal with have many APIs, and we don't use all of them.
So no need to generate code for stuff we're not using.

I may have came up with this example out of necessity, but now I think it's actually good to have it.

hochgi: Also, I don't like the idea of tests depending on stuff outside of the build (other than conventional stuff like library dependencies of course)

But this is how I expect users to actually use it. so… ¯\\\_(ツ)\_/¯

zio-http-gen-sbt-plugin/src/sbt-test/zio-http-sbt-codegen/dynamic/test
```diff
@@ -0,0 +1,6 @@
+# compile should depend on generated sources,
+# thus we'll check it triggers codegen by inspecting expected .class files
+# compiled from expected generated sources
+> compile
+$ exists target/scala-2.13/classes/gov/uspto/ibd/v1/weeklyarchivedata/SearchWeeklyArchiveData.class
+$ exists target/scala-2.13/classes/gov/uspto/ibd/component/ArchiveDataRecord.class
```
hochgi: nit:
might want to also check for existence of `target/scala-2.13/src_managed/oapi/gov/uspto/ibd/api.json`
Just so it can serve as documentation for users where to expect dynamic openapi files to be found.

build.sbt
987Nabil: Why does it not work with 2 M2? I saw other plugins having cross-builds
If it does not, please put this code in an issue, not in a comment.

build.sbt
987Nabil: idk
But this is no option, if the plugin is in the same repo. I actually started myself to work on this issue and and problems with the sbt setup. So I created a different repo. Maybe we should do that.

zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/Format.scala
987Nabil: okay for now

zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/ZioHttpCodegen.scala
987Nabil: this methods should be private?

build.sbt
hochgi: Its solveable. same structure works in playframework and tapir projects.
It's also easier to catch integration issues.
If something changes in the `zio-http-gen` module which breaks the plugin, ideally we would want to catch that right away.

I'll try to spend some more time on this :/

zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/ZioHttpCodegen.scala
hochgi: I don't have any strong opinions about it, if you think such stateless methods in plugins are best to limit access to, then OK, I guess.

build.sbt
hochgi: I don't know why.
maybe those versions are published to another repo, which we need to add another resolver for?
might be as simple as that, I didn't checked.

an issue would be best to put such TODOs

zio-http-gen-sbt-plugin/src/main/scala/zio/http/gen/sbt/ZioHttpCodegen.scala
987Nabil: This is not a plugin to recurse through files

987Nabil:
Do I see it right, that compile always triggers the generation? I'd like it to be opt out

build.sbt
987Nabil: M2 is on maven central. See https://mvnrepository.com/artifact/org.scala-sbt/sbt

hochgi: I'll open an issue then.
should I wait until this is merged?

hochgi:
> Do I see it right, that compile always triggers the generation? I'd like it to be opt out

I assume you want this since codegen could take a lot of time, and you wouldn't want to waste it on every incremental compilation?
If that's the case, this is covered by sbt task caching mechanism I added.
If there's no change in openapi files, then there's no need to re-generate.

Is there any other reason I didn't think of?

build.sbt
hochgi: @987Nabil turns out it works. but only under `++2.12.x`.
Which is expected. When cross version is set to default 2.13, and we try to resolve only 2.12 modules it fails and resorts to trying to download.

So this should work, and we only need carful CI handling for only building/testing the plugin under ++2.12.x

987Nabil (approved):

PR #3215, opened 2024-11-19
Optimise `zio.http.Body.FileBody.asStream` code

guizmaii:

PR #3233, opened 2024-12-05
Further optimise `zio.http.Body.FileBody.asStream` code

guizmaii:
Following https://github.com/zio/zio-http/pull/3215

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Body.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Body.scala
@@ -578,13 +578,18 @@ object Body {
             (fs, size) = r
           } yield ZStream
             .repeatZIOOption[Any, Throwable, Chunk[Byte]] {
-              for {
-                buffer <- ZIO.succeed(new Array[Byte](size))
-                len    <- ZIO.attempt(fs.read(buffer)).mapError(Some(_))
-                bytes  <-
-                  if (len > 0) ZIO.succeed(Chunk.fromArray(buffer.slice(0, len)))
-                  else ZIO.fail(None)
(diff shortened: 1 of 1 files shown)


zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -569,7 +569,7 @@ object Body {
       ZStream.unwrap {
```
kyri-petrou: I don't think we need this right? Since we're already wrapping the for-comprehension in `blocking`

zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -578,13 +578,18 @@ object Body {
             (fs, size) = r
```
kyri-petrou, line 578: Maybe we should change this to `suspendSucceed` instead and use a try-catch block to avoid the additional fold?

zio-http/shared/src/main/scala/zio/http/Body.scala
guizmaii: I think we do as the for comprehensive returns a `ZIO[_, _, ZStream[...]]` 🤔

zio-http/shared/src/main/scala/zio/http/Body.scala
guizmaii: @kyri-petrou is this what you had in mind: https://github.com/zio/zio-http/pull/3234/files? 🤔

guizmaii:
Closed in favour of https://github.com/zio/zio-http/pull/3234

PR #3234, opened 2024-12-05
Further optimise `zio.http.Body.FileBody.asStream` code

guizmaii:
Following https://github.com/zio/zio-http/pull/3215

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Body.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Body.scala
@@ -462,7 +462,8 @@ object Body {
     override def asChunk(implicit trace: Trace): Task[Chunk[Byte]] = zioEmptyChunk
 
     override def asStream(implicit trace: Trace): ZStream[Any, Throwable, Byte] = ZStream.empty
-    override def isComplete: Boolean                                            = true
+
+    override def isComplete: Boolean = true
 
     override def isEmpty: Boolean = true
 
@@ -471,7 +472,8 @@ object Body {
     override private[zio] def unsafeAsArray(implicit unsafe: Unsafe): Array[Byte] = Array.empty[Byte]
 
     override def contentType(newContentType: Body.ContentType): Body = this
-    override def contentType: Option[Body.ContentType]               = None
+
+    override def contentType: Option[Body.ContentType] = None
 
     override def knownContentLength: Option[Long] = Some(0L)
   }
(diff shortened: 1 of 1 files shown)


zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -568,27 +570,32 @@ object Body {
...
+          ZIO.suspendSucceed {
+            try {
               val fs   = new FileInputStream(file)
               val size = Math.min(chunkSize.toLong, file.length()).toInt
 
-              (fs, size)
-            }
-            (fs, size) = r
-          } yield ZStream
-            .repeatZIOOption[Any, Throwable, Chunk[Byte]] {
```
guizmaii, line 580: Replacing `ZStream.repeatZIOOption` with `ZStream.repeatZIOChunkOption` allows us to remove the `.flattenChunks`

Anyway, `repeatZIOOption` is using `repeatZIOChunkOption` underneath:

Screenshot 2024-12-06 at 12 19 16 AM

zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -568,27 +570,32 @@ object Body {
...
-              for {
-                buffer <- ZIO.succeed(new Array[Byte](size))
-                len    <- ZIO.attempt(fs.read(buffer)).mapError(Some(_))
-                bytes  <-
-                  if (len > 0) ZIO.succeed(Chunk.fromArray(buffer.slice(0, len)))
-                  else ZIO.fail(None)
-              } yield bytes
+              Exit.succeed {
+                ZStream.repeatZIOChunkOption {
+                  ZIO.suspendSucceed {
```
guizmaii: @kyri-petrou Are we sure this `ZIO.suspendSucceed` and the `ZIO.attempt(fs.close())` will be run on a blocking thread? 🤔

zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -568,27 +570,35 @@ object Body {
...
+                    else Exit.none
+                  } catch {
+                    case e: Throwable => Exit.fail(e)
+                  }
+                }
+
+              Exit.succeed {
+                // Optimised for our needs version of `ZIO.repeatZIOChunkOption`
+                ZStream
+                  .unfoldChunkZIO(read)(_.map(_.map(_ -> read)))
```
guizmaii, line 593: @kyri-petrou Funnily enough, if we were managing to find a way to write a recursive type in Scala, 

1. we wouldn't need this ugly `(_.map(_.map(_ -> read))`
2. we could write this:
```scala
              type Read = Task[Option[(Chunk[Byte], Read)]]
              val read: Read =
                ZIO.suspendSucceed {
                  try {
                    val buffer = new Array[Byte](size)
                    val len    = fs.read(buffer)
                    if (len > 0) Exit.succeed(Some(Chunk.fromArray(buffer.slice(0, len) -> read)))
                    else Exit.none
                  } catch {
                    case e: Throwable => Exit.fail(e)
                  }
                }

              Exit.succeed {
                ZStream
                  .unfoldChunkZIO(read)(ZIO.identityFn)
                  .ensuring(ZIO.attempt(fs.close()).ignoreLogged)
              }
```

😍

987Nabil: I can imagine, that there is a nice way to do this for Scala 3. But I think we are unable to find a nice way for Scala 2

987Nabil (approved):

PR #3252, opened 2024-12-12
Add state machine based path matching with benchmarks

promisingcoder:
- Implement zero-allocation state machine router
- Add comprehensive benchmarks showing performance improvements:
- Update build configuration for benchmarks
- Configure JMH for performance testing
/claim #2960


987Nabil:
@promisingcoder First of all, pinging multiple ppl to review while you PR is not building, is not nice.
Besides that, I already figured out that this idea is theoretically not possible implement. State machine will always be slower for finding routes.

PR #3261, opened 2024-12-19
Better Netty Logging and Error Reporting.

scottweaver:
Added support to enable internal Netty log output.

Replaced contrived `PrematureChannelException` in `NettyClientDriver` with the actual underlying Netty exception.


987Nabil:
@scottweaver The idea is good and needed. But I think the changes in the traits are not binary compatible.
Mima should give you some hints once it ran 🙂 
Also I'd like @kyri-petrou to have a look.

zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
```diff
@@ -58,17 +58,19 @@ private[zio] final class WebSocketAppHandler(
...
 
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
     dispatch(ChannelEvent.exceptionCaught(cause))
-    onComplete match {
-      case Some(promise) => promise.unsafe.done(Exit.fail(cause))
-      case None          => ()
+    callbacks match {
+      case Some((onComplete, onFailure)) =>
+        onComplete.unsafe.done(Exit.fail(cause))
+        onFailure.unsafe.done(ZIO.succeed(cause))
```
kyri-petrou: Use `Exit.succeed(cause)` instead of `ZIO.succeed` as there's nothing to side-effecting

zio-http/jvm/src/main/scala/zio/http/netty/client/ClientFailureHandler.scala
```diff
@@ -11,12 +11,14 @@ import io.netty.channel.{ChannelHandlerContext, ChannelInboundHandlerAdapter}
...
   onComplete: Promise[Throwable, ChannelState],
+  onFailure: Promise[Nothing, Throwable],
 ) extends ChannelInboundHandlerAdapter {
   implicit private val unsafeClass: Unsafe = Unsafe.unsafe
 
   override def exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable): Unit = {
     val exit = Exit.fail(cause)
     onResponse.unsafe.done(exit)
     onComplete.unsafe.done(exit)
+    onFailure.unsafe.done(ZIO.succeed(cause))
```
kyri-petrou: `Exit.succeed`

zio-http/jvm/src/main/scala/zio/http/netty/client/WebSocketClientInboundHandler.scala
```diff
@@ -42,8 +42,10 @@ private[netty] final class WebSocketClientInboundHandler(
   }
 
   override def exceptionCaught(ctx: ChannelHandlerContext, error: Throwable): Unit = {
-    val exit = Exit.fail(error)
+    implicit val trace = Trace.empty
+    val exit           = Exit.fail(error)
     onResponse.unsafe.done(exit)
     onComplete.unsafe.done(exit)
+    onFailure.unsafe.done(ZIO.succeed(error))
```
kyri-petrou: `Exit.succeed`

zio-http/jvm/src/main/scala/zio/http/netty/client/WebSocketClientInboundHandler.scala
```diff
@@ -42,8 +42,10 @@ private[netty] final class WebSocketClientInboundHandler(
   }
 
   override def exceptionCaught(ctx: ChannelHandlerContext, error: Throwable): Unit = {
-    val exit = Exit.fail(error)
+    implicit val trace = Trace.empty
```
kyri-petrou: You can remove this once you change `ZIO.succeed` to `Exit.succeed`

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyClientDriver.scala
```diff
@@ -74,23 +87,28 @@ final case class NettyClientDriver private[netty] (
...
-          // If onComplete was already set, it means another fiber is already in the process of fulfilling the promises
-          // so we don't need to fulfill `onResponse`
           nettyRuntime.unsafeRunSync {
-            onComplete.interrupt && onResponse.fail(NettyClientDriver.PrematureChannelClosure)
+            // wait for ClientFailureHandler.exceptionCaught to complete onFailure.
+            onFailure.await.flatMap { error =>
+              // If onComplete was already set, it means another fiber is already in the process of fulfilling the promises
+              // so we don't need to fulfill `onResponse`
+              onComplete.interrupt && onResponse.fail(error)
+            }.fork
```
kyri-petrou: I have a really bad feeling about this. We're relying on `onFailure` being always completed, otherwise we're risking memory leaks. Can we ensure that there are no cases where this callback is invoked but `onFailure` was never completed? Perhaps it's better to use `poll` instead and avoid `fork` altogether? In cases that it returns `None`, we can fallback to the error in within the `ChannelFuture` (if present) and finally a generic error as previously if that's not set either.

AFAIK Netty invokes everything within the same channel in a single thread, and the `exceptionCaught` method of the handlers should be invoked before the channel future is closed.

kyri-petrou (commented):
Left some comments below, but my main concern is the failing MiMA checks. Since `zio-http` is in a stable version now, we need to ensure backwards binary compatibility at the very least for classes / methods that are publicly accessible and could be used by users.

zio-http/jvm/src/main/scala/zio/http/netty/client/WebSocketClientInboundHandler.scala
scottweaver: Done

zio-http/jvm/src/main/scala/zio/http/netty/client/ClientFailureHandler.scala
scottweaver: Done

zio-http/jvm/src/main/scala/zio/http/netty/WebSocketAppHandler.scala
scottweaver: Done

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyClientDriver.scala
scottweaver: I will look into safer alternatives.

FYI: At Capital One we are 100Ks of transactions nightly during testing at up to 300 r/s and haven't seen any sign of memory leak afaik.

scottweaver:
> Left some comments below, but my main concern is the failing MiMA checks. Since `zio-http` is in a stable version now, we need to ensure backwards binary compatibility at the very least for classes / methods that are publicly accessible and could be used by users.

The two things I see readily exposed to the end user, correct me id I missed something, are the additions of `enableInternalLogging` and `onFailure` to `ClientDriver.requestOnChannel` and  `enableInternalLogging` added to `ZClient#Config`.

For the later, `Config` has a default value set for the new `enableInternalLogging` so that should limit its impact.

For the former, I would argue that `requestOnChannel` is used exclusively by zio-http internals and is ONLY public so that alternative implementations other than Netty could be created.  

While it is not outside the realm of possibility someone, somewhere has actually done that, I would argue the vast majority of users have never even looked at the `ClientDriver` class much less even considered creating their own custom implementation.

987Nabil:
> > Left some comments below, but my main concern is the failing MiMA checks. Since `zio-http` is in a stable version now, we need to ensure backwards binary compatibility at the very least for classes / methods that are publicly accessible and could be used by users.
> 
> The two things I see readily exposed to the end user, correct me id I missed something, are the additions of `enableInternalLogging` and `onFailure` to `ClientDriver.requestOnChannel` and `enableInternalLogging` added to `ZClient#Config`.
> 
> For the later, `Config` has a default value set for the new `enableInternalLogging` so that should limit its impact.
> 
> For the former, I would argue that `requestOnChannel` is used exclusively by zio-http internals and is ONLY public so that alternative implementations other than Netty could be created.
> 
> While it is not outside the realm of possibility someone, somewhere has actually done that, I would argue the vast majority of users have never even looked at the `ClientDriver` class much less even considered creating their own custom implementation.

While that might be true, I don't want to break compat if it is not absolutely necessary.
For `requestOnChannel`, please add default values, put the values at the end of the parameter list and use [unroll](https://github.com/com-lihaoyi/unroll). And for `Config` actually just the same 🙂

scottweaver:
@987Nabil I went ahead and added Unroll and modified the methods.  

The promise argument bit is mighty ugly, but it is what I came up for a first attempt.  I welcome any input to clean that up.

scottweaver:
I have removed the concerning `await` portion of the code, gonna put the commit comments here as a summary for the new implementation:

_This set of changes improves on the initial pass in a couple of ways._

_There is no longer a need to await on the 'onFailure' promise, reasons provided below._

_All non-websocket channel handling operations now live inside 'ClientInboundHandler'.  This includes the logic for 'exceptionCaught' which is responsible for properly completing 'onComplete', 'onResponse' and 'onFailure' when an exception occurs during Netty channel pipeline processing._

_The 'ChannelInterface.interrupt' implementation now handles completion up of the 'onResponse' and 'onFailure' promises in the case where 'onComplete' is interrupted._

_The close future listener has been completely removed as its responsibilities are covered by the 'ClientInboundHandler' and the 'ChannelInterface'. Interesting discovery for me was that the listener actually gets invoked BEFORE the 'exceptionCaught' method of 'ClientInboundHandler'.  This is what lead be to the decision to completely remove the listener all together._

scottweaver:
~~Well, I "fixed" some of the mima compat issues in my local env.~~

~~Unroll is working fine for `requestOnChannel`, however, it fails to rewrite `ZClient.Config` with a compiler crash.~~

~~I am wondering if Unroll doesn't like the nested case class?~~

Nvm, user error.  I forgot to include the default value :(

zio-http/jvm/src/main/scala/zio/http/netty/client/ClientInboundHandler.scala
```diff
@@ -35,6 +35,7 @@ private[netty] final class ClientInboundHandler(
   jReq: HttpRequest,
   onResponse: Promise[Throwable, Response],
   onComplete: Promise[Throwable, ChannelState],
+  onFailure: Promise[Nothing, Throwable],
```
kyri-petrou: How is this `Promise` used? I can't find any usage of `await` or `poll` on it, so AFAICT we're only completing it without really using it?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyClientDriver.scala
```diff
@@ -73,45 +91,31 @@ final case class NettyClientDriver private[netty] (
         }
       }
       .map { jReq =>
-        val closeListener: GenericFutureListener[ChannelFuture] = { (_: ChannelFuture) =>
-          // If onComplete was already set, it means another fiber is already in the process of fulfilling the promises
-          // so we don't need to fulfill `onResponse`
-          nettyRuntime.unsafeRunSync {
-            onComplete.interrupt && onResponse.fail(NettyClientDriver.PrematureChannelClosure)
-          }(Unsafe.unsafe, trace): Unit
-        }
```
kyri-petrou, line 82: This existed in order to account for cases that the error is not propagated via the pipeline, e.g., if the server drops the connection unexpectedly or if we errored before reaching the `ClientInboundHandler`.

Have you tested what happens in those cases? Are we still fulfilling the promises or are we risking the application hanging?

scottweaver: > This existed in order to account for cases that the error is not propagated via the pipeline, e.g., if the server drops the connection unexpectedly or if we errored before reaching the ClientInboundHandler.

Server dropping a connection results in a `PrematureChannelClosure` exception from Netty.  I have tested this at C1 using Mockserver configured to drop incoming connections.  It even fires correctly in the SSLSpec when there is no valid cert.

Other than that, do you have any other specific cases we can test for?  AFAIK, the Netty pipeline is pretty durable and you can usually depend on the `exceptionCaught` being executed.

scottweaver: Additionally, if you interrupt/fail promises in the listener, this occurs before `exceptionCaught` is called and we are then unable to propagate the underlying exception correctly.

987Nabil:
@scottweaver I tried using unroll on a different PR and had to remove it, since docs fail when the compiler plugin is active 😕 
I see the same error here.

scottweaver:
Yeah, I hit the same issue 😔

scottweaver:
@987Nabil what is the path forward now?

987Nabil:
Manually adding the methods variants that internally use a default value

bejewelled:
Any updates on the status of this PR? Is it still being actively worked on?

Issue #3183, opened 2024-09-29
Add auth to generated OpenAPI


987Nabil:
The auth specified on an `Endpoint` should be reflected in the generated OpenAPI object

melmathari:
/attempt #3183

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm39jluen0010ju03zxm0n378/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

SAIKIRANSURAPALLI:
/attempt #3183

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm39jluen0010ju03zxm0n378/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

ffprivacy:
/attempt #3183

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm39jluen0010ju03zxm0n378/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

ffprivacy:
Digest is for now not implemented by js client SwaggerUIBundle: https://github.com/swagger-api/swagger-ui/issues/8264

palanga:
Any updates on this?

PR #3266, opened 2024-12-31
feat: Encode Auth requirements in Endpoint API

### LINKED ISSUE #2592: Encode Authorization requirements in `Endpoint` API
Currently, we can require the authorization header in the endpoint API, but we do not know, if this is for example Basic Auth or Bearer Auth. This makes it impossible to generate auth schemas for OpenAPI for example.

Two possible solutions might be

- Explicit auth schema on `Endpoint`
- split the generic `Authorization` header codec into specific ones (`HeaderCodec.basicAuth` etc)

mobley-trent:
Closes #2592 
/claim #2592


mobley-trent:
Working on integrating Auth scopes into OpenAPI

zio-http/shared/src/main/scala/zio/http/endpoint/AuthType.scala
```diff
@@ -15,6 +15,13 @@ sealed trait AuthType { self =>
       .asInstanceOf[
         AuthType { type ClientRequirement = ClientReq },
       ]
+
+  private var authScopes: Option[List[String]]  = scala.None
```
987Nabil: There is no need to have a var here. Please create a new subtype of AuthType that can hold a list of scopes and a auth type.

zio-http/shared/src/main/scala/zio/http/endpoint/AuthType.scala
```diff
@@ -15,6 +15,13 @@ sealed trait AuthType { self =>
...
+  def addScopes(scopes: List[String]): AuthType = {
+    authScopes = Some(scopes)
+    self
+  }
+  def getScopes: Option[List[String]]           = authScopes
```
987Nabil: Please remove the Option. There is no difference between empty List and None.

zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -1160,8 +1160,21 @@ object Header {
 
     final case class Basic(username: String, password: Secret) extends Authorization
 
-    object Basic {
+    object Basic extends HeaderType {
       def apply(username: String, password: String): Basic = new Basic(username, Secret(password))
+
+      override def name: String = "basicAuth"
```
987Nabil: This is just wrong. There is no header named "basicAuth" this is "authorization".
Below as well

987Nabil (changes requested):
Might add more later on

zio-http/jvm/src/test/scala/zio/http/endpoint/openapi/OpenAPIGenSpec.scala
```diff
@@ -231,7 +231,7 @@ object OpenAPIGenSpec extends ZIOSpecDefault {
       .outError[NotFoundError](Status.NotFound)
 
   private val endpointWithAuthScopes =
```
987Nabil, line 233: The public API before was okay. Just the internal impl. I wanted you to change. Sorry, I was maybe not clear enough

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -198,7 +198,10 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
   def auth[Auth0 <: AuthType](auth: Auth0): Endpoint[PathInput, Input, Err, Output, Auth0] =
     copy(authType = auth)
 
-  def authScopes: Option[List[String]] = authType.asInstanceOf[AuthType].getScopes
+  def authScopes: List[String] = authType match {
+    case AuthType.ScopedAuth(_, scopes) => scopes
```
987Nabil: This could be recursive.

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -198,11 +198,18 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
...
+  def authScopes(authType: AuthType): List[String] = authType match {
+    case AuthType.ScopedAuth(nestedAuth, scopes) =>
+      scopes ++ authScopes(nestedAuth)
+    case AuthType.Or(auth1, auth2, _)            =>
+      authScopes(auth1) ++ authScopes(auth2)
+    case _                                       =>
+      Nil
   }
 
+  def scopes(scopes: String*): Endpoint[PathInput, Input, Err, Output, AuthType] =
```
987Nabil, line 212: This method should be called with at least one scope and only if the authType is not `AuthType.None`

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -198,11 +198,18 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
   def auth[Auth0 <: AuthType](auth: Auth0): Endpoint[PathInput, Input, Err, Output, Auth0] =
     copy(authType = auth)
 
-  def authScopes: List[String] = authType match {
-    case AuthType.ScopedAuth(_, scopes) => scopes
-    case _                              => Nil
+  def authScopes(authType: AuthType): List[String] = authType match {
```
987Nabil: The recursive call should only be happening in a private method. The public one should have no params. Also to keep the duality to similar methods, it should be named `scopes`

zio-http/shared/src/main/scala/zio/http/endpoint/AuthType.scala
```diff
@@ -15,6 +15,7 @@ sealed trait AuthType { self =>
       .asInstanceOf[
         AuthType { type ClientRequirement = ClientReq },
       ]
+
```
987Nabil, line 18: Please add here methods to add and get scopes. In `Endpoint` there should be just proxies to these.

zio-http/shared/src/main/scala/zio/http/endpoint/AuthType.scala
```diff
@@ -67,5 +67,9 @@ object AuthType {
   ) extends AuthType {
     type ClientRequirement = ClientReq
     override val codec: HttpCodec[HttpCodecType.RequestType, ClientReq] = authType.codec
+
+    def getScopes: List[String] = scopes
```
987Nabil: in the style of zio-http we would call both just `scopes`

987Nabil:
Please fix the bin compat. issue reported by mima

987Nabil (approved):

mobley-trent:
@987Nabil does this also close #3183 ?

cc @jdegoes

PR #3269, opened 2025-01-09
Fix Possible Race Condition

### LINKED ISSUE #3268: Possible Race Condition in NettyConnectionPool
**Describe the bug**
On rare occasions,  `NettConnectionPool` will crash the fiber when attempting to refresh a connection pulled from the pool.

The underlying failure is `java.util.NoSuchElementException: READ_TIMEOUT_HANDLER` (see stacktrace below).

My hypothesis is that this is a case where  a `Channel` that is in the process of being closed and released is pulled from the pool and after the internal Netty logic has removed all handlers from the pipeline.  This is why the `ChannelPipeline.replace` is failing with a `NoSuchElementException`.

My line of thinking is based on this [Stackoverflow comment](https://stackoverflow.com/a/44910867).  It is the only way I could think of that would cause a handler that we know should be on a Channel's pipeline to magically disappear.

I will have a PR out for this soon that wraps the `ChannelPipline.replace` with a `ZIO.attempt` so that failure no longer crashes the fiber and should be correctly retried in the subsequent `NettyConnectionPool` logic if the `replace` fails.

I am also refactoring the logic so that it is always calling `Channel.isOpen` to determine whether or not the Channel in question is useable or not.

**To Reproduce**
Difficult to repro as it is most likely a race condition that only shows up rarely or under heavy load.

**Expected behaviour**
Should never crash the fiber and a fresh, open Channel should be returned if the Channel is closed.

**Stacktrace**
```
Exception in thread "zio-fiber-1467137482,1469468003" java.util.NoSuchElementException: READ_TIMEOUT_HANDLER
at io.netty.channel.DefaultChannelPipeline.getContextOrDie(DefaultChannelPipeline.java:1022)
at io.netty.channel.DefaultChannelPipeline.replace(DefaultChannelPipeline.java:464)
at zio.http.netty.client.NettyConnectionPool$.zio$http$netty$client$NettyConnectionPool$$$refreshIdleTimeoutHandler(NettyConnectionPool.scala:159)
at zio.http.netty.client.NettyConnectionPool$ZioNettyConnectionPool.get$$anonfun$1$$anonfun$2$$anonfun$2(NettyConnectionPool.scala:260)
at scala.Option.fold(Option.scala:263)
at zio.http.netty.client.NettyConnectionPool$ZioNettyConnectionPool.get$$anonfun$1$$anonfun$2(NettyConnectionPool.scala:260)
(removed internal stack info -scott)
at zio.http.HandlerAspect.applyHandler(HandlerAspect.scala:133)
at zio.http.HandlerAspect.applyHandler(HandlerAspect.scala:136)
at zio.http.HandlerAspect.applyHandler(HandlerAspect.scala:133)
at zio.http.HandlerAspect.applyHandler(HandlerAspect.scala:136)
```

**Additional context**

scottweaver:
fixes #3268

Refactor 'refreshIdleTimeoutHandler' so that 'ChannelPipeline.replace' does not crash the fiber on unhandled exceptions.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
@@ -147,9 +147,9 @@ private[netty] object NettyConnectionPool {
    *   true if the handler was successfully refreshed prior to the channel being
    *   closed and the channel is still open.
    *
-   *   true if the `timeout` is None and the channel is still open.
+   * true if the `timeout` is None and the channel is still open.
    *
-   *   false of the channel is closed.
+   * false of the channel is closed.
    */
   private def refreshIdleTimeoutHandler(
     channel: JChannel,
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
@@ -142,23 +142,31 @@ private[netty] object NettyConnectionPool {
   }
 
   /**
-   * Refreshes the idle timeout handler on the channel pipeline.
+   * Attempts to refresh the idle timeout handler on the channel pipeline.
    * @return
    *   true if the handler was successfully refreshed prior to the channel being
-   *   closed
+   *   closed and the channel is still open.
+   *
+   *   true if the `timeout` is None and the channel is still open.
+   *
+   *   false of the channel is closed.
    */
   private def refreshIdleTimeoutHandler(
     channel: JChannel,
-    timeout: Duration,
-  ): Boolean = {
-    channel
-      .pipeline()
-      .replace(
-        Names.ReadTimeoutHandler,
-        Names.ReadTimeoutHandler,
-        new ReadTimeoutHandler(timeout.toMillis, TimeUnit.MILLISECONDS),
-      )
-    channel.isOpen
+    timeout: Option[Duration],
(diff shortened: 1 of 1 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -142,23 +142,31 @@ private[netty] object NettyConnectionPool {
...
+        channel
+          .pipeline()
+          .replace(
+            Names.ReadTimeoutHandler,
+            Names.ReadTimeoutHandler,
+            new ReadTimeoutHandler(timeout.toMillis, TimeUnit.MILLISECONDS),
+          )
+      }
+    }.when(channel.isOpen)
+      .as(channel.isOpen)
```
kyri-petrou, line 169: This "breaks" the current logic for retrying channels from the pool until we succeed. I think a better solution would be to keep the existing code as is and wrap it in a `try-catch` block so that we don't fail requests when this happens

```scala
  private def refreshIdleTimeoutHandler(
    channel: JChannel,
    timeout: Duration,
  ): Boolean = {
  try {
    channel
      .pipeline()
      .replace(
        Names.ReadTimeoutHandler,
        Names.ReadTimeoutHandler,
        new ReadTimeoutHandler(timeout.toMillis, TimeUnit.MILLISECONDS),
      )
    channel.isOpen
  } catch {
    case _: NoSuchElementException => false
  }
```

987Nabil:
@scottweaver @kyri-petrou if there is something left to do here, please reopen

PR #3270, opened 2025-01-09
Schema based header codecs, unified with query codecs (#3232)

### LINKED ISSUE #3232: HttpCodec header constructor based on `Schema`
Should work the same as with query params.
I was thinking that we could even do an equivalent to `queryAll` that takes a case class and looks up all headers that correlate with the field names.

### LINKED ISSUE #3240: Missing cases in query codec for nonempty schemas and transform schemas
For collections we can have now a unified check besed on the Collections interface of Schema.
The transform is just missing.

### LINKED ISSUE #3255: Codec for repeated headers
Currently, we can not define a codec that handles the same header key more than once.
This is a case that is not uncommon and should be supported.

We will need a new HttpCodec subtype. Either
- new wrapper around a header codec
- new variant of header codec (maybe schema based)

There might need to be adjustments in code gens and openAPI/.http gen

987Nabil:
Fix for publish CI job

fixes #3232
/claim #3232

fixes #3255
fixes #3240


zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -64,6 +64,7 @@ object Header {
     type Typed[HV] = HeaderType { type HeaderValue = HV }
   }
 
+  // @deprecated("Use Schema based header codecs instead", "3.1.0")
```
jdegoes: Dead code

zio-http/shared/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -2264,140 +2269,220 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
...
-            namesAndCodecs
-          } else {
-            namesAndCodecs
-          }
+  object Query {
+    def apply[A](name: String, schema: Schema[A]): Query[A, A] = Query(SchemaCodec(Some(name), schema))
+    def apply[A](schema: Schema[A]): Query[A, A]               = Query(SchemaCodec(None, schema))
+  }
+
+  final case class SchemaCodec[A](name: Option[String], schema: Schema[A], kebabCase: Boolean = false) {
```
jdegoes: Shouldn't this be made as private as possible? Including structure and methods? If it's really intended to be user-land, we'll want to review it more carefully and move it to the top level.

zio-http/shared/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -2264,140 +2269,220 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
...
+      fields.map(unlazyField).map {
+        case field if field.schema.isInstanceOf[Schema.Collection[_, _]] =>
+          val elementSchema = field.schema.asInstanceOf[Schema.Collection[_, _]] match {
+            case s: Schema.NonEmptySequence[_, _, _] => s.elementSchema
+            case s: Schema.Sequence[_, _, _]         => s.elementSchema
+            case s: Schema.Set[_]                    => s.elementSchema
+            case _: Schema.Map[_, _]                 => throw new IllegalArgumentException("Maps are not supported")
+            case _: Schema.NonEmptyMap[_, _]         => throw new IllegalArgumentException("Maps are not supported")
+          }
+          val codec         = SchemaCodec(Some(if (!kebabCase) field.name else camelToKebab(field.name)), elementSchema)
```
jdegoes: The capitalization is not important? e.g. `Max-Age` versus `maX-aGE`?

zio-http/shared/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -2264,140 +2269,220 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
...
-      }
+    val recordSchema: Schema.Record[Any] = schema match {
+      case record: Schema.Record[_]                                         =>
+        record.asInstanceOf[Schema.Record[Any]]
+      case s: Schema.Optional[_] if s.schema.isInstanceOf[Schema.Record[_]] =>
+        s.schema.asInstanceOf[Schema.Record[Any]]
+      case _                                                                => null
+    }
 
+    val stringCodec: StringCodec[Any] =
```
jdegoes: Again, make all things private!!

zio-http/shared/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -2264,140 +2269,220 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
...
+                  identity,
+                )
+              case _                                   => null.asInstanceOf[A]
+            }
+        }
+      }
+
+  }
+
+  object SchemaCodec {
```
jdegoes: Ditto about private or public (& locked down) top-level. 

I am not sure `SchemaCodec` is a good name for this so we'll have to do better if you can't make the whole thing private.

zio-http/shared/src/main/scala/zio/http/codec/HttpCodec.scala
```diff
@@ -2409,7 +2494,34 @@ object HttpCodec extends ContentCodecs with HeaderCodecs with MethodCodecs with
     def index(index: Int): Method[A] = copy(index = index)
   }
 
-  private[http] final case class Header[A](name: String, textCodec: TextCodec[A], index: Int = 0)
+  private[http] final case class HeaderCustom[A](codec: SchemaCodec[A], index: Int = 0)
```
jdegoes: Isn't this just the same as name = `None`?

zio-http/shared/src/main/scala/zio/http/codec/StringCodec.scala
```diff
@@ -0,0 +1,394 @@
...
+import zio.stream._
+
+import zio.schema._
+import zio.schema.annotation.simpleEnum
+import zio.schema.codec._
+
+import zio.http.Charsets
+
+object StringCodec {
+  type StringCodec[A] = Codec[String, Char, A]
```
jdegoes: I'd rather put this in a `final case class StringCodec(codec: Codec[String, Char, A])` than have a type alias for it.

Remind me why we need this?

zio-http/shared/src/main/scala/zio/http/codec/StringCodec.scala
```diff
@@ -0,0 +1,394 @@
...
+          case StandardType.UnitType   => true
+          case StandardType.StringType => true
+          case StandardType.BinaryType => true
+          case StandardType.CharType   => true
+          case _                       => false
+        }
+    }
+  }
+
+  implicit def fromSchema[A](implicit schema: Schema[A]): Codec[String, Char, A] = {
```
jdegoes: Can we do this whole thing as `Schema#toString(value)` and `Schema#fromString(string)`?? Why does this even belong here?

zio-http/shared/src/main/scala/zio/http/codec/internal/Atomized.scala
```diff
@@ -25,30 +25,34 @@ private[http] final case class Atomized[A](
...
-      case HttpCodec.AtomTag.Content => copy(content = f(content))
-      case HttpCodec.AtomTag.Query   => copy(query = f(query))
-      case HttpCodec.AtomTag.Header  => copy(header = f(header))
-      case HttpCodec.AtomTag.Method  => copy(method = f(method))
+      case HttpCodec.AtomTag.Status       => copy(status = f(status))
+      case HttpCodec.AtomTag.Path         => copy(path = f(path))
+      case HttpCodec.AtomTag.Content      => copy(content = f(content))
+      case HttpCodec.AtomTag.Query        => copy(query = f(query))
+      case HttpCodec.AtomTag.Header       => copy(header = f(header))
+      case HttpCodec.AtomTag.HeaderCustom => copy(headerCustom = f(header))
```
jdegoes: It's painful to see Header and HeaderCustom when all the other cases just have a single codec!

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -220,171 +222,291 @@ private[codec] object EncoderDecoder {
...
+          val default    = query.codec.defaultValue
+          if (codec.isPrimitive) {
+            val name     = query.nameUnsafe
+            val hasParam = queryParams.hasQueryParam(name)
+            if (
+              (!hasParam || (queryParams
+                .unsafeQueryParam(name) == "" && !emptyStringIsValue(codec.codec.schema))) && hasDefault
+            )
+              default
+            else if (!hasParam)
```
jdegoes: This complicated and fragile code suggests one way to refactor this is as a method on the query codec which returns a sealed trait containing the right information.

In other words, instead of, `if (isA) then extractA else if (isB) then extractB`, you can just have a `thing match { case A(a) => ... }`

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -220,171 +222,291 @@ private[codec] object EncoderDecoder {
...
-                if (validationErrors.nonEmpty) throw HttpCodecError.InvalidEntity.wrap(validationErrors)
-                if (isOptional && decoded == None && emptyStringIsValue(schema.asInstanceOf[Schema.Optional[_]].schema))
-                  Some("")
-                else decoded
+              val validationErrors = codec.codec.erasedSchema.validate(decoded)(codec.codec.erasedSchema)
+              if (validationErrors.nonEmpty) throw HttpCodecError.InvalidEntity.wrap(validationErrors)
+              else decoded
+            }
+
+          } else if (codec.isCollection) {
```
jdegoes: Definitely need to adopt the above pattern to deal with this. Some pre-analysis with pre-collection of required information, packaged into a `sealed trait`.

zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -220,171 +222,291 @@ private[codec] object EncoderDecoder {
...
+                  }
+                case (field, codec)                                                       =>
+                  val value   = queryParams.queryParamOrElse(field.fieldName, null)
+                  val decoded = {
+                    if (value == null || (value == "" && !emptyStringIsValue(codec.schema))) codec.defaultValue
+                    else decodeAndUnwrap(field, codec, value, HttpCodecError.MalformedQueryParam.apply)
+                  }
+                  validateDecoded(codec, decoded)
+              }
+              if (optional) {
```
jdegoes: Ditto for all this.

jdegoes (changes requested):
Very close, though there is some cleanup I'd like to see happen before merging!

zio-http/shared/src/main/scala/zio/http/codec/HttpCodec.scala
987Nabil: RFC says header names are case-insensitive

zio-http/shared/src/main/scala/zio/http/internal/StringSchemaCodec.scala
```diff
@@ -0,0 +1,658 @@
...
+
+private[http] final case class ErrorConstructor(
+  missing: String => HttpCodecError,
+  missingAll: Chunk[String] => HttpCodecError,
+  invalid: Chunk[ValidationError] => HttpCodecError,
+  malformed: (String, DecodeError) => HttpCodecError,
+  invalidCount: (String, Int, Int) => HttpCodecError,
+)
+
+private[http] final case class StringSchemaCodec[A, Target](
```
jdegoes: case classes that store functions are generally better off as `trait`s (interfaces implemented with anonymous classes).

zio-http/shared/src/main/scala/zio/http/internal/StringSchemaCodec.scala
```diff
@@ -0,0 +1,658 @@
...
+  private[http] val unsafeGet: (Target, String) => String,
+  private[http] val getAll: (Target, String) => Chunk[String],
+  private[http] val count: (Target, String) => Int,
+  private[http] val error: ErrorConstructor,
+  private[http] val kebabCase: Boolean,
+  private[http] val defaultValue: A,
+  private[http] val isOptional: Boolean = false,
+  private[http] val isOptionalSchema: Boolean = false,
+) {
+  private[http] val recordFields: Chunk[(Schema.Field[_, _], PrimitiveCodec[Any])] = {
```
jdegoes, line 40: Consider `DynamicValue` to cleanup (it's basically `Json`).

zio-http/shared/src/main/scala/zio/http/internal/StringSchemaCodec.scala
```diff
@@ -0,0 +1,658 @@
...
+      unsafeGet,
+      getAll,
+      count,
+      error,
+      kebabCase,
+      null,
+    )
+}
+
+object StringSchemaCodec {
```
jdegoes: package private

jdegoes (approved):

PR #3276, opened 2025-01-18
Provide a `Scope` per server call (#3197)

### LINKED ISSUE #3197: It is not possible to express a per-request Scope in a Handler that covers the entire request/response lifetime

987Nabil:
fixes https://github.com/zio/zio-http/issues/3197
/claim https://github.com/zio/zio-http/issues/3197

### DIFF
--- a/zio-http-testkit/src/main/scala/zio/http/TestServer.scala
+++ b/zio-http-testkit/src/main/scala/zio/http/TestServer.scala
@@ -98,7 +98,7 @@ final case class TestServer(driver: Driver, bindPort: Int) extends Server {
       _ <- driver.addApp(provided, r)
     } yield ()
 
-  override def install[R](routes: Routes[R, Response])(implicit
+  override def install[R](routes: Routes[Scope & R, Response])(implicit
     trace: zio.Trace,
     tag: EnvironmentTag[R],
   ): URIO[R, Unit] =
--- a/zio-http-testkit/src/main/scala/zio/http/TestServer.scala
+++ b/zio-http-testkit/src/main/scala/zio/http/TestServer.scala
@@ -98,7 +98,7 @@ final case class TestServer(driver: Driver, bindPort: Int) extends Server {
       _ <- driver.addApp(provided, r)
     } yield ()
 
-  override def install[R](routes: Routes[Scope & R, Response])(implicit
+  override private[http] def installInternal[R](routes: Routes[R, Response])(implicit
     trace: zio.Trace,
     tag: EnvironmentTag[R],
   ): URIO[R, Unit] =
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -334,13 +334,16 @@ private[zio] final case class ServerInboundHandler(
         )
       }
 
-    val program = exit.foldCauseZIO(
-      _.failureOrCause match {
-        case Left(resp)                      => writeResponse(resp)
-        case Right(c) if c.isInterruptedOnly => closeChannel()
-        case Right(c)                        => writeResponse(withDefaultErrorResponse(FiberFailure(c)))
-      },
-      writeResponse,
+    val scope   = Scope.unsafe.make
+    val program = scope.use(
+      exit.foldCauseZIO(
+        _.failureOrCause match {
+          case Left(resp)                      => writeResponse(resp)
+          case Right(c) if c.isInterruptedOnly => closeChannel()
+          case Right(c)                        => writeResponse(withDefaultErrorResponse(FiberFailure(c)))
+        },
+        writeResponse,
+      ),
     )
 
     runtime.run(ctx, ensured, preferOnCurrentThread = avoidCtxSwitching)(program)
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -206,7 +206,7 @@ private[zio] final case class ServerInboundHandler(
   private def attemptImmediateWrite(
     ctx: ChannelHandlerContext,
     method: Method,
-    exit: ZIO[Any, Response, Response],
+    exit: ZIO[Scope, Response, Response],
   ): Boolean = {
     exit match {
       case Exit.Success(response) if response ne null =>
@@ -282,7 +282,7 @@ private[zio] final case class ServerInboundHandler(
         ZIO.suspend {
           val nettyChannel     = NettyChannel.make[JWebSocketFrame](ctx.channel())
           val webSocketChannel = WebSocketChannel.make(nettyChannel, queue, handshakeCompleted)
-          webSocketApp.handler.runZIO(webSocketChannel).ignoreLogged.forkDaemon
+          ZIO.scoped(webSocketApp.handler.runZIO(webSocketChannel)).ignoreLogged.forkDaemon
         }
       }
     _                  <- ZIO.attempt {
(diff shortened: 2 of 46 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -92,7 +92,8 @@ private[zio] final case class ServerInboundHandler(
             if (attemptImmediateWrite(ctx, req.method, exit)) {
               releaseRequest()
             } else {
-              writeResponse(ctx, runtime, exit, req)(releaseRequest)
+              val scope = Scope.unsafe.make
+              writeResponse(ctx, runtime, scope.use(exit), req)(releaseRequest)
```
jgranstrom: Can this not close the scope before the response is written? So, for example, if a resource in the Scope is a scoped stream, it can close before being fully written to the response by being closed here when the handler finishes, and not when actually writing it is done

987Nabil: You are right, it needed to be one level more down. I added a test for streaming, that failed with this code and is now green after pulling down the scope one level. Thanks for the review.

zio-http-testkit/src/main/scala/zio/http/TestServer.scala
```diff
@@ -98,7 +98,7 @@ final case class TestServer(driver: Driver, bindPort: Int) extends Server {
       _ <- driver.addApp(provided, r)
     } yield ()
 
-  override def install[R](routes: Routes[R, Response])(implicit
+  override def install[R](routes: Routes[Scope & R, Response])(implicit
```
guersam: What's the purpose of this `Scope` requirement?

I'd prefer hiding it from the `Routes` because it makes users to provide `Scope` that is not actually used.

987Nabil: The Scope has to be here, so that it is "eaten" up by the method. The effect that is returned has no more requirement of scope. That means, it is not possible to use a Scope that extends over a single request from inside a request. That is intentionally

guersam: I see. I misinterpreted the code due to [this line](https://github.com/zio/zio-http/pull/3276/files#diff-4ad96889091924d6e59473328c8a48cdafc18aed92c0f39c92341804a6a5f768R122).

I also wondered if we could do the elimination earlier, `RoutePattern#->`, `Routes.apply` or etc. for example, but it could be error prone and let the compiler do more work.

987Nabil: No, the routes is the right place. We need to close the Scope in the driver, after netty write is over.
If we eliminate the Scope earlier, we would need to close the scope earlier too.
Else you do route.run(..) and your scope is not closed/provided. That would be a bug

guersam: By 'eliminating', I mean something like this:

```scala
trait Route[-Env, +Err] {
  def apply(request: Request): ZIO[Scope & Env, Response, Response] = ???
}

case class RoutePattern[A](...) {
  def ->[Env, Err, I](handler: Handler[Scope & Env, Err, I, Response]): Route[Env, Err] = ???
}
```

Unlike the other environments, `Scope` is a local requirement per request. I think this encoding reveals the relationship more clearly, but there's two downsides that we need to modify every route forming function including Endpoint DSL and the burden to the compiler of the added type calculation per route . In this respect, I think the current design in this PR is still good.

987Nabil: @guersam I tried that out. It complicates the code, has negative impact on type inference and all just to save `Scope` on the type of `Routes` seems like a bad trade off

jgranstrom: > @guersam I tried that out. It complicates the code, has negative impact on type inference and all just to save `Scope` on the type of `Routes` seems like a bad trade off

This was my conclusion on https://github.com/zio/zio-http/pull/3229 as well, I agree with @987Nabil  this is the best approach to not make it very convoluted.

jgranstrom:
FWIW I tested this on the real world case of https://github.com/zio/zio-http/issues/3197 and it looks good.

987Nabil:
@guersam @jgranstrom @kyri-petrou I changed the design after talking to @jdegoes 
please take a look again.

guersam:
I'm excited to see how the new design looks in idiomatic ZIO. It must have involved subtle internal design decisions and thorough changes across a wide range of the codebase. I really appreciate your continuous efforts to improve it!

kyri-petrou (approved):

zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -1051,6 +1058,9 @@ object Handler extends HandlerPlatformSpecific with HandlerVersionSpecific {
   def fromResponseZIO[R, Err](getResponse: ZIO[R, Err, Response]): Handler[R, Err, Any, Response] =
     fromZIO(getResponse)
 
+  def scoped[R]: ScopedPartiallyApplied[R] =
+    new ScopedPartiallyApplied[R]()
```
jdegoes: Make:

```scala
_scopedPartiallyApplied.asInstanceOf[ScopePartiallyApplied[R]]
```

where:

```scala
private val _scopedPartiallyApplied = new ScopedPartiallyApplied[Any]()
```

zio-http/shared/src/main/scala/zio/http/Middleware.scala
```diff
@@ -181,7 +181,7 @@ object Middleware extends HandlerAspects {
           handler { (req: Request) =>
             if (req.headers.contains(header.name)) h(req)
             else h(req.addHeader(make))
-          }
+          }.asInstanceOf[Handler[Env1, Response, Request, Response]]
```
jdegoes: `Handler.scoped` here and elsewhere

zio-http/shared/src/main/scala/zio/http/Route.scala
```diff
@@ -437,7 +437,10 @@ object Route                   {
     location: Trace,
   ) extends Route[Env, Nothing] {
     override def toHandler(implicit ev: Nothing <:< Response, trace: Trace): Handler[Env, Response, Request, Response] =
-      Handler.fromZIO(handler(routePattern).map(_.sandbox)).flatten
+      Handler
+        .fromZIO(handler(routePattern).map(_.sandbox))
+        .flatten
+        .asInstanceOf[Handler[Env, Response, Request, Response]]
```
jdegoes: Ditto

jdegoes:
Ideally this would NOT allow R to extend Scope:

```scala
trait Server {

  /**
   * Installs the given HTTP application into the server.
   */
  def install[R](routes: Routes[R, Response])(implicit trace: Trace, tag: EnvironmentTag[R]): URIO[R, Unit]
```

Using a macro (making this old version private so it's there for backward compatibility only).

PR #3277, opened 2025-01-18
Use HashMap for literals look up (#2960)

### LINKED ISSUE #2960: Improve `Literal` path matching via state machine

987Nabil:
A state machine can not improve literal path matching, since it is only fast for misses, but we need the best performance for hits. This is best done with hash maps. 

This PR adds a benchmark for literal path matching and replaces the current `ListMap` with `HashMap`. This leads to a significant performance increase, as can be seen below

```
Run on MacBook Pro M1 Max
Old implementation:
[info] Benchmark                                   Mode  Cnt   Score    Error  Units
[info] RoutesBenchmark.benchmarkSmallDataZioApi   thrpt    3  45.794 ± 16.348  ops/s
[info] RoutesBenchmark.benchmarkSmallDataZioApi2  thrpt    3  63.488 ± 10.817  ops/s
[info] RoutesBenchmark.notFound1                  thrpt    3  60.807 ± 13.060  ops/s
[info] RoutesBenchmark.notFound2                  thrpt    3  60.809 ± 13.525  ops/s

HashMap instead of ListMap:
[info] Benchmark                                   Mode  Cnt   Score    Error  Units
[info] RoutesBenchmark.benchmarkSmallDataZioApi   thrpt    3  67.768 ±  9.438  ops/s
[info] RoutesBenchmark.benchmarkSmallDataZioApi2  thrpt    3  78.646 ± 15.662  ops/s
[info] RoutesBenchmark.notFound1                  thrpt    3  79.978 ±  3.829  ops/s
[info] RoutesBenchmark.notFound2                  thrpt    3  87.911 ±  9.178  ops/s
```

closes #2960
/claim #2960

### DIFF
--- a/zio-http-benchmarks/src/main/scala/zio/http/benchmark/RoutesBenchmark.scala
+++ b/zio-http-benchmarks/src/main/scala/zio/http/benchmark/RoutesBenchmark.scala
@@ -1,0 +1,77 @@
+package zio.http.benchmark
+
+import java.util.concurrent.TimeUnit
+
+import scala.util.Random
+
+import zio.{Runtime, Unsafe, ZIO}
+
+import zio.http.endpoint.Endpoint
+import zio.http.{Handler, Method, Request, Routes}
+
+import org.openjdk.jmh.annotations._
+
+@State(Scope.Thread)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class RoutesBenchmark {
+
+  val REPEAT_N = 1000
+
+  val paths = ('a' to 'z').inits.map(_.mkString).toList.reverse.tail
+
+  val routes = Routes.fromIterable(paths.map(p => Endpoint(Method.GET / p).out[Unit].implementHandler(Handler.unit)))
+
+  val requests = paths.map(p => Request.get(p))
+
+  def request: Request  = requests(Random.nextInt(requests.size))
+  val smallDataRequests = Array.fill(REPEAT_N)(request)
+
+  def unsafeRun[E, A](zio: ZIO[Any, E, A]): Unit = Unsafe.unsafe { implicit unsafe =>
+    Runtime.default.unsafe
+      .run(zio.unit)
+      .getOrThrowFiberFailure()
+  }
+
+  val paths2 = ('b' to 'z').inits.map(_.mkString).toList.reverse.tail
+
+  val routes2 = Routes.fromIterable(paths2.map(p => Endpoint(Method.GET / p).out[Unit].implementHandler(Handler.unit)))
+
+  val requests2 = requests ++ paths2.map(p => Request.get(p))
+
+  def request2: Request = requests2(Random.nextInt(requests2.size))
+
+  val smallDataRequests2 = Array.fill(REPEAT_N)(request2)
+
+  val routes3 = Routes(
+    Endpoint(Method.GET / "api").out[Unit].implementHandler(Handler.unit),
+    Endpoint(Method.GET / "ui").out[Unit].implementHandler(Handler.unit),
+  )
+
+  val requests3 = Array.fill(REPEAT_N)(List(Request.get("api"), Request.get("ui"))(Random.nextInt(2)))
+
+  @Benchmark
+  def benchmarkSmallDataZioApi(): Unit =
+    for (r <- smallDataRequests) routes.isDefinedAt(r)
+
+  @Benchmark
+  def benchmarkSmallDataZioApi2(): Unit =
+    for (r <- smallDataRequests2) routes2.isDefinedAt(r)
+
+  @Benchmark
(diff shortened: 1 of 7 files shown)


kyri-petrou:
I'm suspecting this speedup is largely due to the number of paths used in the benchmarks (26 and 25 for `benchmarkSmallDataZioApi` and `benchmarkSmallDataZioApi2` respectively). In most cases though the number of paths in the segment will be much smaller, with some segments having just 1-2 paths (e.g., `internal`, `public`).

How are the benchmarking results for a very small number of paths in a segment?

zio-http-benchmarks/src/main/scala/zio/http/benchmark/RoutesBenchmark.scala
```diff
@@ -0,0 +1,67 @@
...
+  val routes2 = Routes.fromIterable(paths2.map(p => Endpoint(Method.GET / p).out[Unit].implementHandler(Handler.unit)))
+
+  val requests2 = requests ++ paths2.map(p => Request.get(p))
+
+  def request2: Request = requests2(Random.nextInt(requests2.size))
+
+  @Benchmark
+  def benchmarkSmallDataZioApi(): Unit =
+    unsafeRun {
+      ZIO.collectAllDiscard(1.to(REPEAT_N).map(_ => routes.run(request)))
```
kyri-petrou: Might be better to use `routes.isDefinedAt` to avoid the noise added by running the ZIO effects

zio-http-benchmarks/src/main/scala/zio/http/benchmark/RoutesBenchmark.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  val REPEAT_N = 1000
+
+  val paths = ('a' to 'z').inits.map(_.mkString).toList.reverse.tail
+
+  val routes = Routes.fromIterable(paths.map(p => Endpoint(Method.GET / p).out[Unit].implementHandler(Handler.unit)))
+
+  val requests = paths.map(p => Request.get(p))
+
+  def request: Request = requests(Random.nextInt(requests.size))
```
kyri-petrou: Perhaps store all requests in an array of size `REPEAT_N` to avoid calling `Random.nextInt` (which is fairly expensive and adds noise to the benchmark)

987Nabil:
> I'm suspecting this speedup is largely due to the number of paths used in the benchmarks (26 and 25 for `benchmarkSmallDataZioApi` and `benchmarkSmallDataZioApi2` respectively). In most cases though the number of paths in the segment will be much smaller, with some segments having just 1-2 paths (e.g., `internal`, `public`).
> 
> How are the benchmarking results for a very small number of paths in a segment?

Do you think we need to? I mean there is probably no difference for this sizes in whatever kind of Map we use.

kyri-petrou:
Hash-based lookups are usually slower than iteration-based ones for for small Maps so I think it'd be good to quantify what the penalty (if any) will be

987Nabil:
> Hash-based lookups are usually slower than iteration-based ones for for small Maps so I think it'd be good to quantify what the penalty (if any) will be

Or we just use Map? It has specialized cases for n <= 4 and defaults to hashmap afterwards, right?

987Nabil:
I changed to `Map` and it seems to be the right choice

```
HashMap
[info] Benchmark                                   Mode  Cnt      Score   Error  Units
[info] RoutesBenchmark.benchmarkSmallDataZioApi   thrpt       40191.769          ops/s
[info] RoutesBenchmark.benchmarkSmallDataZioApi2  thrpt       55464.016          ops/s
[info] RoutesBenchmark.benchmarkSmallDataZioApi3  thrpt       62321.804          ops/s
[info] RoutesBenchmark.notFound1                  thrpt       20618.638          ops/s
[info] RoutesBenchmark.notFound2                  thrpt       22141.172          ops/s

Map
[info] Benchmark                                   Mode  Cnt      Score   Error  Units
[info] RoutesBenchmark.benchmarkSmallDataZioApi   thrpt       39100.894          ops/s
[info] RoutesBenchmark.benchmarkSmallDataZioApi2  thrpt       53377.960          ops/s
[info] RoutesBenchmark.benchmarkSmallDataZioApi3  thrpt       75431.408          ops/s
[info] RoutesBenchmark.notFound1                  thrpt       20761.618          ops/s
[info] RoutesBenchmark.notFound2                  thrpt       21795.605          ops/s
```

jdegoes (approved):

PR #3278, opened 2025-01-19
Revert "[sbt-gen] sbt plugin for openapi to endpoints codegen"

hochgi:
Reverts zio/zio-http#3225

The plugin has moved to https://github.com/zio/zio-http-sbt
and will be published separately from other zio-http modules.


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.10.6
+sbt.version=1.10.0
```
hochgi: this does not need to be reverted

zio-http/jvm/src/test/scala/zio/http/ClientHttpsSpec.scala
hochgi, line 1: these changes to flaky test does not need to be reverted

zio-http/jvm/src/test/scala/zio/http/endpoint/AuthSpec.scala
```diff
@@ -152,7 +151,7 @@ object AuthSpec extends ZIOSpecDefault {
               .catchAllCause(c => ZIO.logInfoCause(c)) <* ZIO.sleep(1.seconds)
             response <- response
           } yield assertTrue(response == "admin")
-        } @@ flaky,
+        },
```
hochgi: if still flaky, should not be reverted

987Nabil (approved):

Issue #3235, opened 2024-12-05
Missing Authorization header is reported as 400 instead of 401


notxcain:
A missing `Authorization` header is reported with 400 instead of 401 status code when served by a route created using the declarative endpoint syntax.

**To Reproduce**
1. Create a simple endpoint using `Endpoint(PathCodec.Get).header(HeaderCodec.authorization)`.
1. Send a request without `Authorization` header.
1. Get a response with 400 status code.

**Expected Behaviour**

According to [the RFC](https://datatracker.ietf.org/doc/html/rfc7235#section-3.1), in such case a response should have 401 Unauthorized code:
> The 401 (Unauthorized) status code indicates that the request has not
   been applied because it lacks valid authentication credentials for
   the target resource.

weili96:
![image](https://github.com/user-attachments/assets/89a78a55-616a-44ed-92f1-0eccb0f543f9)
I also meet this bug !

weili96:
![image](https://github.com/user-attachments/assets/10c68f41-6b58-4242-be86-1c16658ed7a7)
content-type is also error!

987Nabil:
@weili96 without the endpoint definition, Idk what the issue is. Also, if you find a new bug, please open a new issue

weili96:
`
  private val editorViewArticleEndpoint =
    Endpoint((RoutePattern.POST / "article" / "editorView") ?? Doc.p("文章编辑页面-查看文章详情"))
      .auth(AuthType.Bearer)
      .header(HeaderCodec.authorization)
      .in[ArticleId]
      .out[ArticleInfo]
      .outErrors[ServerError](
        HttpCodec.error[BadRequestError](Status.BadRequest),
        HttpCodec.error[DbError](Status.InternalServerError)
      )
      .tag("Article")
  private val editorViewArticleRoute = editorViewArticleEndpoint.implement(
    (authHeader: Authorization, input: ArticleId) =>
      ZIO.serviceWithZIO[ArticleApiHandler](_.editorViewArticle(input, Authorization.render(authHeader)))
  )


`

weili96:
```
private val editorViewArticleEndpoint =
    Endpoint((RoutePattern.POST / "article" / "editorView") ?? Doc.p("文章编辑页面-查看文章详情"))
      .auth(AuthType.Bearer)
      .header(HeaderCodec.authorization)
      .in[ArticleId]
      .out[ArticleInfo]
      .outErrors[ServerError](
[...]

PR #3281, opened 2025-01-22
Make Endpoint respond with a 401 status code if the Authorization header is missing

notxcain:
/claim #3235


zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -340,9 +340,10 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
               case Some(HttpCodecError.CustomError("SchemaTransformationFailure", message))
                   if maybeUnauthedResponse.isDefined && message.endsWith(" auth required") =>
                 maybeUnauthedResponse.get
-              case Some(_) =>
+              case Some(HttpCodecError.MissingAuthorizationHeader) =>
+                Handler.succeed(Response.unauthorized)
```
notxcain, line 391: Let me know if this should be reported the same way as any other missing header.

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -340,9 +340,10 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
...
+              case Some(error)                                     =>
```
notxcain: Not sure why it was not written this way, as it looks like in this case `asHttpCodecError` provides the same value as `cause.defects.head.asInstanceOf[HttpCodecError]`.

zio-http/shared/src/main/scala/zio/http/codec/HttpCodecError.scala
```diff
@@ -33,6 +33,9 @@ object HttpCodecError {
   final case class MissingHeader(headerName: String)                                           extends HttpCodecError {
     def message = s"Missing header $headerName"
   }
+  final case object MissingAuthorizationHeader                                                 extends HttpCodecError {
```
notxcain: I believe introducing a special case for that purpose helps with reasoning. Let me know if you'd like this to be implement using `MissingHeader`.

notxcain:
I suspect `JSClientSpec / HTTP / Get with User Agent` and `AuthSpec / Auth with context and endpoint client with path parameter` are flaky, or at least I can't reproduce it locally with the same JVM and Scala versions.

notxcain:
Hi @987Nabil — could you please take a look at the PR? The suspected tests are indeed flaky, the tests were green before I merged master. And now it failed again.

notxcain:
Hey @jdegoes — could you please take a look at this?

Omar8345:
@notxcain @jdegoes is this still ongoing?

notxcain:
@Omar8345 I did no get any response from the maintainers so I don't really know.

zio-http/shared/src/main/scala/zio/http/codec/HttpCodecError.scala
```diff
@@ -37,6 +36,9 @@ object HttpCodecError {
   final case class MissingHeader(headerName: String)                                           extends HeaderError     {
     def message = s"Missing header $headerName"
   }
+  case object MissingAuthorizationHeader                                                       extends HeaderError     {
```
987Nabil: I think this is not necessary. The `MissingHeader` should be enough, since the header name is a unique id.
I think if we introduce more cases for every such case, the sub types are getting too many.

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
987Nabil: please revert this and change `case Some(_)` to `case _: Some[_]`. Because `Some(error)` is allocating an Option for the pattern match. We can avoid that here easily so we should.

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
987Nabil: The specs say, that we have to return acceptable auth methods.
And we have that data from the Endpoint API

> The server generating a 401 response MUST send
   a WWW-Authenticate header field ([Section 4.1](https://datatracker.ietf.org/doc/html/rfc7235#section-4.1)) containing at least one
   challenge applicable to the target resource.

zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -178,7 +177,12 @@ object Header {
 
     def fromHeadersUnsafe(headers: Headers): HeaderValue =
       fromHeaders(headers).fold(
-        e => throw HttpCodecError.DecodingErrorHeader(name, ReadError(Cause.empty, e)),
+        e => {
```
987Nabil: This is not the correct place to check this I think. It is a generic method, that is not only for the auth header. 
I also don't see how this checks if the auth header is required at all.

987Nabil (changes requested):
Sorry for the late review. I hope you can still work on this :)

987Nabil:
@notxcain I think errors of decoding the header should also lead to 401

notxcain:
> I think errors of decoding the header should also lead to 401

@987Nabil I only have [this](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) link to support my choice, but I believe a malformed (or violating schema) value of a header falls into `Bad Request` and has different semantics.

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
notxcain: @987Nabil the `AuthType` has quite limited information, please see the implementation [draft](https://github.com/zio/zio-http/pull/3281/commits/113c4975eeaafebfa0c39e3f02f9ff90f6a45c00). Any piece of advice here?

notxcain:
Hi @987Nabil — please see check my [comment](https://github.com/zio/zio-http/pull/3281/files#r2137861024).

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
987Nabil: Most things are optional and can be omitted. Digest nonce is generated anyway.
Basic Digest and OAuth need a realm. We can add this to `AuthType` in the same fashion like `ScopedAuth`.
In case there is no realm, we could return no header or a default value. I tend to a default value. Like `"default"` or `"restricted"`

987Nabil:
> > I think errors of decoding the header should also lead to 401
> 
> @987Nabil I only have [this](https://datatracker.ietf.org/doc/html/rfc6750#section-3.1) link to support my choice, but I believe a malformed (or violating schema) value of a header falls into `Bad Request` and has different semantics.

The link says explicitly that malformed is 401 🤔 
> The access token provided is expired, revoked, malformed, or
         invalid for other reasons.  The resource SHOULD respond with
         the HTTP 401 (Unauthorized) status code.  The client MAY
         request a new access token and retry the protected resource
         request.

Issue #3260, opened 2024-12-19
Generated openAPI shows example in application/json instead of specified text/plain


jiramares:
**Describe the bug**
Specifying `text/plain` media type for response causes openAPI shows response example in wrong media  type (stilll uses `application/json`).

**To Reproduce**
Following endpoint specifies response media type to `text/plain`:
```
val ep2 = Endpoint(RoutePattern.GET / "greeting" / string("name"))
    .outCodec(HttpCodec.content[Greeting](MediaType.text.plain))
```
and generated openAPI shows response as (the media type is right `text/plain`):
```
{
  "msg": "string"
}
```

**Expected behaviour**
The response example should be:
```
string
````

**Dependencies:**
```
    "dev.zio" %% "zio" % "2.1.9",
    "dev.zio" %% "zio-http" % "3.0.1"
    "dev.zio" %% "zio-schema" % "1.4.1"
```

**Comment by jiramares** [comment]
The same is also true for error channel:
```
val ep5 = Endpoint(RoutePattern.GET / "greeting3" / nameParam / "exclamation" / boolParam)
        .copy(error = HttpCodec.content[Greeting](MediaType.text.plain) ++ HttpCodec.status(Status.NotFound))
        .outCodec(HttpCodec.content[Greeting](MediaType.text.plain))
```
the example error response is 
```
{
  "msg": "string"
}
```
not plain text
```
string
````

987Nabil:
The examples are currently explicitly created for json. We need to make this dynamic by the media type in the codec

ffprivacy:
/attempt #3260

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm684msme0002i9037ncf6o2p/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

PR #3283, opened 2025-01-24
allow parametrization by the media type of contentExamples

ffprivacy:
/claim #3260


987Nabil:
This is a much bigger issue. I'll takle this fully in a new PR

PR #3285, opened 2025-01-28
added curl logger client aspect.

landlockedsurfer:
Hi all!
There was a message in the Discord the other day asking for a curl command logger.
Here's a first version that should properly handle the most basic options. It does not handle any SSL options yet though.
I used the other `ZClientAspect` instances as a blueprint for the implementation.
Hope this is useful!!

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
+++ b/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
@@ -577,14 +577,9 @@ object ZClientAspect {
             body: Body,
             sslConfig: Option[ClientSSLConfig],
             proxy: Option[Proxy],
-          )(implicit trace: Trace): ZIO[Env & ReqEnv, Err, Response] = {
-
-            val r: ZIO[Env & ReqEnv, Err, Response] =
+          )(implicit trace: Trace): ZIO[Env & ReqEnv, Err, Response] =
               logEffect(formatCurlCommand(version, method, url, headers, body, proxy)) *>
-                oldDriver
-                  .request(version, method, url, headers, body, sslConfig, proxy)
-            r
-          }
+                oldDriver.request(version, method, url, headers, body, sslConfig, proxy)
 
           override def socket[Env1 <: Env](version: Version, url: URL, headers: Headers, app: WebSocketApp[Env1])(
             implicit
--- a/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
+++ b/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
@@ -498,7 +498,7 @@ object ZClientAspect {
    */
   final def curlLogger(
     verbose: Boolean = true,
-    logEffect: String => ZIO[Any, Nothing, Unit] = (m: String) => ZIO.log(m),
+    logEffect: String => UIO[Unit] = (m: String) => ZIO.log(m),
   )(implicit trace: Trace): ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] =
     new ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] {
 
--- a/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
+++ b/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
@@ -492,4 +492,109 @@ object ZClientAspect {
           },
         )
     }
+
+  /**
+   * Client aspect that logs details of web request as curl command
+   */
+  final def curlLogger(
+    verbose: Boolean = true,
+    logEffect: String => ZIO[Any, Nothing, Unit] = (m: String) => ZIO.log(m),
+  )(implicit trace: Trace): ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] =
+    new ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] {
+
+      def formatCurlCommand(
+        version: Version,
+        method: Method,
+        url: URL,
+        headers: Headers,
+        body: Body,
+        proxy: Option[Proxy],
+      ): String = {
+        val versionOpt = version match {
+          case Version.Default  => Chunk.empty
+          case Version.Http_1_0 => Chunk("--http1.0")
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
```diff
@@ -492,4 +492,109 @@ object ZClientAspect {
           },
         )
     }
+
+  /**
+   * Client aspect that logs details of web request as curl command
+   */
+  final def curlLogger(
+    verbose: Boolean = true,
+    logEffect: String => ZIO[Any, Nothing, Unit] = (m: String) => ZIO.log(m),
```
jivanic-demystdata: 🟢 Suggestion
```suggestion
    logEffect: String => UIO[Unit] = (m: String) => ZIO.log(m),
```

zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
```diff
@@ -492,4 +492,109 @@ object ZClientAspect {
...
+            sslConfig: Option[ClientSSLConfig],
+            proxy: Option[Proxy],
+          )(implicit trace: Trace): ZIO[Env & ReqEnv, Err, Response] = {
+
+            val r: ZIO[Env & ReqEnv, Err, Response] =
+              logEffect(formatCurlCommand(version, method, url, headers, body, proxy)) *>
+                oldDriver
+                  .request(version, method, url, headers, body, sslConfig, proxy)
+            r
+          }
```
jivanic-demystdata: 🟢 Suggestion
```suggestion
          )(implicit trace: Trace): ZIO[Env & ReqEnv, Err, Response] =
              logEffect(formatCurlCommand(version, method, url, headers, body, proxy)) *>
                oldDriver.request(version, method, url, headers, body, sslConfig, proxy)
```

jivanic-demystdata (commented):
LGTM 

Let's wait for Nabil's review :)

zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
```diff
@@ -492,4 +492,104 @@ object ZClientAspect {
...
+  /**
+   * Client aspect that logs details of web request as curl command
+   */
+  final def curlLogger(
+    verbose: Boolean = true,
+    logEffect: String => UIO[Unit] = (m: String) => ZIO.log(m),
+  )(implicit trace: Trace): ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] =
+    new ZClientAspect[Nothing, Any, Nothing, Body, Nothing, Any, Nothing, Response] {
+
+      def formatCurlCommand(
```
guizmaii, line 505: @landlockedsurfer Do you think we could put this capacity to transform a Request to a `curl` String as a method of `Request` and use this method here?

That'd allow people to "toString" their requests in a nice way if they need to (to log it at any moment, for example)

zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
```diff
@@ -492,4 +492,104 @@ object ZClientAspect {
...
+        body: Body,
+        proxy: Option[Proxy],
+      ): String = {
+        val versionOpt = version match {
+          case Version.Default  => Chunk.empty
+          case Version.Http_1_0 => Chunk("--http1.0")
+          case Version.Http_1_1 => Chunk("--http1.1")
+        }
+        val verboseOpt = if (verbose) Chunk("--verbose") else Chunk.empty
+        val requestOpt = method match {
```
987Nabil, line 519: Can be simplified to 
```suggestion
        val requestOpt = Chunk(s"--request ${method.name}")
```

zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
```diff
@@ -492,4 +492,104 @@ object ZClientAspect {
...
+          case Method.CONNECT      => Chunk("--request CONNECT")
+          case Method.TRACE        => Chunk("--request TRACE")
+          case Method.CUSTOM(name) => Chunk(s"--request $name")
+          case Method.ANY          => Chunk("--request GET")
+        }
+        val headerOpt  = Chunk.fromIterable(headers.map(h => s"--header '${h.headerName}:${h.renderedValue}'"))
+        val bodyOpt    = body match {
+          case Body.empty => Chunk.empty
+          case body       => {
+            Chunk(s"--data '${body.asString.map(_.replace("'", "'\\''"))}'")
```
987Nabil, line 536: body.asString returns an effect. Printing an effect is not what you want.
Also, not all bodies can just be transformed to a string.
You need to check `body.isCompleted` to know if a it has data that you are allowed to read.
Else you might consume a stream of the body during logging, and it is not available anymore when we try to send it over the wire.
See the requestLogging client aspect for reference

987Nabil:
@landlockedsurfer please run `fmt` before pushing :)

guizmaii:
@landlockedsurfer Any news?

987Nabil:
Feel free to reopen if it is fixed

landlockedsurfer:
Hey guys!
Sorry for the long wait... but I've worked on fixing the issues you mentioned above.
Somehow reopen this pull request is not enabled for me... maybe because I've force pushed the curl logger branch, but I've created another pull request here:
https://github.com/zio/zio-http/pull/3394

zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
landlockedsurfer: @guizmaii Yes! That's definitely a good idea and very useful! I've added the method `zio.http.ZClientAspect.CurlLogger.formatCurlCommand` here:
https://github.com/landlockedsurfer/zio-http/blob/curl-logger/zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala#L600

zio-http/shared/src/main/scala/zio/http/ZClientAspect.scala
landlockedsurfer: Thanks @987Nabil for spotting those issues! 🙏
The body  is now only converted to a string if `body.isCompleted` is `true`. Also `body.asChunk` is now correctly evaluated to get the result of the effect.

PR #3293, opened 2025-02-06
Add Object Mapper Example

khajavi:


987Nabil:
Is what you are talking about here really a DTO? Or more a db model.
I mean we don't do spring style DTO aka modify the item to auto save in db, right? 
I am just thinking about naming

khajavi:
Not sure I got your main point since I'm not familiar with Spring! The
pattern I'm explaining here is indeed aligned with objects that are not
part of the domain, they are used for transferring data between layers
without exposing the underlying model.

987Nabil:
I would call it `DBModel` instead of `DTO`

khajavi:
Can you share your thoughts on this? Why do you suggest that?

guersam:
As far as I understood, `PersonDTO` represents the HTTP API payload from the outside without ownership, such as a webhook, and `Person` is the domain model that can be processed internally, e.g. being stored in DB.

@khajavi Am I right? If so, `DTO` might be misleading to the users who are familiar with Spring.

khajavi:
@guersam Right!
I've found this diagram from [this article](https://www.baeldung.com/java-dto-pattern):

image

The mapper I'm referring to is similar to what this diagram reflects.

Also, about DTOs in Spring, I found [this article](https://medium.com/@vishamberlal/understanding-data-transfer-objects-dto-in-spring-boot-ac06b575a1d5), and this explained the DTO the same as what I meant to show in this example!

> Am I right? If so, DTO might be misleading to the users who are familiar with Spring.

So why is that misleading?

guersam:
@khajavi Thanks for clarifying.

Although I'm not that familiar with Spring, I overheard that many Spring developers confuse DTO with JPA entity (persistence model), domain object, or value objects. I guess that's why @987Nabil thought `PersonDTO` was for databases at the first glance.

I think it could be sufficient to add some explanation in the comment with the reference, preferably more authentic one that's also mentioned in the article you've linked.
https://martinfowler.com/eaaCatalog/dataTransferObject.html

987Nabil (approved):

PR #3298, opened 2025-02-11
CI: Fix scala_steward && Add dependabot to maintain GHA workflows

guizmaii:
(Copied from ZIO repo)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/scala-steward.yml
+++ b/.github/workflows/scala-steward.yml
@@ -1,10 +1,14 @@
 name: Scala Steward
 
-# This workflow will launch everyday at 00:00
+# This workflow will launch every day at 00:00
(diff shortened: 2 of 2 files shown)


987Nabil:
@khajavi I think I will merge this, but can we get this into zio-sbt? I think we can also generate dependabot settings.
Maybe with adjustable schedule time. So we don't run into the OSS runner limit for gh actions.

khajavi:
@987Nabil 
ZIO‌ SBT has been configured with [renovate](https://github.com/zio/zio-sbt/blob/cbf9cdec295a6caa294c1b748b67d84b9de23eab/renovate.json), so GitHub actions will be updated as soon as a new release is available. The `zio-sbt-ci` module generates updated Github actions for each release.
No need for individual repositories; updating GitHub Actions via tools like Renovate or Dependabot suffices.

guizmaii:
@khajavi I'd prefer we keep scala-steward over Renovate. Also, as already discussed, I think we should remove the `zio-sbt-ci` module from zio-sbt. IMO, it adds too much complication to just write these yaml files. it's not flexible enough for the different zio projects and maintainers, with their different needs. IMO, it's over-engineering. Copy/paste is good enough for this.

987Nabil (approved):

PR #3299, opened 2025-02-11
CI: Fix release-drafter

guizmaii:
(Copied from ZIO repo)

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   update_release_draft:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -0,0 +1,14 @@
+name: Release Drafter
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  update_release_draft:
+    runs-on: ubuntu-20.04
```
987Nabil: There is a deadline one ubuntu 20 for this year. Can we just use latest?

987Nabil (changes requested):

.github/workflows/release-drafter.yml
guizmaii: Done :)

987Nabil (approved):

PR #3300, opened 2025-02-12
Fixes: #3243

### LINKED ISSUE #3243: Nested routes serve routes on all nested paths
**Describe the bug**
The feature of nesting routes is exposing paths on both the original path, and on the nested path. 
 https://zio.dev/zio-http/reference/routing/routes#nested-routes 

**To Reproduce**
I found it by curl-ing my app, but here is a reproducer using both the direct request syntax and calling the server directly
```scala
//> using dep "dev.zio::zio-http:3.0.1"
import zio.*
import zio.http.*
import zio.http.codec.PathCodec

object HelloWorld extends ZIOAppDefault:
  val routed = PathCodec.literal("foo") / Routes(
    Method.GET / "hello" -> handler(Response.text("World"))
  )
  val hello = Request.get("http://localhost:8080/hello")
  val fooHello = Request.get("http://localhost:8080/foo/hello")
  override def run = {
    for {
      res1 <- routed.runZIO(hello).debug //Gives OK
      res2 <- routed.runZIO(fooHello).debug //Gives OK
      p <- Server.serve(routed).forkDaemon
      res3 <- Client.batched(hello).debug //Gives OK
      res4 <- Client.batched(fooHello).debug //Gives OK
      _ <- p.interrupt
    } yield ()
  }.provide(Server.default ++ Client.default)

```

**Expected behaviour**
Only the requests to `fooHello` respond with 200 OK, with the requests to `hello` giving 404.

deekshatomer:
/claim #3243 

This PR fixes the issue with nested routes not resolving correctly in ZIO-HTTP. The problem was caused by incorrect route matching logic, which led to failures when handling deeply nested paths.

Fix Details

Updated the route-matching logic to correctly handle nested paths.
Refactored HttpOptions.scala (or any other files you modified) to ensure proper request handling.
Added test cases to validate the fix.
Issue Reference

Fixes #3243

Testing

Verified locally by running test cases.
Ensured no breaking changes by running all existing tests.


guizmaii:
@deekshatomer Still not good.

987Nabil:
Please open PR's if they are reviewable

PR #3303, opened 2025-02-16
More efficient root path look up (#3302)

### LINKED ISSUE #3302: Replace `ListMap` usage in `RoutePattern.Tree`
`ListMap` is the worst Scala collection when it comes to performance. And the list like features of the collection are not utilized.
We should add a benchmark and replace it with a more suitable data structure.

987Nabil:
**Reasoning**
Most http apps don't use custom http methods. In fact most probably only use GET, POST, PUT and DELETE.
Having an efficient lookup for the default methods, seems the most reasonable solution.

Current impl.
```
[info] Benchmark                              Mode  Cnt      Score       Error  Units
[info] MethodLookupBenchmark.customMethods   thrpt    3  17133.273 ± 13370.029  ops/s
[info] MethodLookupBenchmark.defaultMethods  thrpt    3  18215.269 ± 17379.965  ops/s
```

Map instead of ListMap
```
[info] Benchmark                              Mode  Cnt      Score      Error  Units
[info] MethodLookupBenchmark.customMethods   thrpt    3  22974.932 ± 1760.853  ops/s
[info] MethodLookupBenchmark.defaultMethods  thrpt    3  25215.348 ± 4809.175  ops/s
```

New filed per default HTTP method
```
[info] Benchmark                              Mode  Cnt      Score      Error  Units
[info] MethodLookupBenchmark.customMethods   thrpt    3  31862.950 ± 3407.335  ops/s
[info] MethodLookupBenchmark.defaultMethods  thrpt    3  32393.768 ±  765.915  ops/s

```

fixes #3302
/claim #3302

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
@@ -770,19 +770,22 @@ object PathCodec {
   ) {
     self =>
     def ++[A1 >: A](that: SegmentSubtree[A1]): SegmentSubtree[A1] = {
-      val newLiterals          = mergeMaps(self.literals, that.literals)(_ ++ _)
-      val newOthers            = mergeMaps(self.others, that.others)(_ ++ _)
-      val newLiteralCollisions = mergeLiteralCollisions(
-        self.literalsWithCollisions ++ that.literalsWithCollisions,
-        newLiterals.keySet,
-        newOthers.keys,
-      )
-      SegmentSubtree(
-        Map(newLiterals.toList: _*),
-        ListMap(newOthers.toList: _*),
-        newLiteralCollisions,
-        self.value ++ that.value,
-      )
+      if (that eq null) self
+      else {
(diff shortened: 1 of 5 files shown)


zio-http/shared/src/main/scala/zio/http/codec/PathCodec.scala
```diff
@@ -770,19 +770,22 @@ object PathCodec {
...
+      else {
+        val newLiterals          = mergeMaps(self.literals, that.literals)(_ ++ _)
+        val newOthers            = mergeMaps(self.others, that.others)(_ ++ _)
+        val newLiteralCollisions = mergeLiteralCollisions(
+          self.literalsWithCollisions ++ that.literalsWithCollisions,
+          newLiterals.keySet,
+          newOthers.keys,
+        )
+        SegmentSubtree(
+          Map(newLiterals.toList: _*),
```
kyri-petrou, line 783: You can probably use `Map.from` but you might need to do something about Scala 2.12.

kyri-petrou (approved):
LGTM

PR #3312, opened 2025-02-17
Align Client SSE support to HTML specification

disordered:
This should (mostly) align SSE support on the `Client` to HTML [specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream).

Fixes:
 - Use UTF-8 for encoding (server) and decoding
 - Ignore comment and non-field data (usually used for keep-alive)
 - Ignore non-integer `retry` values
 - Combine successive `data` field values
 - Parse events with fields specified in any order
 - Do not emit events without data
 - Do not emit incomplete events
 - Allow field overrides within event
 - Support field names without colon
 - Handling of first space after field name

Breaking:
 - `retry` is now `Duration` parsed as milliseconds

### Handling of `retry`
HTML specification leaves a lot of room to interpretation for edge cases on how to dispatch events. For browsers it says to not emit events if `data` is not specified, however, chromium based browsers consume such events and adjust their retry frequency accordingly. So, this implementation emits such events too with `data = ""`. It is not obvious, however, that non-string codec has to handle this case.

### Handling of various line endings
Chromium browsers and Postman handle all versions (Windows, Linux, OSX) of line endings. However, the specification explicitly states only `\n` should be treated as separator. I considered aligning with the browser implementation, but `ZPipeline.splitLines` only supports Windows and Linux line endings. It could probably be achieved through some sort of chunk and `String.linesIterator` mapping.


zio-http/jvm/src/test/scala/zio/http/ClientServerSentEventSpec.scala
```diff
@@ -0,0 +1,294 @@
...
+
+  def assertEvents(
+    streamData: String,
+    expectedEvents: Chunk[ServerSentEvent[String]],
+  ): ZIO[Client with Server, Throwable, TestResult] =
+    eventStream(streamData).runCollect.map(actual => assertTrue(actual == expectedEvents))
+
+  override def spec: Spec[TestEnvironment with Scope, Any] =
+    suite("ServerSentEventClientSpec")(
+      test("Line breaks only payload does not emit events") {
```
disordered, line 26: Postman will emit empty events, chromium does not. Specification says not to.

zio-http/jvm/src/test/scala/zio/http/ClientServerSentEventSpec.scala
```diff
@@ -0,0 +1,294 @@
...
+        val events =
+          """|
+             |
+             |
+             |""".stripMargin
+        assertEvents(events, Chunk.empty)
+      },
+      test("Successive line breaks do not emit extra events") {
```
disordered, line 34: Similar to the other test, Postman inserts empty events.

zio-http/jvm/src/test/scala/zio/http/ClientServerSentEventSpec.scala
```diff
@@ -0,0 +1,294 @@
...
+      test("Event type can be unset") {
+        // IDEs tend to trim lines on multiline strings, so build event from collection
+        val events = Chunk("event: ", "event:", "event").map { field =>
+          s"""|event: sse
+              |data: data
+              |$field""".stripMargin
+        }.mkString("", "\n\n", "\n\n")
+        assertEvents(events, Chunk.fill(3)(ServerSentEvent(data = "data")))
+      },
+      test("Event type without data is not emitted") {
```
disordered, line 124: Postman emits the event, Chromium does not. Specification says not to emit.

zio-http/jvm/src/test/scala/zio/http/ClientServerSentEventSpec.scala
```diff
@@ -0,0 +1,294 @@
...
+        assertEvents(events, Chunk.fill(3)(ServerSentEvent(data = "data")))
+      },
+      test("Event retry without data is emitted") {
+        val events =
+          """|retry: 1
+             |
+             |""".stripMargin
+        assertEvents(events, Chunk.single(ServerSentEvent(data = "", retry = Some(1.millisecond))))
+      },
+      test("Event retry that is not an integer is not emitted") {
```
disordered, line 189: Postman will emit events with provided `retry` value. Browser ignores them. Specification says to ignore.

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -42,12 +46,12 @@ final case class ServerSentEvent[T](
   data: T,
   eventType: Option[String] = None,
   id: Option[String] = None,
-  retry: Option[Int] = None,
+  retry: Option[Duration] = None,
```
987Nabil, line 47: Please add a constructor, that takes `Option[Int]` for compatibility.
Also adjust the docs of the param here.

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -42,12 +46,12 @@ final case class ServerSentEvent[T](
...
 ) {
 
   def encode(implicit binaryCodec: BinaryCodec[T]): String = {
     val sb = new StringBuilder
-    binaryCodec.encode(data).asString.linesIterator.foreach { line =>
+    binaryCodec.encode(data).asString(StandardCharsets.UTF_8).linesIterator.foreach { line =>
```
987Nabil: Please use `zio.http.Charsets`

987Nabil (changes requested):
Small changes

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
disordered: Changed

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
disordered: Added it, though, I realised I missed an edge case, when a negative int value is passed. So, this constructor ends up silently filtering it out as well.

EDIT: Removed again.
The compiler cannot decide which constructor to use when optional values are not provided, e.g. `ServerSentEvent("some data")`. Named constructor would work, but doesn't address backwards compatibility.

987Nabil: Please add back the constructor with the Int, but without default values.

disordered: Done

987Nabil (approved):

987Nabil (approved):

Issue #3186, opened 2024-10-02
Strange handling of empty json array (possibly zio-schema bug?)


russwyte:
**Describe the bug**
Routes that produce JSON don't correctly handle empty collections. The resulting JSON should always have a field with an empty JSON array - but instead the field is omitted altogether.
**To Reproduce**
See this [gist](https://gist.github.com/russwyte/1cf305ddf861ffd62a8fbb86b8e2695a)
result of run is:
```
unexpected: {"bars":[1,2,3]}
expected: {"bars":[1,2,3]}
unexpected: {}
expected: {"bars":[]}
```
**Additional context**
This causes breakage in codecs when lists are not optional and are empty

The gist also shows a work-around that might shed light on a possible fix?

russwyte:
This [is the issue](https://github.com/zio/zio-http/blob/f29c69929745fc3e53cfefc1e1b90911f9662153/zio-http/shared/src/main/scala/zio/http/codec/BinaryCodecWithSchema.scala#L42-L44). It seems like the default should be false - as empty collections are valid data.

russwyte:
Adding this works - but this seems like it should be the default as it is in zio-json and elsewhere
```scala
  val routes = Routes(
    unexpected.implement(foo => ZIO.succeed(foo)),
    expected.implement(foo => ZIO.succeed(foo))
  ) @@ CodecConfig.withConfig(CodecConfig(ignoreEmptyCollections = false))
  ```

jdegoes:
@987Nabil Should we change the default here? I too would expect empty collections to be represented in responses (but missing collections to be accepted in requests -- notice the duality).

987Nabil:
Yeah. I think we should

notxcain:
Just got bitten by it. It's indeed quite unexpected.

Ronyonka:
/attempt #3186

<div id="algora-attempt" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cm685bajc000kl5033p7qwncc/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

Tvaroh:
Hey, has this been released? Still drops the empty collection field under `zio-http` `3.2.0`.

PR #3341, opened 2025-02-19
Preserve empty collections in JSON by default (#3186)

guersam:
/claim #3186

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/RoundtripSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/RoundtripSpec.scala
@@ -74,9 +74,9 @@ object RoundtripSpec extends ZIOHttpSpec {
     implicit val schema: Schema[PostWithAge] = DeriveSchema.gen[PostWithAge]
   }
 
-  case class Outs(ints: List[Int])
+  case class Ints(ints: List[Int])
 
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/test/scala/zio/http/endpoint/RoundtripSpec.scala
```diff
@@ -551,16 +551,32 @@ object RoundtripSpec extends ZIOHttpSpec {
...
         )
       },
+      test("default CodecConfig accepts missing collections in request") {
+        val api = Endpoint(POST / "test").in[Ints].out[Int]
+        testEndpointCustomRequestZIO(
+          api.implement(ints => ZIO.succeed(ints.ints.length)).toRoutes,
+          Request.post("/test", Body.fromString("""{}""")),
+          response => response.body.asString.map(s => assertTrue(s == "0")),
+        )
+      },
```
guersam, line 569: To ensure the duality mentioned in the comment below

https://github.com/zio/zio-http/issues/3186#issuecomment-2465995247
> I too would expect empty collections to be represented in responses (but missing collections to be accepted in requests -- notice the duality).

987Nabil:
@guersam thanks for the try, but to fulfill the requirements, we need changes in zio schema. We want to have the deserialization and serialization part be configurable independently. This is implemented in zio-json and I will add this soonish to schema. And then integrate it in http.
That's why I am closing this PR

Issue #2431, opened 2023-09-08
Configuration of ZIO HTTP


TomTriple:
**What is not easy to do right now?**
There is no single source (website, markdown...) where one can find out about all config options in ZIO HTTP. One needs to inspect the code for the various configs we have e.g. `zio.http.ConnectionPoolConfig.config` or `zio.http.Server.Config.config` or `zio.http.netty.Config.config`. 

**Describe the solution you'd like**
We should auto-generate all config options (probably in the build process) into a markdown page entitled `Configuration` and with the help of `zio.config`. That works already: 
```
val config = zio.http.ConnectionPoolConfig.config // compose with other configs
val docs = generateDocs(config).toTable.toGithubFlavouredMarkdown // generateDocs comes from zio.config
println(docs) // prints auto-generated content directly from inspecting the config instance
```
Also we could generate an example config file including default values from all given configs. 

**Why is it important?**
Easier configuration, user friendlyness and always up to date docs

**Additional context**
zio.config: https://zio.dev/zio-config/dive-into-zio-config#documentations-using-config

guersam:
/attempt #2431

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cln0iijxm000ml10f95gb5k8k/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

guersam:
As the built-in `zio.Config` does not provide bidirectional config rendering, generating documentations for configs is more difficult than before.

Currently I'm writing an ad-hoc table generator from scratch for ZIO HTTP.

vivasvan1:
/attempt #2431

<details id="algora-options">
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cln0iijxm000ml10f95gb5k8k/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

pragmaxim:
/attempt https://github.com/zio/zio-http/issues/2431

<div id="algora" />

<details>
  <summary>Options</summary>
  <ul>
    <li>
      <a href="https://console.algora.io/api/bounties/cln0iijxm000ml10f95gb5k8k/cancel-attempt">
        Cancel my attempt
      </a>
    </li>
  </ul>
</details>

Kishan-Kulkarni:
[...]

PR #3367, opened 2025-02-27
docs: Config reference

igneus:
For a start of #2431 here's the Config reference page generated using the approach outlined in the issue description.

Unfortunately the contents generated this way are more confusing than helpful (even the FieldName table cells are often empty, the other contents, if available, are rather enigmatic). And for `ZClient.Config` attempt to generate the table results in stack overflow.

/claim #2431


igneus:
`ZClient.Config` is not included, as an attempt to generate its documentation results in stack overflow (zio/zio-config#1545).

Please note that the output is affected by zio/zio-config#1542 , which is also source of the broken links reported by docusaurus. PR with a fix is linked to the issue.

987Nabil:
@igneus I just hit publish on zio-config. Can you update it once it is released?

igneus:
@987Nabil Thanks, done.

`sbt compileDocs` still emits some warnings (caused by this PR), because the generated markdown contains links to anchors which are only created when producing HTML (=> are not available when markdown link targets are being checked). If preferable, this could be fixed by generating our own anchors instead of relying on the auto-generated ones. But I would prefer not to, as it would involve markdown with inline HTML like

```markdown
# <a name="custom-heading-anchor"></a>Heading Text
```

and that's rather ugly.

987Nabil (approved):

PR #3394, opened 2025-03-11
Curl logger

landlockedsurfer:
Hi all!

Sorry for the delay...! Here are the fixes for the issues raised in the this pull request:
https://github.com/zio/zio-http/pull/3285

As mentioned by @guizmaii  it is really nice to have an explicit formatter method for the any `zio.http.Request`, so I added one.
I also fixed the issues discovered by @987Nabil :
Check if the body can be consumed using `body.isComplete` and the bug were an effect is returned instead of the result of the effect. A second test explicitly checks for this issue.

It might be a good idea to start reorganizing the client aspects into a dedicated package, but I'm pretty bad with code reorganizations. 😓

Hope this is useful!

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/ZClientAspectSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ZClientAspectSpec.scala
@@ -100,6 +100,29 @@ object ZClientAspectSpec extends ZIOHttpSpec {
           extractStatus(response) == Status.Ok,
         ),
       ),
+      test("curl request logger") {
+
+        for {
+          port       <- Server.install(routes)
+          baseClient <- ZIO.service[Client]
+          client = baseClient
+            .url(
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/test/scala/zio/http/ZClientAspectSpec.scala
```diff
@@ -100,6 +103,57 @@ object ZClientAspectSpec extends ZIOHttpSpec {
           extractStatus(response) == Status.Ok,
         ),
       ),
+      suite("curl request logger")(
```
987Nabil, line 110: Can you add one with a streaming body to check that it is not logged?

landlockedsurfer: That's a good idea! I've just added a new commit that implements a test for the logging of a streaming request body.

987Nabil (approved):

PR #3412, opened 2025-03-16
Improve `From#toQueryParams` performance (#3289)

### LINKED ISSUE #3289: Building a `Body` from `Form` is too slow
Form -> QueryParams reallocates **heavily**:

https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/Form.scala#L162-L167
https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/internal/QueryModifier.scala#L43

Likely necessary to introduce a `builder` internally to avoid generating so many (single element) collections

987Nabil:
fixes #3289 
/claim #3289

Old:
```
[info] Benchmark                                   Mode  Cnt     Score     Error  Units
[info] FormToQueryBenchmark.fromToQueryBenchmark  thrpt    3  1402.546 ± 455.149  ops/s
```

```
New:
[info] Benchmark                                   Mode  Cnt     Score      Error  Units
[info] FormToQueryBenchmark.fromToQueryBenchmark  thrpt    3  5090.872 ± 1630.801  ops/s
```

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Form.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Form.scala
@@ -160,11 +160,10 @@ final case class Form(formData: Chunk[FormField]) {
   }
 
   def toQueryParams: QueryParams =
-    formData.foldLeft(QueryParams.empty) {
-      case (acc, FormField.Text(k, v, _, _)) => acc.addQueryParam(k, v)
-      case (acc, FormField.Simple(k, v))     => acc.addQueryParam(k, v)
-      case (acc, _)                          => acc
-    }
+    QueryParams(formData.collect {
+      case FormField.Text(k, v, _, _) => (k, Chunk(v))
+      case FormField.Simple(k, v)     => (k, Chunk(v))
+    }: _*)
 
   /**
    * Encodes the form using URL encoding, using the default charset.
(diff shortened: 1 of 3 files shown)


zio-http/shared/src/main/scala/zio/http/Form.scala
```diff
@@ -160,11 +160,10 @@ final case class Form(formData: Chunk[FormField]) {
...
 
   def toQueryParams: QueryParams =
-    formData.foldLeft(QueryParams.empty) {
-      case (acc, FormField.Text(k, v, _, _)) => acc.addQueryParam(k, v)
-      case (acc, FormField.Simple(k, v))     => acc.addQueryParam(k, v)
-      case (acc, _)                          => acc
-    }
+    QueryParams(formData.collect {
+      case FormField.Text(k, v, _, _) => (k, Chunk(v))
+      case FormField.Simple(k, v)     => (k, Chunk(v))
```
guizmaii: Have you tried not using the `unapply`? 
```scala
    QueryParams(formData.collect {
      case t: FormField.Text   => (t.k, Chunk(t.v))
      case s: FormField.Simple => (s.k, Chunk(s.v))
    }: _*)
```

That's a change @kyri-petrou did a lot in ZIO codebase to improve perfs

Also, minor, you change use `Chunk.single` here. It'll avoid a `if` call for each element of `formData`

987Nabil: You are right. Should be even better since it does not wrap in option instances. I'll make the change an check the benchmark again

987Nabil:
```
[info] Benchmark                                   Mode  Cnt     Score     Error  Units
[info] FormToQueryBenchmark.fromToQueryBenchmark  thrpt    3  5574.717 ± 140.199  ops/s
```

PR #3421, opened 2025-03-20
fix: refactor addQueryParam methods to return self instead of concatenating

### LINKED ISSUE #3420: Query Parameter Duplication When Using Schema-Derived Parameters
**Describe the bug**
Query parameters get duplicated when using the `addQueryParam` methods that work with Schema-derived types. This appears to happen because `StringSchemaCodec.queryFromSchema(...).encode(value, queryParameters)` both mutates the input query parameters and returns them, causing parameters to be added multiple times.

**To Reproduce**
1. Create case classes with Schema derivation:
```scala
import zio.http.Request
import zio.schema.{derived, Schema}

case class Foo(
    a: Int,
    b: Int
) derives Schema

case class Bar(
    c: Int,
    d: Int
) derives Schema
```

2. Use `addQueryParam` with case classes:
```scala
// Results in: a=1&a=1&b=2&b=2&c=3&d=4
Request.get("https://example.com")
  .addQueryParam(Foo(1, 2))
  .addQueryParam(Bar(3, 4))
```

3. Or use `addQueryParam` with key-value pairs:
```scala
// Results in even worse duplication: a=1&a=1&a=1&a=1&a=1&a=1&a=1&a=1&b=2&b=2&b=2&b=2&c=3&c=3&d=4
Request.get("https://example.com")
  .addQueryParam("a", 1)
  .addQueryParam("b", 2)
  .addQueryParam("c", 3)
  .addQueryParam("d", 4)
```

**Expected behaviour**
The parameters should be added without duplication:
- First example should produce: `a=1&b=2&c=3&d=4`
- Second example should also produce: `a=1&b=2&c=3&d=4`

- ZIO HTTP version: 3.1.0
- Scala version: 3.6.4
- SBT version: 1.10.11

**Additional context**
The issue appears to be in these methods:

```scala
def addQueryParam[T](key: String, value: T)(implicit schema: Schema[T]): A =
  self ++ StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, key).encode(value, queryParameters)

def addQueryParam[T](value: T)(implicit schema: Schema[T]): A =
  self ++ StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, null).encode(value, queryParameters)
```

The `encode` method on `StringSchemaCodec` appears to both mutate the target `queryParameters` and return the query parameters which subsequently get added again via the `++` operator. This double-addition causes the duplication.

Possible solutions:
1. Modify `encode` to either mutate or return, not both
2. Change the implementation of `addQueryParam` to handle this correctly

mhriemers:
Fixes #3420

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/internal/QueryModifier.scala
+++ b/zio-http/shared/src/main/scala/zio/http/internal/QueryModifier.scala
@@ -54,11 +54,15 @@ trait QueryModifier[+A] { self: QueryOps[A] with A =>
       }),
     )
 
-  def addQueryParam[T](key: String, value: T)(implicit schema: Schema[T]): A =
-    self ++ StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, key).encode(value, queryParameters)
+  def addQueryParam[T](key: String, value: T)(implicit schema: Schema[T]): A = {
+    StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, key).encode(value, queryParameters)
+    self
+  }
 
-  def addQueryParam[T](value: T)(implicit schema: Schema[T]): A =
-    self ++ StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, null).encode(value, queryParameters)
+  def addQueryParam[T](value: T)(implicit schema: Schema[T]): A = {
+    StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, null).encode(value, queryParameters)
+    self
+  }
 
   /**
    * Removes the specified key from the query parameters.


mhriemers:
Alternative fix would be:
```scala
  def addQueryParam[T](key: String, value: T)(implicit schema: Schema[T]): A =
    self ++ StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, key).encode(value, QueryParams.empty)

  def addQueryParam[T](value: T)(implicit schema: Schema[T]): A =
    self ++ StringSchemaCodec.queryFromSchema(schema, ErrorConstructor.query, null).encode(value, QueryParams.empty)
```

mhriemers:
This is how `Header`s handle them:
```scala
  def addHeader[T](name: String, value: T)(implicit schema: Schema[T]): A =
    updateHeaders(StringSchemaCodec.headerFromSchema(schema, ErrorConstructor.header, name).encode(value, _))

  def addHeader[T](value: T)(implicit schema: Schema[T]): A =
    updateHeaders(StringSchemaCodec.headerFromSchema(schema, ErrorConstructor.header, null).encode(value, _))
```

which does function correctly:
```scala
import zio.http.Request
import zio.schema.{derived, Schema}

case class Foo(
    a: Int,
    b: Int
) derives Schema

case class Bar(
    c: Int,
    d: Int
) derives Schema

// val res0: zio.http.Request = Request(Default,GET,https://example.com,Headers((a,1), (b,2), (c,3), (d,4)),Body.empty,None,None)
Request
  .get("https://example.com")
  .addHeader(Foo(1, 2))
  .addHeader(Bar(3, 4))

// val res1: zio.http.Request = Request(Default,GET,https://example.com,Headers((a,1), (b,2), (c,3), (d,4)),Body.empty,None,None)
Request
  .get("https://example.com")
  .addHeader("a", 1)
  .addHeader("b", 2)
  .addHeader("c", 3)
  .addHeader("d", 4)
```

987Nabil:
Hey @mhriemers thanks for the report and attempt to fix, but the fix was not quite right.
I opened #3422 an close your PR in favor of it.

mhriemers:
@987Nabil Great, thanks! That was also one of my proposed solutions. Is there a timeframe in which this will be released?

987Nabil:
I'd like to make a release before Monday

PR #3424, opened 2025-03-20
Add `bodyAs` to `Response` (#1374)

### LINKED ISSUE #1374: Decode response body to case class
It would be nice to have the following code works:
```scala
  val url = "http://sports.api.decathlon.com/groups/water-aerobics"

  val program = for {
    res  <- Client.request(url)
    data <- res.bodyAs[MyCaseClass]
    _    <- Console.printLine(data)
  } yield ()
```

987Nabil:
fixes #1374
/claim #1374


guizmaii (approved):

guizmaii:
I have this exact operator in my project as an extension function 🙂🚀

Issue #3441, opened 2025-04-01
when an Endpoints implements function with options will only be called when all options are present


tont0r:
**Describe the bug**
I have an Endpoint that should have query params as Optional. If I do not specify all the query parameters that are optional, it will result in a 500. I have tried all combinations of @optionalField and Option[String] for the type.

**To Reproduce**
Steps to reproduce the behaviour:
```
final case class QueryParams(
    @optionalField
    language: String,
    @optionalField
    location: String
  )

val getInfo =
    Endpoint(Method.GET / "testEndpoint")
      .query[QueryParams]
      .out[String]
```

In this curl, only location is being passed, which results in a 500. If both location and language are passed, I will get a 200 and the `implements` function will be called.
```
curl -v -X 'GET' \
  'http://localhost:8080/testEndpoint?location=1' \
  -H 'accept: application/json' \


* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET /testEndpoint?location=1 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> accept: application/json
>
* Request completely sent off
< HTTP/1.1 500 Internal Server Error
< content-length: 0
```

**Expected behaviour**
I should get a 200 when the implements is called.

**Additional context**

tont0r:
https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/internal/StringSchemaCodec.scala#L123

It seems this is where the error is being thrown.
https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/QueryParams.scala#L137

the parameter is missing and its not finding it. even when its flagged as an optional

PR #3443, opened 2025-04-02
Fix Issue #3441

mwalden:
When an Endpoints implements function with options will only be called when all options are present. Tested with a case class, optional parameters. Validated with tests.


987Nabil:
@mwalden Thanks for the digging. The solution needs to be somewhat different though. Closing in favor of #3445 
If you find more bugs or have feature ideas, your future PRs will be welcomed 🙂

PR #3445, opened 2025-04-02
Make QueryParams#unsafeQueryParam return null for missing key (#3441)

### LINKED ISSUE #3441: when an Endpoints implements function with options will only be called when all options are present

987Nabil:
fixes #3441

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/QueryParams.scala
+++ b/zio-http/shared/src/main/scala/zio/http/QueryParams.scala
@@ -133,8 +133,10 @@ object QueryParams {
     override def updateQueryParams(f: QueryParams => QueryParams): QueryParams =
       f(self)
 
-    override private[http] def unsafeQueryParam(key: String): String =
-      underlying.get(key).get(0)
+    override private[http] def unsafeQueryParam(key: String): String = {
+      val value = underlying.get(key)
+      if (value == null || value.isEmpty) null else value.get(0)
+    }
 
     override def valueCount(name: CharSequence): Int =
       if (underlying.containsKey(name)) underlying.get(name.toString).size() else 0
--- a/zio-http/shared/src/main/scala/zio/http/internal/StringSchemaCodec.scala
+++ b/zio-http/shared/src/main/scala/zio/http/internal/StringSchemaCodec.scala
@@ -218,7 +218,7 @@ private[http] trait StringSchemaCodec[A, Target] {
             target0 = addAll(target0, values.map { v => (name, codec.encode(v)) })
           case _                   =>
             val encoded = codec.encode(value)
-            target0 = add(target0, name, encoded)
+            if (encoded != null) target0 = add(target0, name, encoded)
         }
       }
       target0


kyri-petrou (approved):

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -133,8 +133,10 @@ object QueryParams {
     override def updateQueryParams(f: QueryParams => QueryParams): QueryParams =
       f(self)
 
-    override private[http] def unsafeQueryParam(key: String): String =
-      underlying.get(key).get(0)
+    override private[http] def unsafeQueryParam(key: String): String = {
+      val value = underlying.get(key)
+      if (value == null) null else value.get(0)
```
mwalden: given that this is a list, you are still open to throwing a null pointer.

mwalden:
there were tests in the initial PR that were closed. to be safe, we should probably include them too.

https://github.com/zio/zio-http/pull/3443

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
987Nabil: Why? If it is null, I return null. If it is the empty list, get 0 returns null. That is the behaviour I want. Did I miss something?

987Nabil:
The tests already exists in the RoundtripSpec. It was green, because the value was added when it is encoded as null. Now it is not added, and without the fix in QueryParams it fails

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
mwalden: sorry i meant that on value.get(0). we may want a check around that. 
edit:
sorry did you say .get(0) would return null on an empty list? Wouldnt it return out of bounds?

987Nabil: fixed

Issue #3438, opened 2025-03-31
Request#addCookie(s) replaces Cookie instead of combining them


myazinn:
**Describe the bug**
Introduced in #3410
`Request#addCookie(s)` on invocation just adds `Header.Cookie` to other headers. It's fine when there's only one invocation per request, but when you call it multiple times in a chain, only the latest cookie arg ends up in a request. 

**To Reproduce**
Steps to reproduce the behaviour:
Here's the test that should pass, but fails
https://scastie.scala-lang.org/NExUWtaSQQuUajPEc2TXEw

Perhaps it's more clear what happens with good old println's
https://scastie.scala-lang.org/BKnG4gZSQbCBpRWC1M6IPQ

**Expected behaviour**
Multiple invocations of `Request#addCookie` / `Request#addCookies` should result in a request which contains all of the provided cookies, not just the latest argument.

hcwilhelm:
Hi, I was looking a bit into this issue and found the following. 

Multiple invocations of `Request#addCookie` / `Request#addCookies` do indeed result in a `Request` with multiple `Cookie Headers` it is not just the latest argument! First, I wasn't 100% sure if multiple Cookie headers are allowed in HTTP, but according to the RFC it is allowed in HTTP/2 ([RFC 7540](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2.5)),

```
suite("addCookie") (
      test("add cookie to request") {
        val c1 = Cookie.Request("n1", "c1")
        val c2 = Cookie.Request("n2", "c2")
        val c3 = Cookie.Request("n3", "c3")
        val c4 = Cookie.Request("n4", "c4")

        val request =
          Request
            .get(URL.root)
            .addCookie(c1)
            .addCookie(c2)
            .addCookies(c3, c4)

        println(request.cookies)
        println(request.rawHeaders("cookie"))

        assertTrue(request.cookies == Chunk(c1, c2, c3, c4))
      }
    )
```

Println Output 

```
Chunk(Request(n1,c1),Request(n2,c2),Request(n3,c3),Request(n4,c4))
Chunk(n1=c1,n2=c2,n3=c3; n4=c4)
```

I found that the `Request#cookies` method is not correctly implemented so I fixed that with 

```
  /**
   * Returns all cookies from the request.
   */
  def cookies: Chunk[Cookie] =
    headers(Header.Cookie).flatMap(_.value)
[...]

PR #3446, opened 2025-04-02
Fix multiple cookie header netty conversion

hcwilhelm:
Fix for #3438


987Nabil:
Thanks for the PR. I was tho working on a fix in parallel. I will close this PR in favor of mine (https://github.com/zio/zio-http/pull/3447) since HTTP/1.1 does not allow multiple cookie headers and we do not even support HTTP/2 for now.

Issue #3448, opened 2025-04-03
query parameters with `Option` type are in runtime always decoded as `Some`


gregor-rayman:
**Describe the bug**
When a query parameter is defined as `HttpCodec.query[Option[String]]("key")`, the value is decoded as `Some(Some("test"))` instead of `Some("test")` if the URL contain the parameter value `test`

**To Reproduce**
Here a test to add to `QueryParamterSpec.scala`

```scala
test("multiple options query parameters") {
      check(Gen.int, Gen.alphaNumericString, Gen.int) { (userId, key, value) =>
        val testRoutes = testEndpoint(
          Routes(
            Endpoint(GET / "users" / int("userId"))
              .query(HttpCodec.query[Option[String]]("key"))
              .query(HttpCodec.query[Option[Int]]("value"))
              .out[String]
              .implementHandler {
                Handler.fromFunction { case (userId, key, value) =>
                  s"path(users, $userId, $key, $value)"
                }
              },
          ),
        ) _
        testRoutes(s"/users/$userId", s"path(users, $userId, None, None)") &&
        testRoutes(s"/users/$userId?key=&value=", s"path(users, $userId, None, None)") &&
          testRoutes(s"/users/$userId?key=", s"path(users, $userId, Some(), None)") &&
          testRoutes(s"/users/$userId?key=$key", s"path(users, $userId, ${Some(key)}, None)") &&
          testRoutes(s"/users/$userId?key=&value=$value", s"path(users, $userId, Some(), ${Some(value)})") &&
          testRoutes(
            s"/users/$userId?key=$key&value=$value",
            s"path(users, $userId, ${Some(key)}, ${Some(value)})",
          )
      }
    },
```

PR #3451, opened 2025-04-05
Fix for optional query string parameters

landlockedsurfer:
Hi all,
Took a closer look at issue #3448 and I think I found the problem.
I've added two simple examples to to decode an optional query string parameter to `QueryParamsSpec`.
Hope this is useful!!

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/QueryParamsSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/QueryParamsSpec.scala
@@ -325,6 +325,8 @@ object QueryParamsSpec extends ZIOHttpSpec {
             single.queryOrElse[Int](typed, default) == 1,
             queryParams.queryOrElse[Int](invalidTyped, default) == default,
             queryParams.queryOrElse[Int](unknown, default) == default,
+            single.query[Option[Int]](typed) == Right(Some(1)),
+            single.query[Option[Int]]("notGiven") == Right(None),
             queryParams.query[Chunk[Int]](typed) == Right(Chunk(1, 2)),
             queryParams.query[Chunk[Int]](invalidTyped).isLeft,
             queryParams.query[Chunk[Int]](unknown) == Right(Chunk.empty),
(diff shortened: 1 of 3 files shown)


gregor-rayman:
It looks good :) 

Could you please also add the following test to `QueryParameterSpec`?

```scala
  test("multiple options query parameters") {
      check(Gen.int, Gen.alphaNumericString, Gen.int) { (userId, key, value) =>
        val testRoutes = testEndpoint(
          Routes(
            Endpoint(GET / "users" / int("userId"))
              .query(HttpCodec.query[Option[String]]("key"))
              .query(HttpCodec.query[Option[Int]]("value"))
              .out[String]
              .implementHandler {
                Handler.fromFunction { case (userId, key, value) =>
                  s"path(users, $userId, $key, $value)"
                }
              },
          ),
        ) _
        testRoutes(s"/users/$userId", s"path(users, $userId, None, None)") &&
          testRoutes(s"/users/$userId?key=", s"path(users, $userId, Some(), None)") &&
          testRoutes(s"/users/$userId?key=$key", s"path(users, $userId, ${Some(key)}, None)") &&
          testRoutes(s"/users/$userId?key=&value=$value", s"path(users, $userId, Some(), ${Some(value)})") &&
          testRoutes(
            s"/users/$userId?key=$key&value=$value",
            s"path(users, $userId, ${Some(key)}, ${Some(value)})",
          )
      }
    },```

landlockedsurfer:
Hi @gregor-rayman ,
I've just added the mentioned test to this branch also.

gregor-rayman:
> Hi @gregor-rayman , I've just added the mentioned test to this branch also.

Thanks a lot. :) 

It seems that there is an issue with Scala 3. I can try to have a look later today.

landlockedsurfer:
@gregor-rayman Thanks for pointing out the compile error with Scala 3! Hopefully this is fixed now.

987Nabil (approved):

PR #3482, opened 2025-05-03
Add tests and docs for new CodecConfig options (#3477)

### LINKED ISSUE #3477: Integrate new zio-schema-json options into CodecConfig
zio-schema-json added new options to configure the json codec. We should make them configurable through `CodecConfig` in the EndpointAPI

987Nabil:
fixes #3477
/claim #3477

### DIFF
--- a/docs/reference/endpoint.md
+++ b/docs/reference/endpoint.md
@@ -534,6 +534,63 @@ In the above example, we mapped over the input type of the `endpoint` and transf
 
 The `transformOut` and `transformError` methods work similarly to the `transformIn` method.
 
+## CodecConfig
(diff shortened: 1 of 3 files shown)


kyri-petrou (commented):
I might be missing something but the PR doesn't seem to be aligned with / fixing the linked issue.
Was there a change in the decision to integrate it directly with the EndpointAPI and simply apply the config to routes instead?

987Nabil:
@kyri-petrou the `CodecConfig` is a Endpoint API concept. It has no effect on the features of the low level API.
I added already the config parameters in #3341 
I could have made it one PR and claim two bounties, but to claim automatically I just added tests and docs here.

kyri-petrou (approved):

PR #3489, opened 2025-05-11
Introduce format annotation to control string-types in openapi spec

### LINKED ISSUE #3488: Allow setting format and validations in open-api spec for all properties of type string
When generating API clients, I'd like to automatically handle conversion of object properties of type string based on the "format" property in the open-api spec, i.e. based on https://json-schema.org/draft/2020-12/json-schema-validation#name-vocabularies-for-semantic-c
The most relevant examples are date, time and date-time.

To that end I suggest introducing a new `format` annotation that can be stored in the annotations field of Schema instances.

I could post-process my spec and manually add the formats when generating the spec

jdsalchow:
# Add support for string format annotations

## Issue
Fixes #3488

## Description
This PR introduces a new `format` annotation that can be used to specify the format of string properties in OpenAPI schemas. This allows for automatic handling of string-based object properties according to the OpenAPI specification's format property, as defined in [JSON Schema Validation](https://json-schema.org/draft/2020-12/json-schema-validation#name-vocabularies-for-semantic-c).

## Implementation Details
- Added a `format` annotation that allows overriding the default format in the schema
- Don't ignoring validation annotations for string types other than string
- Set default `StringFormat` for various format types:
  - LocalDate -> date
  - Instant | OffsetDateTime | ZonedDateTime ->  date-time
  - OffsetTime -> time
  - Period | Duration -> Duration
  - UUID -> UUID

## Example Usage
```scala
final case class EmailData(
  @format(Email) email: String, 
  received: Option[java.time.LocalDate]
) derives Schema
```

This generates an OpenAPI schema with the appropriate format annotations:
```json
{
  "email": {
    "type": "string",
    "format": "email"
  },
  "received": {
    "type": ["string", "null"],
    "format": "date"
  }
}
```

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -1,6 +1,6 @@
 package zio.http.endpoint.openapi
 
-import scala.annotation.{nowarn, tailrec}
+import scala.annotation.{StaticAnnotation, nowarn, tailrec}
 
 import zio._
 import zio.json.ast.Json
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -712,68 +712,82 @@ object JsonSchema {
...
-          case StandardType.StringType         =>
-            JsonSchema.String.fromValidation(
+          case StandardType.UnitType                               => JsonSchema.Null
+          case StandardType.StringType | StandardType.DayOfWeekType | StandardType.MonthType |
+              StandardType.MonthDayType | StandardType.YearType | StandardType.YearMonthType | StandardType.ZoneIdType |
+              StandardType.ZoneOffsetType | StandardType.LocalDateTimeType | StandardType.LocalTimeType |
+              StandardType.CurrencyType | StandardType.BinaryType | StandardType.CharType =>
+            JsonSchema.String.fromValidationAndFormat(
               annotations.collect { case zio.schema.annotation.validate(v) => v }.headOption,
+              annotations.collect { case format(stringFormat) => stringFormat }.headOption,
```
987Nabil, line 722: I don't think an annotation is the right solution.
Especially not one that lives in this file.

I think right solution is to have a mapping of `StandardType` to a format.
If this is not precise enough, then we need something in zio schema to make it configurable

987Nabil:
Please reopen if you want to keep working on this

PR #3503, opened 2025-05-20
Handler: catch NonFatal exceptions in by-name parameters

hearnadam:


zio-http/jvm/src/test/scala/zio/http/HandlerSpec.scala
```diff
@@ -28,6 +28,11 @@ import zio.test._
 @nowarn("msg=dead code")
 object HandlerSpec extends ZIOHttpSpec with ExitAssertion {
 
+  private def catches[R, E, In, Out](handler: => Handler[R, E, In, Out]): TestResult =
+    ZIO.suspendSucceed(handler(null).exit).flatMap(exit => assertTrue(exit.isDie))
+
+  private def throws: Nothing = throw new RuntimeException
+
```
hearnadam:
```suggestion
```

zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -1248,7 +1251,7 @@ object Handler extends HandlerPlatformSpecific with HandlerVersionSpecific {
           try {
             Exit.fromEither(f(in))
           } catch {
-            case error: Throwable => Exit.die(error)
+            case NonFatal(error) => Exit.die(error)
```
987Nabil: I think this might be the reason, that the performance went down. Since it allocates options. Can you change it to only do a boolean check via the `NonFatal.apply`

987Nabil (changes requested):

zio-http/shared/src/main/scala/zio/http/Handler.scala
hearnadam: Sure, happy to change it. Though if this is an issue I would consider it a bug. `NonFatal.unapply` is specialized: https://github.com/scala/scala/pull/10601

hearnadam:
@987Nabil I am fairly confident this is caused by the new visibility of log messages involving Netty's Fatals. Somehow it seems new exceptions are being propagated through channels and being logged. Is it possible to disable Netty's logging?

987Nabil:
We can take a look I guess. But I am hesitant to merge this, before we understand what is going on for sure.
@kyri-petrou maybe you have an idea?

kyri-petrou:
I'm 99% sure that this is due to one of the usages of `ZIO.suspendSucceed`. The extra ZIO itself is not what's causing the regression, but the fact that now we're now evaluating the effect in ZIO's threadpool causing context switching between netty's and ZIO's threadpools. Check whether adding the following code [here](https://github.com/zio/zio-http/blob/188ff2ed4b980981961e49f133d50bba59778ff1/zio-http-example/src/main/scala/example/PlainTextBenchmarkServer.scala#L38) reverts the regression for that benchmark:

```scala
.avoidContextSwitching(true)
```

Currently we're delegating anything that is not an `Exit` to ZIO's threadpool without trying to synchronously evaluate it first. This is to avoid having CPU-intensive work (or potentially blocking if ZIO's executor is the virtual-thread based executor is used) being done in Netty's threadpool.

See https://github.com/zio/zio-http/issues/2577 for more info on this behaviour (especially https://github.com/zio/zio-http/issues/2577#issuecomment-2063854315)

zio-http-example/src/main/scala/example/SimpleEffectBenchmarkServer.scala
```diff
@@ -31,8 +31,10 @@ object SimpleEffectBenchmarkServer extends ZIOAppDefault {
       ),
   )
 
-  private val config = Server.Config.default
-    .port(8080)
+  private val config =
+    Server.Config.default
+      .port(8080)
+      .logWarningOnFatalError(false)
```
hearnadam:
```suggestion
      .logWarningOnFatalError(false)
      .avoidContextSwitching(true)
```

zio-http-example/src/main/scala/example/PlainTextBenchmarkServer.scala
```diff
@@ -34,8 +34,10 @@ object PlainTextBenchmarkServer extends ZIOAppDefault {
       ),
   )
 
-  private val config = Server.Config.default
-    .port(8080)
+  private val config =
+    Server.Config.default
+      .port(8080)
+      .logWarningOnFatalError(false)
```
hearnadam:
```suggestion
      .logWarningOnFatalError(false)
      .avoidContextSwitching(true)
```

hearnadam:
@987Nabil ready for review. Thanks Kyri for the help!

987Nabil (approved):

PR #3525, opened 2025-06-01
Implement dependencyless query de-/encoding (#3521)

### LINKED ISSUE #3521: Replace QueryParamEncoding with own impl
It is netty based but we want netty in its own module. We should impl. the http spec here ourselves

987Nabil:
fixes #3521 
/claim #3521


zio-http/shared/src/main/scala/zio/http/Path.scala
```diff
@@ -159,7 +169,7 @@ final case class Path private[http] (flags: Path.Flags, segments: Chunk[String])
   /**
    * Checks if the path is equal to "".
    */
-  def isEmpty: Boolean = segments.isEmpty && (flags == Flags.none)
+  def isEmpty: Boolean = self == Path.empty || segments.isEmpty && (flags == Flags.none)
```
kyri-petrou: I haven't checked the implementation of `Path#equals` but if it's not cheap is it worth maybe adding a `.isEmpty` method on it or using `eq` instead?

zio-http/shared/src/main/scala/zio/http/Path.scala
```diff
@@ -122,10 +122,20 @@ final case class Path private[http] (flags: Path.Flags, segments: Chunk[String])
...
     else if (self == Path.root) "/"
     else segments.mkString(if (hasLeadingSlash) "/" else "", "/", if (hasTrailingSlash) "/" else "")
 
+  private[http] def encodeBuilder: StringBuilder = {
+    val sb = new StringBuilder(256)
+    if (hasLeadingSlash) sb.append('/')
+    segments.foreach { segment =>
+      sb.append(segment)
+      if (hasTrailingSlash || segment != segments.last) sb.append('/')
+    }
```
kyri-petrou: Maybe use a while loop to avoid creating the closure

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -184,10 +182,10 @@ object QueryParams {
         case Some(previous) =>
           val combined = new util.ArrayList[String]()
           combined.addAll(previous)
-          combined.addAll(entry.getValue)
+          combined.addAll(entry.getValue.asScala.filterNot(_.isBlank).asJava)
```
kyri-petrou: Are we expecting blank values? Might be worth to check if `entry.getValue` contains blank values first to avoid the 3 collection conversions/allocations?

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -184,10 +182,10 @@ object QueryParams {
...
           result.replace(entry.getKey, combined)
         case None           =>
-          result.put(entry.getKey, entry.getValue)
+          result.put(entry.getKey, entry.getValue.asScala.filterNot(_.isBlank).asJava)
```
kyri-petrou: As above

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -86,9 +86,7 @@ trait QueryParams extends QueryOps[QueryParams] {
    * Normalizes the query parameters by removing empty keys and values.
    */
   def normalize: QueryParams =
-    QueryParams.fromEntries(seq.filter { entry =>
-      entry.getKey.nonEmpty && entry.getValue.asScala.nonEmpty
-    }: _*)
+    QueryParams.fromEntries(seq.filter { entry => entry.getKey.nonEmpty }: _*)
```
kyri-petrou: Might be worth to check if there's anything to filter out first and return the same object if nothing is empty?

zio-http/shared/src/main/scala/zio/http/internal/QueryGetters.scala
```diff
@@ -108,7 +108,8 @@ trait QueryGetters[+A] { self: QueryOps[A] =>
    * Retrieves the first query parameter value having the specified name.
    */
   def queryParam(key: String): Option[String] =
-    if (hasQueryParam(key)) Some(queryParams(key).head) else None
+    if (hasQueryParam(key)) if (queryParams(key).isEmpty) Some("") else Some(queryParams(key).head)
```
kyri-petrou: 1. `queryParams(key)` is called twice unnecessarily
2. `Some("")` can be stored in a val to avoid allocating it each time

Also I have a feeling that it might be possible to avoid `hasQueryParam` and the allocation of `Chunk` altogether by adding something like `getHeadUnsafe` on `QueryParams` that returns a null in case it doesn't exist

zio-http/shared/src/main/scala/zio/http/internal/QueryModifier.scala
```diff
@@ -36,7 +36,13 @@ trait QueryModifier[+A] { self: QueryOps[A] with A =>
    * Adds the specified key/value pair to the query parameters.
    */
   def addQueryParam(key: String, value: String): A =
-    addQueryParams(key, Chunk(value))
+    if (value.isEmpty)
+      addQueryParam(key)
+    else
+      addQueryParams(key, Chunk(value))
```
kyri-petrou: Use `Chunk.single` as it's faster than `apply` for single value Chunks

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
```diff
@@ -17,12 +17,307 @@
...
-}
+private[http] object QueryParamEncoding {
+  def decode(queryStringFragment: String, charset: Charset): QueryParams = {
+    if (queryStringFragment == null || queryStringFragment.isEmpty) {
+      QueryParams.empty
+    } else {
+      val length = queryStringFragment.length
+
+      // Count & characters more idiomatically using Scala's count function
+      val paramCount = 1 + queryStringFragment.count(_ == '&')
```
kyri-petrou: This will box the `Char`s in the string, better use a while loop

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
kyri-petrou: Also might be worth to pre-size `util.ArrayList[String]()`?

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
```diff
@@ -17,12 +17,307 @@
...
+      return baseUri.toString
+    }
+
+    // Estimate additional capacity needed for query parameters
+    // Initial guess: 1 char for ? + average of 20 chars per param pair
+    val additionalCapacity = 1 + (queryParams.seq.size * 20)
+    baseUri.ensureCapacity(baseUri.length + additionalCapacity)
+
+    var isFirst = true
+    queryParams.seq.foreach { entry =>
```
kyri-petrou: Don't use `foreach` with local vars, it moves them to the heap instead of keeping them instack memory. Prefer using a while-loop instead

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
```diff
@@ -17,12 +17,307 @@
...
+            val decoded = (digit1 << 4) | digit2
+            result.append(decoded.toChar)
+            i += 3
+          } else {
+            // Invalid hex encoding, treat % as literal
+            result.append('%')
+            i += 1
+          }
+        } catch {
+          case _: IndexOutOfBoundsException =>
```
kyri-petrou: Is this a common thing or a very unlikely fallback? Throwing exceptions is really expensive + it destroys the JVMs JIT optimizations

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -344,20 +344,23 @@ object URL {
...
-        url.queryParams.normalize,
+    val path =
+      QueryParamEncoding.encode(
+        url.path.encodeBuilder,
+        url.queryParams,
         Charsets.Http,
-      ) + url.fragment.fold("")(f => "#" + f.raw)
+      )
+
+    val fragment = url.fragment.fold("")(f => "#" + f.raw)
```
kyri-petrou: It seems that this is only used to do `path + fragment` so maybe use patmat instead and simply return `path` in one case and `s"${path}#{f.raw}"` in the other

zio-http/shared/src/main/scala/zio/http/Path.scala
987Nabil: This is the isEmpty :) eq is good

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
987Nabil: I check where we call this. Since it is private, I went for optimizing the call sides.
So I expect no more blanks. I general I like to optimize/normalize on creation. Makes life easier.

987Nabil:
@kyri-petrou I added benchmarks to compare with the old netty impl
Decoding looks good enough. Encoding seems still to be significantly too slow. I am taking a look

```
Benchmark                                Mode  Cnt    Score     Error  Units
QueryEncodingBenchmark.nettyDecoding    thrpt    3  275.658 ±  63.229  ops/s
QueryEncodingBenchmark.nettyEncoding    thrpt    3  717.565 ± 444.026  ops/s
QueryEncodingBenchmark.zioHttpDecoding  thrpt    3  267.109 ±  57.085  ops/s
QueryEncodingBenchmark.zioHttpEncoding  thrpt    3  559.516 ±  54.185  ops/s
```

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -86,9 +87,71 @@ trait QueryParams extends QueryOps[QueryParams] {
...
   def normalize: QueryParams =
-    QueryParams.fromEntries(seq.filter { entry =>
-      entry.getKey.nonEmpty && entry.getValue.asScala.nonEmpty
-    }: _*)
+    self match {
+      case _ if self.isEmpty                        => self
+      case JavaLinkedHashMapQueryParams(underlying) =>
+        var needsNormalization = false
+        val it                 = underlying.entrySet().iterator()
+        while (it.hasNext) {
```
kyri-petrou: I think you can add `!needsNormalization` here right

zio-http/shared/src/main/scala/zio/http/QueryParams.scala
```diff
@@ -86,9 +87,71 @@ trait QueryParams extends QueryOps[QueryParams] {
...
+        val seq                = self.seq
+        val it                 = seq.iterator
+        var needsNormalization = false
+        while (it.hasNext && !needsNormalization) {
+          val entry = it.next()
+          if (entry.getKey.isEmpty || entry.getValue == null || entry.getValue.isEmpty) {
+            needsNormalization = true
+          } else {
+            var i = 0
+            while (!needsNormalization && i < entry.getValue.size()) {
```
kyri-petrou: `!needsNormalization` seems redundant

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -344,21 +345,27 @@ object URL {
...
-        url.queryParams.normalize,
+    val path =
+      QueryParamEncoding.encode(
+        url.path.encodeBuilder,
+        url.queryParams,
         Charsets.Http,
-      ) + url.fragment.fold("")(f => "#" + f.raw)
+      )
+
+    def fromAbsURL(abs: Location.Absolute, path: String) = {
```
kyri-petrou, line 355: Scala's string interpolator is terrible performance-wise because it initializes the string builder with an initial size equal to the number of constant chars.

If this is in the hotpath I recommend using a StringBuilder and initializing it with a pre-calculated initial value

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -344,21 +345,27 @@ object URL {
...
     url.kind match {
-      case Location.Relative      => path(true)
-      case abs: Location.Absolute =>
-        val path2 = path(false)
-        abs.portIfNotDefault match {
-          case None             => s"${abs.scheme.encode}://${abs.host}$path2"
-          case Some(customPort) => s"${abs.scheme.encode}://${abs.host}:$customPort$path2"
-        }
+      case Location.Relative if url.fragment.isEmpty      => path
+      case Relative                                       => s"$path#${url.fragment.get.raw}"
```
kyri-petrou: As above

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
```diff
@@ -17,12 +17,311 @@
...
+            isFirst = false
+          } else {
+            baseUri.append('&')
+          }
+          baseUri.append(encodeComponent(key, charset))
+          baseUri.append('=')
+        } else {
+          // Handle key with one or more values
+          import scala.jdk.CollectionConverters._
+          values.asScala.foreach { value =>
```
kyri-petrou: I think you can use an iterator on `values` with a while loop

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
```diff
@@ -17,12 +17,311 @@
...
+    if (isUtf8) {
+      // Optimized UTF-8 path
+      while (j < len) {
+        val c = component.charAt(j)
+        if (
+          (c >= 'a' && c <= 'z') ||
+          (c >= 'A' && c <= 'Z') ||
+          (c >= '0' && c <= '9') ||
+          c == '-' || c == '.' || c == '_' || c == '~' || c == '*'
+        ) {
```
kyri-petrou: I think this is also used above. Maybe we can put it into a method? The JVM will be able to JIT-compile it and inline it

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
```diff
@@ -17,12 +17,311 @@
...
+          result.append(Character.forDigit((byte3 >> 4) & 0xf, 16).toUpper)
+          result.append(Character.forDigit(byte3 & 0xf, 16).toUpper)
+        }
+        j += 1
+      }
+    } else {
+      // Generic path for other charsets
+      val bytes = component.getBytes(charset)
+
+      bytes.foreach { b =>
```
kyri-petrou: Haven't checked but this might be boxing the `Byte`

987Nabil:
@kyri-petrou I added a thread local for StringBuffer to reuse instead of allocate (it is what netty is doing).
It made performance of encoding significantly better

```scala
[info] Benchmark                                Mode  Cnt    Score     Error  Units
[info] QueryEncodingBenchmark.nettyDecoding    thrpt    3  271.907 ± 292.971  ops/s
[info] QueryEncodingBenchmark.nettyEncoding    thrpt    3  777.109 ± 170.925  ops/s
[info] QueryEncodingBenchmark.zioHttpDecoding  thrpt    3  278.308 ± 141.437  ops/s
[info] QueryEncodingBenchmark.zioHttpEncoding  thrpt    3  661.459 ± 157.933  ops/s
```

I will look at your comments now

zio-http/shared/src/main/scala/zio/http/URL.scala
987Nabil: It potentially is, since it is used to build client requests.

zio-http/shared/src/main/scala/zio/http/internal/QueryParamEncoding.scala
987Nabil: Used while with index access. This is a Java list, most likely ArrayList. So get is O(1) and I save any kind of allocation.

987Nabil:
@kyri-petrou you can take a look again.
Here the latest benchmarks. The gap closes. Idk if we should invest more time in optimizing now. I did not use a profiler yet.
But it does not look too bad.

```
Benchmark                                Mode  Cnt    Score     Error  Units
QueryEncodingBenchmark.nettyDecoding    thrpt    3  311.339 ±  62.625  ops/s
QueryEncodingBenchmark.nettyEncoding    thrpt    3  806.052 ± 125.475  ops/s
QueryEncodingBenchmark.zioHttpDecoding  thrpt    3  259.053 ± 194.042  ops/s
QueryEncodingBenchmark.zioHttpEncoding  thrpt    3  752.642 ± 501.157  ops/s
```

Second run, same code. Probably we should check again when merged on the main build. The runner should have less fluctuations then my machine.

```
Benchmark                                Mode  Cnt    Score     Error  Units
QueryEncodingBenchmark.nettyDecoding    thrpt    3  293.061 ± 407.580  ops/s
QueryEncodingBenchmark.nettyEncoding    thrpt    3  796.414 ± 248.773  ops/s
QueryEncodingBenchmark.zioHttpDecoding  thrpt    3  281.032 ±  53.932  ops/s
QueryEncodingBenchmark.zioHttpEncoding  thrpt    3  766.250 ± 388.666  ops/s
```

zio-http/shared/src/main/scala/zio/http/internal/ThreadLocals.scala
```diff
@@ -0,0 +1,24 @@
+package zio.http.internal
+
+private[http] object ThreadLocals {
+  // We need some java api and the Scala StringBuilder has no public access to the underlying
+  // Java StringBuilder in 2.12
+  // Since this is only used internally that should be fine.
+  val StringBuilder: ThreadLocal[java.lang.StringBuilder] =
```
kyri-petrou, line 7: Just be very careful not to use this in any kind of re-entrant method / way otherwise it'll make a mess. Although that shouldn't really be an issue if the test coverage is good

kyri-petrou (approved):

zio-http/shared/src/main/scala/zio/http/internal/ThreadLocals.scala
987Nabil: Yeah I know. It must not be called in an effect, future etc. So far it is only called in pure functions. Netty had the same thing. But yes, this has to be handled with care. The performance boost is worth having it tho

PR #3548, opened 2025-06-12
New default implementation for text-based bodies (#3529)

### LINKED ISSUE #3529: Replace BodyEncoding with own impl

987Nabil:
fixes #3529
/claim #3529


zio-http/jvm/src/main/scala/zio/http/netty/NettyBodyWriter.scala
```diff
@@ -79,6 +79,9 @@ private[netty] object NettyBodyWriter {
       case AsciiStringBody(asciiString, _) =>
         writeArray(asciiString.array(), isLast = true)
         None
+      case sb: StringBody                  =>
+        writeArray(sb.bytes, isLast = true)
```
kyri-petrou, line 83: Shouldn't this call `unsafeAsArray`?

zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -650,4 +650,51 @@ object Body {
 
   private val zioEmptyChunk = Exit.succeed(Chunk.empty[Byte])
 
+  private[zio] final case class StringBody(
+    data: String,
+    charset: Charset = Charsets.Http,
+    override val contentType: Option[Body.ContentType] = None,
+  ) extends UnsafeBytes {
+
+    var bytes: Array[Byte] = null
```
kyri-petrou, line 659: This should be private right?

987Nabil: Yes

kyri-petrou: Also I'm not sure whether caching makes sense here. Is there a chance the method will be invoked less than once or more than once?

987Nabil: I did this mainly, because this mimics the behavior of nettys AsciiString class, that keep the bytes cached.
If the body is used by us, we call it at least once.

More then once, maybe if we use known size. I will check. Maybe remove it

guizmaii: @987Nabil Did you forget to make it private? 🤔

kyri-petrou: @987Nabil if it's invoked at least once, does it make sense to store it as a String in the class to begin with? Can't we simply make the conversion to `Array[Byte]` and store it as such instead?

987Nabil: A `Body` should be short lived anyway. If you do logging or similar, it might be helpful to still have a String ref and not create a new string? Even if only we call it, having the byte conversion done a little earlier or later does not seem to matter for me. Does it?
I don't see a harm in its current state. If only we call it, the body is GC-able so soon afterwards that it should not matter. And when user creates the body it should usually be at the end of the user land call stack.
Wdyt @kyri-petrou. Did I miss something?

kyri-petrou: It doesn't make a difference as long as we don't cache the `Array[Byte]` in a var, because in that case the class contains an extra field which increases memory pressure (especially for large responses).

PR #3550, opened 2025-06-13
Replace Netty based RFC 1123 date encoding (#3527)

### LINKED ISSUE #3527: Replace DateEncoding with own impl

987Nabil:
fixes #3527
/claim #3527


plokhotnyuk:
[Here](https://github.com/disneystreaming/smithy4s/blame/4bfa08b0306a75cae85d03bfb6e245462824a2f7/modules/core/src-jvm-native/smithy4s/Timestamp.scala#L22) is an efficient implementation of parsing/formatting for different timestamp formats including RFC 1123. Most performance tricks were ported from jsoniter-scala routines for JSON parsing and serialization of different data time values.

Java's `DateTimeFormatter` is ~10x times slower and affects scalability by eating memory bandwidth greatly. Even when using caching for formatting it introduces much bigger tail latency during refreshing of formatted values.

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
```diff
@@ -16,11 +16,226 @@
...
+private[http] object DateEncoding {
+  private val formatter = DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC)
+  private val `'0'`     = 48 // ASCII code for '0'
+
+  def encodeDate(date: ZonedDateTime): String =
+    encodeRFC1123(date)
+
+  def decodeDate(date: String): Option[ZonedDateTime] = {
+    try {
+      if (date.head.isUpper && date.charAt(3) == ',' && date.charAt(4) == ' ')
```
guizmaii: `date.head` throws if `date` is empty.
Should we maybe prefer to do:
```scala
if (date.isEmpty) None 
else if (date.charAt(0).isUppe && ...) 
  ...
```
?
Probably a good change for performances

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
```diff
@@ -16,11 +16,226 @@
...
+      case second                => stringBuilder.append(second).append(" GMT")
+    }
+
+    stringBuilder.toString
+  }
+
+  private def decodeRFC1123(date: String): Option[ZonedDateTime] = {
+    {
+      // TODO consider trie-hard approach
+      val c0 = date.head
```
guizmaii: Prefer `date.charAt(0)`? It avoids an `if (string.isEmpty)` check present in the `.head` method

Related: https://github.com/zio/zio-http/pull/3550/files#r2155802128

987Nabil:
@guizmaii Implemented some benchmarks

Before your suggested changes
```
Benchmark                                      Mode  Cnt        Score         Error  Units
DateEncodingBenchmark.nettyDecoding           thrpt    3   439288.041 ± 1092690.136  ops/s
DateEncodingBenchmark.nettyDecodingInvalid    thrpt    3   427059.216 ±  895231.682  ops/s
DateEncodingBenchmark.nettyEncoding           thrpt    3   635390.868 ± 1403521.676  ops/s
DateEncodingBenchmark.zioHttpDecoding         thrpt    3  2317595.133 ± 2830576.722  ops/s
DateEncodingBenchmark.zioHttpDecodingInvalid  thrpt    3  2443173.737 ±  219913.014  ops/s
DateEncodingBenchmark.zioHttpEncoding         thrpt    3  1095266.738 ±  378289.638  ops/s
```

After
```
Benchmark                                      Mode  Cnt         Score         Error  Units
DateEncodingBenchmark.nettyDecoding           thrpt    3    471698.844 ±  401495.065  ops/s
DateEncodingBenchmark.nettyDecodingInvalid    thrpt    3    433314.955 ±  208199.739  ops/s
DateEncodingBenchmark.nettyEncoding           thrpt    3    596804.828 ± 1649578.829  ops/s
DateEncodingBenchmark.zioHttpDecoding         thrpt    3   2327353.872 ± 2900047.041  ops/s
DateEncodingBenchmark.zioHttpDecodingInvalid  thrpt    3  17922976.451 ± 3374509.274  ops/s
DateEncodingBenchmark.zioHttpEncoding         thrpt    3   1054091.409 ± 1728102.469  ops/s
```

The invalid case has explicitly strings with invalid length.

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
987Nabil: I checked instead, that length is 28 or 29 since only these are valid.
Theoretically 27 is valid (year before 1k) But I think it is okay to ignore

nau:
Cool job! Although why not make these changes in Netty code for everyone to benefit?

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
```diff
@@ -16,11 +16,227 @@
...
-import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+import java.time.{ZoneOffset, ZonedDateTime}
 
-private[http] trait DateEncoding {
-  def encodeDate(date: ZonedDateTime): String
-  def decodeDate(date: String): Option[ZonedDateTime]
-}
+private[http] object DateEncoding {
+  private val formatter = DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC)
```
guizmaii: Is this `formatter` never used? 🤔

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
```diff
@@ -16,11 +16,227 @@
...
+  private val `'0'`     = 48 // ASCII code for '0'
+
+  def encodeDate(date: ZonedDateTime): String =
+    encodeRFC1123(date)
+
+  def decodeDate(date: String): Option[ZonedDateTime] = {
+    try {
+      if (date.length != 29 && date.length != 28) return None
```
guizmaii: @plokhotnyuk Is there any cost of calling `date.length` twice? Or the compiler and/or JIT will optimise this? 🤔

guizmaii:
@nau 

> Although why not make these changes in Netty code for everyone to benefit?

I can be wrong but I think the idea is probably to have this code in pure Scala so that it compiles to JS and Native and reduces the dependency on Netty

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
987Nabil: Yes. Forgot to delete it agin

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
987Nabil: String length is just a call to an internal field of the string impl. I don't think there is a real cost for this call.

987Nabil:
@nau @guizmaii There are two reasons.
1. We want to be independent of Netty. Not only for JS/Native, but also to be able to offer different backends in the JVM.
2. Nettys formatter supports not only RFC1123 dates. We can optimize here by failing for any other date then RFC1123 dates

zio-http/shared/src/main/scala/zio/http/internal/DateEncoding.scala
```diff
@@ -16,11 +16,228 @@
...
-  def decodeDate(date: String): Option[ZonedDateTime]
-}
+private[http] object DateEncoding {
+  private val `'0'` = 48 // ASCII code for '0'
+
+  def encodeDate(date: ZonedDateTime): String =
+    encodeRFC1123(date)
+
+  def decodeDate(date: String): Option[ZonedDateTime] = {
+    if (date.length != 29 && date.length != 28) return None
```
guizmaii, line 29: @987Nabil 
Shouldn't it be `||` instead of `&&`? 🤔

Issue #3540, opened 2025-06-07
`Request.query` doesn't support `Optional(Transform(..))` schemas


nartamonov:
**Describe the bug**

`Request.query[Option[T]](String)` doesn't work if the schema of `T` is derived from transformed `Schema[String]`  (with `transformOrFail` combinator).

To reproduce:

```scala
import zio.http.*
import zio.schema.Schema
import zio.test.*

object QueryBug2 extends ZIOSpecDefault {

  final case class PhoneNumber private (value: String)

  object PhoneNumber {
    def fromString(s: String): Option[PhoneNumber] =
      if (s.forall(_.isDigit)) Some(PhoneNumber(s)) else None

    implicit val schema: Schema[PhoneNumber] =
      Schema[String].transformOrFail(
        s => fromString(s).toRight(s"Invalid phone number: $s"),
        p => Right(p.value)
      )
  }

  val spec = suite("QueryBug")(
    // it works!
    test("decode mandatory query param") {
      val req = Request.get(URL.empty).addQueryParam("phone", "1234567890")
      assertTrue(
        req.query[PhoneNumber]("phone") == Right(PhoneNumber("1234567890")),
      )
    },

    // it doesn't work!
    // Runtime exception: java.lang.IllegalArgumentException: Unsupported schema Optional(Transform(Primitive(string,Chunk()), ...)))
    test("decode optional query param") {
      val req = Request.get(URL.empty).addQueryParam("phone", "1234567890")
      assertTrue(
        req.query[Option[PhoneNumber]]("phone") == Right(Some(PhoneNumber("1234567890"))),
      )
    }
  )
}
```

**Expected behaviour**
I expect that if `query[T]` works for some type `T` then `query[Option[T]]` should also work for the same type `T`.

987Nabil:
Fix is already on main

PR #3556, opened 2025-06-17
Fix: Support Request.query[Option[T]] for Transformed Schemas (transformOrFail)

Brijeshthummar02:
/claim #3540

This PR fixes a bug where Request.query[Option[T]] would fail if T is a value class or uses a transformed schema (e.g., via transformOrFail). The schema matching logic now recursively unwraps Optional and Transform schemas, enabling correct decoding of optional query parameters for value classes and validated types. Includes a test to verify the fix.


987Nabil:
This is already fixed on main

PR #3557, opened 2025-06-17
feat(ssl): add support and tests for ClientSSLCertConfig.FromClientCertContent

pjendrusik:
feat(ssl): add support and tests for ClientSSLCertConfig.FromClientCertContent

- Introduced FromClientCertContent to allow passing cert/key as strings
- Added tests for successful SSL handshake using certs loaded from resources
- Added test for invalid cert/key handling via env stream input

Passing the cert and key via environment variables lets you configure ClientSSLCertConfig.FromClientCertContent without writing files.


987Nabil:
@pjendrusik please check the build errors

Issue #3168, opened 2024-09-25
`JsonSchema.fromZSchema` includes refs when `refType = SchemaStyle.Inline`


guersam:
**Describe the bug**
`JsonSchema.fromZSchema` includes refs when `refType = SchemaStyle.Inline`.

**To Reproduce**
Steps to reproduce the behaviour:
1. Define nested case classes with Schema
2. Convert the outer one's schema using `JsonSchema.fromZSchema` with `refType = SchemaStyle.Inline`
3. Serialize the generated JsonSchema into a JSON string

[scastie](https://scastie.scala-lang.org/guersam/coXzfIumR1qgp4DOpc95aw/14)

**Expected behaviour**

All properties are inlined i.e. there are no `$ref`

**Actual behavior**

Nested object is still referenced by name

```json
{
  "type" : 
    "object",
  "properties" : {
    "bar" : {
      "$ref" : "#/components/schemas/Bar"
    }
  },
  "required" : [
    "bar"
  ]
}
```

**Additional context**

It's needed for [structured output](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/structured-outputs?tabs=rest) when using OpenAI.

guersam:
I tried simply replacing the hardcoded `refType` arguments in `JsonSchema.fromZSchema` with the outer `refType`, but it causes another bug that flattens nested case classes in `properties` field.

PR #3559, opened 2025-06-19
Add `JsonSchema.fromZSchemaInlineDeepOrFail` (#3168)

guersam:
Also reorganized JsonSchema tests in a dedicated suite with more cases verifying the same behavior with the other schema styles.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -1577,9 +1577,9 @@ object JsonSchema {
   }
 
   /**
-   * Builds a fully inlined `JsonSchema`, eliminating all `\$ref` occurrences. If a
-   * recursive definition is encountered, the process short-circuits and returns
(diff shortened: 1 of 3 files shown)


guersam:
Despite the current regressions, there's a problem with this approach; cannot support recursive structures.

Inline schema is necessary in the use cases like OpenAI client lacking support for `#ref`. It would be better to have a fallible version of `fromZSchema` returning `Left` when a recursion is detected.

@987Nabil what do you think?

guersam:
Updated with the alternative approach mentioned above.

987Nabil (approved):

guersam:
@987Nabil  I just discovered [OpenAI structured output](https://platform.openai.com/docs/guides/structured-outputs?api-mode=responses#recursive-schemas-are-supported) now supports recursive schemas with `$ref`!  Probably I need to revert the PR not to pollute the public API. Sorry for the insufficient research.

987Nabil:
Hey @guersam I think this is not in a release yet or? Then just revert it.

guersam:
No it isn't. Opened a reverting PR #3584 .

PR #3569, opened 2025-06-29
Custom Cookie encoding (#3526)

### LINKED ISSUE #3526: Replace CookieEncoding with own impl

987Nabil:
fixes #3526
/claim #3526

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Body.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Body.scala
@@ -656,8 +656,6 @@ object Body {
     override val contentType: Option[Body.ContentType] = None,
   ) extends UnsafeBytes {
 
-    var bytes: Array[Byte] = null
-
     override def asArray(implicit trace: Trace): Task[Array[Byte]] =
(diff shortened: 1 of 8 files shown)


zio-http/shared/src/main/scala/zio/http/Body.scala
```diff
@@ -656,7 +656,7 @@ object Body {
     override val contentType: Option[Body.ContentType] = None,
   ) extends UnsafeBytes {
 
-    var bytes: Array[Byte] = null
+    private var bytes: Array[Byte] = null
```
kyri-petrou: We should keep either the `data: String` or `bytes: Array[Byte]` as a class field to reduce memory footprint. I don't really see a benefit for keeping both of them as we're increasing the memory footprint of `StringBody` and we keep reference on both the `String` and the `Array[Byte]` which increases GC pressure.

987Nabil: I kept the String. When we log, it is better. When we convert to byte to send over the wire, I doubt it makes a difference when we do it.

Issue #3188, opened 2024-10-10
`text/plain` response body encoding silent failure


lemony312:
**Describe the bug**
I am using the Endpoint definition to accept some body and return a case class.
I defined the schema and some json encoders.
If a user makes a request with `Accept: plain/text` the route will return a 500 with no body.
(I had to copy the endpoint code and add log statements to see what the actual error was and it showed that is TextBinary encoder failing - see screenshot)

**To Reproduce**
Steps to reproduce the behaviour:
```scala
  case class MyItem(id: String, name: String)
  object MyItem {
    implicit val schema: zio.schema.Schema[MyItem] = zio.schema.DeriveSchema.gen

    implicit val encoder: JsonEncoder[MyItem] = JsonCodec.jsonEncoder(schema)
    implicit val decoder: JsonDecoder[MyItem] = JsonCodec.jsonDecoder(schema)
  }

  val getEndpoint =
    Endpoint(
      RoutePattern.GET / "api" / "item" / PathCodec.string("itemId") ?? Doc.p("Get Item")
    ).out[MyItem]


  val getItemHandler = handler { (itemId: String) =>
    ZIO.succeed(MyItem(itemId, "name"))
  }
  val getItemRoute = getEndpoint.implementHandler(getItemHandler)

  val routes = Routes(getItemRoute)
  Server.serve(routes)
```

`$ curl localhost:8080/api/item/id -H 'accept:text/plain'`

This will return a 500 with no body (and NO LOGS) on the server.

if you explicity set text plain 
```scala
    implicit val codec: zio.json.JsonCodec[MyItem] = JsonCodec.jsonCodec(schema)
    implicit val zioHttpContentCodec: HttpContentCodec[MyItem] = {
      val supportPlainText =
        HttpContentCodec(
          ListMap(
            MediaType.text.plain ->
              BinaryCodecWithSchema(
                zio.schema.codec.JsonCodec.zioJsonBinaryCodec[MyItem],
                schema
              )
          )
        )
      HttpContentCodec.fromSchema ++ supportPlainText
    }
```
then you can curl this way

as a side note, the client sent a list of accept content including application/json
Yet the server decided to pick the first from the list, fail silently, and not retry another content type.

Another note - if you define the route explicitly (not using endpoint) - it will not fail! 
**Expected behaviour**
better: Either a server log or some information 
best: Support text/plain automatically 

**Screenshots**
[...]

PR #3571, opened 2025-06-29
fix: silent encoding failures

brndt:
/claim #3188


zio-http/shared/src/main/scala/zio/http/codec/internal/EncoderDecoder.scala
```diff
@@ -80,29 +80,20 @@ private[codec] object EncoderDecoder {
...
-
-        try {
-          encoded = current.encodeWith(config, value, outputTypes)(f)
-
-          i = singles.length // break
-        } catch {
-          case error: HttpCodecError =>
-            // TODO: Aggregate all errors in disjunction:
-            lastError = error
+    ): IO[Throwable, Z] = {
```
987Nabil: We can't change public method signatures. Also, throwing instead of using IO is here probably the bette option. The costs of ZIO is probably too hight for this. Either might have been fine.

zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala
```diff
@@ -347,12 +364,26 @@ final case class Endpoint[PathInput, Input, Err, Output, Auth <: AuthType](
             original(value)
               .asInstanceOf[ZIO[Env, Err, Output]]
               .foldZIO(
-                success = output => Exit.succeed(endpoint.output.encodeResponse(output, outputMediaTypes, config)),
-                failure = error => Exit.succeed(endpoint.error.encodeResponse(error, outputMediaTypes, config)),
+                success = output =>
+                  ZIO
+                    .attempt(endpoint.output.encodeResponse(output, outputMediaTypes, config))
+                    .logError
```
brndt: I'm not sure if it's a good way of logging, I can't find the proper way of doing it.

987Nabil: Since this is internals, we should keep this not beautiful but efficient. And I think that means not using ZIO attempt but try catch instead. @kyri-petrou would you agree?

987Nabil: And then we can log like ZIO.logError(..) *> effect

brndt: I'll try your approach to see how it looks

brndt: done

987Nabil (approved):

987Nabil:
Tanks @brndt !

PR #3596, opened 2025-07-27
Optimize `ServerSentEvent::encode` code && put the `event:` part first

guizmaii:
The `event:` part doesn't need, per spec, to be first but it seems to be a common practice to put it first

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/ServerSentEventSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ServerSentEventSpec.scala
@@ -144,15 +144,10 @@ object ServerSentEventSpec extends ZIOHttpSpec {
       test("empty string data") {
         val event  = ServerSentEvent(data = "")
         val result = event.encode
-        assertTrue(
-          result ==
-            """data:
-              |
-              |""".stripMargin,
-        )
+        assertTrue(result == "data: \n\n")
       },
       test("eventType with carriage returns gets flattened") {
-        val event  = ServerSentEvent(data = "data", eventType = Some("message\r\nwith\rreturns"))
+        val event  = ServerSentEvent(data = "data", eventType = Some("message\nwith\nreturns"))
         val result = event.encode
         assertTrue(
           result ==
--- a/zio-http/jvm/src/test/scala/zio/http/ServerSentEventSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/ServerSentEventSpec.scala
@@ -146,6 +146,11 @@ object ServerSentEventSpec extends ZIOHttpSpec {
         val result = event.encode
         assertTrue(result == "data: \n\n")
       },
+      test("empty string data with eventType") {
+        val event  = ServerSentEvent(data = "", eventType = Some("message"))
+        val result = event.encode
+        assertTrue(result == "event: message\ndata: \n\n")
+      },
       test("eventType with carriage returns gets flattened") {
         val event  = ServerSentEvent(data = "data", eventType = Some("message\nwith\nreturns"))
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/test/scala/zio/http/ServerSentEventSpec.scala
```diff
@@ -39,8 +44,124 @@ object ServerSentEventSpec extends ZIOHttpSpec {
...
+            """data: data
+              |id: id with newlines
+              |
+              |""".stripMargin,
+        )
+      },
+      test("empty string data") {
+        val event  = ServerSentEvent(data = "")
+        val result = event.encode
+        assertTrue(result == "data: \n\n")
```
guizmaii, line 147: This PR changes how the empty `data` SSE event is sent. 
Before, it was sending an empty string while now, it sends this `data: ` empty event.
AFAIU the spec (https://html.spec.whatwg.org/multipage/server-sent-events.html), that's what is expected (but I can be wrong. The spec is a bit obscure)

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -48,15 +48,48 @@ final case class ServerSentEvent[T](
...
+
+    val dataLines: Array[String] = dataString.split("\n")
+
+    val initialCapacity: Int =
+      (
+        (6 + dataString.length + dataLines.length) // 6 for "data: ", the data itself, and the newlines
+          + eventType.fold(0)(_ => 24) // 24 because 7 for "event: ", 1 for the newline, 16 for the event type itself
+          + id.fold(0)(_ => 21)        // 21 because 4 for "id: ", 1 for the newline, 16 for the id itself
+          + retry.fold(0)(_ => 24)     // 24 because 7 for "retry: ", 1 for the newline, 16 for the retry value
+          + 1                          // for the final newline
```
guizmaii: @plokhotnyuk, does my calculation make sense here?

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -125,13 +156,7 @@ object ServerSentEvent {
               event.copy(eventType = Some(line.replaceFirst("event: ?", "")).filter(_.nonEmpty))
             case Some("event")  => event.copy(eventType = None)
             case Some("retry:") =>
-              event.copy(retry =
-                Some(line.replaceFirst("retry: ?", ""))
-                  .filter(_.nonEmpty)
```
guizmaii, line 130: `.toIntOption` already check if the `String` is empty

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
plokhotnyuk: Folds will allocate for lambdas and introduce boxing/unboxing

guizmaii: Fixed. Thanks

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -125,13 +167,8 @@ object ServerSentEvent {
...
-                  .flatMap(retry => Try(retry.toInt).toOption)
-                  .filter(_ >= 0)
-                  .map(_.milliseconds),
-              )
+              import scala.collection.compat._
```
guizmaii, line 164: Needed for Scala 2.12...

987Nabil (approved):

PR #3614, opened 2025-08-10
Fixed the auth header missing error to 401 from 400

### LINKED ISSUE #3235: Missing Authorization header is reported as 400 instead of 401

Skyiesac:
This PR fixes a bug where missing Authorization headers in declarative endpoint syntax were incorrectly returning HTTP 400 (Bad Request) instead of the RFC-compliant HTTP 401 (Unauthorized) status code.

Closes #3235
/claim #3235

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/MissingAuthHeaderSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/MissingAuthHeaderSpec.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2021 - 2023 Sporta Technologies PVT LTD & the ZIO HTTP contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Skyiesac:
Hey @jdegoes  @987Nabil  @kyri-petrou  @vigoo ,
Could you review this PR when you get a chance? If everything looks good, kindly merge it.

987Nabil:

```scala
```

PR #3616, opened 2025-08-10
Better templating

### LINKED ISSUE #3613: Improve dom representing structure
Currently a element can not be accessed for attributes after creation. Or can have attributes added after creation. 
Assigning an attribute a value returns a `Html` but often `Dom` is needed. I have yet to see any benefit of two such generic types at all. Probably it should be reduced to one. Elements and attributes should have a more specific type and a related interface instead of just being `Html`/`Dom`

987Nabil:
fixes #3613


zio-http/shared/src/main/scala/zio/http/template2/package.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  implicit def stringToModifier(s: String): Modifier = Text(s)
+  implicit def optToModifier(opt: Option[Modifier]): Modifier = opt.getOrElse(Empty)
+  implicit def itrToModifier(seq: Iterable[Dom]): Modifier = Fragment(seq)
+
+}
+
+trait LowPriorityTemplateImplicits {
+  implicit def stringToDom(s: String): Dom = Text(s)
+  implicit def optStrToDom(opt: Option[String]): Dom = opt.map(Text).getOrElse(Empty)
```
guizmaii:
```suggestion
  implicit def optStrToDom(opt: Option[String]): Dom = opt.fold(Empty)(Text.apply)
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+}
+
+private[template2] object RenderState {
+  case object NoIndentation extends RenderState {
+    def separator: String = ""
+    def indent: RenderState = this
+  }
+
+  final case class WithIndentation(level: Int, spaces: Int) extends RenderState {
+    def separator: String = "\n" + (" " * (level * spaces))
```
guizmaii:
```suggestion
    val separator: String = "\n" + (" " * (level * spaces))
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+            val textContent = if (escapeContent) singleText.renderInternal(state) else singleText.content
+            s"<$tag$attributeString>$textContent</$tag>"
+          case _ =>
+            val childrenHtml = children.map(
+              child =>
+                if (escapeContent) child.renderInternal(childState)
+                else child.renderInternalRaw(childState)
+              ).mkString(childState.separator)
+            s"<$tag$attributeString>${childState.separator}$childrenHtml${state.separator}</$tag>"
+        }
```
guizmaii:
```suggestion
        // Handle single text child specially (no extra newlines)
        children match {
          case Vector(singleText: Text) =>
            val textContent = if (escapeContent) singleText.renderInternal(state) else singleText.content
            s"<$tag$attributeString>$textContent</$tag>"
          case _ =>
            val childState = state.indent
            val childrenHtml = children.map(
              child =>
                if (escapeContent) child.renderInternal(childState)
                else child.renderInternalRaw(childState)
              ).mkString(childState.separator)
            s"<$tag$attributeString>${childState.separator}$childrenHtml${state.separator}</$tag>"
        }
```

zio-http/shared/src/main/scala/zio/http/template2/HtmlElements.scala
```diff
@@ -0,0 +1,258 @@
...
+  val u: Dom.Element = Dom.element("u")
+  val ul: Dom.Element = Dom.element("ul")
+  val `var`: Dom.Element = Dom.element("var")
+  val video: Dom.Element = Dom.element("video")
+  val wbr: Dom.Element = Dom.element("wbr")
+
+  // Convenience methods for common patterns
+  def text(content: String): Dom.Text = Dom.text(content)
+  def raw(content: String): Dom.RawHtml = Dom.raw(content)
+  val empty: Dom = Dom.empty
```
guizmaii:
```suggestion
  def empty: Dom = Dom.empty
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+    def indent: RenderState = this
+  }
+
+  final case class WithIndentation(level: Int, spaces: Int) extends RenderState {
+    def separator: String = "\n" + (" " * (level * spaces))
+    def indent: RenderState = WithIndentation(level + 1, spaces)
+  }
+
+  def noIndentation: RenderState = NoIndentation
+  def withIndentation: RenderState = WithIndentation(0, 2)
```
guizmaii:
```suggestion
  val withIndentation: RenderState = WithIndentation(0, 2)
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+
+  // Attribute value types
+  sealed trait AttributeValue extends Product with Serializable
+  object AttributeValue {
+    final case class StringValue(value: String) extends AttributeValue {
+      override def toString: String = value
+    }
+    final case class BooleanValue(value: Boolean) extends AttributeValue {
+      override def toString: String = value.toString
+    }
```
guizmaii: What about having:
```scala
sealed trait BooleanValue extends AttributeValue
object BooleanValue {
  // Could potentially be `private[template2]`
  case object True extends BooleanValue {
    override def value: Boolean = true
    override def toString: String = "true"
  }
  // Could potentially be `private[template2]`
  case object False extends BooleanValue {
    override def value: Boolean = false
    override def toString: String = "false"
  } 
}
```

then in `BooleanAttribute`, you'd have:
```scala
  final case class BooleanAttribute(name: String, enabled: Boolean = true) extends Attribute {
    def value: AttributeValue =  if (enabled) AttributeValue.BooleanValue.True else AttributeValue.BooleanValue.False
  }
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+  }
+
+  // Attribute value types
+  sealed trait AttributeValue extends Product with Serializable
+  object AttributeValue {
+    final case class StringValue(value: String) extends AttributeValue {
+      override def toString: String = value
+    }
+    final case class BooleanValue(value: Boolean) extends AttributeValue {
+      override def toString: String = value.toString
```
guizmaii: Avoids need for boxing
```suggestion
      override def toString: String = String.valueOf(value)
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    implicit def fromString(value: String): AttributeValue = StringValue(value)
+    implicit def fromBoolean(value: Boolean): AttributeValue = BooleanValue(value)
+    implicit def fromInt(value: Int): AttributeValue = StringValue(value.toString)
```
guizmaii:
```suggestion
    implicit def fromInt(value: Int): AttributeValue = StringValue(String.valueOf(value))
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+    implicit def fromDouble(value: Double): AttributeValue = StringValue(value.toString)
```
guizmaii:
```suggestion
    implicit def fromDouble(value: Double): AttributeValue = StringValue(String.valueOf(value))
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,476 @@
...
+  }
+
+  // Factory methods
+  def element(tag: String): Element.Generic = Element.Generic(tag)
+  def script: Element.Script = Element.Script()
+  def style: Element.Style = Element.Style()
```
guizmaii:
```suggestion
  val script: Element.Script = Element.Script()
  val style: Element.Style = Element.Style()
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
987Nabil: That is not correct. This object means no indentation should be used. So the impl. is correct.

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,509 @@
...
+    sealed trait BooleanValue extends AttributeValue with Serializable {
+      val value: Boolean
+    }
+
+    final case class StringValue(value: String) extends AttributeValue {
+      override def toString: String = value
+    }
+
+    object BooleanValue {
+      def apply(value: Boolean): BooleanValue = if (value) True else False
```
guizmaii, line 477: Nice!

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,509 @@
...
+      private[template2] case object True  extends BooleanValue {
+        override val toString: String = "true"
```
guizmaii: As it's a constant
```suggestion
        override def toString: String = "true"
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,509 @@
...
+        val value: Boolean            = true
```
guizmaii:
```suggestion
        def value: Boolean            = true
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -0,0 +1,509 @@
...
+      }
+      private[template2] case object False extends BooleanValue {
+        override val toString: String = "false"
+        val value: Boolean            = false
```
guizmaii:
```suggestion
        override def toString: String = "false"
        def value: Boolean            = false
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
guizmaii: I'm not sure to understand why it's not correct? 🤔

987Nabil: Sorry, commented the wrong comment

PR #3619, opened 2025-08-12
Add support for a custom fallback handler for TestClient

merijn:
This simplifies detecting/handling accesses of missing routes during tests.

### DIFF
--- a/zio-http-testkit/src/main/scala/zio/http/TestClient.scala
+++ b/zio-http-testkit/src/main/scala/zio/http/TestClient.scala
@@ -12,6 +12,7 @@ import zio._
 final case class TestClient(
   behavior: Ref[Routes[Any, Response]],
   serverSocketBehavior: Ref[WebSocketApp[Any]],
+  missingRouteHandler: Ref[Handler[Any, Response, Request, Response]],
 ) extends ZClient.Driver[Any, Scope, Throwable] {
 
   /**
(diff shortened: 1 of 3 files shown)


zio-http-testkit/src/main/scala/zio/http/TestClient.scala
```diff
@@ -12,6 +12,7 @@ import zio._
 final case class TestClient(
   behavior: Ref[Routes[Any, Response]],
   serverSocketBehavior: Ref[WebSocketApp[Any]],
+  missingRouteBehavior: Ref[Option[Routes[Any, Response]]],
```
987Nabil: This type is too powerful. There should never be multiple routes, but only one. And not any one, but only one with the route pattern `Method.ANY / trailing`. I think this should be a Ref[Handler[...]]. With a default value that logs a warning but returns a 200.

987Nabil (changes requested):

zio-http-testkit/src/main/scala/zio/http/TestClient.scala
merijn: I went to quickly implement this, but ran into where existing tests that defined their own catch-all route break, since two routes for the same is an error and it ends up picking the last one (i.e. the default fallback).

So that leaves two options:
 1) Add the default route, fix the tests, and require users to rewrite their own catch all routes via `setFallbackHandler` (this should be pretty trivial)
 2) make it `Ref[Option[Handler[...]]`, similar to the current version to avoid breaking any existing tests

I have no strong opinions either way, so I'm not sure what you'd prefer? @987Nabil

987Nabil: @merijn This should only be a matter of the order of adding routes. Routes on the RHS should override the LHS, so if you put the default catch all to the left, old tests should work as before.

merijn: @987Nabil ok, should be fixed now. I decided to use NotFound instead of 200 for the default result value, as that's the current default for unspecified routes.

987Nabil (approved):

PR #3653, opened 2025-08-23
Add helper functions for creating Handler/Routes

Kalin-Rudnicki:


Kalin-Rudnicki:
derp, I didn't see `Handler.scoped`

PR #3663, opened 2025-08-29
Bearer Token: JWT Authentication Guide

khajavi:
## Demo Preview

image


987Nabil:
@khajavi I found some concerning code/behavior of jwt-scala. And I think we have to replace this. Could you make sure, that you use it in examples only in a way that docs would not break if we remove the dependency?

khajavi:
@987Nabil 

> I found some concerning code/behavior of jwt-scala.

Can you please clarify what you mean?

> And I think we have to replace this. Could you make sure, that you use it in examples only in a way that docs would not break if we remove the dependency?

I've to use mdoc to enforce that.

khajavi:
@jdegoes 
This PR is ready for your review!

It contains two articles
1. Part One: JWT Authentication
2. Part Two: JWT Authentication with Refresh Tokens

jdegoes (approved):

Issue #3639, opened 2025-08-18
Unexpected Route Ordering


dylanowen:
**Describe the bug**

While the solution for in #3337 makes sense for newly added routes it leads to confusing behavior when adding a collection of routes. Now non duplicate routes also take precedence in reverse order.

**To Reproduce**

```scala
val example = Routes(
    Method.GET / string("test") / "expectedA" -> handler { (test: String, req: Request) =>
      Response.text(s"expectedA: $test")
    },
    Method.GET / trailing -> handler { (path: Path, req: Request) =>
      Response.text(s"Remainder: $path")
    },
  )
```

`GET http://localhost:4000/value/expectedA` returns `Remainder: /value/expectedA` when I would expect the `expectedA` route to take precedence.

If we reverse the definition we get the expected behavior:

```scala
val example = Routes(Chunk(
    Method.GET / string("test") / "expectedA" -> handler { (test: String, req: Request) =>
      Response.text(s"expectedA: $test")
    },
    Method.GET / trailing -> handler { (path: Path, req: Request) =>
      Response.text(s"Remainder: $path")
    }).reverse
  )
```

**Additional context**

The cause can be found here:

https://github.com/zio/zio-http/blob/d8917d5db5037ed8348ec16195e014920aa41065/zio-http/shared/src/main/scala/zio/http/Routes.scala#L324-L331

We reverse the routes before checking for duplicates.

I believe this is also unexpected by other parts of the ZIO Http library. When the cors middleware is applied, we put the `OPTIONS` route at the front:

https://github.com/zio/zio-http/blob/d8917d5db5037ed8348ec16195e014920aa41065/zio-http/shared/src/main/scala/zio/http/Middleware.scala#L174-L177

The solution might be as easy as just reversing the duplicate routes instead of all of them.

987Nabil:
The issue here is something else. The order should not matter for the trailing route. Trailing should always be less specific.
To start with routes in reverse is intentional. Since the most recent and therefore dominating route should be the one added last.

Harshvardhan-91:
I think the correct fix is to modify the uniqueRoutes method to preserve the original order for unique routes while still allowing the last duplicate to win

PR #3664, opened 2025-08-29
feat: enhance route handling with precedence and uniqueness checks

Harshvardhan-91:
#3639


987Nabil:
Broken build, wrong fix

PR #3667, opened 2025-08-31
Add template2 migration bridge (implicit conversions)

### LINKED ISSUE #3611: Templating improvement
The templating is very rudimentary. We should improve it quite a bit.
If this is not possible in a backwards compatible way, we don't need to wait for 4.x but could add a new package like `template2` that tries to keep source compatibility, so it can be a drop in replacement for user code. And for places where we use it internally we could just use implicit conversions.

notcodesid:
Fixes: #3611

### Summary

Introduce an additive package ``zio.http.template2`` that supplies implicit conversions so existing entry points—``Response.html`` and ``Handler.html``, seamlessly accept ``template2`` DOM values. This enables incremental adoption of the improved templating API without breaking existing code.

### Motivation

Per #3611, fully retrofitting the templating API can be tricky. The proposed path is to add a new package (``template2``) and provide implicit conversions so current call sites keep working while users opt into the new DOM API. This PR implements that approach.

#### New: ``zio/http/template2/package.scala``

Provides implicits bridging template2 DOM → legacy ``template.Html`` so ``Response.html(...) / Handler.html(...)`` continue to work unchanged when ``import zio.http.template2._ `` is in scope.

#### Docs:
Brief Scaladoc notes in ``Handler.scala`` and ``Response.scala`` indicating that ``template2`` DOM values are accepted via the above implicits when imported.

#### No example/demo changes.
This PR intentionally keeps changes minimal and text-only. (If maintainers prefer, example updates can follow in a separate PR.)

### Usage
```
import zio.http._
import zio.http.template2._

// build your DOM using the new template2 API
val dom = /* template2 DOM value */

// Works with existing APIs:
val res: Response = Response.html(dom)
val h: Handler[Any, Nothing, Request, Response] = Handler.html(dom)
```

### Backward Compatibility
- Source-compatible: Legacy ``zio.http.template._`` remains fully supported.
- No runtime behavior changes: The bridge only provides conversions; it does not alter semantics.

### Testing
- Compiles end-to-end with ``sbt clean compile``.
- A minimal spec verifies that:

``Response.html`` and ``Handler.html`` accept ``template2`` DOM values with only ``import zio.http.template2._`` in scope.
- Legacy ``template._ `` usages continue to type-check.

### DIFF
--- a/zio-http-example/src/main/scala/example/HtmlTemplating.scala
+++ b/zio-http-example/src/main/scala/example/HtmlTemplating.scala
@@ -3,51 +3,126 @@
 package example
 
 import zio._
-
 import zio.http._
+import zio.http.template2._
 
 object HtmlTemplating extends ZIOAppDefault {
-  // Importing everything from `zio.html`
-  import zio.http.template._
 
-  def routes: Routes[Any, Response] = {
-    // Html response takes in a `Html` instance.
-    Handler.html {
+  // ===== UPDATED TO USE TEMPLATE2 (Issue #3611 Resolution) =====
+  // The template system has been improved from rudimentary to powerful!
+  // This example now demonstrates template2 features with compile-time validation
 
-      // Support for default Html tags
+  // Basic CSS styles (template2 supports advanced interpolation too)
+  val pageStyles = """
+    .container {
+      max-width: 800px;
+      margin: 2rem auto;
+      padding: 2rem;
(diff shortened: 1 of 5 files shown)


notcodesid:
@987Nabil hey, this is ready for a first pass. Adds zio.http.template2 with implicit conversions; no runtime changes. Happy to trim examples or add a minimal spec per your preference. 🙏

PR #3674, opened 2025-09-04
Polish template2 API (#3611)

### LINKED ISSUE #3611: Templating improvement

987Nabil:
fixes #3611
/claim #3611


zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -154,6 +154,22 @@ object Dom {
 
   def boolAttr(name: String, enabled: Boolean = true): BooleanAttribute = BooleanAttribute(name, enabled)
 
+  def multiAttr(name: String): PartialMultiAttribute =
+    PartialMultiAttribute(name, AttributeSeparator.Space)
```
guizmaii, line 158: Reuse the other `multiAttr` method maybe?

```suggestion
    multiAttr(name, AttributeSeparator.Space)
```

zio-http/shared/src/main/scala/zio/http/template2/Dom.scala
```diff
@@ -154,6 +154,22 @@ object Dom {
...
+
+  def multiAttr(
+    name: String,
+    values: Iterable[String],
+    separator: AttributeSeparator = AttributeSeparator.Space,
+  ): CompleteAttribute =
+    CompleteAttribute(name, AttributeValue.MultiValue(values.toVector, separator))
+
+  def multiAttr(name: String, separator: AttributeSeparator, values: String*): CompleteAttribute =
+    CompleteAttribute(name, AttributeValue.MultiValue(values.toVector, separator))
```
guizmaii, line 171: Same here
```suggestion
    multiAttr(name, values.toVector, separator)
```

guizmaii (approved):
LGTM

PR #3687, opened 2025-09-10
Fix regex validation in the OpenAPI

minedeljkovic:
This PR fixes what was probably a copy/paste error.
Without this fix a negated regex validation is documented as a non-negated regex validation, while the non-negated regex validation is not documented at all.


987Nabil:
@minedeljkovic Could you please add a test

987Nabil (approved):

987Nabil (approved):

PR #3699, opened 2025-09-14
Remove back-and-forth conversion from String/CharSequence, which causes incompatibility in ScalaJS

Kalin-Rudnicki:


987Nabil:
I wrote a fix for Scala.js [here](https://github.com/scala-js/scala-js/pull/5239) but will merge this for now and reevaluate solutions after the Scala.js release

PR #3703, opened 2025-09-17
feat: SSE comments as heartbeats and make sure they're encoded correctly

### LINKED ISSUE #3686: SSE comments
According to https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#examples sending Server Sent Event comments is the preferred way to keep a connection alive (heartbeats) since comments are discarded by the clients. This makes it hard to use in a GraphQL world leveraging SSE subscriptions

It would make sense if [heartbeat](https://github.com/zio/zio-http/blob/main/zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala#L199) instead produced SSE comments since those are ignored by the clients using them.

I've interrupted the actual response with a middleware appending SSE comments as ZStream[Any, Throwable, Byte] instead to make it work today but that causes some issues when merging streams etc.

SvenW:
Fixes #3686

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
+++ b/zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
@@ -53,11 +53,12 @@ final case class ServerSentEvent[T](
       }
 
     val dataLines: Array[String] = dataString.split("\n")
+    val isComment                = dataString.startsWith(":")
 
     val initialCapacity: Int =
       (
         // 6 for "data: ", the data itself, and the newlines
-        (6 + dataString.length + dataLines.length)
+        ((if (isComment) 0 else 6) + dataString.length + dataLines.length)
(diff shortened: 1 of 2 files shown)


guizmaii (approved):
LGTM

zio-http/shared/src/main/scala/zio/http/ServerSentEvent.scala
```diff
@@ -81,7 +82,8 @@ final case class ServerSentEvent[T](
       sb.append('\n')
     }
     dataLines.foreach { line =>
-      sb.append("data: ").append(line).append('\n')
+      if (isComment) sb.append(line).append('\n')
```
987Nabil, line 85: I think this is wrong. If you do this, you would not send the comment over the wire and that was the whole point of this or not?

987Nabil (changes requested):

987Nabil (approved):

PR #3708, opened 2025-09-24
Add `fileName` to OpenAPI.Info

atyagi-demyst:
Adding a separate `fileName` field to `OpenAPI.Info` to specify the JSON file name; if it is not provided, the `OpenAPI.Info.title` can be used instead.


987Nabil:
Please open an issue or write in discord what your use case is. These changes are neither bin compatible nor do they fit in the OpenAPI data model, which reflects the json structure of the spec.

guizmaii:
@987Nabil I work with @atyagi-demyst. Basically, we need to be able to configure the name of the generated file. Using the title doesn't work for what our company is asking us

PR #3720, opened 2025-10-01
add support for multiple CSP directives

### LINKED ISSUE #3719: support for multiple Content-Security-Policy directives via typed headers
The current implementation of the `ContentSecurityPolicy` header only supports a single directive.
This header typically includes multiple directives.

I'd like for the CSP header to support multiple directives.

There are currently two workarounds;
1. Using `Header.Custom` and providing the full CSP as the value
2. Providing multiple CSP headers, each containing a single directive; however, [this is tricky and comes with gotchas](https://content-security-policy.com/examples/multiple-csp-headers/)

**Additional context**
[Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP)
[CSP Quick Reference Guide](https://content-security-policy.com/)

dontgitit:
resolves https://github.com/zio/zio-http/issues/3719

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Header.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Header.scala
@@ -2224,483 +2224,503 @@ object Header {
   }
 
   // scalafmt: { maxColumn = 180 }
-  sealed trait ContentSecurityPolicy extends Header {
+  final case class ContentSecurityPolicy(directives: List[ContentSecurityPolicy.Directive]) extends Header {
     override type Self = ContentSecurityPolicy
     override def self: Self                                          = this
     override def headerType: HeaderType.Typed[ContentSecurityPolicy] = ContentSecurityPolicy
   }
 
-  // TODO: Should we make deprecated types deprecated in code?
   object ContentSecurityPolicy extends HeaderType {
     override type HeaderValue = ContentSecurityPolicy
 
     override def name: String = "content-security-policy"
 
-    final case class SourcePolicy(srcType: SourcePolicyType, src: Source) extends ContentSecurityPolicy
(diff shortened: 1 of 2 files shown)


zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -2224,483 +2225,503 @@ object Header {
...
-        override def &&(other: Source): Source = other
+      sealed trait Source {
+        self =>
+        def &&(other: Source): Source =
+          if (other == Source.none) self else Source.Sequence(self, other)
       }
 
-      final case class Host(uri: URI) extends Source
+      object Source {
+        case object none extends Source {
```
dontgitit, line 2383: I wonder if while we're at it, this should be changed to `None` to be consistent with all the others?

987Nabil: We can't change this. That would be a breaking change

zio-http/shared/src/main/scala/zio/http/Header.scala
```diff
@@ -2224,483 +2224,503 @@ object Header {
   }
 
   // scalafmt: { maxColumn = 180 }
-  sealed trait ContentSecurityPolicy extends Header {
+  final case class ContentSecurityPolicy(directives: List[ContentSecurityPolicy.Directive]) extends Header {
```
987Nabil, line 2227: This change is not binary compatible. We can't do that.

987Nabil (changes requested):

zio-http/shared/src/main/scala/zio/http/Header.scala
dontgitit: I'm not sure how to implement this change without breaking binary compatibility...

dontgitit:
implemented [here](https://github.com/zio/zio-http/pull/4008)

PR #3727, opened 2025-10-06
Added ZClient.requiringConfig, which has consistent types across platforms

Kalin-Rudnicki:
The way `ZClientPlatformSpecific` defines its layers, with layers of the same names but different types on different platforms, is a bit unfortunate for a consuming library which is also cross platform, and does not want to deal with platform-specific things within zio-http.

The only layer which has a consistent signature is `default`.

There is no consistent layer across platforms which has the signature `ZLayer[ZClient.Config, Throwable, Client]`. This MR adds such a layer builder.

### DIFF
--- a/zio-http/js/src/main/scala/zio/http/ZClientPlatformSpecific.scala
+++ b/zio-http/js/src/main/scala/zio/http/ZClientPlatformSpecific.scala
@@ -7,6 +7,9 @@ import zio.http.internal.FetchDriver
 
 trait ZClientPlatformSpecific {
 
+  def requiringConfig: ZLayer[ZClient.Config, Throwable, Client] =
+    live
+
   def customized: ZLayer[Config with ZClient.Driver[Any, Scope, Throwable], Throwable, Client] = {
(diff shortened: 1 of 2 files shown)


zio-http/js/src/main/scala/zio/http/ZClientPlatformSpecific.scala
```diff
@@ -7,6 +7,9 @@ import zio.http.internal.FetchDriver
 
 trait ZClientPlatformSpecific {
 
+  def requiringConfig: ZLayer[ZClient.Config, Throwable, Client] =
```
Kalin-Rudnicki, line 10: I would have ideally like to have called this something like `customized` or `configured`, but both of those are already taken by type-mismatching platform-specific overrides. Open to naming suggestions if something else is preferable.

987Nabil: You could do fromConfig, withConfig or even configured, since the new one has no params

PR #3753, opened 2025-10-17
Redirect Netlify Domain to The ziohttp.com Domain

khajavi:
Currently, Google directs the non-primary domain of the ZIO HTTP project: zio-http.netlify.app, so we need to instruct and redirect it to the primary one: ziohttp.com.

image


khajavi:
@jdegoes This is ready for your review
I think the failing CI is not caused by these changes.

jdegoes (approved):

Issue #709, opened 2021-12-28
Support Http Range header on request for Files


ashprakasan:
**Is your feature request related to a problem? Please describe.**

The Range HTTP request header indicates the part of a document that the server should return. Several parts can be requested with one Range header at once, and the server may send back these ranges in a multipart document.

**Describe the solution you'd like**

Example requesting 3 ranges from files -
`Range: bytes=200-1000, 2000-6576, 19000-`

Send only those parts of the document in Response.

**Describe alternatives you've considered**

Currently, the range is hardcoded as follows - 
` ctx.write(new DefaultFileRegion(raf.getChannel, 0, fileLength))`
Manipulate the positions as per request headers instead.

**Additional context**

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range

BenraouaneSoufiane:
@jdegoes Does all zio bounties are reserved to @987Nabil?

francoo98:
/attempt #709
I'm gonna modify the Handler.scala and Body.scala to create a response from a file

ritik4ever:
/attempt #709

## Implementation Plan

I'll implement HTTP Range request support (RFC 7233) with the following approach:

1. **Create Range Parser** (`zio/http/headers/Range.scala`) - Parse `Range: bytes=` headers into typed models supporting single ranges (bytes=0-499), suffix ranges (bytes=-500), and multiple ranges (bytes=0-499,1000-1499)

2. **Create RangeSupport Handler** (`zio/http/RangeSupport.scala`) - Core logic to:
   - Detect Range header in requests
   - Validate ranges against file size
   - Return 206 Partial Content with Content-Range for single ranges
   - Return 206 with multipart/byteranges for multiple ranges
   - Return 416 Range Not Satisfiable for invalid ranges

3. **Add Header Types** - Define Range, Content-Range, and Accept-Ranges headers in the headers package

4. **Replace DefaultFileRegion** - Update existing file serving code to use RangeSupport.serveFile() instead of hardcoded `DefaultFileRegion(raf.getChannel, 0, fileLength)`

5. **Add Tests** - Comprehensive test suite covering all range types, error cases, and multipart responses

This enables features like video seeking, resume downloads, and parallel chunk fetching while maintaining backward compatibility (requests without Range headers get full file with 200 OK).

Michael-A-Kuykendall:
[...]

PR #3769, opened 2025-10-28
Add native HTTP Range request support for partial content delivery (#709)

jxnnyo:


987Nabil:
Does not build for two weeks

jxnnyo:
Why has this been closed?

PR #3831, opened 2025-12-01
JSON Schema gen

QuinnHarris:
This PR introduces a new JsonSchema.jsonSchema(schema: Schema[_]) function that generates a fully compliant JSON Schema document.

The existing JsonSchema encoder already supports nearly all the structures required for generating JSON Schema. However, OpenAPI-style output expects definitions under components/schemas, while JSON Schema requires them under the top-level "$defs" key.  This makes necessary changes to support that.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -434,8 +434,7 @@ object JsonSchema {
     def toSerializableSchema: SerializableJsonSchema = {
       root.toSerializableSchema.copy(
         schema = Some("https://json-schema.org/draft/2020-12/schema"),
-        defs = Some(children.map { case (key, schema) => key -> schema.toSerializableSchema },
-      ),
+        defs = Some(children.map { case (key, schema) => key -> schema.toSerializableSchema }),
       )
     }
 
--- a/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
+++ b/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
@@ -18,6 +18,7 @@ import zio.http.endpoint.openapi.JsonSchema.MetaData
 
 @nowarn("msg=possible missing interpolator")
 private[openapi] case class SerializableJsonSchema(
+  @fieldName("$schema") schema: Option[String] = None,
   @fieldName("$ref") ref: Option[String] = None,
   @fieldName("type") schemaType: Option[TypeOrTypes] = None,
   format: Option[String] = None,
@@ -49,7 +50,7 @@ private[openapi] case class SerializableJsonSchema(
   exclusiveMaximum: Option[Either[Boolean, Either[Double, Long]]] = None,
   uniqueItems: Option[Boolean] = None,
   minItems: Option[Int] = None,
-  @fieldName("$defs") defs: Map[String, SerializableJsonSchema] = Map.empty,
+  @fieldName("$defs") defs: Map[String, SerializableJsonSchema] = Map.empty
 ) {
   def asNullableType(nullable: Boolean): SerializableJsonSchema = {
     import SerializableJsonSchema.typeNull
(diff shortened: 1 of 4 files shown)


zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -418,10 +420,87 @@ object JsonSchema {
...
+   * @return
+   *   A [[JsonSchemaRoot]] representing a valid JSON Schema document.
+   */
+  def jsonSchema(schema: Schema[_]): JsonSchemaRoot = {
+    val schemaRef = SchemaRef.jsonSchema(SchemaStyle.Compact)
+    val s         = fromZSchemaMultiple(schema, schemaRef)
+    JsonSchemaRoot(s.root, s.children.map { case (key, schema) => key.replace(schemaRef.path, "") -> schema })
+  }
+
+  @deprecated("use fromZSchemaMultiple", "4.0")
```
987Nabil: Do we really need to deprecate the old things? Anyhow, all deprecations should be for 3.8

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -814,23 +903,30 @@ object JsonSchema {
...
-     * and uses the type name to help to generate schema names.
+     * Generates schema names using the type name
      * @see
      *   SchemaStyle.Reference for full package path schema names to avoid name
      *   collisions.
      */
     case object Compact extends SchemaStyle
   }
 
+  case class SchemaRef(path: java.lang.String, style: SchemaStyle) {
```
987Nabil: final and not path as a string. The type of path should be a sealed trait with two options. Idk about the name

987Nabil (changes requested):
I am okay with getting this in, but json schema impl will eventually move to zio-blocks.
@plokhotnyuk you might be interested in this change.

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
QuinnHarris: Created SchemaSpec similar to SchemaStyle to replace the path

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
QuinnHarris: I created the second function to maintain binary compatibility to pass mima test.  These functions seem very internal so I question if maintaining compatibility is really needed.  I marked deprecated with the idea that can be completely removed and seems possible nobody is using this directly.

987Nabil: This does not matter. They are public and will stay public until at least 4.0

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -485,17 +485,19 @@ object JsonSchema {
...
-    val s         = fromZSchemaMultiple(schema, schemaRef)
-    JsonSchemaRoot(s.root, s.children.map { case (key, schema) => key.replace(schemaRef.path, "") -> schema })
+    val s = fromZSchemaMultiple(schema, SchemaRef(SchemaSpec.JsonSchema, SchemaStyle.Compact))
+    JsonSchemaRoot(
+      s.root,
+      s.children.map { case (key, schema) => key.replace(SchemaSpec.JsonSchema.path, "") -> schema },
+    )
   }
 
   @deprecated("use fromZSchemaMultiple", "4.0")
```
987Nabil: you missed to adjust the version

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -918,13 +922,24 @@ object JsonSchema {
...
+    /**
+     * Generates schema references for an JSON Schema document
+     */
+    case object JsonSchema extends SchemaSpec { override def path: java.lang.String = "#/$defs/" }
   }
-  object SchemaRef                                                 {
-    def openApi(style: SchemaStyle)    = SchemaRef("#/components/schemas/", style)
-    def jsonSchema(style: SchemaStyle) = SchemaRef("#/$defs/", style)
+
+  case class SchemaRef(spec: SchemaSpec, style: SchemaStyle) {
```
987Nabil: final

zio-http/shared/src/main/scala/zio/http/endpoint/openapi/JsonSchema.scala
```diff
@@ -918,13 +922,24 @@ object JsonSchema {
     case object Compact extends SchemaStyle
   }
 
-  case class SchemaRef(path: java.lang.String, style: SchemaStyle) {
-    def inline: SchemaRef  = copy(style = SchemaStyle.Inline)
-    def compact: SchemaRef = copy(style = SchemaStyle.Compact)
+  sealed trait SchemaSpec extends Product with Serializable { def path: java.lang.String }
```
987Nabil: i think we can make path package private

987Nabil (changes requested):

987Nabil (approved):

987Nabil:
Sorry, I made some PR foo. Code is okay, but I will leave the PR I opened unmerged until I want to make the 3.8 release. I will probably make some bug fixes before that.

PR #3838, opened 2025-12-06
Implement HappyEyeballs for NettyConnectionPool (#1996)

### LINKED ISSUE #1996: Add support for Happy Eyeballs

987Nabil:
fixes #1996
/claim #1996

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
@@ -16,7 +16,7 @@
 
 package zio.http.netty.client
 
-import java.net.InetSocketAddress
+import java.net.{Inet6Address, InetAddress, InetSocketAddress}
 import java.util.concurrent.TimeUnit
 
 import zio._
@@ -35,6 +35,8 @@ import io.netty.handler.timeout.{ReadTimeoutException, ReadTimeoutHandler}
 private[netty] trait NettyConnectionPool extends ConnectionPool[JChannel]
 
 private[netty] object NettyConnectionPool {
+
+  private val HappyEyeballsDelay: Duration = 250.millis
 
   protected def createChannel(
     channelFactory: JChannelFactory[JChannel],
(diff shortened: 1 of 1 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+            addresses.head,
+            channelFactory,
+            eventLoopGroup,
+            location,
+            initializer,
+            connectionTimeout,
+            localAddress,
+          ).onError(_ => queue.offer(())),
+          addresses.tail.zipWithIndex.map { case (address, index) =>
+            ZIO.sleep(HappyEyeballsDelay * index.toLong).raceFirst(queue.take).ignore *>
```
987Nabil:
```suggestion
            ZIO.sleep(HappyEyeballsDelay * index.toDouble).raceFirst(queue.take).ignore *>
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -36,6 +36,8 @@ private[netty] trait NettyConnectionPool extends ConnectionPool[JChannel]
 
 private[netty] object NettyConnectionPool {
 
+  private val HappyEyeballsDelay: Duration = 250.millis
```
kyri-petrou, line 39: Maybe add a comment that this is according to spec recommendations

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+  /**
+   * Returns a sequence of connection attempts with their delays. Per RFC 8305,
+   * we start with IPv6, then after firstAddressFamilyDelay we try IPv4, then
+   * alternate between families.
+   */
+  private def sortAddresses(resolvedHosts: Chunk[InetAddress]): Chunk[InetAddress] = {
+    val (ipv6Addresses, ipv4Addresses) = resolvedHosts.partition(_.isInstanceOf[Inet6Address])
+    val ipv6Iter                       = ipv6Addresses.iterator
+    val ipv4Iter                       = ipv4Addresses.iterator
+    val builder                        = ChunkBuilder.make[InetAddress]()
```
kyri-petrou:
```suggestion
    val builder                        = ChunkBuilder.make[InetAddress](resolvedHosts.size)
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+        .handler(initializer)
+      localAddress.foreach(bootstrap.localAddress)
+
+      val channelFuture = bootstrap.connect()
+      val ch            = channelFuture.channel()
+      Scope.addFinalizer {
+        NettyFutureExecutor.executed {
+          channelFuture.cancel(true)
+          ch.close()
+        }.when(ch.isOpen).ignoreLogged
```
kyri-petrou:
```suggestion
        }.whenDiscard(ch.isOpen).ignoreLogged
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+    resolvedHosts: Chunk[InetAddress],
+    channelFactory: JChannelFactory[JChannel],
+    eventLoopGroup: JEventLoopGroup,
+    location: URL.Location.Absolute,
+    initializer: ChannelInitializer[JChannel],
+    connectionTimeout: Option[Duration],
+    localAddress: Option[InetSocketAddress],
+  )(implicit trace: Trace): ZIO[Scope, Throwable, JChannel] = {
+
+    if (resolvedHosts.isEmpty) {
```
kyri-petrou, line 201: Very minor nit (feel free to ignore) in order to make this a bit more readable, you can pat-mat on `resolvedHosts.size` instead of using if statements. e.g.,

```scala
resolvedHosts.size match {
  case 0 => ...
  case 1 => ...
  case _ => ....
}
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+        connectionTimeout,
+        localAddress,
+      )
+    } else {
+      val addresses = sortAddresses(resolvedHosts)
+      for {
+        queue   <- Queue.bounded[Unit](requestedCapacity = 1)
+        channel <- ZIO.raceAll(
+          connectToAddress(
+            addresses.head,
```
kyri-petrou, line 220: If we're going to be using `head` and `tail` it might be better for `sortAddresses` to return a `List` instead. Creating a list with a `ListBuffer` is very cheap

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+        eventLoopGroup,
+        location,
+        initializer,
+        connectionTimeout,
+        localAddress,
+      )
+    } else {
+      val addresses = sortAddresses(resolvedHosts)
+      for {
+        queue   <- Queue.bounded[Unit](requestedCapacity = 1)
```
kyri-petrou: A `Promise` seems to be a better fit for this usecase

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+      localAddress.foreach(bootstrap.localAddress)
+
+      val channelFuture = bootstrap.connect()
+      val ch            = channelFuture.channel()
+      Scope.addFinalizer {
+        NettyFutureExecutor.executed {
+          channelFuture.cancel(true)
+          ch.close()
+        }.when(ch.isOpen).ignoreLogged
+      } *> NettyFutureExecutor.executed(channelFuture).as(ch)
```
kyri-petrou: I believe this should run uninterruptibly 

```suggestion
      }.uninterruptible *> NettyFutureExecutor.executed(channelFuture).as(ch)
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,136 @@ private[netty] object NettyConnectionPool {
...
+                channelFactory,
+                eventLoopGroup,
+                location,
+                initializer,
+                connectionTimeout,
+                localAddress,
+              )
+          },
+        )
+      } yield channel
```
kyri-petrou, line 253: I feel like there is a race condition / connection leak here. In the case that two attempts connect at the same time (within the netty connection pool), we won't be closing the 2nd connection until the `Scope` is closed down. We need to guarantee that only the "winner' connection remains open by the end of this operation

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: No, I made a mistake, there is a loc missing. But there can potentially multiple errors, so it need to be fulfilled multiple times. So queue is right

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+          connectToAddress(
+            addresses.head,
+            channelFactory,
+            eventLoopGroup,
+            location,
+            initializer,
+            connectionTimeout,
+            localAddress,
+          ).onExit {
+            case e: Exit.Success[JChannel] => successful.update(channels => channels :+ e.value)
```
kyri-petrou, line 228: I don't think this will work; Interruption might occur right between the moment that we create the channel and the evaluation of the next effect; before we reach this point.

We probably need to pass the ref into `connectToAddress` itself and add the channel to the list (potentially unsafely) at some uninterruptible point

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+              ).onExit {
+                case e: Exit.Success[JChannel] => successful.update(channels => channels :+ e.value)
+                case _: Exit.Failure[_]        => lastFailed.offer(())
+              }
+          },
+        )
+        channels   <- successful.get
+        channel    <- channels.headOption match {
+          case ch: Some[JChannel] => ZIO.succeed(ch.value)
+          case None               => ZIO.fail(new RuntimeException("All connection attempts failed"))
```
kyri-petrou, line 250: Isn't it better to partition by open channels and pick the first one? And then close all other ones?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+        }
+        _          <- ZIO.foreachDiscard(channels.tail)(ch => ZIO.attempt(ch.close()))
```
kyri-petrou, line 252: There's a potential leak here again in case that closing one of them fails. I think this is better:

```suggestion
        _          <- ZIO.foreachDiscard(channels.tail)(ch => ZIO.ignore(ch.close()))
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: I am not sure what you mean. That channels might contain one that is not open? If so, how should that happen? Connection problem directly after establishing?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: Is onExit not called always? Even on interrupt?

PR #3839, opened 2025-12-06
Documentation of Template2

### LINKED ISSUE #3157: give me some examples how to use zio-http with templates engine like thymeleaf or twirl
I want to migrate my play-framework app to zio-http, thare are lots of frontend code that written with html templates.

support to integrate zio-http with thymeleaf and twirl.

Just don't know if there are alternatives

**Additional context**
I have lots of code written with `jekyll`(a static site generation engine) and `play-framework + twirl `, I want to make my sites dynamic rather than static.

khajavi:
also resolved #3157

### DIFF
--- a/docs/reference/template.md
+++ b/docs/reference/template.md
@@ -392,6 +392,8 @@ def userCard(user: User, showEmail: Boolean): Dom.Element = {
 }
 ```
 
+Please note that in the above example, the `user.avatar.map(url => img(src := url))` is an `Option[Dom]`. If the avatar is `None`, it will not render anything in that place.
+
 If we need to create a custom element that is not predefined, we can use the `Dom.element` method:
 
 ```scala mdoc:compile-only
--- a/docs/reference/template.md
+++ b/docs/reference/template.md
@@ -1,5 +1,5 @@
 ---
-id: index
+id: template
 title: "Template DSL"
 sidebar_label: Template
 ---
--- a/docs/reference/template.md
+++ b/docs/reference/template.md
@@ -583,7 +583,7 @@ object User {
 
 Now you can use this Twirl template inside a ZIO HTTP route as follows:
 
-```scala mdoc:compile-only
+```scala
 import zio.http._
 import zio.schema.codec.JsonCodec.schemaBasedBinaryCodec
 
--- a/docs/reference/template.md
+++ b/docs/reference/template.md
@@ -100,11 +100,11 @@ div(
 )
 ```
 
-There are many predefined partial attributes available for use. However, if you need to create a custom one, you can use the `custom()` helper. For example:
+There are many predefined partial attributes available for use. However, if you need to create a custom one, you can use the `attr` helper. For example:
 
 ```scala mdoc:compile-only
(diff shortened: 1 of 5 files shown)


khajavi:
@jdegoes This is ready for your review!

docs/reference/template.md
```diff
@@ -0,0 +1,602 @@
...
+case class User(name: String, email: String, avatar: Option[String])
+val showEmail = true
+```
+
+```scala mdoc:compile-only
+def userCard(user: User, showEmail: Boolean): Dom.Element = {
+  div(`class` := "user-card")(
+    h3(user.name),
+    if (showEmail) p(user.email) else Dom.empty,
+    user.avatar.map(url => img(src := url))
```
987Nabil, line 373: You should make explicit in text that this is option of Dom

docs/reference/template.md
```diff
@@ -0,0 +1,602 @@
...
+  )
+}
+```
+
+If we need to create a custom element that is not predefined, we can use the `Dom.element` method:
+
+```scala mdoc:compile-only
+Dom.element("custom-tag")(custom("x-property") := "value")("Content here")
```
987Nabil: I'd prefer Dom.attr over custom. I think it is more explicit

docs/reference/template.md
```diff
@@ -0,0 +1,602 @@
...
+// Renders as: 
+//<style type="text/css">
+//  body {
+//    margin: 0;
+//    font-family: sans-serif;
+//  }
+//</style>
+```
+
+To inline a CSS file from the resource directory, use `inlineResource`:
```
987Nabil, line 456: This also exists for script

987Nabil (changes requested):
For template2 as well as twirl an example with the endpoint API would be nice. template2 has an HttpContentCodec. That one could also be used by twirl. Like call transform and transform a twirl `Html` or `String` to a `Dom`

docs/reference/template.md
khajavi: Okay, I'll fix it.
But what about these aliases:

```scala
  def aria(name: String): Dom.PartialAttribute   = Dom.attr(s"aria-$name")
  def custom(name: String): Dom.PartialAttribute = Dom.attr(name)
  def data(name: String): Dom.PartialAttribute   = Dom.attr(s"data-$name")
```

 What are their use cases, and are they going to be deprecated??

987Nabil: Custom yes. Data no. Araia, I am not sure.

docs/reference/template.md
khajavi: It is already documented!

docs/reference/template.md
khajavi: Done!

docs/reference/template.md
khajavi: Done!

khajavi:
@987Nabil Added examples for Endpoint API.

khajavi:
@987Nabil CI is green!

PR #3844, opened 2025-12-08
Improve client timeout options (#2383)

### LINKED ISSUE #2383: Client does not handle broken servers
When a HTTP server accepts the connection, streams back a response header (or at least some bytes) but then stops before sending a body, zio-http client (`3.0.0-RC2+47-60c31e2b-SNAPSHOT`) doesn't do anything any more (dead-lock?).

This server behavior was observed in production after replacing the zio-kafka HTTP client with the JVM's HTTP client. The server we need to talk to has the described behavior roughly every 10 minutes. Since this is only correlation, not causation, it is possible that behavior of the zio-kafka client has a different cause.

To reproduce one can use an intentionally evil HTTP server. For example https://github.com/kiewic/heyhttp. Such a server is also great for unit testing.

Instead of deadlocking, I expect the client to timeout after some time and replace the connection in the pool.

987Nabil:
fixes #2383

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
@@ -17,6 +17,8 @@
 package zio.http.netty
 
 import java.io.IOException
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicReference
 
 import scala.collection.mutable
 
@@ -28,15 +30,18 @@ import zio.http.netty.NettyBody.UnsafeAsync
 import io.netty.buffer.ByteBufUtil
 import io.netty.channel.{ChannelHandlerContext, SimpleChannelInboundHandler}
 import io.netty.handler.codec.http.{HttpContent, LastHttpContent}
+import io.netty.util.concurrent.ScheduledFuture
 
-private[netty] abstract class AsyncBodyReader extends SimpleChannelInboundHandler[HttpContent](true) {
+private[netty] abstract class AsyncBodyReader(timeoutMillis: Option[Long])
+    extends SimpleChannelInboundHandler[HttpContent](true) { self =>
   import zio.http.netty.AsyncBodyReader._
 
-  private var state: State               = State.Buffering
-  private val buffer                     = new mutable.ArrayBuilder.ofByte()
-  private var previousAutoRead: Boolean  = false
-  private var readingDone: Boolean       = false
-  private var ctx: ChannelHandlerContext = _
+  private var state: State                    = State.Buffering
+  private val buffer                          = new mutable.ArrayBuilder.ofByte()
+  private var previousAutoRead: Boolean       = false
+  private var readingDone: Boolean            = false
+  private var ctx: ChannelHandlerContext      = _
+  private var timeoutTask: ScheduledFuture[_] = _
 
   private def result(buffer: mutable.ArrayBuilder.ofByte): Chunk[Byte] = {
     val arr = buffer.result()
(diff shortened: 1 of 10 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
```diff
@@ -121,12 +177,52 @@ private[netty] abstract class AsyncBodyReader extends SimpleChannelInboundHandle
...
+      // Reschedule timeout for next chunk if not the last message
+      if (readMore && !isLast) {
+        timeoutMillis.foreach { timeoutMillis =>
+          val task = ctx
+            .channel()
+            .eventLoop()
+            .schedule(
+              new Runnable {
+                override def run(): Unit = {
+                  AsyncBodyReader.this.synchronized {
```
kyri-petrou: This is a bit confusing (I've actually never seen this syntax before). Maybe just use `this.synchronized`?

zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
```diff
@@ -74,7 +117,13 @@ private[netty] abstract class AsyncBodyReader extends SimpleChannelInboundHandle
...
 
   override def handlerRemoved(ctx: ChannelHandlerContext): Unit = {
-    val _ = ctx.channel().config().setAutoRead(previousAutoRead)
+    val _           = ctx.channel().config().setAutoRead(previousAutoRead)
+    // Cancel any pending timeout task
+    val currentTask = timeoutTask.get()
+    if (currentTask != null) {
+      currentTask.cancel(false)
+      timeoutTask.set(null)
+    }
```
kyri-petrou: If you wrap this in `synchronized` you can simply use a var for `timeoutTask` instead of an `AtomicReference`.

zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
```diff
@@ -121,12 +177,52 @@ private[netty] abstract class AsyncBodyReader extends SimpleChannelInboundHandle
...
+                    }
+                  }
+                }
+              },
+              timeoutMillis,
+              TimeUnit.MILLISECONDS,
+            )
+          timeoutTask.set(task)
+        }
+      }
```
kyri-petrou: Isn't this the same code as above? Maybe extract it all into a method if that's the case?

zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
```diff
@@ -138,10 +234,27 @@ private[netty] abstract class AsyncBodyReader extends SimpleChannelInboundHandle
 
   override def channelInactive(ctx: ChannelHandlerContext): Unit = {
     this.synchronized {
+      // Cancel timeout task
+      val currentTask = timeoutTask.get()
+      if (currentTask != null) {
+        currentTask.cancel(false)
+        timeoutTask.set(null)
+      }
```
kyri-petrou: This is being used everywhere and doesn't have any dependency on the local vals. Maybe extract it into a method to keep things DRY

zio-http/jvm/src/main/scala/zio/http/netty/AsyncBodyReader.scala
987Nabil: `this.synchronized` has a different meaning. Cos `this` would be the `new Runnable`. Thats why we use `self =>` in Scala. I will change it so self.

PR #3870, opened 2025-12-19
Allow : and @ unescaped in URL path and query

QuinnHarris:
RFC 3986 explicitly allows both `:` and `@` to appear unescaped in URL **path segments** and **query strings** (`pchar`):

```
pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
```

However, zio-http currently percent-encodes these characters even when they appear in contexts where they are treated as data rather than delimiters.

This PR updates URL rendering to allow `:` and `@` to remain unescaped in path and query segments both using the `encodeComponentInto` function.

I specifically required this change when using google validateAddress API with URL  `https://addressvalidation.googleapis.com/v1:validateAddress` that does not work when it is percent-encoded.

**References**

* RFC 3986 §3.3 (Path)
* RFC 3986 §3.4 (Query)

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/URLSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/URLSpec.scala
@@ -160,6 +160,11 @@ object URLSpec extends ZIOHttpSpec {
           val expected = Right(host + channels)
           assertTrue(actual == expected)
         },
+        test("does not escape valid characters") {
+          check(HttpGen.nonEmptyPath) { path =>
+            assertTrue(URL(path = path).encode == path.encode)
+          }
+        }
       ),
       suite("builder")(
         test("creates a URL with all attributes set") {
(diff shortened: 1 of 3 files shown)


987Nabil:
@QuinnHarris please fix this

```
[error] /home/runner/work/zio-http/zio-http/zio-http/jvm/src/test/scala/zio/http/URLSpec.scala:165:24: type mismatch;
[error]  found   : Option[zio.http.URL.Fragment]
[error]  required: Option[a.Fragment]
[error]             assertTrue(URL(path = path).encode == path.encode)
[error]                        ^
[error] /home/runner/work/zio-http/zio-http/zio-http/jvm/src/test/scala/zio/http/URLSpec.scala:165:23: not found: value location
[error]             assertTrue(URL(path = path).encode == path.encode)
[error]                       ^
```

987Nabil (approved):

PR #3874, opened 2025-12-27
feat: Add HTTP Range header support for file serving

### LINKED ISSUE #709: Support Http Range header on request for Files

Exceluyi:
## Summary

This PR adds support for HTTP Range requests (RFC 9110 §14) when serving static files. Clients can now request specific byte ranges of files, enabling features like resumable downloads and video seeking.

### New API

Added `Handler.fromFileWithRange` - a handler that serves files with Range header support:

```scala
Handler.fromFileWithRange(ZIO.succeed(file), request)
```

### Behavior

| Request | Response |
|---------|----------|
| No Range header | `200 OK` with full file + `Accept-Ranges: bytes` |
| Valid Range (e.g., `bytes=0-99`) | `206 Partial Content` + `Content-Range` header |
| Invalid Range (beyond file size) | `416 Range Not Satisfiable` + `Content-Range: bytes */<size>` |
| Suffix Range (e.g., `bytes=-100`) | `206 Partial Content` (last N bytes) |
| Prefix Range (e.g., `bytes=100-`) | `206 Partial Content` (from offset to end) |

## Usage Example

```scala
val app = Routes(
  Method.GET / "files" / string("name") -> handler { (name: String, req: Request) =>
    Handler.fromFileWithRange(ZIO.succeed(new File(s"/data/$name")), req)
  }
)
```

/fixes #709 
/ckaim #709

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/StaticFileServerSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/StaticFileServerSpec.scala
@@ -20,7 +20,7 @@ import java.io.File
 
 import zio._
 import zio.test.Assertion._
-import zio.test.TestAspect.{mac, os, sequential, unix, withLiveClock}
+import zio.test.TestAspect.{ignore, mac, os, sequential, unix, withLiveClock}
 import zio.test.{assertTrue, assertZIO}
 
 import zio.http.internal.{DynamicServer, RoutesRunnableSpec, serverTestLayer}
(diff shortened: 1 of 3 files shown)


987Nabil:
feature already merged

PR #3876, opened 2026-01-01
Implement HappyEyeballs for NettyConnectionPool (#1996)

### LINKED ISSUE #1996: Add support for Happy Eyeballs

987Nabil:
fixes https://github.com/zio/zio-http/issues/1996
/claim https://github.com/zio/zio-http/issues/1996

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
@@ -128,7 +128,8 @@ private[netty] object NettyConnectionPool {
   }
 
   /**
-   * Attempts to connect to a single address.
+   * Attempts to connect to a single address. Adds a finalizer to close the
+   * channel when the Scope is closed.
    */
   private def connectToAddress(
     host: InetAddress,
@@ -138,7 +139,34 @@ private[netty] object NettyConnectionPool {
     initializer: ChannelInitializer[JChannel],
     connectionTimeout: Option[Duration],
     localAddress: Option[InetSocketAddress],
-  )(implicit trace: Trace): ZIO[Scope, Throwable, JChannel] = {
+  )(implicit trace: Trace): ZIO[Scope, Throwable, JChannel] =
+    connectToAddressUnmanaged(
+      host,
+      channelFactory,
+      eventLoopGroup,
+      location,
+      initializer,
+      connectionTimeout,
+      localAddress,
+    ).tap { ch =>
+      Scope.addFinalizer {
+        NettyFutureExecutor.executed(ch.close()).whenDiscard(ch.isOpen).ignoreLogged
+      }
+    }
+
+  /**
+   * Attempts to connect to a single address without adding a finalizer. The
+   * caller is responsible for closing the channel.
+   */
+  private def connectToAddressUnmanaged(
+    host: InetAddress,
+    channelFactory: JChannelFactory[JChannel],
+    eventLoopGroup: JEventLoopGroup,
+    location: URL.Location.Absolute,
+    initializer: ChannelInitializer[JChannel],
+    connectionTimeout: Option[Duration],
+    localAddress: Option[InetSocketAddress],
+  )(implicit trace: Trace): ZIO[Any, Throwable, JChannel] = {
     ZIO.suspend {
       val bootstrap = new Bootstrap()
         .channelFactory(channelFactory)
@@ -150,12 +178,11 @@ private[netty] object NettyConnectionPool {
 
       val channelFuture = bootstrap.connect()
       val ch            = channelFuture.channel()
-      Scope.addFinalizer {
-        NettyFutureExecutor.executed {
-          channelFuture.cancel(true)
-          ch.close()
-        }.whenDiscard(ch.isOpen).ignoreLogged
-      }.uninterruptible *> NettyFutureExecutor.executed(channelFuture).as(ch)
+
+      NettyFutureExecutor
+        .executed(channelFuture)
+        .onInterrupt(ZIO.succeed(channelFuture.cancel(true)) *> ZIO.succeed(ch.close()).ignore)
+        .as(ch)
     }
   }
 
@@ -164,7 +191,7 @@ private[netty] object NettyConnectionPool {
    * we start with IPv6, then after firstAddressFamilyDelay we try IPv4, then
    * alternate between families.
    */
-  private def sortAddresses(resolvedHosts: Chunk[InetAddress]): List[InetAddress] = {
+  private[client] def sortAddresses(resolvedHosts: Chunk[InetAddress]): List[InetAddress] = {
     val (ipv6Addresses, ipv4Addresses) = resolvedHosts.partition(_.isInstanceOf[Inet6Address])
     val ipv6Iter                       = ipv6Addresses.iterator
     val ipv4Iter                       = ipv4Addresses.iterator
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+            channelFactory,
+            eventLoopGroup,
+            location,
+            initializer,
+            connectionTimeout,
+            localAddress,
+          ).onExit {
+            case e: Exit.Success[JChannel] => successful.update(channels => channels :+ e.value)
+            case _: Exit.Failure[_]        => lastFailed.offer(()).unit
+          },
```
kyri-petrou: Maybe put all of this in a function and reuse it to make the code a bit more readable:

```scala
        connect0 = connectToAddress(
          _,
          channelFactory,
          eventLoopGroup,
          location,
          initializer,
          connectionTimeout,
          localAddress,
        ).onExit {
          case e: Exit.Success[JChannel] => successful.update(channels => channels :+ e.value)
          case _: Exit.Failure[_]        => lastFailed.offer(()).unit
        }
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+                case _: Exit.Failure[_]        => lastFailed.offer(()).unit
+              }
+          },
+        )
+        channels   <- successful.get
+        channel    <- channels.collectFirst { case c if c.isOpen => c } match {
+          case ch: Some[JChannel] => ZIO.succeed(ch.value)
+          case None               => ZIO.fail(new RuntimeException("All connection attempts failed"))
+        }
+        _          <- ZIO.foreachDiscard(channels.tail)(ch => ZIO.ignore(ch.close()))
```
kyri-petrou: There's no guarantee that the channel we selected via the `collectFirst` will be at the head of the list (since the first one might fail the `c.isOpen` condition)

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+            location,
+            initializer,
+            connectionTimeout,
+            localAddress,
+          ).onExit {
+            case e: Exit.Success[JChannel] => successful.update(channels => channels :+ e.value)
+            case _: Exit.Failure[_]        => lastFailed.offer(()).unit
+          },
+          addresses.tail.zipWithIndex.map { case (address, index) =>
+            ZIO.sleep(HappyEyeballsDelay * index.toDouble).raceFirst(lastFailed.take).ignore *>
```
kyri-petrou: I'm not sure if the usage of a Queue here is intended or if it's a bug.

With a Queue, only one of the fibers will be "interrupted" via `raceFirst(lastFailed.take)` because only a single fiber can poll the item from the queue. Is this a bug? If this is the intended operation, then why do we want to "interrupt" a single connection attempt only?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
kyri-petrou: Don't we need `raceFirst(lastFailed.take).ignore` here as well?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,150 @@ private[netty] object NettyConnectionPool {
...
+          connectToAddress(
+            addresses.head,
+            channelFactory,
+            eventLoopGroup,
+            location,
+            initializer,
+            connectionTimeout,
+            localAddress,
+          ).onExit {
+            case e: Exit.Success[JChannel] => successful.update(channels => channels :+ e.value)
```
kyri-petrou: Previously I described how there's an issue here but perhaps my comment wasn't clear enough. What I meant was that since the release of the connections is managed by the `Scope`, you might end up in a case that the connection is acquired but not released until the Scope is closed, even if that connection lost the race.

Hope this reproducer helps to understand the issue:

```scala
import zio.*

final class Foo(val i: Int) {
  def acquired() = println(s"acquired: $i")
  def released() = println(s"released: $i")
}

  val sleepRandom =
    ZIO.randomWith(_.nextIntBounded(100).flatMap(i => ZIO.sleep(i.millis)))

  def resource(i: Int) = {
    val res = ZIO.acquireRelease(ZIO.succeed {
      val foo = new Foo(i)
      foo.acquired()
      foo
    })(f => ZIO.succeed(f.released()))

    (res.map(_.i) <* sleepRandom <* ZIO.debug(s"initialized: $i")).onInterrupt(ZIO.debug(s"interrupted: $i"))
  }
  def run              = {
    val rs = (1 to 4).map(resource).toList
    val f  =
      for {
        won <- ZIO.raceAll(rs.head, rs.tail)
        _   <- ZIO.sleep(10.millis)
        _   <- ZIO.debug(s"doing some stuff with resource: $won")
        _   <- ZIO.sleep(5.seconds)
        _   <- ZIO.debug("Finished work")
      } yield ()
    ZIO.scoped(f)
  }
}
```

This prints the following which is really problematic, because `released: 2` should be printed _before_  the "Finished work" message.
```
acquired: 2
acquired: 1
initialized: 1
interrupted: 2
doing some stuff with resource: 1
Finished work
released: 1
released: 2
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,188 @@ private[netty] object NettyConnectionPool {
...
+
+        channels <- successful.get
+
+        channel <- channels.find(_.isOpen) match {
+          case Some(ch) => ZIO.succeed(ch)
+          case None     => ZIO.fail(new RuntimeException("All connection attempts failed"))
+        }
+
+        _ <- ZIO.foreachDiscard(channels.filterNot(_ eq channel)) { ch =>
+          ZIO.succeed(ch.close()).ignore
```
kyri-petrou:
```suggestion
          ZIO.ignore(ch.close())
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,188 @@ private[netty] object NettyConnectionPool {
...
+          channelFactory,
+          eventLoopGroup,
+          location,
+          initializer,
+          connectionTimeout,
+          localAddress,
+        )
+
+      for {
+        failedHub  <- Hub.dropping[Unit](requestedCapacity = 1)
```
kyri-petrou: I might have missed the previous response for this, but isn't a `Promise` better suited for this? Since we only care about fulfilling it once I think it's a better implementation for this, wdut?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,188 @@ private[netty] object NettyConnectionPool {
...
+        .withOption[Integer](ChannelOption.CONNECT_TIMEOUT_MILLIS, connectionTimeout.map(_.toMillis.toInt))
+        .handler(initializer)
+      localAddress.foreach(bootstrap.localAddress)
+
+      val channelFuture = bootstrap.connect()
+      val ch            = channelFuture.channel()
+
+      NettyFutureExecutor
+        .executed(channelFuture)
+        .onInterrupt(ZIO.succeed(channelFuture.cancel(true)) *> ZIO.succeed(ch.close()).ignore)
```
kyri-petrou:
```suggestion
        .onInterrupt(ZIO.ignore { channelFuture.cancel(true);  ch.close() })
```

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,188 @@ private[netty] object NettyConnectionPool {
...
+            case Exit.Failure(_)  => failedHub.publish(()).unit
+          }
+
+        _ <- ZIO.scoped {
+          for {
+            failedSub <- failedHub.subscribe
+            _         <- ZIO.raceAll(
+              attemptConnect(addresses.head),
+              addresses.tail.zipWithIndex.map { case (address, index) =>
+                ZIO.sleep(HappyEyeballsDelay * (index + 1).toDouble).raceFirst(failedSub.take).ignore *>
```
kyri-petrou: Is this the right behaviour according to the spec? In the case that a single connection attempt fails, should all the other ones be cancelled?

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
```diff
@@ -106,38 +108,188 @@ private[netty] object NettyConnectionPool {
...
+   */
+  private def happyEyeballsConnect(
+    resolvedHosts: Chunk[InetAddress],
+    channelFactory: JChannelFactory[JChannel],
+    eventLoopGroup: JEventLoopGroup,
+    location: URL.Location.Absolute,
+    initializer: ChannelInitializer[JChannel],
+    connectionTimeout: Option[Duration],
+    localAddress: Option[InetSocketAddress],
+  )(implicit trace: Trace): ZIO[Scope, Throwable, JChannel] = {
```
kyri-petrou, line 232: I feel like we need to control interruption more closely here (using `uninterruptibleMask`) and allow it only during `NettyFutureExecutor.executed(channelFuture)`. Otherwise we're risking interruption creeping in between effect evaluations and leaking connections

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: Done — applied `ZIO.ignore(ch.close())` as suggested.

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: Good call — switched to `Promise.make[Nothing, Unit]`. The one-shot semantics are actually what we want here: per RFC 8305 Section 5, a failure should trigger the next staggered attempt to start immediately, but should NOT cancel or affect other running attempts. `Promise.succeed(()).unit` is idempotent, so subsequent failures are no-ops. This also let me remove the `ZIO.scoped` wrapper that was only needed for `Hub.subscribe`.

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: Done — applied `ZIO.ignore { channelFuture.cancel(true); ch.close() }` as suggested.

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: Good question — per RFC 8305 Section 5: "Starting a new connection attempt does not affect previous attempts" and "Once one of the connection attempts **succeeds**, all other connections attempts that have not yet succeeded SHOULD be canceled." So failures should NOT cancel other attempts. The previous Hub-based approach was incorrect because `Hub` broadcasts to ALL subscribers simultaneously, meaning a single failure would unblock ALL waiting delays at once. Switched to `Promise` which has one-shot semantics — only the first failure unblocks the next waiting delay, subsequent failures are no-ops. The `raceAll` handles cancelling losers on success.

zio-http/jvm/src/main/scala/zio/http/netty/client/NettyConnectionPool.scala
987Nabil: Good point — added `ZIO.uninterruptibleMask { restore => ... }` wrapping the entire `connectToAddressUnmanaged` body, with `restore()` applied only to `NettyFutureExecutor.executed(channelFuture)`. This follows the same pattern used in `ZioNettyConnectionPool.get` and ensures the channel creation + `onInterrupt` handler registration happen in an uninterruptible region, preventing leaks.

PR #3877, opened 2026-01-02
WIP: Implementing support for HTTP Range header (#709)

### LINKED ISSUE #709: Support Http Range header on request for Files

francoo98:
This pr is a draft to fix #709.
Atm, it only supports one range in the Range header. Example: Range: bytes=0-5000
/claim #709


francoo98:
Do you have any thoughts on fixing the MIMA check? I changed the definition of `fromFile` in `Body.scala`.
Current:
`fromFile(file: java.io.File, chunkSize: Int = 1024 * 4)(implicit trace: Trace): ZIO[Any, Nothing, Body]`
This pr:
`fromFile(file: java.io.File, chunkSize: Int = 1024 * 4, start: Long = 0, end: Option[Long] = None)(implicit trace: Trace): ZIO[Any, Nothing, Body]`
Should I modify the definition or create a new method `fromRangedFile`?

987Nabil:
Already merged

PR #3879, opened 2026-01-06
Add `RequestStore.getMany` functions

guizmaii:


987Nabil (commented):
This is a lot of code for a feature that I find questionable. 
Why do you need to get back multiple values? Why don't you put a case class or tuple in the store in the first place? 
It is also not faster then you making multiple calls yourself. 
This seems to me to be code that optimizes an edge case/anti pattern.

guizmaii:
@987Nabil 

> Why do you need to get back multiple values? 

We set values at different times during the request execution, so when we need to get the values at the end to generate a kind of report/audit log, dependending on the execution path, we might have only some of these values

It's a lot of code but a lot of dumb code, and I add the missing tests for all the `RequestStore` functions 🙂

987Nabil:
Still seems to me like an edge case and I think the code might be fitting more as extension methods in your codebase

guizmaii:
Hum, IMO, more people will adopt the same pattern as this `RequestStore` is very useful for it: putting different values in different places in the code path and later retrieving the values set during the request execution to, for example, log something.

Very easy way to implement Audit logs, for example

PR #3889, opened 2026-01-16
Implement full HTTP Range request support to Handler.fromFile/ZIO (#709)

Michael-A-Kuykendall:
This PR implements comprehensive support for HTTP Range requests in `Handler.fromFile`, resolving issue #709.

### Features
- **Full Range Support**: Handles single ranges (`bytes=0-100`), multi-part ranges (`bytes=0-10, 20-30`), suffix ranges (`bytes=-100`), and open-ended ranges (`bytes=100-`).
- **Large File Support (>2GB)**: Includes specific handling for files larger than 2GB, using `Long` for offsets and bypassing `Int` limitations in the `Content-Range` header model.
- **RFC Compliance**: 
  - Implements **Weak ETags** (`W/"..."`) and strict `If-Range` header parsing (normalizing quotes and prefixes).
  - Correctly handles `HEAD` requests (calculates `Content-Length`/headers without processing the body).
- **Bug Fix**: Fixes regex patterns in `Header.scala` to correctly parse `*/total` and `start-end/*` formats.

### Testing
Added a new test suite `RangeStaticFileServerSpec` that verifies:
- Single, Multi-part, and Suffix ranges.
- `If-Range` ETag handling (match vs mismatch).
- **Large File Test**: A specific test case generating a 3GB sparse file to prove `Content-Range` headers do not overflow.

/claim #709


987Nabil:
Already on master. And there is also functionality missing here for the static file middleware

PR #3919, opened 2026-01-23
Optimise CI: reduce build jobs from 12 to 6

guizmaii:
## Summary

Optimise CI by reducing the test matrix from 12 to 6 jobs (50% reduction).

The `sbt-github-actions` plugin generates a full Scala × JVM matrix (3 × 4 = 12 jobs), but most issues are Scala-specific OR JVM-specific, not both. The plugin lacks the flexibility to customise this matrix, so we replace it with a manual workflow that provides better control.

## New Test Strategy

This test strategy is adopted from the [ZIO project](https://github.com/zio/zio/blob/series/2.x/.github/workflows/ci.yml).

| Job | Scala | JVM | Count |
|-----|-------|-----|-------|
| `test` | 2.12.x, 2.13.x, 3.x | temurin@17 | 3 |
| `testJvms` | 2.13.x | temurin@21, graalvm@17, graalvm@21 | 3 |
| **Total** | | | **6** |

This provides equivalent coverage: Scala-specific issues are caught by testing all Scala versions, JVM-specific issues are caught by testing all JVMs.

## Other Improvements

- Consolidate 18 JMH benchmark jobs into 1 sequential job
- Add `ci` aggregator job using alls-green pattern
- Use `release` event trigger instead of tag-based triggers
- Separate `lint` job for format checking

## Test plan
- [ ] CI passes on this PR
- [ ] Verify 6 test jobs instead of 12
- [ ] Verify all functionality preserved (tests, MiMa, scoverage, benchmarks, publishing)


987Nabil:
@guizmaii I am not sure that is a really good idea. I expect you want to do this to reduce load on the free runners. I talked to John and we might get some extra runners. Also we will get rid of 2.12 soon. The only thing I would merge for sure is reduce the benchmark jobs

987Nabil:
I will find some solution to optimize the builds

PR #3920, opened 2026-01-23
Fix ClassCastException when applying HandlerAspect to routes with path parameters

Godzilla675:
/claim #3141

## Summary
This PR fixes the `ClassCastException` that occurs when combining a route with path parameters and a `HandlerAspect` middleware.

## Changes
- Added `@@` operators to the `Route` trait to apply aspects after path parameter decoding
- Updated `RouteSpec.scala` with a regression test (issue #3141)

## Root Cause
The `Handler.@@` operator assumes input is `Request`, but applied to a parameterized route, the input is a Tuple.
The fix applies the aspect at the `Route` level where we can handle this correctly.


987Nabil:
https://github.com/zio/zio-http/issues/3141 is not fixable in 3.x

PR #3945, opened 2026-02-02
Fixed the auth header missing error to 401 from 400

Exceluyi:
/claim #3235


987Nabil:

```scala
```

Exceluyi:
understood

PR #3948, opened 2026-02-10
Return 401 for missing/malformed Authorization header

### LINKED ISSUE #3235: Missing Authorization header is reported as 400 instead of 401

buildingvibes:
## Summary

Fixes #3235. When an endpoint requires authentication via `.header(HeaderCodec.authorization)` or `.auth(AuthType.Bearer)`, a request without a valid Authorization header now returns **401 Unauthorized** instead of 400 Bad Request, per [RFC 7235 Section 3.1](https://www.rfc-editor.org/rfc/rfc7235#section-3.1).

### `Endpoint.scala`
- Added `isAuthorizationError` helper that detects when a codec error is related to the Authorization header (handles `MissingHeader`, `MissingHeaders`, `MalformedHeader`, and `DecodingErrorHeader`)
- Added a case in the `catchAllCause` error handler that returns 401 with `WWW-Authenticate` header for auth-related errors

### `AuthType.scala`
- Added `wwwAuthenticateHeader` method that generates the appropriate `WWW-Authenticate` challenge value (e.g., `Bearer realm="restricted"`)

### Tests
- Missing Authorization header → 401 Unauthorized with WWW-Authenticate header
- Malformed Authorization header → 401 Unauthorized  
- Missing non-auth header → 400 Bad Request (regression guard)

## RFC Compliance
- **RFC 7235 §3.1**: "The server generating a 401 (Unauthorized) response MUST send a WWW-Authenticate header field"
- **RFC 6750 §3.1**: Malformed/invalid Authorization headers should also receive 401

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/endpoint/Authorization401Spec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/endpoint/Authorization401Spec.scala
@@ -1,0 +1,79 @@
+package zio.http.endpoint
+
+import zio.test._
+
+import zio.http._
+import zio.http.codec.HeaderCodec
+
+object Authorization401Spec extends ZIOHttpSpec {
+
+  override def spec =
+    suite("Authorization 401 Spec")(
+      test("Missing Authorization header with Bearer auth returns 401 Unauthorized") {
+        val endpoint = Endpoint(Method.GET / "secure")
+          .out[String]
+          .auth(AuthType.Bearer)
+        val route    = endpoint.implementHandler(handler((_: Unit) => "ok"))
+        val request  = Request(method = Method.GET, url = url"/secure")
+        for {
+          response <- route.toRoutes.runZIO(request)
+        } yield assertTrue(
+          response.status == Status.Unauthorized,
+          response.header(Header.WWWAuthenticate).isDefined,
+        )
+      },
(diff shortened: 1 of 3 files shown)


buildingvibes:
recheck

buildingvibes:
Hi! Just checking in - all CI checks are passing (40+ jobs across Scala 2.12/2.13/3.3 on multiple JVMs), CLA is signed, and the implementation follows the existing patterns for error handling in Endpoint.scala. This fixes #3235 where missing/malformed Authorization headers return 400 instead of the correct 401. Happy to address any feedback!

987Nabil:

```scala
```

PR #3953, opened 2026-02-25
CI: Add JDK 25

guizmaii:
## Summary

- Add `temurin@25` to the CI build matrix to test against JDK 25
- Use Scala version wildcards instead of hardcoded versions in CI matrices and conditions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
           - graal_graalvm@21
           - temurin@17
           - temurin@21
+          - temurin@25
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -26,12 +26,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.12.21, 2.13.18, 3.3.7]
+        scala: [2.12.x, 2.13.x, 3.x]
```
senia-psm, line 29: This file is generated by https://github.com/sbt/sbt-github-actions
I suspect you have to configure it this way:
```scala
ThisBuild / githubWorkflowJavaVersions += JavaSpec.temurin("25")
```

PR #3983, opened 2026-03-02
perf: cache headerType on Header.Custom via null-init field

guizmaii:
## Summary
- `Header.Custom.headerType` is a `def` that allocates a new anonymous `HeaderType` on every call. The default `headerName` and `renderedValue` delegate to `headerType`, causing unnecessary allocation when these fields are accessed.
- Cache the `HeaderType` instance using a private null-init field so it is only allocated once per `Custom` header instance.
- Note: the primary hot path (incoming request headers via `Headers.Native`) is unaffected — this fixes the path for user-constructed headers.

## Test plan
- [x] All 50 existing HeaderSpec tests pass
- [x] Added test verifying `headerType` is cached (referential identity via `eq`) and returns correct values
- [x] `sbt 'zioHttpJVM/testOnly zio.http.HeaderSpec'` — 51 tests pass

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/HeaderSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/HeaderSpec.scala
@@ -400,19 +400,14 @@ object HeaderSpec extends ZIOHttpSpec {
     Headers(Header.Accept(MediaType.application.json), Header.ContentType(MediaType.application.json))
 
   private val customHeaderSpec = suite("Custom header")(
-    test("headerName returns lowercase name without allocating headerType") {
-      val h = Header.Custom("X-Request-Id", "abc-123")
+    test("headerType is cached and returns correct values") {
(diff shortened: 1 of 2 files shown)


987Nabil:
I would tend to make headerType a lazy val

guizmaii:
@987Nabil 

> I would tend to make headerType a lazy val

The issue I have with lazy val is the overhead it introduces. For simple computations, which is likely the case here, recomputing is faster than having to deal with the lazy val synchronisation mechanism

With Scala3 `@threadUnsafe lazy val` I'd agree with you, but because we have Scala2 compatibility needs, I'd prefer to avoid using `lazy val`

987Nabil:
@guizmaii private null init field

guizmaii:
@987Nabil Simple and effective idea! I love it! Thanks! I changed the code :)

987Nabil (approved):

PR #3997, opened 2026-03-05
perf: reduce per-request heap allocations in HTTP hot path

987Nabil:
## Summary

Reduces per-request heap allocations in zio-http's server-side hot path (`ServerInboundHandler → Handler → Response → Netty write`) through four surgical, binary-compatible changes.

Motivated by the observation that zio-http allocates unnecessary intermediate objects on every HTTP request in TechEmpower-style benchmarks.

### 1. `StringBody.knownContentLength` — zero-alloc byte counting
- **Before**: Called `data.getBytes(charset).length` which allocates a full byte array just to measure length
- **After**: For UTF-8, counts bytes by iterating chars (1–4 bytes per codepoint with surrogate pair handling). For ISO-8859-1, returns `data.length` directly. Falls back to `getBytes()` for other charsets
- **Impact**: Eliminates one byte array allocation per response body on the `Content-Length` computation path

### 2. `URL.decodeOrNull` — avoid `Either` allocation in `makeZioRequest`
- Adds `private[http] def decodeOrNull(rawUrl: String): URL` that returns `URL | null` instead of `Either[Exception, URL]`
- Replaces `URL.decode(uri).getOrElse(URL.empty)` in both `makeZioRequest` call sites in `ServerInboundHandler`
- **Impact**: Eliminates `Either`/`Right` wrapper allocation per HTTP request

### 3. `Handler.orElse` — lazy fallback evaluation
- **Before**: Evaluated both `self(in)` and `that(in)` eagerly in a tuple before pattern matching
- **After**: Evaluates `self(in)` first; only calls `that(in)` when self fails with a typed error
- **Impact**: Avoids unnecessary computation and side-effects of fallback handlers when the primary handler succeeds

### 4. `Response.json`/`Response.text` overloads with `extraHeaders`
- Adds new overloads: `Response.json(data, extraHeaders)` and `Response.text(data, extraHeaders)`
- Avoids the pattern `Response.text(msg).addHeader(h)` which creates an intermediate `Response` copy
- Purely additive — existing methods unchanged

## Verification

- [x] `sbt fmt` — no formatting changes needed
- [x] `sbt testOnly BodySpec URLSpec HandlerSpec ResponseSpec` — **146 tests passed, 0 failed**
- [x] `sbt mimaChecks` — **no binary incompatibilities**

## Scope Investigation (not implemented)

Investigated `Scope.unsafe.make` in `ServerInboundHandler.writeResponse` — it allocates 2 small objects per request on the slow path (~64-96 bytes). Concluded that skipping it is **unsafe** because there's no way to statically determine whether a handler uses `Scope.addFinalizer`, and using `Scope.global` would silently discard finalizers causing resource leaks. Recommend deferring to 4.x where a type-level marker could distinguish scoped from unscoped handlers.


zio-http/shared/src/main/scala/zio/http/Handler.scala
```diff
@@ -449,17 +449,15 @@ sealed trait Handler[-R, +Err, -In, +Out] { self =>
...
-          case (Exit.Failure(cause), _) if cause.isDie         =>
-            Exit.die(cause.dieOption.get)
-          case (Exit.Failure(cause), other) if cause.isFailure =>
-            other
-          case (self, other)                                   =>
-            self.orElse(other)
+      override def apply(in: In1): ZIO[Scope & R1, Err1, Out1] = {
+        val s = self(in)
+        s match {
+          case s @ Exit.Success(_)                    => s
```
guizmaii, line 455:
```suggestion
          case s: Exit.Success[..., ...]                    => s
```

zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -309,6 +309,16 @@ object URL {
     }
   }
 
+  private[http] def decodeOrNull(rawUrl: String): URL = {
+    try {
+      val uri = new URI(rawUrl)
+      val url = if (uri.isAbsolute) fromAbsoluteURI(uri) else fromRelativeURI(uri)
```
guizmaii: If you go this way, I think it's be interesting to also have `fromAbsoluteURIOrNull` and `fromRelativeURIOrNull` as, for example, `fromAbsoluteURI` allocates quite a few `Option`, especially this one: `Option(uri.getPort)` which introduces boxing

PR #3999, opened 2026-03-06
call remove element function in SSE `executeScript`

illegalprime:
As it stands the function is not actually called (at least for me?) and the [demo](https://ziohttp.com/reference/datastar-sdk/) populates a ton of noisy script tags.


987Nabil:
Tests are broken, I integrated it in #4001 so closing this one

PR #4007, opened 2026-03-08
feat(middleware): add response-derived labels to metric middleware

### LINKED ISSUE #2536: Metric middleware labels based on response
Currently we only allow passing labels when providing the metric middleware (https://github.com/zio/zio-http/blob/b43dbc478e1f387c6de60bdab0b99dfbc1de3e6f/zio-http/src/main/scala/zio/http/Middleware.scala#L260). 

What do you think about making this effectful and allow providing these based on request+response? It can be useful when you want to segment metrics on, for example, cache hit or similar only known at response time.

987Nabil:
## Summary

Closes #2536

Extends the metric middleware to accept a pure function callback that derives metric labels from the response, enabling users to add response-derived labels (like status code class, custom headers, etc.) to their HTTP metrics.

### New API

```scala
// With response labels
Middleware.metrics(
  extraLabels = response => Set(
    MetricLabel("status_class", (response.status.code / 100).toString + "xx")
  )
)

// Full overload with all metric names + response labels
Middleware.metrics(
  pathLabelMapper = ...,
  totalRequestsName = ...,
  requestDurationName = ...,
  responseLabels = response => Set(MetricLabel("status", response.status.code.toString))
)
```

### Design Decisions

- **Pure functions only** (`Response => Set[MetricLabel]`) — no effectful callbacks to avoid per-request ZIO overhead
- **Zero overhead** when no callback is provided — existing `Middleware.metrics()` path remains unchanged
- **New overloads** — existing signatures preserved for backward compatibility

### Changes

- `Middleware.scala`: Added 2 new `metrics` overloads with `responseLabels` parameter (~100 lines)
- `MetricsSpec.scala`: Added tests for response-derived labels, status class labeling, and backward compatibility

### Binary Compatibility

New method overloads only — fully backward compatible. No MiMa filters needed.

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/Middleware.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Middleware.scala
@@ -430,7 +430,7 @@ object Middleware extends HandlerAspects {
       for {
         _   <- requestsTotal.tagged(labels).increment
         _   <- concurrentRequests.tagged(requestLabels).decrement
-        end <- Clock.nanoTime
+        end <- ZIO.succeed(java.lang.System.nanoTime())
         took = end - start
         _ <- requestDuration.tagged(labels).update(took / nanosToSeconds)
       } yield ()
@@ -441,7 +441,7 @@ object Middleware extends HandlerAspects {
 
       HandlerAspect.interceptHandlerStateful(Handler.fromFunctionZIO[Request] { req =>
         for {
-          start <- Clock.nanoTime
+          start <- ZIO.succeed(java.lang.System.nanoTime())
           _     <- concurrentRequests.tagged(requestLabels).increment
         } yield ((start, requestLabels), (req, ()))
       })(Handler.fromFunctionZIO[((Long, Set[MetricLabel]), Response)] { case ((start, requestLabels), response) =>
--- a/zio-http/shared/src/main/scala/zio/http/Middleware.scala
+++ b/zio-http/shared/src/main/scala/zio/http/Middleware.scala
@@ -459,6 +459,110 @@ object Middleware extends HandlerAspects {
     }
   }
 
+  /**
+   * Creates middleware that will track metrics, with additional labels derived
+   * from the response.
+   *
+   * @param responseLabels
+   *   A pure function that derives additional metric labels from the response
+   *   (e.g. status class, specific headers). Invoked once per response.
+   * @param concurrentRequestsName
+   *   Concurrent HTTP requests metric name.
+   * @param totalRequestsName
+   *   Total HTTP requests metric name.
+   * @param requestDurationName
+   *   HTTP request duration metric name.
+   * @param requestDurationBoundaries
+   *   Boundaries for the HTTP request duration metric.
+   * @param extraLabels
+   *   A set of extra labels all metrics will be tagged with.
+   */
+  def metrics(
+    responseLabels: Response => Set[MetricLabel],
+    pathLabelMapper: String => String,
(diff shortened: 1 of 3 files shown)


zio-http/shared/src/main/scala/zio/http/Middleware.scala
```diff
@@ -459,6 +459,107 @@ object Middleware extends HandlerAspects {
...
+
+    def report(
+      start: Long,
+      requestLabels: Set[MetricLabel],
+      labels: Set[MetricLabel],
+    )(implicit trace: Trace): ZIO[Any, Nothing, Unit] =
+      for {
+        _   <- requestsTotal.tagged(labels).increment
+        _   <- concurrentRequests.tagged(requestLabels).decrement
+        end <- Clock.nanoTime
```
987Nabil: Use java system time instead. No need for the effect overhead

zio-http/shared/src/main/scala/zio/http/Middleware.scala
```diff
@@ -459,6 +459,107 @@ object Middleware extends HandlerAspects {
...
+        took = end - start
+        _ <- requestDuration.tagged(labels).update(took / nanosToSeconds)
+      } yield ()
+
+    def aspect(routePattern: RoutePattern[_])(implicit trace: Trace): HandlerAspect[Any, Unit] = {
+      val requestLabels = labelsForRequest(routePattern)
+
+      HandlerAspect.interceptHandlerStateful(Handler.fromFunctionZIO[Request] { req =>
+        for {
+          start <- Clock.nanoTime
```
987Nabil: same

zio-http/shared/src/main/scala/zio/http/Middleware.scala
```diff
@@ -459,6 +459,107 @@ object Middleware extends HandlerAspects {
...
+  )(implicit trace: Trace): Middleware[Any] = {
+    val requestsTotal: Metric.Counter[RuntimeFlags] = Metric.counterInt(totalRequestsName)
+    val concurrentRequests: Metric.Gauge[Double]    = Metric.gauge(concurrentRequestsName)
+    val requestDuration: Metric.Histogram[Double]   = Metric.histogram(requestDurationName, requestDurationBoundaries)
+    val nanosToSeconds: Double                      = 1e9d
+
+    def labelsForRequest(routePattern: RoutePattern[_]): Set[MetricLabel] =
+      Set(
+        MetricLabel("method", routePattern.method.render),
+        MetricLabel("path", routePattern.pathCodec.render),
```
987Nabil: paths might contain IDs or other values that need to be cleaned up to not pollute the metric. So /api/basket/497d395d becomes /api/basket/{basketId}. We need to be able to give the user a way to this

zio-http/shared/src/main/scala/zio/http/Middleware.scala
987Nabil: Fixed — replaced `Clock.nanoTime` with `ZIO.succeed(System.nanoTime())` in both locations.

zio-http/shared/src/main/scala/zio/http/Middleware.scala
987Nabil: Added `pathLabelMapper: String => String` parameter to both overloads. In the convenience overload it defaults to `identity`. Applied in `labelsForRequest` as `pathLabelMapper(routePattern.pathCodec.render)`.

PR #4011, opened 2026-03-08
feat(datastar): add dispatchEvent for firing custom DOM events from backend

### LINKED ISSUE #3846: Simple way to fire events from the BE
`DatastarEvent` and `ServerSentEventGenerator` should have a simple way to send a custom event.
Something like event name and object with schema for the payload. Also with js event options and from which element the event should be sent. With a default of document.

987Nabil:
## Summary

Closes #3846

Adds `dispatchEvent` convenience methods to `DatastarEvent` and `ServerSentEventGenerator` that generate JavaScript to dispatch `CustomEvent` instances in the browser, with type-safe payloads via ZIO Schema.

### Changes

- **`DispatchEventOptions`** case class with source element (`Option[CssSelector]`), `bubbles` (default `true`), `cancelable` (default `false`), `composed` (default `false`), `autoRemove`, `eventId`, and `retryDuration`
- **`DatastarEvent.dispatchEvent`** — 5 overloads supporting typed payloads (`T <: Product: Schema`), raw `Js` payloads, optional CSS selector for target element, and full options
- **`ServerSentEventGenerator.dispatchEvent`** — 4 overloads wrapping `DatastarEvent.dispatchEvent` into `ZIO[Datastar, Nothing, Unit]`
- **10 new tests** — 8 in `DatastarEventSpec` (basic usage, custom selector, all options, escaping, nested payload, raw Js, source overload, SSE format) and 2 in `ServerSentEventGeneratorSpec`

### Generated JavaScript

```javascript
// Default (document target):
document.dispatchEvent(new CustomEvent('event-name', {detail:{...}, bubbles:true, cancelable:false, composed:false}))

// With CSS selector target:
(function(){var el=document.querySelector('#my-el');if(el)el.dispatchEvent(new CustomEvent('event-name', {detail:{...}, bubbles:true, cancelable:false, composed:false}))})()
```

### DIFF
--- a/zio-http-datastar-sdk/src/test/scala/zio/http/datastar/ServerSentEventGeneratorSpec.scala
+++ b/zio-http-datastar-sdk/src/test/scala/zio/http/datastar/ServerSentEventGeneratorSpec.scala
@@ -3,9 +3,15 @@ package zio.http.datastar
 import zio._
 import zio.test._
 
+import zio.schema.{DeriveSchema, Schema}
+
 import zio.http.template2._
 
 object ServerSentEventGeneratorSpec extends ZIOSpecDefault {
+  case class Payload(name: String, value: Int)
+  implicit val payloadSchema: Schema[Payload] = DeriveSchema.gen[Payload]
+  case class Info(msg: String)
+  implicit val infoSchema: Schema[Info]       = DeriveSchema.gen[Info]
 
   override def spec = suite("ServerSentEventGeneratorSpec")(
     suite("EventType")(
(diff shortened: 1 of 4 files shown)


zio-http-datastar-sdk/src/test/scala/zio/http/datastar/ServerSentEventGeneratorSpec.scala
```diff
@@ -591,6 +597,35 @@ object ServerSentEventGeneratorSpec extends ZIOSpecDefault {
...
+      },
+      test("should send dispatchEvent with options") {
+        for {
+          datastar <- Datastar.make
+          queue   = datastar.queue
+          options = DispatchEventOptions(source = Some(selector"#target"), bubbles = false, cancelable = true)
+          _ <- ServerSentEventGenerator.dispatchEvent("custom", Info("hi"), options).provide(ZLayer.succeed(datastar))
+          event <- queue.take
+        } yield assertTrue(
+          event.data.contains("querySelector('#target')"),
```
987Nabil: never string contains tests. only exact matching

PR #4015, opened 2026-03-09
perf: use Scheme.unsafe.decode in URL parsing hot path

987Nabil:
## Summary

Follow-up to #3997 — eliminates one `Option` allocation in the URL parsing hot path.

### Changes

**Use `Scheme.unsafe.decode` directly in `fromAbsoluteURIOrNull`**
- `Scheme.decode(uri.getScheme).orNull` wraps in `Some(scheme)` then unwraps with `.orNull`
- `Scheme.unsafe.decode` returns the value directly — same behavior, no `Option` allocation
- Saves 1 `Some(Scheme)` allocation per request with absolute URLs

### DIFF
--- a/zio-http/shared/src/main/scala/zio/http/URL.scala
+++ b/zio-http/shared/src/main/scala/zio/http/URL.scala
@@ -20,7 +20,7 @@ import java.net.{MalformedURLException, URI}
 
 import scala.util.control.NonFatal
 
-import zio.{Config, ZIO, durationInt}
+import zio.{Config, Unsafe, ZIO, durationInt}
 
 import zio.http.URL.Location.Relative
 import zio.http.URL.{Fragment, Location}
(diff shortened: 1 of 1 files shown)


zio-http/shared/src/main/scala/zio/http/URL.scala
```diff
@@ -321,13 +321,13 @@ object URL {
     val connection = URL.Location.Absolute(scheme, host, portOpt)
     val path2      = Path.decodeRaw(rawPath)
     val path3      = if (rawPath.nonEmpty) path2.addLeadingSlash else path2
-    URL(path3, connection, QueryParams.decode(uri.getRawQuery), Fragment.fromURI(uri))
```
hearnadam, line 324: If these are always None, they won't be allocating.

987Nabil: You're right — `Option(null)` returns the `None` singleton, no allocation. Reverted the fragment change. Kept only the `Scheme.unsafe.decode` optimization which does avoid a `Some` boxing.

Issue #4024, opened 2026-03-13
bug(datastar): `astToExpression` doesn't escape single quotes in string values


987Nabil:
## Description

`SignalUpdate.astToExpression` wraps string values in single quotes but does not escape single quotes within the string. This produces invalid JavaScript expressions.

## Example

Given a signal value containing a single quote (e.g., `O'Brien`):

```scala
val signal = Signal[String]("name")
val update = signal := "O'Brien"
update.toExpression // Produces: {name: 'O'Brien'} — broken JS!
```

The expected output should escape the inner quote:

```
{name: 'O\'Brien'}
```

## Current code

```scala
private[signal] def astToExpression(ast: zio.json.ast.Json): String = ast match {
  // ...
  case zio.json.ast.Json.Str(value) => s"'$value'"  // No escaping!
  // ...
}
```

## Expected

```scala
case zio.json.ast.Json.Str(value) => s"'${value.replace("'", "\\'")}'"
```

Or alternatively, use backtick template literals which only need `${}` escaping.

## File

`zio-http-datastar-sdk/src/main/scala/zio/http/datastar/signal/Signal.scala` (in `SignalUpdate` companion object)

Issue #4023, opened 2026-03-13
bug(datastar): `Signal.ref` incorrectly restricts to primitive types and has race condition


987Nabil:
## Description

`Signal.ref` has two issues:

### 1. Incorrect primitive type restriction

`Signal.ref` throws a `RuntimeException` if the signal's schema is not a `Schema.Primitive`. This is incorrect — Datastar's `$signalName` expression syntax works for **any** type, not just primitives. A signal referencing a complex object (e.g., `$user`) is perfectly valid in Datastar expressions.

### 2. Race condition on mutable `var ref0`

The caching mechanism uses a mutable `var ref0` without synchronization. In concurrent code, multiple threads could read/write `ref0` simultaneously, leading to undefined behavior.

## Current code

```scala
private var ref0: String = null

def ref: String = {
  if (ref0 == null && schema.isInstanceOf[Schema.Primitive[_]]) ref0 = name.ref
  else if (!schema.isInstanceOf[Schema.Primitive[_]])
    throw new RuntimeException(s"Signal.ref is only supported for primitive types, got: $schema")
  ref0
}
```

## Expected

Remove the primitive type restriction and fix the caching:

```scala
lazy val ref: String = name.ref
```

Using `lazy val` is thread-safe (Scala guarantees it) and removes the unnecessary type restriction. `SignalName.ref` is already a `val` so the lazy is just for consistency with the `Signal` being a case class.

## File

`zio-http-datastar-sdk/src/main/scala/zio/http/datastar/signal/Signal.scala`

Issue #4022, opened 2026-03-13
bug(datastar): `DatastarRequestCancellation` has invalid enum values and case mismatch


987Nabil:
## Description

`DatastarRequestCancellation` has two problems:

### 1. Invalid `Cleanup` value

The current implementation uses `Auto` and `Disabled`, but also accepts arbitrary strings as `Custom(Js(...))`. The previous implementation had a `Cleanup` variant — however, the Datastar protocol defines only:

```
requestCancellation?: 'auto' | 'disabled' | AbortController
```

There is no `cleanup` value in the Datastar specification.

**Source**: [Datastar `FetchArgs` type](https://github.com/starfederation/datastar/blob/main/library/src/plugins/actions/fetch.ts)

### 2. PascalCase vs lowercase mismatch

The `Schema` transform encodes/decodes `"Auto"` and `"Disabled"` (PascalCase), but Datastar expects lowercase `"auto"` and `"disabled"`.

## Current code

```scala
implicit val schema: Schema[DatastarRequestCancellation] = Schema[String].transform[DatastarRequestCancellation](
  {
    case "Auto"     => Auto
    case "Disabled" => Disabled
    case other      => Custom(Js(other))
  },
  {
    case Auto          => "Auto"      // Bug: should be "auto"
    case Disabled      => "Disabled"  // Bug: should be "disabled"
    case Custom(value) => value.value
  },
)
```

## Expected

Encode/decode should use lowercase strings matching the Datastar protocol:

```scala
{
  case "auto"     => Auto
  case "disabled" => Disabled
  case other      => Custom(Js(other))
},
{
  case Auto          => "auto"
  case Disabled      => "disabled"
  case Custom(value) => value.value
},
```

## File

`zio-http-datastar-sdk/src/main/scala/zio/http/datastar/model/DatastarRequest.scala`

Issue #4021, opened 2026-03-13
bug(datastar): `DatastarRetry` missing `Always` and `Never` enum values


987Nabil:
## Description

`DatastarRetry` only defines `Auto` and `Error`, but the Datastar protocol supports 4 values:

```
'auto' | 'error' | 'always' | 'never'
```

**Source**: [Datastar `FetchArgs` type](https://github.com/starfederation/datastar/blob/main/library/src/plugins/actions/fetch.ts) — `retry?: 'auto' | 'error' | 'always' | 'never'`

## Current code

```scala
sealed trait DatastarRetry
object DatastarRetry {
  case object Auto  extends DatastarRetry
  case object Error extends DatastarRetry
  // Missing: Always, Never
}
```

## Expected

```scala
sealed trait DatastarRetry
object DatastarRetry {
  case object Auto   extends DatastarRetry
  case object Error  extends DatastarRetry
  case object Always extends DatastarRetry
  case object Never  extends DatastarRetry
}
```

The `Schema` transform should also be updated to handle `"always"` and `"never"` string values.

## File

`zio-http-datastar-sdk/src/main/scala/zio/http/datastar/model/DatastarRequest.scala`

PR #4025, opened 2026-03-13
fix: Datastar SDK — readSignals, enum values, Signal.ref, quote escaping

### LINKED ISSUE #4020: bug(datastar): `readSignals` reads GET signals from header instead of query parameter
## Bug Description

`readSignals` for GET requests reads signals from the `datastar` **header**, but the Datastar JS client sends them as a **query parameter** `?datastar={...}`.

## Current Behavior (wrong)

```scala
// DatastarPackageBase.scala
def readSignals[T: Schema](request: Request): IO[String, T] =
  if (request.method == Method.GET) {
    ZIO.fromEither {
      request
        .header[String]("datastar")  // ← reads from HEADER
```

## Expected Behavior

For GET requests, Datastar sends signals as a query parameter, not a header.

From the Datastar source ([`fetch.ts:113-115`](https://github.com/starfederation/datastar/blob/main/library/src/plugins/actions/fetch.ts)):

```typescript
if (method === 'GET') {
  queryParams.set('datastar', body)  // ← sends as QUERY PARAMETER
} else {
  req.body = body                    // ← body for POST/PUT/PATCH/DELETE
}
```

The `Datastar-Request: true` header is just a boolean flag indicating the request comes from Datastar — it does NOT contain the signal data.

## Fix

```scala
def readSignals[T: Schema](request: Request): IO[String, T] =
  if (request.method == Method.GET) {
    ZIO.fromEither {
      request
        .queryParam("datastar")        // ← read from query parameter
        .toRight(new Exception("Missing 'datastar' query parameter"))
        .left.map(_.getMessage())
        .flatMap(_.fromJson[T](zio.schema.codec.JsonCodec.jsonDecoder(Schema[T])))
    }
  } else {
    request.body.asJson[T].mapError(_.getMessage)
  }
```

## Test Fix

`ReadSignalsSpec` also needs to be updated — it currently adds signals as a header in test requests, which matches the buggy implementation:

```scala
// Current (wrong — matches buggy code):
val request = Request
  .get(URL.root / "test")
  .addHeader(Header.Custom("datastar", jsonData))

// Should be:
val request = Request
  .get((URL.root / "test").withQueryParams(QueryParams("datastar" -> jsonData)))
```

## Additional Bugs Found

While reviewing the datastar SDK, the following additional bugs were identified:

1. **`DatastarRetry` is missing 2 of 4 valid values** — Datastar supports `'auto' | 'error' | 'always' | 'never'` but the SDK only models `Auto` and `Error`.
2. **`DatastarRequestCancellation` has invalid `Cleanup` value** — Datastar only supports `'auto' | 'disabled' | AbortController`. Also uses PascalCase (`"Auto"`) but Datastar expects lowercase (`"auto"`).
3. **`Signal.ref` throws for non-primitive types** — Datastar `$signalName` works for any type. The runtime restriction is incorrect.
4. **`SignalUpdate.astToExpression` doesn't escape `'` in strings** — generates broken JS for values containing single quotes.

987Nabil:
## Summary

Fix 5 bugs in the Datastar SDK to align with the [official Datastar JS SDK](https://github.com/starfederation/datastar/blob/main/library/src/plugins/actions/fetch.ts) behavior.

### 1. `readSignals` reads GET signals from header instead of query parameter (#4020)
The Datastar JS SDK sends signals as a `datastar` query parameter for GET requests, not as a header. The Scala SDK was reading from the header, causing GET signal reads to always fail.

### 2. `DatastarRetry` missing `Always` and `Never` enum values (#4021)
The JS SDK supports `retry: 'auto' | 'error' | 'always' | 'never'`, but the Scala SDK only had `Auto` and `Error`.

### 3. `DatastarRequestCancellation` invalid PascalCase enum values (#4022)
The JS SDK uses lowercase `"auto"` and `"disabled"`, but the Scala SDK was encoding/decoding `"Auto"` and `"Disabled"`, causing client-side mismatches.

### 4. `Signal.ref` incorrect primitive restriction + race condition (#4023)
`Signal.ref` threw a runtime exception for non-primitive types (e.g., case classes) and used an unsynchronized mutable `var`. Replaced with `lazy val` delegating to `name.ref`, which is thread-safe and works for all types.

### 5. `astToExpression` single-quote escaping bug (#4024)
String values containing single quotes would produce invalid JavaScript expressions. Added proper escaping.

## Files Changed
- `DatastarPackageBase.scala` — `readSignals` GET path
- `DatastarRequest.scala` — `DatastarRetry` + `DatastarRequestCancellation`
- `Signal.scala` — `ref` field + `astToExpression`
- `ReadSignalsSpec.scala` — Updated GET tests to use query params
- `DatastarRequestSpec.scala` — Updated assertion to match lowercase enum

## Testing
All 406 tests pass (`sbt zioHttpDatastarSdk/test`).

Fixes #4020, #4021, #4022, #4023, #4024

### DIFF
--- a/zio-http-datastar-sdk/src/main/scala/zio/http/datastar/DatastarPackageBase.scala
+++ b/zio-http-datastar-sdk/src/main/scala/zio/http/datastar/DatastarPackageBase.scala
@@ -962,9 +962,8 @@ trait DatastarPackageBase extends Attributes {
     if (request.method == Method.GET) {
       ZIO.fromEither {
         request
-          .header[String]("datastar")
-          .left
-          .map(_.getMessage())
+          .queryParam("datastar")
+          .toRight("Missing 'datastar' query parameter")
           .flatMap(_.fromJson[T](zio.schema.codec.JsonCodec.jsonDecoder(Schema[T])))
       }
     } else {
--- a/zio-http-datastar-sdk/src/main/scala/zio/http/datastar/model/DatastarRequest.scala
+++ b/zio-http-datastar-sdk/src/main/scala/zio/http/datastar/model/DatastarRequest.scala
@@ -204,13 +204,13 @@ object DatastarRequestCancellation {
 
   implicit val schema: Schema[DatastarRequestCancellation] = Schema[String].transform[DatastarRequestCancellation](
     {
-      case "Auto"     => Auto
-      case "Disabled" => Disabled
+      case "auto"     => Auto
+      case "disabled" => Disabled
       case other      => Custom(Js(other))
     },
     {
-      case Auto          => "Auto"
-      case Disabled      => "Disabled"
+      case Auto          => "auto"
+      case Disabled      => "disabled"
       case Custom(value) => value.value
     },
   )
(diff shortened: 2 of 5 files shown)


987Nabil:
Splitting into individual PRs per bug fix for cleaner review.

Issue #3551, opened 2025-06-13
Move Netty to separate module


987Nabil:

987Nabil:
This is an internal bounty

Issue #3472, opened 2025-04-30
Split into multiple modules


987Nabil:
Currently, the zio-http artifact contains a lot of different parts of zio-http exclusively. They are not separate maven artifacts.

We want to change this, to support future changes/features.

There should be at least these modules that are published into maven. 

1. core
2. endpoint
3. netty

Should we have client and server in different modules?

987Nabil:
Internal bounty

glitchdoescode:
i want to work on this

sangwaboi:
Hi @987Nabil, I’d like to take this on and claim the bounty.

Proposing my PR (keeping backward compatibility for users of the existing `zio-http` artifact):
 
Publish three new artifacts:-
  - dev.zio %% zio-http-core (JS/JVM): core HTTP primitives and abstractions
  - dev.zio %% zio-http-endpoint (JS/JVM): declarative endpoints + OpenAPI/GRPC
  - dev.zio %% zio-http-netty (JVM): Netty-based client/server drivers + shading
 
 Keep dev.zio %% zio-http as a meta artifact depending on {core, endpoint, netty} so current users can continue to add only `zio-http`.

Implementation will be:-
- Extracting `zio.http.endpoint/**` to `zio-http-endpoint` crossProject.
- Extracting `zio.http.netty/**` to `zio-http-netty` (JVM), and moving shading rules here.
- Decouple core from Netty (e.g., move default client/server layers currently importing `Netty*` into the Netty module; core exposes abstractions ONLY).
- Update `testkit`, `cli`, `gen`, `docs`, `example` dependencies accordingly.
- Add MiMa where applicable and a migration note in the docs.


Also, do you want an additional split for client/server (e.g., `zio-http-client` cross, `zio-http-server` JVM), or should we defer that to a follow-up?

Please assign me to the issue; I’ll reference this plan and iterate based on your feedback.

987Nabil:
@sangwaboi I tried this and it is not possible. This will be part of the next major release and is no bounty open to the public

sangwaboi:
Thank you for the reply, @987Nabil. I totally understand. I thought I’d give it a try once since I’ve been studying the codebase for a while
.
I’d still love to contribute, regardless of whether there's a bounty.

987Nabil:
There is no work for you here. This will need to be part of a bigger effort and needs some insight and decisions power you simply do not have. This is a a task that only maintainers can do. Sorry

ck3960250:
[...]

PR #4034, opened 2026-03-14
feat: move Netty implementation to separate zio-http-netty module (#3…

singhharsh1708:
## Move Netty to Separate Module (#3551)

This PR implements sub-issue **#3551** as part of the artifact modularization effort described in **#3470**.

The goal of this change is to separate the Netty implementation from the main `zio-http` artifact so that transport-specific code is modular and can evolve independently.

---

### Summary

A new `zio-http-netty` module has been introduced to host all Netty-based server and client implementations. Previously, these sources lived inside the `zio-http/jvm` project.

By moving Netty into its own artifact, we reduce coupling between core HTTP abstractions and the Netty transport layer, making the project easier to maintain and extend.

---

#### 1. New `zio-http-netty` sbt project

* Introduced a new module:

  * `zio-http-netty`
* This module now contains all Netty-related implementations.

#### 2. Source migrations

All sources under:

```
zio-http/jvm/src/main/scala/zio/http/netty
```

were moved to:

```
zio-http-netty/src/main/scala/zio/http/netty
```

This includes all Netty server, client, model, and socket related code (25+ files).

#### 3. Public API additions

New API objects were introduced in the `zio.http` package to provide access to Netty implementations.

**NettyServer**

* `NettyServer.live`
* `NettyServer.customized`
* `NettyServer.default`
* `NettyServer.defaultWithPort()`

**NettyClient**

* `NettyClient.live`
* `NettyClient.customized`
* `NettyClient.default`
* `NettyClient.configured()`

These APIs provide a clear entry point for using the Netty driver while keeping the core module transport-agnostic.

#### 4. Platform-specific stubs

The following platform-specific components were updated:

* `ServerPlatformSpecific`
* `DriverPlatformSpecific`
* `ZClientPlatformSpecific`

These now throw `UnsupportedOperationException` when used without the Netty module.

They were implemented as **`def` instead of `val`** to avoid class initialization failures during JVM startup.

Additionally:

* `Server.default` was changed to a `lazy val` to prevent premature initialization.

#### 5. Test migrations

All Netty integration tests were moved to the new module:

```
zio-http-netty/src/test
```

This ensures that Netty-specific functionality is tested only within the Netty module.

---

### Test Results

All existing tests were executed after the refactor.

```
zioHttpJVM/test
1191 passed
0 failed
```

```
zioHttpNetty/test
552 passed
```

A few failures were observed in `NettyConnectionPoolSpec` timeout tests.
These failures appear to be **pre-existing flaky tests**, as they also occur on the original branch before this refactor.

---

### Outcome

This change:

* Decouples Netty transport implementation from the core HTTP abstractions
* Establishes the `zio-http-netty` module as the Netty driver implementation
* Maintains backward compatibility through the new `NettyServer` and `NettyClient` APIs
* Keeps the project aligned with the broader modularization goals of **#3470**

---

### Next Steps

This PR focuses only on the Netty module extraction.
Future steps in the modularization effort may include:

* moving the endpoint DSL to a separate module
* continuing refactoring of encoding implementations
* further refining module boundaries
/claim #3472


singhharsh1708:
Thanks for the review! All comments have been addressed and the branch has been updated.

### Fixes applied

1. **Platform-specific stubs**

   * Replaced direct `throw new UnsupportedOperationException(...)` calls with:

     ```scala
     ZLayer.fail(new UnsupportedOperationException(...))
     ```
   * This keeps failures inside ZIO’s typed error channel instead of surfacing as fiber defects.
   * Updated in:

     * `ServerPlatformSpecific`
     * `DriverPlatformSpecific`
     * `ZClientPlatformSpecific`
   * Removed now-unnecessary `@nowarn` annotations.

2. **Invalid port generator**

   * Fixed invalid TCP port range in `HttpGen.scala`.
   * Updated:

     ```scala
     Gen.int(0, 65536)
     ```

     →

     ```scala
     Gen.int(1, 65535)
     ```
   * Applied in both `zio-http-netty` and `zio-http` test generators.

3. **Scaladoc example fix**

   * `Server.defaultConfig` was incorrect.
   * Updated the example in `NettyServer.scala` to use:

     ```scala
     ZLayer.succeed(Server.Config.default)
     ```

4. **TLS test credentials documentation**

   * Added `zio-http-netty/src/test/resources/README.md` explaining that all included TLS keys/certs are **non-production test credentials** used only for automated testing.

5. **Examples updated to use Netty server**

   * Since `Server.default` is now a failing stub when Netty is not present, examples were updated to use `NettyServer.default`.
   * Fixed the remaining examples that were missed in the initial refactor:

     * `ClientServer`
     * `ClientServerExample`
     * `Interrupt`
     * `MultipartFormData`
     * `SSEServer`
     * `ServerSentEventEndpoint`
     * `ServerSentEventAsJsonEndpoint`
     * `endpoint/CliExamples`

All fixes have been pushed to the PR branch.

PR #4044, opened 2026-03-20
Apply `HandlerAspect` on `Route` after path decoding

### LINKED ISSUE #3141: ClassCassException when using HandlerAspect and Path parameters.

Exceluyi:
Fixes #3141 
## Summary

Fixes `HandlerAspect` usage on single routes with path parameters.

The original issue expression:

```scala
Method.GET / "base" / string("id") -> handler(...) @@ aspect
```

applies `@@` to the `Route`, not directly to the `Handler`. The previous fix changed `Handler` internals and broke MiMa, but the bug is really at the `Route` layer.

## Changes

- add `Route.@@` overloads for `Middleware`, `HandlerAspect[Env, Unit]`, and contextual `HandlerAspect[Env, Ctx]`
- apply route aspects through `Route.transform`, so they run on the request-only handler after path parameters have already been decoded
- add a `Route.@@[Env0]` overload that forwards to `Routes.ApplyContextAspect`

/claim #3141

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/HandlerAspectSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/HandlerAspectSpec.scala
@@ -45,6 +45,40 @@ object HandlerAspectSpec extends ZIOSpecDefault {
           bodyString <- response.body.asString
         } yield assertTrue(bodyString == "1 test")
       },
+      test("HandlerAspect with context works for routes with a path parameter") {
+        case class WebSession(id: Int)
+
+        val sessionAspect: HandlerAspect[Any, Option[WebSession]] =
+          HandlerAspect.interceptIncomingHandler(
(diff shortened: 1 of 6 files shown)


Exceluyi:
changed my approach completely at https://github.com/zio/zio-http/pull/4044/commits/89055ca432026c870b53c0e7f9942cf4f4e6d870 because it broke other code

987Nabil:
This issue is not solvable in 3.x. It can't be done and keep bin compat. And your approach is on Route not Handler and does not solve anything

PR #4062, opened 2026-03-25
Add Mill 1.1.4 build

987Nabil:
## Summary

- Add Mill 1.1.4 build alongside existing sbt build
- All 8 core modules compile with `./mill __.compile` on both Scala 2.13.18 and 3.3.7
- Remove `unroll-plugin`/`unroll-annotation` (not needed for new major version)

### Modules defined

| Module | Platform | Dependencies |
|--------|----------|-------------|
| `core` | JVM + JS | zio, zio-streams, zio-schema |
| `server` | JVM + JS | core |
| `client` | JVM + JS | core |
| `endpoint` | JVM + JS | core, client |
| `fetchClient` | JS only | client |
| `nettyCore` | JVM only | core, client, netty |
| `nettyServer` | JVM only | server, nettyCore |
| `nettyClient` | JVM only | client, nettyCore |

### Usage

```bash
./mill __.compile                    # compile all
./mill 'core.jvm[2.13.18].compile'  # compile specific
./mill 'core.jvm[3.3.7].test'       # test specific
```

### Notes

- sbt build is kept for now (dual build period)
- Source directories unchanged — Mill build points to existing `shared/`, `jvm/`, `js/` layout
- `scala-2`/`scala-3` version-specific source dirs handled via `crossScalaVersion` conditional
- CI not yet updated for Mill — still uses sbt


russwyte:
Just curious - what is the motivation? (not complaining in the least)

987Nabil:
I think mill is simpler (to maintain) and most likely faster. And sbt 2 takes so long. And mill means I can get rid of Scala 2.12

PR #4069, opened 2026-03-27
Fix hanging request streaming under concurrent load

davidlar:
After updating zio-http to 3.9.0+ one of our services started to hang for a lot of requests. That service uses request streaming to proxy large payloads to different storage backends.

Tracked down the bug to a commit that fixed "prevent response header interleaving on keep-alive connections".
The problem was that auto read was manipulated outside of the netty event loop thread and this caused a race problem with request streaming (also changes auto read).

Created a test case as a reproducer of the problem as well as a fix that handles auto read in the event loop thread.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
@@ -196,13 +196,26 @@ private[zio] final case class ServerInboundHandler(
             ctx.writeAndFlush(jResponse)
             NettyBodyWriter.writeAndFlush(response.body, contentLength, ctx) match {
               case Some(bodyTask) =>
-                val channelConfig    = ctx.channel().config()
-                val previousAutoRead = channelConfig.isAutoRead
-                if (previousAutoRead) channelConfig.setAutoRead(false)
-                Some(bodyTask.ensuring(ZIO.succeed {
-                  if (previousAutoRead) {
-                    channelConfig.setAutoRead(true)
-                    ctx.channel().read()
+                // Disable auto-read while the response body is being streamed to prevent
+                // the next request's headers from being read and interleaved.
+                // All autoRead manipulation must happen on the event loop thread to avoid
+                // racing with AsyncBodyReader (used for request streaming)
+                val channel          = ctx.channel()
+                val previousAutoRead = new java.util.concurrent.atomic.AtomicBoolean(true)
+                channel
+                  .eventLoop()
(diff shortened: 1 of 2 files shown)


zio-http/jvm/src/main/scala/zio/http/netty/server/ServerInboundHandler.scala
```diff
@@ -196,13 +196,26 @@ private[zio] final case class ServerInboundHandler(
...
+                val channel          = ctx.channel()
+                val previousAutoRead = new java.util.concurrent.atomic.AtomicBoolean(true)
+                channel
+                  .eventLoop()
+                  .execute(() => {
+                    val prev = channel.config().isAutoRead
+                    previousAutoRead.set(prev)
+                    if (prev) channel.config().setAutoRead(false): Unit
+                  })
+                Some(bodyTask.ensuring(ZIO.async[Any, Nothing, Unit] { cb =>
```
guizmaii, line 212: Why do you need `ZIO.async` here?

davidlar:
```
channelConfig.setAutoRead(true)
ctx.channel().read()
``` 
needs to happen in the netty event loop thread. Without ZIO.async it will run in the ZIO executor thread and the race problem will remain.

guizmaii: How does `ZIO.async` allow it to run on Netty's threads? 🤔

davidlar: Well it's actually `channel.eventLoop().execute` that schedules the code block to be run on nettys event loop thread. Then when the code block is fully executed in netty, cb(Exit.unit) is called to to resume the zio fiber. 

Only when the code block is completed, it's safe to process the next request. That's why we need ZIO.async instead of e.g. ZIO.succeed. ZIO.succeed will not wait for the code block to complete

guizmaii: Indeed! My bad. I missed that the `cb(Exit.unit)` call was inside the `channel.eventLoop().execute`
Thanks for your explanations 🙂

davidlar: Thanks for reviewing! And I could probably have been more clear :-)

987Nabil (approved):

PR #4112, opened 2026-04-30
Add `AuthType.Cookie(name)` for endpoint cookie authentication

guizmaii:
`AuthType.Custom(HeaderCodec.cookie)` already let an endpoint accept auth via a cookie header, but it returned the entire `Header.Cookie` value (a `NonEmptyChunk[Cookie.Request]`) and required the user to hand-roll the lookup-by-name plus the encode-side codec. This adds a typed shortcut that does both:

```scala
Endpoint(Method.GET / "download" / fileId)
  .auth(AuthType.Bearer | AuthType.Cookie("session"))
  .unauthorizedStatus(Status.Unauthorized)
```

The handler / context middleware receives the cookie value as a `String`. If the named cookie is missing, the codec fails to decode and the endpoint responds with `unauthorizedStatus` -- matching the existing behaviour for `Basic`/`Bearer`/`Digest` when the `Authorization` header is absent. Hooked into `Endpoint`'s missing-header handler via a small `authHeaderNames` helper that walks `Or` / `WithStatus` / `ScopedAuth` wrappers, so composed auth types behave consistently too.


guizmaii:
### Code review

Found 2 issues:

1. **Named cookie missing while `Cookie` header is present leaks a generic codec error instead of returning `unauthorizedStatus`.** The Scaladoc on `AuthType.Cookie` promises that *"if the named cookie is missing from the request, the codec fails to decode and the endpoint responds with `AuthType.unauthorizedStatus`"*. There are two failure paths, and only one is handled:
   - `Cookie` header entirely absent → `HttpCodecError.MissingHeader("cookie")` → caught by the new `authHeaderNames` branch ✓
   - `Cookie` header present but the named cookie not in it → `transformOrFailLeft` returns `Left(s"Missing cookie '$name'")` → becomes `CustomError("SchemaTransformationFailure", "Missing cookie 'session'")` → does **not** match the existing `endsWith(" auth required")` gate, so it falls through to the generic codec-error response.

   The existing `Basic`/`Bearer`/`Digest` cases avoid this by formatting their codec errors as e.g. `"Bearer auth required"`. Easiest fix: change the `Left` message to end with `" auth required"` (e.g. `s"Cookie '$name' auth required"`). The new tests don't cover this path — both "missing cookie" tests omit the `Cookie` header entirely, so the `SchemaTransformationFailure` branch is never exercised.

   https://github.com/zio/zio-http/blob/b758aa68e15d220c151301dc0c5768dac771f120/zio-http/shared/src/main/scala/zio/http/endpoint/AuthType.scala#L106-L113

   https://github.com/zio/zio-http/blob/b758aa68e15d220c151301dc0c5768dac771f120/zio-http/shared/src/main/scala/zio/http/endpoint/Endpoint.scala#L501-L510

2. **`AuthType.Cookie` is not handled in `OpenAPIGen`, so endpoints using it produce no security scheme/requirement.** Both `loop` functions (in `endpointSecurity` and the `httpSecuritySchemes` builder) match on `Basic | Bearer | Digest`, `Custom`, `Or`, `WithStatus`, `ScopedAuth`, with a `case _ => Nil`/`ListMap.empty` catch-all. `AuthType.Cookie` silently falls through, so generated OpenAPI specs lose security info that the existing `AuthType.Custom(HeaderCodec.cookie)` workaround already produces (verified by `OpenAPIGenSpec`'s "cookie" apiKey test). Adding a `case AuthType.Cookie(name) => ...` to both loops (using `SecurityScheme.ApiKey` with `In.Cookie`, mirroring `customAuthSchemes`' existing cookie branch) would close this gap.

   https://github.com/zio/zio-http/blob/b758aa68e15d220c151301dc0c5768dac771f120/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPIGen.scala#L756-L775

   https://github.com/zio/zio-http/blob/b758aa68e15d220c151301dc0c5768dac771f120/zio-http/shared/src/main/scala/zio/http/endpoint/openapi/OpenAPIGen.scala#L1059-L1087

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #4124, opened 2026-05-08
Remove sbt build support from v4.x

987Nabil:
## Summary
- remove sbt build definitions, example sbt projects, and sbt-only grpc tooling from the v4.x branch
- switch local tooling, profiling, and GitHub workflows to mill-only paths
- update docs and published website references so build instructions no longer point at sbt

## Verification
- ./mill __.compile
- ./mill 'core.jvm[2.13.18].test'
- ./mill 'core.jvm[3.3.7].test'
- ./mill profiling.assembly
- ./build.sh


swoogles:
Is there a link to any discussion around this decision? I'm surprised, but congrats lihaoyi 😄

PR #4139, opened 2026-05-11
Fix HandlerAspect with path params and request input

### LINKED ISSUE #3141: ClassCassException when using HandlerAspect and Path parameters.

ritwikareddykancharla:
Fixes #3141

## Summary

This fixes the `ClassCastException` that happens when a context-producing `HandlerAspect` is applied to a route handler that accepts both path parameters and the incoming `Request`.

Previously, `Handler.@@` coerced aspect-wrapped handlers down to `Request => Response`. That loses the original route input shape, so handlers with inputs like `(String, Request)` could be invoked with only a `Request` and fail at runtime.

## Changes

- Preserve the original handler input when applying `HandlerAspect`.
- Add request extraction/replacement support for route inputs that end with `Request`, such as `(A, Request)` and `(A, B, Request)`.
- Keep context injection behavior unchanged for plain `Request` handlers.
- Add a regression test covering a path parameter, `Request`, and `withContext` together.

## Verification

Ran locally with a portable JDK 17 and the repo's sbt 1.12.11 launcher:

```text
zioHttpJVM/testOnly zio.http.HandlerAspectSpec
4 tests passed. 0 tests failed. 0 tests ignored.
```

### DIFF
--- a/zio-http/jvm/src/test/scala/zio/http/HandlerAspectSpec.scala
+++ b/zio-http/jvm/src/test/scala/zio/http/HandlerAspectSpec.scala
@@ -45,6 +45,29 @@ object HandlerAspectSpec extends ZIOSpecDefault {
           bodyString <- response.body.asString
         } yield assertTrue(bodyString == "1 test")
       },
+      test("HandlerAspect with context preserves path parameters and request input") {
+        final case class WebSession(id: Int)
+
+        val handlerAspect: HandlerAspect[Any, Option[WebSession]] =
+          HandlerAspect.interceptIncomingHandler(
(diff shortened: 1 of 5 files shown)


ritwikareddykancharla:
Closing in favor of #4138, which already addresses #3141 and has passing CI.

PR #4178, opened 2026-06-22
fix: fail WebSocket client connection when the server rejects the upgrade

### LINKED ISSUE #3464: WebSocket client: request should fail if server doesn't return "Switching Protocols"
Hi,

Today I was trying to connect to a WebSocket API using `client.url(url).socket(…)`, but the server decided that it doesn't want to play the WebSocket game today and sent me a "Bad Request" response. ZIO-HTTP doesn't care: it simply doesn't launch the WebSocket app and returns the "Bad Request" response to me.
I think the `socket` method should raise an error for any response other than "Switching Protocols". After all, the whole purpose of the `socket` method is to launch a WebSocket app, and if the method cannot fulfil its purpose, it should raise an error.

henricook:
### What

`client.url(url).socket(app)` (and `WebSocketApp#connect`) used to return successfully with whatever HTTP response the server sent, even when the server refused the WebSocket upgrade. For example, if the server replied with `400 Bad Request`, the WebSocket app was silently not started and the response was handed back as a success, leaving the caller unaware that the connection never became a WebSocket.

### Fix

`WebSocketClientInboundHandler` now only continues the upgrade when the server responds with `101 Switching Protocols`. For any other status it fails the connection (and the handshake-completion promise) with a `WebSocketHandshakeException` describing the status that was received instead.

### Tests

Added a `WebSocketSpec` test that points the client at a server returning `400 Bad Request` and asserts that `connect` fails rather than returning the response.

Fixes #3464.

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/WebSocketClientInboundHandler.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/WebSocketClientInboundHandler.scala
@@ -24,7 +24,8 @@ import zio.http.internal.ChannelState
 import zio.http.netty.NettyResponse
 
 import io.netty.channel.{ChannelHandlerContext, SimpleChannelInboundHandler}
-import io.netty.handler.codec.http.FullHttpResponse
+import io.netty.handler.codec.http.websocketx.WebSocketHandshakeException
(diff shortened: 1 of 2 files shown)


henricook:
Thanks @987Nabil !

PR #4232, opened 2026-07-26
Regenerate expired mTLS test certificates

guizmaii:
## Problem

CI is red on every PR since 2026-07-22: `DualSSLSpec`'s *"succeed when client has the server certificate and client certificate is configured"* fails on all matrix jobs with a `PrematureChannelClosureException` (TLS handshake aborted).

Root cause: `zio-http/jvm/src/test/resources/client.crt` and `client_other.crt` expired on **2026-07-22**, so the server now rejects the client certificate during the mTLS handshake. (Last green `main` run was 2026-07-20.)

## Solution

Regenerate the certificate chain with 100-year validity so this can't recur:

- New CA `ca.pem` (the old CA's private key was never committed, so re-signing under it wasn't possible). Same subject as before.
- `client.crt` re-issued from the **existing** `client.key`, signed by the new CA, with the exact subject the test asserts (`O=client1,ST=Some-State,C=AU`).
- `client_other.crt` re-issued self-signed from the existing `client_other.key` (it must remain untrusted by the server for the negative test).

Private keys are unchanged. `ServerClientJKSMutualSSLSpec` uses the separate JKS stores under `jks_keystore_truststore/` and is unaffected.

Commands used, for the next time:

```bash
openssl genrsa -out ca.key 2048
openssl req -new -x509 -days 36500 -key ca.key -out ca.pem \
  -subj "/C=NL/ST=Noord-Holland/L=Amsterdam/O=Internet Widgits Pty Ltd"
openssl req -new -key client.key -subj "/C=AU/ST=Some-State/O=client1" -out client.csr
openssl x509 -req -in client.csr -CA ca.pem -CAkey ca.key -CAcreateserial -days 36500 -out client.crt
openssl req -new -x509 -days 36500 -key client_other.key \
  -subj "/C=AU/ST=Some-State/O=Internet Widgits Pty Ltd" -out client_other.crt
```

Verified locally: all 3 `DualSSLSpec` tests pass.


khajavi:
@guizmaii While working on the documentation, I ran into the same issue and fixed it on the fly without checking the existing PRs first. I only noticed your PR afterward. My apologies for the duplicate work! https://github.com/zio/zio-http/pull/4242

PR #4282, opened 2026-09-02
Add in-memory PEM bytes client SSL config variants

### LINKED ISSUE #4281: Support in-memory certificate/key bytes for client SSL config

`ClientSSLConfig` and `ClientSSLCertConfig` can only reference certificates and private keys by filesystem path (`FromCertFile`, `FromClientCertFile`) or classpath resource (`FromCertResource`, `FromClientCertResource`). There is no way to hand the client certificate/key material that already lives in memory.

This is a problem when the material comes from a secret manager / vault, an environment variable, or is generated at runtime. Today you have to write the bytes to a temporary file just to point the config at a path, which defeats the purpose of keeping private keys off disk.

Add byte-based variants alongside the existing ones:

- `ClientSSLConfig.FromCertBytes(certBytes)`: server trust material.
- `ClientSSLCertConfig.FromClientCertBytes(certBytes, keyBytes)`: client cert + key.
- `ClientSSLCertConfig.FromClientCertBytesWithPassword(certBytes, keyBytes, keyPassword)`: same, for a password-encrypted key.

`ClientSSLConverter` would feed these to Netty's `SslContextBuilder` via `ByteArrayInputStream`. The key-bearing variants should redact their key and password in `toString` so nothing sensitive lands in logs.

- Writing secrets to a temporary file and using the existing `FromCertFile` variants. This puts private keys on disk, which is exactly what we want to avoid.

**Additional context**
N/A

shuch3ng:
Adds byte-based alternatives to the existing file and resource based client SSL configs, so callers can supply certificates and keys that are already in memory (e.g. loaded from a vault or secret manager) instead of requiring an on-disk path or classpath resource.

Closes https://github.com/zio/zio-http/issues/4281

### DIFF
--- a/zio-http/jvm/src/main/scala/zio/http/netty/client/ClientSSLConverter.scala
+++ b/zio-http/jvm/src/main/scala/zio/http/netty/client/ClientSSLConverter.scala
@@ -16,7 +16,7 @@
 
 package zio.http.netty.client
 
-import java.io.{File, FileInputStream, InputStream}
+import java.io.{ByteArrayInputStream, File, FileInputStream, InputStream}
 import java.security.KeyStore
 import javax.net.ssl.{KeyManagerFactory, TrustManagerFactory}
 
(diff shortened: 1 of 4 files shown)


guizmaii (approved):
LGTM

@987Nabil could you have a look, please? 
@shuch3ng is working with me. It's something we need at work

PR #4313, opened 2026-09-14
Update Scala 3 to 3.9.0

guizmaii:
Updates the Scala 3 cross-build from 3.3.7 to 3.9.0.

## Heads-up for maintainers: this drops the 3.3 LTS line

`main` currently publishes `_3` artifacts built with the LTS compiler. Artifacts built with 3.9 cannot be consumed by a 3.3 compiler (TASTy is not forward compatible), so this is a support-policy change, not a routine patch bump. `v4.x` already made the same move (#4306). If dropping 3.3 is not wanted, the alternative is cross-building 3.3 *and* 3.9 rather than replacing — happy to redo it that way.

## What the bump actually required

**`-source:3.3`.** Scala 3.9 defaults to a much later source level. Without the pin, the migration warnings alone come to ~780: `_` as a wildcard type argument (520), `with` as a type operator (120), `x: _*` splices (56), implicit arguments passed without `using` (74). None of them are fixable in sources that also cross-compile on 2.12/2.13, where `?`, `&` and `x*` are not available. The pin keeps language semantics at 3.3 while using the 3.9 compiler.

**Per-version CI warning flag.** `-Xfatal-warnings` is a deprecated alias for `-Werror` on Scala 3, and the deprecation warning it emits is itself promoted to an error by the `-Werror` it enables — the build fails before compiling a line. `stdOptions` now picks `-Werror` on 3.x and `-Xfatal-warnings` on 2.x.

**unroll 0.1.12 → 0.2.0.** This is the substantive part. Scala 3.9 ships its own `scala.annotation.unroll` in the standard library and expands it in the compiler (`UnrollDefinitions`), which means:

- `unroll-annotation` 0.1.12 and `scala-library` 3.9.0 both put `scala.annotation.unroll` on the classpath, and whichever wins decides the build's behaviour;
- the built-in expansion applies stricter rules and rejects `ClientDriver.requestOnChannel` outright (`Cannot unroll parameters of method requestOnChannel because it can be overridden` — `@unroll` is not allowed on an overridable method);
- the stdlib annotation is `@experimental` under SIP-61, so using it would require `-experimental` here and in every downstream project.

unroll 0.2.0 renames the annotation to `com.lihaoyi.unroll`, which sidesteps all three: the collision disappears, the built-in phase never fires, and the plugin stays in charge of the expansion on both Scala 2 and Scala 3. `unroll-plugin_3` is binary cross-published at 0.2.0 and loads under 3.9.0.

**It cannot go to 0.3.0.** From 0.3.0 the plugin is cross-published per *full* Scala version (`unroll-plugin_3.8.1`, `unroll-plugin_2.13.18`, …) and the latest Scala 3 build is 3.8.1 — there is no 3.9.0 one. Scala Steward will keep proposing 0.3.0 and it will keep being wrong until upstream publishes it.

## Warnings that are new in 3.9

- `Middleware.scala` — type ascription on a tuple pattern is no longer supported.
- `Handler.scala` — two type tests on erased type arguments, now matched with wildcards. The tests are only used as predicates there, so nothing needs refining and no suppression is required.
- `SegmentCodec.Combinable` — `private[codec]`, so its implicits are not accessible where they are used. Scala 3.10 will stop finding them; the object is now public.
- `zio-http-example` — `scala.App` is deprecated since 3.8.

## Verification

| | 2.12.21 | 2.13.18 | 3.9.0 |
|---|---|---|---|
| `Test/compile` with the CI warning flag | ✅ | ✅ | ✅ |
| `zioHttpJVM/test` | — | ✅ 1799 | ✅ 1811 |
| `mimaChecks` (vs `dev.zio:zio-http:3.11.5`) | — | ✅ | ✅ |
| `fmtCheck` | — | ✅ | — |

Binary compatibility was also checked directly: every member generated at the twelve `@unroll` sites (`ClientDriver`, `NettyClientDriver`, `SSLConfig`, `Server.Config`, `DnsResolver.Config`, `ZClient.Config`) has the same JVM signature as in the published `zio-http_2.13:3.11.5` jar.


guizmaii:
> (Even thinking about 2.13)

+1

guizmaii:
@987Nabil 

>  think we should split 3.3 and 3.9 and publish 3.9 as a new artifact.

How do you want to do that? FYI, most of ZIO libs have upgrade to Scala 3.9, without keeping a 3.3 compatible branch

987Nabil:
@guizmaii 
https://discord.com/channels/629491597070827530/733728086637412422/1550041461691650110
I think it might have been a mistake to go to 3.9 directly in the other repos too.

And the blocks setup is an example of how

